PR #120, opened 2016-03-06
Added Retina display support

dscso:


jonoomph:
Thanks for the fix, and sorry for the delay! Much appreciated! I don't have a retina display, so I wasn't able to test this fully, but it doesn't seem to break anything for me. =)

Issue #139, opened 2016-03-29
import of the image fails for unknown reason


MilanV:
When I try to add a JPEG image to the project, I get the error popup "myfile.jpg is not a valid video, audio, or image file." and nothing is imported. More on that, openshot-qt prints this message to the console:

 main_window:INFO Imported media file /home/milan/myfile.jpg

instead of an error message. So I cannot know what the problem is. All I know is that I was not able to find any JPEG image which openshot-qt can import now. It worked with previous Beta versions.

OS: Linux, Debian stable
openshot: 2.0.7 - either from Ubuntu PPA or the AppImage

jonoomph:
@MilanV Can you please attach a small, example JPEG file (that you don't mind sharing with the Internet). =) I'm not able to reproduce this. It sounds like maybe some JPEG related library is missing from my AppImage, or missing from your system (not sure).

MilanV:
I'm sorry I'm not able to send any such picture because I've found it's even
more magic than I thought at first. When I copy the "problematic" image to
another location or to another name in the same directory, openshot-qt loads it
without any problem.

Can I send you any useful debug messages from my openshot run?

vchekan:
@jonoomph Ha! I got it!
My file name is all in capital letters ("DSC01171.JPG"). If I make a copy of it in small letters, it opens successfully.

jonoomph:
I'm not able to reproduce your steps... but I did find a bug related to **ignore_image_sequence_paths**. 
Please visit www.openshot.org/download, click '**Other Downloads**', and install the latest daily build of OpenShot. _It might take a couple hours for the latest version to appear._

If you still have issues, please attach a simple example (zip file with a few images in your sequence that cause the issue) and I'll take a closer look. Thanks!

vchekan:
@jonoomph I've just committed a fix for this one. There is a member which is not initialized before possible use. It is tricky to reproduce because the logic before this usage can be short-circuted if some conditions for Image Sequence are not met. For example, if "Check for previous or next image" is false, it will not do farther checks. So it can be reproduced only if there are pics in the folder which are a picture sequence.

jonoomph:
Thanks! Appreciate the debugging. This should be good and fixed now. =) Closing the issue now, but please reopen if you see any other strangeness.

PR #367, opened 2016-10-06
Fix for #139

vchekan:


jonoomph:
Thanks for the pull request.. but this was actually just fixed and committed right before you sent this. =) Closing this request out.

PR #641, opened 2017-05-14
Fixing bug in SpaceWars 3D title

eM-Ce:
Fixing the bug, that the main title was showing from the begining and was covering the blue one.


jonoomph:
Fixed the issue upstream

PR #662, opened 2017-05-22
Convert REAMDE to an AsciiDoc document

TheDcoder:
Hello! Thanks for the wonderful project 😄.
I decided to do my part and make the README more appealing by converting it to a [AsciiDoc](http://asciidoc.org/) document, it is better than markdown because it's well defined and has a good processor called [AsciiDoctor](http://asciidoctor.org/) which can make beautiful documentation.

Please have a look at the README.adoc file at my fork's repository to see the converted document.


peanutbutterandcrackers:
Wow. This is actually a whole lot better than the current one. If you're still active, could you please also make it a lot more shinier and add the OpenShot logo in, too? Perhaps do your style of designing too, and not just porting the .md to asciidoctor. I think you can do this even better (even though I like this one already).

@DylanC - Don't you think that [this (the readme)](https://github.com/TheDcoder/openshot-qt) is a lot better than what we currently have? I think, with a few more decorations, we should switch to using this rom the current one.

@ferdnyc - Any thoughts regarding this? o.O We could really use a shinier README. :smile:

peanutbutterandcrackers:
Even if we want to stick to markdown, I think the current readme does require some make-over, though.

ferdnyc:
Yeah, I saw this a while ago. The README needs updating, no question. It
still contains Launchpad URLs.

However, this is Github, and Markdown is the standard for READMEs here. As
a format it's "fine", Github have developed a rich suite of tools for
dealing with it, including online editing, and I honestly don't see any
compelling arguments _for_ converting to AsciiDoc. It feels like rebellion
just for the sake of it, or maintaining a Betamax-format video collection
because it's "clearly superior!"

DylanC:
@peanutbutterandcrackers - Yes, its much, much better! Markup really is the default for GitHub. I use it on my own project. It could use a screenshot maybe up top but thats around the only improvement I would do. 

@ferdnyc - Great job! Much appreciated. :)

TheDcoder:
I converted the readme to the AsciiDoc format out of excitement when I started using it. I agree with @ferdnyc that it just seems to be rebellious to choose AsciiDoc over traditional GitHub markdown at this point.

At the moment I am busy with my own personal projects and work so I don't have time to improve this pull-request anymore, but everyone is free to fork my version or start from scratch and improve it :smile:

ferdnyc:
> I converted the readme to the AsciiDoc format out of excitement when I started using it.

Hey, we've all been there. Ask me about the period in the mid-1990s where I wrote _everything_ in Perl, even stupid one-off scripts that could've and should've been 10 lines of bash scripting. :roll_eyes: (Actually, don't ask, because there's a danger I might actually tell you.)

peanutbutterandcrackers:
@TheDcoder - Hope you come back around to helping out OpenShot, soon. :)

jonoomph:
Sounds like markdown is what we need, but thanks again. I'm going to close this one for now, until we convert to Markdown. =)

PR #896, opened 2017-09-12
Upgrade to Angular 1.6.6

DylanC:


jonoomph:
I tested out the patch and found some strange issues with the timeline. Opening an existing project, clips don't appear to snap at first... not until dropping a clip, and then dragging it again does the snapping start to work. Not sure why upgrading Angular would cause this behavior though.

jonoomph:
Snapping also appears broken when debugging in Chrome with a few test clips. Hmmm, not sure what is going on there.

DylanC:
@jonoomph - I reverted one small change to the clip.js. It should be purely an Angular upgrade now with no code changes. Snapping seems to work for me but I will let you try it out.

PR #1188, opened 2018-01-07
Add checks for invalid file names before file export

GregoryMason:
• Check if the user is trying to save the file with a name which contains invalid characters \/:*?"<>|
• Check if the user is trying to save a file without any name/empty string (Whilst technically not a problem, it can lead to files which cause problems for other programs, so not a good idea to let it through).

NOTE: been a while since I used python, but it compiles without error, so I think I got that right (couldn't find the unit tests referred to in the contributing guidelines though). Also, I've only had limited time to understand how the project is coded, so there's a chance I've done something slightly differently from how it should be done.

### DIFF
--- a/src/windows/export.py
+++ b/src/windows/export.py
@@ -605,6 +605,37 @@ class Export(QDialog):
         self.export_button.setEnabled(False)
         self.exporting = True
 
+        # Check if file name contains invalid characters
+        invalidFileChars = set("\\/:*?\"<>|")
+        if any((char in invalidFileChars) for char in self.txtFileName.text()):
+            # Alert user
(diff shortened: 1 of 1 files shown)


DylanC:
@GregoryMason - I'm no Python expert either but it looks good to me.

@jonoomph - Can you review this pull request when you have time?

jonoomph:
Hi! This looks good, however we need to incorporate translations before I can merge this.

jonoomph:
This one still needs some work. We need to make all visible text as translatable, and verify this logic will work cross platform (windows, mac, and linux). Going to close PR for now, but feel free to reopen when this is ready for another review. =)

Issue #1262, opened 2018-02-08
Documentation tables don't wrap


ferdnyc:
The Sphinx formatting for the [online documentation](http://www.openshot.org/files/user-guide/) makes heavy use of ReStructuredText tables, which unfortunately do not use wrapping on internal text. This means that when the documentation is browsed in a narrower view, the tables have to be scrolled horizontally to view their contents.

After a fair bit of experimentation with the possibilities for RST formatting, I found that converting the images to figures with a caption and definition list was probably the best option for layout. (I tried both a custom-titled admonition and a sidebar, but neither really worked out the way I'd hope.)

I'm not thrilled with the result myself, to be perfectly honest, but ReStructuredText doesn't give a lot of formatting options. (I have to wonder if wholesale conversion to MarkDown might not work out better, but for now I worked within the flexibility of the existing RST format.) Ultimately I felt that it was "better than before", which is probably the best I can hope for.

Making that change, this:

![screenshot from 2018-02-08 17-33-25](https://user-images.githubusercontent.com/538020/36002508-6d7b8ebc-0cf8-11e8-90aa-c6228042687d.png)

becomes this:

![screenshot from 2018-02-08 17-33-37](https://user-images.githubusercontent.com/538020/36002518-74b6efd2-0cf8-11e8-86e6-789e4711cf6b.png)

I'm happy to update all of the documentation files, convert all of the image legends, and submit a PR. But I wouldn't want to go through all of that without getting buy-in that it's something desirable that would be accepted, so I thought I'd raise this here first. Further discussion of the exact formatting is eagerly encouraged.

(There's also the issue of the non-legend tables, but for now I'm leaving that aside. Perhaps the same solution would work for them.)

DylanC:
@ferdnyc - Looks good to me. What do you think @jonoomph ?

ferdnyc:
BTW, the caption ("_Figure: Animation.1_") seen in the second screenshot is both completely free-form and optional. I threw it in because I figured it might be useful to refer to the figures by name from the body text, at some point. I was also hoping to come up with a format for the legend that explicitly labeled it as corresponding to the named figure, but failed in that goal.

I chose the "Chaptername.#" format simply because the chapters aren't currently numbered, and I figured this was simpler than imposing a numbering that would have to be manually adjusted whenever chapters are added, removed, or reordered.

peanutbutterandcrackers:
[...]

PR #1272, opened 2018-02-12
Enable text wrapping in table cells (addresses #1262)

ferdnyc:
Here's the change promised in #1262, to enable text wrapping in table cells (disabled by default in the Read The Docs theme for Sphinx). It's two simple changes and one... not so simple. It turns out, having table cells wrap is a deceptively complex thing. (And I now understand the theme's reluctance to just enable wrapping by default, though I can't say I like their solution of forcing side-scrolling tables everywhere any better.)

The issue, basically, is how the cell widths are balanced by Sphinx. Sometimes it makes OK decisions. Other times, the columns are wildly imbalanced, requiring intervention by the author to get the tables sized correctly. So, after enabling wrapping, I went through the entire manual, looked at all of the tables, and manually adjusted them to what I felt was a reasonable balance by manually specifying widths for some or all columns.

Doing that is the difference between this (the default output, with wrapping enabled):

screenshot from 2018-02-11 22-28-13

And this (manually adjusted to make the second column wider):

screenshot from 2018-02-11 22-28-26

Note that that's the "full-width" formatting, or the "starting point" for column sizing. Even with explicit column widths, Sphinx's tables are flexible and will resize. With the browser window reduced to half the screen width, those become (respectively):

screenshot from 2018-02-11 22-28-45

screenshot from 2018-02-11 22-28-52

On my phone (Samsung Galaxy S6, Chrome browser) they end up looking identical, because the columns compress to the same state regardless of the specified widths.

(Honestly, while I know that one of these is a screenshot from my "unadjusted" export of the docs, and the other is the output with my column-width fixes applied, I'm really not sure _which_ of these is which! They are, as I said, identical.)

screenshot_20180211-224233

screenshot_20180211-224211

Really, I think it does quite a good job, especially once the widths are tweaked as needed.

Now, here's the slightly bad news. The trick to doing this, as I explain in the commit message for the third (large) commit, is to use reStructuredText's explicit `.. table:` directive, as it can be provided with a `:widths:` argument specifying column widths. When using `.. table:`, the table syntax is the same, but the table gets nested below the directive, requiring a 5-space indent. So this table (from the repo HEAD):

```rst
==  ==================  ============
#   Name                Description
==  ==================  ============
1   Main Toolbar        Contains buttons to open, save, and export your video project.
2   Function Tabs       Switch between Project Files, Transitions, and Effects.
3   Project Files       All audio, video, and image files that have been imported into your project.
4   Preview Window      This is the area that the video will playback on the screen.
5   Edit Toolbar        This toolbar contains buttons used for snapping, inserting markers, and jumping between markers.
6   Zoom Slider         This slider will adjust the time-scale of your timeline.
7   Play-head / Ruler   The ruler shows the time-scale, and the red line is the play-head. The play-head represents the current playback position.
8   Timeline            The timeline visualizes your video project, and each clip and transition in your project.
9   Filter              Filter the list of items shown (project files, transitions, and effects) by using these buttons and filter textbox. Enter a few letters of what you are looking for, and the results will be shown.
==  ==================  ============
```

Becomes this (after my patch):

```rst
.. table::
     :widths: 5 22 73
     
     ==  ==================  ============
     #   Name                Description
     ==  ==================  ============
     1   Main Toolbar        Contains buttons to open, save, and export your video project.
     2   Function Tabs       Switch between Project Files, Transitions, and Effects.
     3   Project Files       All audio, video, and image files that have been imported into your project.
     4   Preview Window      This is the area that the video will playback on the screen.
     5   Edit Toolbar        This toolbar contains buttons used for snapping, inserting markers, and jumping between markers.
     6   Zoom Slider         This slider will adjust the time-scale of your timeline.
     7   Play-head / Ruler   The ruler shows the time-scale, and the red line is the play-head. The play-head represents the current playback position.
     8   Timeline            The timeline visualizes your video project, and each clip and transition in your project.
     9   Filter              Filter the list of items shown (project files, transitions, and effects) by using these buttons and filter textbox. Enter a few letters of what you are looking for, and the results will be shown.
     ==  ==================  ============
```
It's not ideal, but it's not unworkable. Using `.. table::` does also give us the option of converting to `.. list-table::` format, which dumps the ASCII art and turns that code into:

```rst
.. list-table::
     :widths: 5 22 73
     :header-rows: 1

     * - #
       - Name
       - Description
     * - 1
       - Main Toolbar
       - Contains buttons to open, save, and export your video project.
     * - 2
       - Function Tabs
       - Switch between Project Files, Transitions, and Effects.
```
...and so on.

There are definite advantages to that format, especially as the column contents get more complex, and _especially_ especially if they end up containing RST markup. (That's something else I'd like to explore at a later date, as I have some thoughts in that area.) So, if we want to convert the doc tables to `.. list-table::` formatting, then that's a possibility for a future PR. The rendered table is the same, so it's purely an issue of source readability.

Oh, I made one other change (as you'll see in the second commit), to replace the ASCII arrows (`->`) used at a couple of points in the tables (and body text) with real Unicode → arrows. That was purely an aesthetic decision, as word-wrapping would cause the ASCII arrows to get split in half sometimes. (e.g:)
> File menu -
> \> Import Media...

which will instead now be (worst-case):

> File menu →
> Import Media...

### DIFF
--- a/doc/_templates/layout.html
+++ b/doc/_templates/layout.html
@@ -1,4 +1,5 @@
 {% extends "!layout.html" %}
+{% set css_files = css_files + ["_static/tablefix.css"] %}
 
 {% block footer %}
     {{ super() }}
--- a/doc/animation.rst
+++ b/doc/animation.rst
@@ -31,15 +31,18 @@ Overview
 
 .. image:: _static/animation-overview.jpg
 
-==  ==================  ============
-#   Name                Description
-==  ==================  ============
-1   Green Property      When the play-head is on a key frame, the property appears green
-1   Blue Property       When the play-head is on an interpolated value, the property appears blue
-2   Value Slider        Click and drag your mouse to adjust the value (this automatically creates a key frame if needed)
-3   Play-head           Position the play-head over a clip where you need a key frame
-4   Key frame Markers   Small green tick marks are drawn at all key frame positions (on a clip)
-==  ==================  ============
+.. table::
+     :widths: 5 32
+
+     ==  ==================  ============
+     #   Name                Description
+     ==  ==================  ============
+     1   Green Property      When the play-head is on a key frame, the property appears green
+     1   Blue Property       When the play-head is on an interpolated value, the property appears blue
+     2   Value Slider        Click and drag your mouse to adjust the value (this automatically creates a key frame if needed)
+     3   Play-head           Position the play-head over a clip where you need a key frame
+     4   Key frame Markers   Small green tick marks are drawn at all key frame positions (on a clip)
+     ==  ==================  ============
 
 Key Frames
 ----------
@@ -50,13 +53,16 @@ the properties again. All animations require at least 2 key frames, but can supp
 
 To adjust the **interpolation mode**, right click on the small graph icon next to a property value.
 
-==================  ============
-Name                Description
-==================  ============
-Bézier              Interpolated values use a quadratic curve, and ease-in and ease-out
-Linear              Interpolated values are calculated linear (each step value is equal)
-Constant            Interpolated values stay the same until the next key frame, and jump to the new value
-==================  ============
+.. table::
+     :widths: 12
+
+     ==================  ============
+     Name                Description
+     ==================  ============
+     Bézier              Interpolated values use a quadratic curve, and ease-in and ease-out
+     Linear              Interpolated values are calculated linear (each step value is equal)
+     Constant            Interpolated values stay the same until the next key frame, and jump to the new value
+     ==================  ============
 
 For more info on clip properties, see :ref:`clip_properties_ref`. For more info on preset animations, see :ref:`clip_presets_ref`.
 For more info on creating key frames for location and scale, see :ref:`clip_transform_ref`.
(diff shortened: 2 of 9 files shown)


peanutbutterandcrackers:
@frednyc - Awesome, man! I hope Mr. Thomas gets back to this soon. And even more than that, I hope you keep on contributing to this project. You're the man, good sir! 👍

DylanC:
@ferdnyc - Thanks for the PR. Hopefully @jonoomph can review this soon.

ferdnyc:
> There are definite advantages to that format, especially as the column contents get more complex, and especially especially if they end up containing RST markup. (That's something else I'd like to explore at a later date, as I have some thoughts in that area.)

Which I shall set aside, as it seems that this isn't the time to be worrying about the documentation.

peanutbutterandcrackers:
@ferdnyc - I think OpenShot could use any help in any aspect, to become a better Video Editor.
Mr. Thomas has neither made any commits nor has replied to my mails. I think he'll be back to review this PR soon, though. :)

peanutbutterandcrackers:
@ferdnyc - Good sir, you'd perhaps want to go through the remaining half of the tables. :smile:

ferdnyc:
There's no reason the remaining tables need to be modified, we don't want to specify column widths on them. They could be modified just for consistency's sake, I suppose, even though they don't need to use any `..table` features. But I'd want to make a case for rewriting _all_ of the tables in `..list-table` form, instead.

As I said, one of the advantages of `..list-table` is that it makes the table code much cleaner and more readable when the table contains RST markup. Sphinx [supports markup](http://www.sphinx-doc.org/en/stable/markup/inline.html) for things like `:guilabel:` for UI components, `:kbd:` for keystrokes, and `:menuselection:` for, well, menu selections. The OpenShot manual currently doesn't use any of those, but I think it should. In both the prose and (where appropriate) the tables.

But embedding those directives in the current ASCII-art tables would be... ugly, to say the least. With `..list-table` formatting, it's far more readable.

peanutbutterandcrackers:
@ferdnyc - I think you should provide some screenshots - samples that will be convincing enough to make a move towards list-table, good sir. :)
[And I really appreciate your efforts. You're awesome! Honestly.]

ferdnyc:
Makes sense. I'll file a new Issue with the proposal. Give me a few days.
End of the week hopefully, end of the weekend at the latest.

peanutbutterandcrackers:
Whoa! @ferdnyc is a Contributor, too, now. Awesome! Now you get to label issues and stuff.

Looking forward to your proposal. :+1:

ferdnyc:
> Whoa! @ferdnyc is a Contributor, too, now. Awesome! Now you get to label issues and stuff.

Heh. No, turns out that's just Github's auto-label indicating I've had code committed to the project. No special powers, just a piece of flare.

peanutbutterandcrackers:
Oh... I see. Hmm.

PR #1341, opened 2018-03-06
Use source path for images, not destination

ferdnyc:
Before I get to the table changes, a tweak I made in building the docs locally.

Currently, the docs source references images using their output location in the `_build` directory, which only becomes valid _after_ the build process is complete, when the `html_static_path` directive in `conf.py` causes `sphinx-build` to copy the image files into `_build/html/_static`. This means that, during build time, `sphinx-build` will spew tons of error messages about missing image files:

>% make html
>sphinx-build-3 -b html -d _build/doctrees   . _build/html
>Running Sphinx v1.6.6
>loading pickled environment... done
>building [mo]: targets for 0 po files that are out of date
>building [html]: targets for 12 source files that are out of date
>updating environment: 0 added, 9 changed, 0 removed
>reading sources... [100%] transitions                                           
>**$PWD/animation.rst:33: WARNING: image file not readable: _static/animation-overview.jpg
>$PWD/animation.rst:80: WARNING: image file not readable: _static/curve-presets.jpg
>$PWD/animation.rst:90: WARNING: image file not readable: _static/file-properties.jpg**

...and so on, for every image in the docs. The abundance of these (expected, ignorable) errors can make it difficult to spot any _genuine_, problematic errors output during the build process, as they get lost in the noise.

I originally worked around the issue by symlinking `_static` to `_build/html/_static` in the documentation directory, but that caused _two_ copies of every image to be written into the output directory when building the docs.

The reason for that was, when images are referenced using a _valid_ source path, `sphinx-build` will **automatically** copy them into to `_build/html/_images/` (rather than `_build/html/_static/`). It also adjusts the path in the generated output HTML accordingly. So, as soon as I made `_static` a valid source path, `sphinx-build` started duplicating all of the images, and the copies in `_build/html/_static/` became redundant.

So, this patch leaves `sphinx-build` to do its thing automatically, by specifying the path to all images as `images/filename.png` instead of `_static/filename.png`. `images` is removed from the `html_static_paths` list, so that they are not also copied into `_build/html/_static/`.

Beyond allowing for a cleaner build process with better error-checking (since the existence of all referenced image files will be verified during the copying process, and any missing images reported), the only effect of this change is the migration of non-theme images (screenshots, figures) from the `_static` subdirectory to the `_images` subdirectory, in the generated output HTML. This may potentially require minor adjustment to existing webserver configs when serving the docs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 .idea
 .settings
 idea/*
+doc/_build
+**/__pycache__/
--- a/doc/animation.rst
+++ b/doc/animation.rst
@@ -29,7 +29,7 @@ values at certain points on a clip, and OpenShot does the hard work of interpola
 Overview
 --------
 
-.. image:: _static/animation-overview.jpg
+.. image:: images/animation-overview.jpg
 
 .. table::
      :widths: 5 32
@@ -76,7 +76,7 @@ the expected value and then back (producing a bounce effect).
 
 To choose a curve preset, right click on the small graph icon next to a key frame.
 
-.. image:: _static/curve-presets.jpg
+.. image:: images/curve-presets.jpg
 
 Image Sequences
 ---------------
@@ -86,7 +86,7 @@ drag and drop one of them into OpenShot, and you will be prompted to import the 
 To adjust the frame rate of the animation, right click and choose **File Properties** in the **Project Files** panel,
 and adjust the frame rate. Once you have set the correct frame rate, drag the animation onto the timeline.
 
-.. image:: _static/file-properties.jpg
+.. image:: images/file-properties.jpg
 
 .. table::
      :widths: 5 25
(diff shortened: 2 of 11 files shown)


ferdnyc:
Oh, right, I forgot: And I snuck in a couple of `.gitignore` additions, to excude the entire `docs/_build` output tree, and **all** `__pycache__` directories anywhere in the repo.

peanutbutterandcrackers:
Awesome! :+1:

jonoomph:
Thanks! Looks good. I'm going to attempt to deploy this right now, and fix server paths as needed. =)

ferdnyc:
Awesome. It really does make the documentation-building process much
cleaner, if the only errors displayed are REAL errors. (And makes it much
less likely you'll miss one in all the noise, as I have a few times.)

jonoomph:
Nice, deployed and looks good! https://www.openshot.org/static/files/user-guide/index.html

Issue #1329, opened 2018-03-03
Request for New Target: Chromebook


gary9872:
I was helping out someone recently with some video work and they presentations were at school. Neither of the two projects worked (the audio). The culprit was the chromebooks, which was the standard I believe throughout that entire school system.

So to that end I am requesting to add Chromebook  under Device > Target >

using webm VP8 or VP9 for video, and Vorbis for audio as per google support page.

https://support.google.com/chromebook/answer/183093?hl=en

If you like I can submit a diff patch against 2.4.1 or git for it 

Thanks,

Gary

peanutbutterandcrackers:
@DylanC - I think you should take a look at this. He has made an offer we can not refuse.

@gary9872 - Hey! Thanks! I'm so glad that you are helping OpenShot become a better Video Editor. Please do stick around and help us out with other bugs and stuffs too. We need more developers here. Currently, it's just Mr. Thomas alone. We really appreciate the help. :)

DylanC:
@gary9872 - If you could submit a PR that would be great! Might take a while to get it reviewed but at least it will get in after a while.

peanutbutterandcrackers:
Yeah! You go, @gary9872 !!! :smile:

gary9872:
Add to openshot-qt/src/presets directory

chromebook.xml

https://pastebin.com/G6sHeCRh

peanutbutterandcrackers:
@gary9872 - Actually, if you could make a PR, that would probably land you a credit (at least in the git logs) of doing so... Are you sure you do not want to make a PR?

Also, could you please, if you decide to make a PR, take care of #201 ?

Thanks!

DylanC:
I'm adding the file here in case the pastebin disappears or anything like that. (Note its supposed to be just .xml but for uploading purposes I added the .txt to the end.
[chromebook.xml.txt](https://github.com/OpenShot/openshot-qt/files/1784099/chromebook.xml.txt)

@peanutbutterandcrackers - I can do a PR for this if you like.

peanutbutterandcrackers:
@DylanC - Well, if @gary9872 does not want to, then please, by all means. :)
Would you like to take care of #201, too, please?

DylanC:
@peanutbutterandcrackers , @gary9872 - PR added. I should be able to get around to #201 too.

peanutbutterandcrackers:
@DylanC - Awesome!

@gary9872 - I hope you stick around and make some PRs too, good sir!
[...]

PR #1344, opened 2018-03-06
Add preset xml file for chromebook target

DylanC:
This should take care of #1329

### DIFF
--- a/src/presets/chromebook.xml
+++ b/src/presets/chromebook.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!DOCTYPE openshot-export-option>
+<export-option>
+        <type translatable="True">Device</type>
+        <title translatable="True">Chromebook</title>
+        <videoformat>webm</videoformat>
+        <videocodec>libvpx</videocodec>
+        <audiocodec>libvorbis</audiocodec>
+        <audiochannels>2</audiochannels>
+        <audiochannellayout>3</audiochannellayout>
+        <videobitrate
+         low="384 kb/s"
+         med="5 Mb/s"
+         high="15.00 Mb/s"></videobitrate>
+        <audiobitrate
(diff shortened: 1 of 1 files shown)


peanutbutterandcrackers:
:smile: 
25vxez

ferdnyc:
The standard/default Chromebook samplerate is 44.1kHz, not 48kHz? That's surprising and unfortunate.

DylanC:
@ferdnyc - We should be able to still update this anyway.

DylanC:
@ferdnyc - I found this:
https://developer.chrome.com/native-client/devguide/coding/audio

This page has the following information:
- sample rate: 44,100 Hz or 48,000 Hz
- bit depth: 16
- channels: 2 (stereo)

Should this PR be updated to 48Hz?

ferdnyc:
I guess not, since (now that I've checked) **all** of the existing presets use 44.1kHz except for `avchd.xml`, `dvd_pal.xml`, and `dvd_ntsc.html`. I've gotten used to thinking of 48kHz as sort of "standard video rate", since it's not only what DVD uses, but AFAIK all broadcast/cable is sampled at that rate as well (if not higher). But it makes sense to follow OpenShot convention, which seems to be 44.1kHz. If that were going to be changed, probably better to change the whole preset collection (or most of it) all at once.

ferdnyc:
Although, I noticed that even the two youtube presets use 44.1kHz, when [YouTube's recommended settings article](https://support.google.com/youtube/answer/1722171) explicitly calls for either 48kHz or 96kHz audio. So, maybe it's worth considering an audit of the presets' samplerate settings.

DylanC:
@ferdnyc - Yes, I think its worth some more investigation and a review of our current sample rates.

jonoomph:
Merged, thanks! I agree, we probably need to update a bunch of these presets, and add some other common missing ones.

Issue #96, opened 2016-02-15
Wish - Remember where you got the last clip


poetimp:
When I add a clip to OpenShot I often go back and get something else from the same folder. Yes, I know I can add it all at the same time and not have to go back but that would require forethought :) Anyway, When I go to add another clip I have to traverse my way all the way to my deeply nested directory every time. It would be nice if it remember where it got the last clip and start me there.

jarlarntzen:
:+1: Completely agree to this.

schuko24:
would be very useful, I agree

peanutbutterandcrackers:
@ferdnyc - This one has to be implemented in the source, I believe. Or do you thin PyQt5 has got something for this too? o.O Just curious.

peanutbutterandcrackers:
@DylanC - As this has already been implemented (thanks to the awesome @ferdnyc), I think it is safe to close this issue, too. :)

DylanC:
@peanutbutterandcrackers - Thanks! @ferdnyc is indeed awesome! :)

PR #1366, opened 2018-03-12
Track file-import path, the same way export path is tracked (issue #96)

ferdnyc:
OpenShot currently forces the open-file dialog for "Import files..." to open at the location of the project file every time. That's is fine for users who store their video files with the project file, which is probably good practice. But for users who don't, it creates frustration.

This change adds `import_path` tracking to the project file, the same way `export_path` is currently tracked, and attempts to use the stored `import_path` (if present) as the starting location of both "Import File..." and the  missing-file recovery process. So if the user imports from their project file location, nothing will really change for them. For other users, it will save them a lot of navigation.

Implements #96.

### DIFF
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -893,9 +893,11 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
 
     def check_if_paths_are_valid(self):
         """Check if all paths are valid, and prompt to update them if needed"""
-        # Get project folder
+        # Get import path or project folder
         starting_folder = None
-        if self.current_filepath:
+        if self._data["import_path"]:
+            starting_folder = os.path.join(self._data["import_path"])
+        elif self.current_filepath:
             starting_folder = os.path.dirname(self.current_filepath)
 
         # Get translation method
@@ -914,9 +916,10 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
                 starting_folder = QFileDialog.getExistingDirectory(None, _("Find directory that contains: %s" % file_name_with_ext), starting_folder)
                 log.info("Missing folder chosen by user: %s" % starting_folder)
                 if starting_folder:
-                    # Update file path
+                    # Update file path and import_path
                     path = os.path.join(starting_folder, file_name_with_ext)
                     file["path"] = path
+                    get_app().updates.update(["import_path"], os.path.dirname(path))
                 else:
                     log.info('Removed missing file: %s' % file_name_with_ext)
                     self._data["files"].remove(file)
--- a/src/classes/timeline.py
+++ b/src/classes/timeline.py
@@ -76,7 +76,7 @@ class TimelineSync(UpdateInterface):
         """ This method is invoked by the UpdateManager each time a change happens (i.e UpdateInterface) """
 
         # Ignore changes that don't affect libopenshot
-        if len(action.key) >= 1 and action.key[0].lower() in ["files", "history", "markers", "layers", "export_path", "scale"]:
+        if len(action.key) >= 1 and action.key[0].lower() in ["files", "history", "markers", "layers", "export_path", "import_path", "scale"]:
             return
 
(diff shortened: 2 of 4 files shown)


ferdnyc:
I coded this against the v2.4.1 release, because when I tried the current git HEAD "Import File..." wasn't working for me. Every time I attempted to import a file I got the "Foo is not a valid video, audio, or image file." message (which was not true) and nothing was imported. Perhaps I needed the "newer ffmpeg version" mentioned in one commit?

peanutbutterandcrackers:
Wow. Awesome! This will be so very useful. Thanks! [I hope to test this out ASAP. ]

peanutbutterandcrackers:
I just merged this branch to the openshot:master branch, locally (had some merge conflicts that I had to resolve) but it works just great.

Could you perhaps update this so that it will be ready-to-merge so that Mr. Thomas will not have to deal with the conflicts?

ferdnyc:
Problem is, I can't make these changes to the current master because, as I said, I couldn't get it to import files _at all_. (Even unmodified.) And I have no idea what was causing that. (OpenShot:master ran without problems for you? Curiouser and curiouser.)

Regardless, point is if I rebased to the master branch, I wouldn't be able to test my changes — I'd have to be writing/submitting blind, _hoping_ the committed code runs OK simply because the same changes worked as modifications to the v2.4.1 source. Not real comfortable submitting untested code. But, here's a patch version against the master branch HEAD. (It just took `git apply -C2` instead of the default `-C3`, for it to apply cleanly.)

[master_1366.patch.txt](https://github.com/OpenShot/openshot-qt/files/1802175/master_1366.patch.txt)

ferdnyc:
Having bisected the code, I can now state definitively that the rotation commit breaks OpenShot's ability to import any files, for me.

> % git bisect good                        
> 3accac8bed9d07bc612cae7044a49c713496227f is the first bad commit
>commit 3accac8bed9d07bc612cae7044a49c713496227f
>Author: Jonathan Thomas <jonathan@openshot.org>
>Date:   Sat Feb 3 01:58:55 2018 -0600
>
>    Auto-rotate thumbnails for Readers with 'rotate' metadata. This requires a slightly newer version of FFmpeg, otherwise the 'rotate' tag is never found.

I currently have `ffmpeg-libs-3.3.6-1.fc27.x86_64` installed, which may be older than the "slightly newer" version required? If so, we should probably figure out what that required version is, so that things like RPM spec files can be updated with the proper version dependency.

peanutbutterandcrackers:
`$ ffmpeg --version` says that I have `ffmpeg version 2.8.11-0ubuntu0.16.04.1 Copyright (c) 2000-2017 the FFmpeg developers`.

peanutbutterandcrackers:
Do you think that this issue (that commit breaking import) needs to be brought to Mr. Thomas's attention? Can something be done in the source?

ferdnyc:
Well, I don't know, I don't know what's causing it.

I do notice that there's a commit in libopenshot that seems to correspond

> commit f2b0f3a0f4612bd34080d9e4e62cc271a718ecc1
> Author: Jonathan Thomas <jonathan@openshot.org>
> Date:   2018-02-03 01:57:18 -0600

>    Adding metadata from format, audio stream, and video streams to ReaderBase.info, which in some cases includes the 'rotate' metadata added by certain cameras, and audio metadata like title, album, artist, copyright, dates, etc... Auto-Rotates any Clip with Reader metadata 'rotate' attribute.

So the issue may just be that I was testing `openshot-qt` with `libopenshot-0.1.9` release, when it really requires a later version with that commit, I guess you'd call it `libopenshot-0.1.10-dev` or whatever.

ferdnyc:
> So the issue may just be that I was testing openshot-qt with libopenshot-0.1.9 release, when it really requires a later version with that commit, I guess you'd call it libopenshot-0.1.10-dev or whatever.

That was it exactly. After building libopenshot from its git HEAD, and running the openshot-qt HEAD against that version, file importing functionality was restored. I'll file an issue, I guess. And then rework this against master.

PR #1372, opened 2018-03-12
Track file-import path, the same way export path is tracked (issue #96)

ferdnyc:
OpenShot currently forces the open-file dialog for "Import files..." to open at the location of the project file every time. That's is fine for users who store their video files with the project file, which is probably good practice. But for users who don't, it creates frustration.

This change adds `import_path` tracking to the project file, the same way `export_path` is currently tracked, and attempts to use the stored `import_path` (if present) as the starting location of both "Import File..." and the missing-file recovery process. So if the user imports from their project file location, nothing will really change for them. For other users, it will save them a lot of navigation.

Implements #96.

### DIFF
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -893,9 +893,11 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
 
     def check_if_paths_are_valid(self):
         """Check if all paths are valid, and prompt to update them if needed"""
-        # Get project folder
+        # Get import path or project folder
         starting_folder = None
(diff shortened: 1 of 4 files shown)


ferdnyc:
I was worried that changing the project file format might cause issues, but even with this patch older project files load without issue, and openshot-qt 2.4.1 release has no problems loading project files containing `import_path` (though it ignores it). So it looks like there are no concerns there.

jonoomph:
Nice! I like this one a bunch!

Issue #920, opened 2017-09-21
Inconsistent use of navigation icons in Time Line navigation bar


itsharrie:
System Details:

Operating System / Distro: Linux Mint 18.2 64bit
OpenShot Version: 2.4.0

**Issue Description and steps to reproduce:**
In the **Video Preview** frame the orange icons **|<** and **>|** are used to jump to start and end of the clip respectivily.
The same orange icons are used in the **Time Line** navigation bar to jump to previous and next marker. This is inconsistent and may confuse the user.

Proposed solution:
1. In the **Time Line** navigation bar use icons **'<** and **>'** (replace **'** with the marker icon used in the **Time Line**) for jumping to previous and next marker respectivily. Grey out these controls if no markers present.
2. To match the style of the **Time Line** navigation bar to the **Video Preview** navigation bar, the **Add Marker** control may be positioned inbetween the **Jump to Start** and **Jump to End** controls:**'<   V   >'**
3. Change the color of the three controls in the **Time Line** navigation bar as well as the marker icons in the **Time Line** to (light) blue to emphasize functional consistency.

ferdnyc:
I'm going to close this one, because:

1. If there are no markers (or even if there are), when a clip is selected, Prev/Next Marker _will_ jump to the start/end of the selected clip; the clip boundaries are also treated as markers. So, graying them out when no (explicit) markers are present would not be desirable.
1. That also means that, in the presence of no markers, the functionality actually _does_ pretty well mirror the Preview transport controls that use the same icons.
1. The icons used are stock theme icons, they're not re-colorable (and not necessarily orange). They're also not easily customizable for the same reason.

ferdnyc:
(Oh, right, and I already fixed the fact that the **same** icons were used in both places, with #1381. The Preview toolbar wasn't using the correct skip-start/skip-end media transport icons.)

PR #1381, opened 2018-03-14
Icon/pointer fixes

ferdnyc:
Two changes to the application chrome:

1. Switch the "Jump to Start/End" icons to `media-skip-forward`/`-backward`. The icons previously being used were `go-first` and `go-last` from the navigation set, which are also used for the marker-skipping controls on the timeline toolbar. Displaying identical icons that have two completely different meanings leads to confusion. (Addresses #920 .)
    
    So in Gnome Shell on Fedora (where OpenShot uses platform icons), the controls look like this:    
    screenshot from 2018-03-14 06-44-50

1. I re-drew the razor-tool cursor image `razor_line_with_razor` in Inkscape as an SVG, then exported the PNG from there. I found that with the previous cursor, the dotted blue line was difficult to make out against the application's dark theme.The new one's lines are a bit cleaner, it's slightly taller, and it uses a _much_ lighter blue for the dotted line. (It's practically white, I may have overdone it a bit).
    
    Before & After:
    openshot-oldrazor
    openshot-newrazor
    (It's hard to see but the knife part is outlined in dark gray, so it doesn't disappear against light colors.)


peanutbutterandcrackers:
Awesome!

jonoomph:
I'm a bit worried this might result in missing icons on Windows and Mac, so we'll need to do some testing once this version is done building installers.

jonoomph:
Oops, hit enter too quick, haha. Thanks again for a great PR!!!

ferdnyc:
> I'm a bit worried this might result in missing icons on Windows and Mac, so we'll need to do some testing once this version is done building installers.

That's a good point, and certainly something I didn't (couldn't) test, but I'm _hoping_ it'll be OK. The icon files were already present in the source tree (`src/images/Humanity/actions/16/media-skip-{backwards,forwards}.svg`), _and_ were already listed in `src/images/openshot.qrc`, so it was just a matter of saying "Hey, use these!" in the UI file. And, certainly, `media-skip-*` are pretty universally-standard icons, every bit as much as `go-prev` and `go-next`. But I imagine it would be `openshot.qrc` that the build servers would be working, off to determine the icon file set to package (and/or actual usage in the source code), yes?

peanutbutterandcrackers:
OpenShot certainly looks a *lot* better with this PR merged, though. Awesome work, @ferdnyc!

I hope this will be all right on Windows and Macs too...

Issue #1402, opened 2018-03-22
Can I produce video in 4/1 aspect ratio?


PRORACING-SK:
All video formats are only 16/9 or 4/3. But I need video in 4/1 !!! How to...?

N3WWN:
Hi @PRORACING-SK !

Give this page a look to create a custom profile:  https://www.openshot.org/static/files/user-guide/profiles.html#custom-profile

I haven't tried to create a profile with a custom aspect ratio, but this would be where I would start if I did. 😀

PRORACING-SK:
Thanks N3WWN. Something works but something not. Rendered video is in right resolution but the preview window is still 16:9 or 4:3 and you don't see how the result will be during work. :-(

![image](https://user-images.githubusercontent.com/37664665/37787124-901beb5c-2dfe-11e8-9c30-df9ffc116386.png)

N3WWN:
Just to be certain, are you setting your custom profile for the project (using the "filmstrip" icon or the `File`->`Choose Profile` menu) or are you setting the custom profile in the `Video Profile` optin in the `Export Video` dialog?

Assuming that you're setting the project's profile, I'm looking at the code now to see if I can tackle this issue.

PRORACING-SK:
Of course I changed both.

N3WWN:
I think I've got the preview aspect ratio working now, but won't be able to submit a PR today.  Perhaps tomorrow.

gary9872:
@PRORACING-SK Can you post your custom profile file (inline or pastebin) if you feel it would be useful for others. Curious? what application is there for 4:1?

PRORACING-SK:
I'm happy to hear you solved the problem. How can I get the changes to work? I downloaded the installer today and reinstalled it but with no success. It was still the same. 

We are going to use the outdoor LED screen with the aspect ratio 4:1 as you can see on the picture. We will use it to display racing results and advertising videos of our sponsors. 

![image](https://user-images.githubusercontent.com/37664665/37863609-d7804390-2f60-11e8-9b4a-c3cb34c4a84d.png)


Enclosed my profile-file. Can you check if it is okay please?

[proracing-4-1.txt](https://github.com/OpenShot/openshot-qt/files/1844327/proracing-4-1.txt)

peanutbutterandcrackers:
[...]

Issue #1144, opened 2017-12-20
how to fix aspect ratio in preview for 3:4 VHS DVD video


Tbouricius:
**System Details:**
* Operating System / Distro: ?
* OpenShot Version: ?
* _Please attach log files if crash_

**Issue Description and steps to reproduce:**

DylanC:
@Tbouricius - Is the aspect ratio still 4:3 ?

Issue #1093, opened 2017-12-01
Wish: Following features:


peanutbutterandcrackers:
1. Color grading capabilities
2. Optional frame-by-frame view in the timeline for more precise cutting
     It would be nice if it went from 1 second in timeline, to frame-by-frame.
3. Move-able, drag-able, click-able, delete-able keyframes
4. Aspect Ratio Manipulation
5. Ability to check for marker's properties i.e. the exact time/position of where the marker is placed.
6. Ability to change colur of how the clip looks in the timeline. Not the color of the video (although it would be nice too) but the color of the clip marker in the timeline. That would help organize the work even more, IMHO.
7. Themes?
8. The ability to save custom presets

Thank you, Mr. Jonathan Thomas and other developers for this wonderful software. Your hard work is very much appreciated. May God Bless you.

With much appreciation,
From Nepal :)

ferdnyc:
> 2. Optional frame-by-frame view in the timeline for more precise cutting
>   It would be nice if it went from 1 second in timeline, to frame-by-frame.

Is it... **not** frame-by-frame? I mean, The scrubbing controls (arrow keys) move frame-by-frame, no matter what, and at "1 second" zoom (on my system) each frame looks like it's about _three_ pixels wide (at 25fps), even without fullscreening OpenShot. So, each individual frame is already represented by a distinct, unique position on the timeline.

Or are you referring to the tickmarks on the ruler, which are still only placed at half-second intervals? At "1 second" zoom, the ruler probably _could_ contain tickmarks for individual frames. Or at least every second (odd) frame, with the spaces between representing the even-numbered ones.

peanutbutterandcrackers:
It is frame-by-frame. But yeah, I am referring to the tickmark thingies. But still, having the time-line show everything frame-by-frame would allow one to have a lot more control. For one, a picture could be added on top-of a frame rather easily - to produce a lot of cool stuffs like the shooting star memes and stuff. Secondly, keyframes would be more identifiable. So, that should help a user out with the editing.

Kdenlive also has frame-by-frame timeline view. Pretty neat, actually. :)

peanutbutterandcrackers:
Haha. I know.

Here are some screenshots:
![screenshot from 2018-03-15 12 09 21](https://user-images.githubusercontent.com/26764547/37447362-e8bc96d6-2849-11e8-88fc-8e48f60faf37.png)
[...]

Issue #835, opened 2017-08-13
Aspect Ratio incorrect for Video Preview.


trosanelli:
**System Details:**
* Operating System / Distro: ? Ubuntu 17.04
* OpenShot Version: ? 2.3.4
* _Please attach log files if crash_

**Issue Description and steps to reproduce:**
I set the video profile to HD 1080p 25 fps. I place my videos clips in the project file. The video clips are a .MTS format. When I add them to the track or preview them from the project files. The video is elongated like the aspect ratio is off.

DylanC:
@trosanelli - Any luck with the latest daily builds?

vitaly-zdanevich:
The same for me too, mpeg video. You can get example file here
https://cloud.mail.ru/public/4ksS/zt6oPGPaH

trosanelli:
DylanC - I was unaware of the daily builds. I will try it out and let you know.

DylanC:
@trosanelli - Thanks. Do let us know.

trosanelli:
@DylanC - I just checked it with the daily build and still does the same thing. I was using Openshot 1.4.3 and it worked fine when I upgraded to the current version I started having the problems.

DylanC:
@trosanelli  - No problem. Thanks for the feedback.

PR #1407, opened 2018-03-23
[WIP] Check in changes which alter the preview window to apply the proper a…

N3WWN:
…spect ratio.  Now, each time a project is loaded, the Profile selection dialog is opened in order for the user to verify the selected profile.  Upon closing of the Profile dialog, the preview window aspect ratio is updated.  Modified SetAspectRatio() in video_widget.py instead of creating a new function since the existing function does not appear to be used anywhere.

I worked this specifically for https://github.com/OpenShot/openshot-qt/issues/1402, but this PR should (partially or completely) resolve all of : 

https://github.com/OpenShot/openshot-qt/issues/456
https://github.com/OpenShot/openshot-qt/issues/731
https://github.com/OpenShot/openshot-qt/issues/835
https://github.com/OpenShot/openshot-qt/issues/1093 (item # 4)
https://github.com/OpenShot/openshot-qt/issues/1144
https://github.com/OpenShot/openshot-qt/issues/1402


DylanC:
@N3WWN - Awesome, thanks! :) Might take care of many issues at once.

ferdnyc:
> Now, each time a project is loaded, the Profile selection dialog is opened in order for the user to verify the selected profile.

If I'm interpreting that correctly (and apologies if I'm not), the Profile dialog will be displayed _every_ time a project file is opened? To what purpose?

When I save a project file with a profile I've selected... well, then the whole _purpose_ of storing it there is so that I **don't** have to select that information again. Certainly not every time I load the project! Why do I need to keep "verifying" what I've already selected? (And if I _want_ to verify the profile selection, well, the current profile is displayed in the titlebar of the main OpenShot window.

If that's not enough for people, and they're ending up working under the wrong profile because they're not reading the window title (or maybe they don't see titlebars, because they're fullscreened or whatever), then why not move the profile selector _itself_ into the main interface? 

That `Profile:` dropdown list from the Choose Profile dialog window (the only interactive element there, other than the Close button) could be placed **directly** on OpenShot's toolbar, in place of the current Choose Profile button. (Though I think it'd be better moved to the right, after all the other buttons — or even onto its own, separate, movable toolbar segment.)

That way, if the user wants to verify the selected Profile, they just look at what's selected in the dropdown list. If they want to select a different profile, they just _open_ the list.

I'd prefer that, myself, over having that already-kind-of-pointless little one-widget dialog window being unnecessarily thrown in front of me all the time. No?

(The Choose Profile dialog could still be retained in the interface, as an alternate "advanced mode" profile selector. It does break out useful info like the selected profile's dimensions, frame-rate, aspect ratio, and field ordering details, if the user wants to examine those in detail.)

N3WWN:
@ferdnyc I have to admit a bit of lack of understanding in how OpenShot works which led to this "feature".  😕  I was unable to determine how to alter the project profile properties without going through this dialog.  I tried many, many potential avenues to do it, but could not quite get it to work properly.  

I figured that someone would see this, know exactly what the code is trying to do, and go, "oh, you have to do it this way".  😁  

If you can point me in the right direction on how to adjust the project profile settings without going through this dialog, I'd love to hear it and I'll update my PR to include the change.

This is a "getting my feet wet" contribution for OpenShot.  My intention was not to be this kludgy... as I mentioned, I figured I'd get called out on it during review and I'm glad you've done that.

ferdnyc:
Gotcha. Well, that _should_ be solvable without opening the dialog, if necessary. The project loader already sets the profile (and causes display updates like setting the window title), it could repaint the video preview too.

But there's a larger issue, in that resizing the preview doesn't appear to work as intended. It shapes the preview to the output aspect ratio, yes, but unfortunately that comes at the expense of previewing what will actually be output.

In OpenShot 2.4.1, if I select the NTSC profile and import 16:9 video, the video will be shown at 16:9 in the preview:

screenshot from 2018-03-26 19-26-24

That _seems_ wrong, but it's arguably **correct** because with the default Best Fit scaling, the video will be output letterboxed. So the resulting video will be exactly as shown, except it will have black bars on the top and bottom.

With these changes, though, the preview is resized to 4:3, which previews the 16:9 video distorted to 4:3 aspect, despite the fact that it will _actually_ be output letterboxed with no distortion:

screenshot from 2018-03-26 19-27-30

By the same token, if I reimport my exported 4:3 video, the default OpenShot preview shows it with bars on all four sides:

screenshot from 2018-03-26 19-26-14

Again, weird but arguably **correct** — or, at least, correct aspect ratio — since the output video will _actually_ be 4:3, with the preview's left- and right-side bars chopped off.

But with these changes, the preview doesn't match the output video here, either:

screenshot from 2018-03-26 19-32-09

I think having the preview window reflect the output aspect precisely would be a good thing, but it doesn't appear that resizing the widget is the way to achieve that unfortunately. The preview renderer appears to expect the default 16:9-aspect preview image, and generates previews already-formatted to fit that size. Resizing the widget only distorts the preview.

jonoomph:
I'm worried about merging this one, due to the forcing a profile dialog on every project. We should probably come up with a better or less obvious way to accomplish this same thing.

ferdnyc:
@jonoomph : You know the preview code better than any of us, can you offer any insight on how intrinsic the locked 16:9 aspect ratio is? In my testing of this patch, unfortunately it seemed like resizing the widget resulted in distorted previews, since the video was still 16:9 regardless of output profile (and then became distorted by the resizing).

AFAICT the preview renderer expects a fixed 16:9 preview viewport, and ignores the target aspect entirely. (It previews 16:9 video full-frame even when it's going to be output at 4:3 with letterboxing, for instance.) So I'm wondering how deeply that fixed preview aspect is baked into the code, and where it'd need to be changed to make non-16:9 previews possible. (With the eventual goal being to conform the preview's aspect to the selected output profile.) Any insights you can offer?

DylanC:
@N3WWN - Closing since this relates to master and not develop. Please re-submit if you plan to get this in. Thanks.

Issue #1327, opened 2018-03-02
Zooming in to within 100 milliseconds


Gunnar-Vestergaard:
**System Details:**
* Operating System / Distro: Windows 7
* OpenShot Version: 2.4.1

**Issue Description and steps to reproduce:**
I need to do this: I have an mp4 video that has an audio track that is far out of sync with its own video track. I need to slide that audio track by about 40 ms (milliseconds) or 80 ms or 120 ms or whatever. But the user interface in OpenShot v2.4.1 won't let me zoom in to more than 1 second.
How can I do this in OpenShot?
Gunnar

gary9872:
Click on your audio clip then >  Go to properties ( left bottom pane) and adjust the position. You should be able to adjust 1/100 of a second.

Cheers,

Gary

peanutbutterandcrackers:
Also, we do have plans to improve the time-line zoom in the future. Please do wait for future versions. :)

DylanC:
@Gunnar-Vestergaard - Already have an enhancement open at #891. Hopefully this can be implemented in the future. 

(Dup of #891)

ferdnyc:
Closing as duplicate of #891.

Issue #1024, opened 2017-11-13
Feature Request: Frame-accurate video


OctagonHex:
**System Details:**
* Operating System / Distro: Win 7 x64
* OpenShot Version: 2.4.1

I would like to be able to work on videos with single-frame accuracy. This is needed to analyse slow-motion videos and technical editing. To support this, I it would be great if OpenShot would add these features:
- Controls to advance or rewind in single-frame increments (i.e. Go +1 Frame, Go -1 Frame)
- If I zoom into the video-timeline, the maximum zoom is 1 second. Please increase this to at least 0.1 second. It would be great, you zoomed in enough then the frames are displayed as individual pictures.
-Display the current position in frames instead of fractional seconds.
- The cut-tool should snap-in on frame beginnings and ends.

DylanC:
Dup of #891

peanutbutterandcrackers:
You can go +1 frame and -1 frame by using the arrowed keys on your keyboard. But yeah, frame-by-frame mode would be super awesome!

Issue #891, opened 2017-09-11
Improve zoom level more


DylanC:
It has been suggested to improve the lowest level of zoom to 0.1s instead of 1s. Something to consider in the future perhaps.

peanutbutterandcrackers:
@DylanC Does this mean frame-by-frame mode for the timeline like KdenLive has?

DylanC:
@peanutbutterandcrackers - I think just to improve the maximum zoom level in general.

peanutbutterandcrackers:
@DylanC I'm so eagerly waiting for frame-by-frame mode. And a new version of OpenShot too. And a tutorial on how to actually use alpha-mask effect in OpenShot... among others.

schorhr:
Hello :-) I love OpenShot, I want to use it for stop-motion animation with students. But zooming into single frames, even see their preview in the time-line, would be a massive improvement.

Thank you <3

mikelacsa:
The same was requested in [#1024](https://github.com/OpenShot/openshot-qt/issues/1024)

ferdnyc:
And in #1327, and probably others; it's kind of a perennial request.

I explain in https://github.com/OpenShot/openshot-qt/pull/1411#issuecomment-376017445 why I don't _personally_ think there's any actual benefit to doing it. Usually the people requesting it don't have any arguments for _why_ — they just "think it would be a good idea". (I'm still willing to be convinced by some argument I haven't considered, but so far nobody's offered one.)

Sometimes the request is coupled, as it was in #1024, with a request that individual frames be displayed in the Timeline. While I agree that sub-second zoom would be necessary _**for that feature**_, that feature is the only real reason I can see for implementing sub-second zoom. And, more importantly, that feature isn't happening.)

muddipper:
Frame-by-frame will be huge improvement because massively increase accuracy to the editing. Is just that simple.
One second 'usually' contain 24, 25, 30, etc depends on the videos. So sliding 1 frame each press instead of 1 second (24 frames) will be amazing.

SuslikV:
@muddipper 
> So sliding 1 frame each press...

_Left Arrow_ and _Right Arrow_ keys on keyboard. Now, you know how to.

[...]

Issue #652, opened 2017-05-17
Ruler isn't fully drawn if timeline length exceeds maximum canvas pixel dimension (32767px)


N3WWN:
**System Details:**
* Operating System / Distro: Fedora 25
* OpenShot Version: OpenShot-v2.3.1-x86_64.AppImage

**Issue Description and steps to reproduce:**
I have a ~35  minute long project and I noticed that the time scale disappears from the top of the timeline under certain conditions:

6sec zoom:  time scale disappears after about 32m45s (1965s = 6 x 327s)
5sec zoom:  time scale disappears after about 27m17s (1637s = 5 x 327s)
4sec zoom:  time scale disappears after about 21m50s (1310s = 4 x 327s)
3sec zoom:  time scale disappears after about 16m22s (982s = 3 x 327s)
2sec zoom:  time scale disappears after about 10m55s (655s = 2 x 327s)
1sec zoom:  time scale disappears after about 5m27s (327s)

It looks like there may be 200 "points" used to display the time scale.  327 seconds would place the "points" from 0 to 65400.  If these points are determined by a 16 bit integer variable, that variable would overflow at 327.675s with a 1s zoom.

I haven't had chance to look at the code, but the solution may be to use a 32 bit integer to store the points.

Let me know if you want any screenshots or additional info.

Thanks!

-Rich

sknaumov:
I saw this problem too on Windows 10 and OpenShow 2.3.1 and 2.3.2.

DylanC:
@N3WWN - Thanks for you feedback. I will take a look at this. Please do provide some screenshots.

N3WWN:
@DylanC 

No problem at all!  I'd love to be able to contribute to the project, but not sure where I could begin.  Do you have a design document that will let me get my head around what the software is doing, all the moving parts (or at least some of them), etc so I can do more than just report issues?

As requested, attached are screenshots showing the issue I'm reporting.

I downloaded a royalty-free video and just placed it on the timeline a bunch of times to get the video length necessary.

1sec zoom: time scale disappears after about 5m27s (327s):
![screen sharing picture september 7 2017 at 09 15 47 edt](https://user-images.githubusercontent.com/19431804/30165200-3d523200-93ad-11e7-9636-6a05687fa9a8.png)

2sec zoom: time scale disappears after about 10m55s (655s = 2 x 327s):
[...]

PR #1411, opened 2018-03-24
Exponential zoom levels (and only 14 of them, instead of 800)

ferdnyc:
This patch reins in the vast, and not _entirely_ beneficial, number of zoom levels in OpenShot, by making the zoom level roughly exponential: each zoom-in halves the previous scale, each zoom-out doubles it — _except_ that whole numbers of seconds, minutes, and hours are used, rather than exponential numbers of seconds each time.

Long and short, there are now 14 zoom levels: 
* 1, 2, 4, 8, 16, 32 seconds; followed by
* 1, 2, 4, 8, 16, 32 minutes; followed by
* 1, 2 hours 
...which is now the maximum zoom. 

(Previous maximum was 800 seconds, so it also allows zooming out _much_ farther than before. Handy if, say, you've tossed a 2-hour-long movie file onto the timeline as a single clip.)

All conversions between zoom-factor and scale (in seconds) are done in the utility functions `zoomToSeconds()` and `secondsToZoom()`, found in new file `src/classes/conversion.py`. This file imports the math library as it has to compute logarithms (base 2) in `secondsToZoom()`.

Honestly, while it _feels_ like the zoom levels "should be" too coarse now, and that going from 800 fine-grained settings to only 14 big jumps would be limiting, in practice it just... isn't, at least in my testing.

There really doesn't seem to be _any_ reason why zooming in at 45 seconds is preferable to either 32 seconds or 60 seconds, in practice. For the most part, you either want to be zoomed in _suuuper_-close, or you want to be able to back out as quickly as possible. Which is exactly what exponential zooming provides.

Oh, yes, I locked the zoom levels in to whole units (minutes, hours) instead of just jumping by exponential seconds continuously, for two reasons:
1. The timescale ruler formats nicely.
    (Tickmarks will be at 00:01:00, 00:02:00, etc., for example, instead of 00:01:04, 00:01:08 like they would be if the scale jumped to 64 seconds.)
2. The scale factors can _potentially_ be expressed as "1 minute", "8 minutes", "1 hour", etc.
    **HOWEVER**, this is not currently done in this patch, so after "32 seconds" it goes to "60 seconds", etc. and the final zoom level (2 hours) will be displayed as "7200 seconds".

### DIFF
--- a/src/classes/conversion.py
+++ b/src/classes/conversion.py
@@ -1,0 +1,49 @@
+"""
+ @file
+ @brief This file deals with value conversions
+ @author Jonathan Thomas <jonathan@openshot.org>
+ @author Frank Dana <ferdnyc AT gmail com>
+
+ @section LICENSE
+
+ Copyright (c) 2008-2018 OpenShot Studios, LLC
+ (http://www.openshotstudios.com). This file is part of
+ OpenShot Video Editor (http://www.openshot.org), an open-source project
+ dedicated to delivering high quality video editing and animation solutions
+ to the world.
+
+ OpenShot Video Editor is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ OpenShot Video Editor is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with OpenShot Library.  If not, see <http://www.gnu.org/licenses/>.
+ """
+
+import math
+
+def zoomToSeconds(zoomValue):
+    """ Convert zoom factor (slider position) into scale-seconds """
+
+    if (zoomValue <= 5):
+        # Under 1 minute, convert to seconds (exponential)
+        return 2**zoomValue
+    elif (zoomValue <= 11):
+        # convert to N minutes (exponential), in seconds
+        return 60 * (2**(zoomValue-6))
+    else:
+        # Over 1 hour, convert to N hours (exponential), in seconds
+        return 3600 * (2**(zoomValue-12))
+
+
+def secondsToZoom(scaleValue):
+    """ Convert a number of seconds to a timeline zoom factor """
+
+    return int(math.log(scaleValue,2))
+
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -48,6 +48,7 @@ from classes.timeline import TimelineSync
 from classes.query import File, Clip, Transition, Marker, Track
 from classes.metrics import *
 from classes.version import *
+from classes.conversion import zoomToSeconds, secondsToZoom
 from images import openshot_rc
 from windows.views.files_treeview import FilesTreeView
 from windows.views.files_listview import FilesListView
@@ -1963,17 +1964,19 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         self.timelineToolbar.addSeparator()
 
         # Get project's initial zoom value
-        initial_scale = get_app().project.get(["scale"]) or 20
+        initial_scale = get_app().project.get(["scale"]) or 16
+        # Round non-exponential scale down to next lowest power of 2
+        initial_zoom = secondsToZoom(initial_scale)
 
         # Setup Zoom slider
         self.sliderZoom = QSlider(Qt.Horizontal, self)
         self.sliderZoom.setPageStep(2)
-        self.sliderZoom.setRange(1, 800)
-        self.sliderZoom.setValue(initial_scale)
+        self.sliderZoom.setRange(0, 15)
+        self.sliderZoom.setValue(initial_zoom)
         self.sliderZoom.setInvertedControls(True)
         self.sliderZoom.resize(100, 16)
 
-        self.zoomScaleLabel = QLabel(_("{} seconds").format(self.sliderZoom.value()))
+        self.zoomScaleLabel = QLabel( _("{} seconds").format(zoomToSeconds(self.sliderZoom.value())) )
 
         # add zoom widgets
         self.timelineToolbar.addAction(self.actionTimelineZoomIn)
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -1971,7 +1971,7 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         # Setup Zoom slider
         self.sliderZoom = QSlider(Qt.Horizontal, self)
         self.sliderZoom.setPageStep(2)
-        self.sliderZoom.setRange(0, 15)
+        self.sliderZoom.setRange(0, 13)
         self.sliderZoom.setValue(initial_zoom)
         self.sliderZoom.setInvertedControls(True)
         self.sliderZoom.resize(100, 16)
--- a/src/windows/views/timeline_webview.py
+++ b/src/windows/views/timeline_webview.py
@@ -47,6 +47,7 @@ from classes.logger import log
 from classes.query import File, Clip, Transition, Track
 from classes.waveform import get_audio_data
 from classes.thumbnail import GenerateThumbnail
+from classes.conversion import zoomToSeconds, secondsToZoom
 
 try:
     import json
@@ -193,8 +194,8 @@ class TimelineWebView(QWebView, updates.UpdateInterface):
         # Reset the scale when loading new JSON
         if action.type == "load":
             # Set the scale again (to project setting)
-            initial_scale = get_app().project.get(["scale"]) or 20
-            get_app().window.sliderZoom.setValue(initial_scale)
+            initial_scale = get_app().project.get(["scale"]) or 16
+            get_app().window.sliderZoom.setValue(secondsToZoom(initial_scale))
 
     # Javascript callable function to update the project data when a clip changes
     @pyqtSlot(str)
@@ -2442,8 +2443,11 @@ class TimelineWebView(QWebView, updates.UpdateInterface):
     def update_zoom(self, newValue):
         _ = get_app()._tr
 
+        # Convert slider value (passed in) to a scale (in seconds)
+        newScale = zoomToSeconds(newValue)
+
         # Set zoom label
-        self.window.zoomScaleLabel.setText(_("{} seconds").format(newValue))
+        self.window.zoomScaleLabel.setText(_("{} seconds").format(newScale))
 
         # Determine X coordinate of cursor (to center zoom on)
         cursor_y = self.mapFromGlobal(self.cursor().pos()).y()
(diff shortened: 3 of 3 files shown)


ferdnyc:
I'm not sure why buildkite is failing this. If it's because of that merge commit I did to catch up my master branch, I can fix that and re-submit a cleaner pull request. If it's because two commits touch the same line in the same file (I modified the slider range in 728aff7 and then again in be48611), then that's just... really lame. :laughing:

peanutbutterandcrackers:
It seems that using `git remote add upstream <upstream url>`
and running `git fetch upstream`
and `git merge upstream/master`
is a lot better way to sync your fork with the upstream.

But you might already know this....

ferdnyc:
Yeah, I _should_ know it, but I was in the GUI, and had a different branch checked out in the terminal, and blah blah blah I got lazy and tried to avoid doing the work the right way, which ended up creating _more_ work for myself.

Aaaaanyway, I reset my master to the origin/master HEAD, then rebased this branch to it. buildkite seems happier now. Once again I'm doing things to appease software. There's gotta be some sort of lesson or cautionary tale or danger sign there.

ferdnyc:
No reaction/input, positive **or** negative, regarding the actual _change_, though? From anyone? Hrm.

DylanC:
@ferdnyc - I'm definitely in favour of improving the zoom and so are many other users according to the issues reported. Not saying all info in the tickets need to be taken on board but maybe check if you change does/does not satisfy certain issues here:

#652, #1024, #1327

Note: My ticket and some others want 100ms as the minimum. #891 (and #1327) Maybe consider 100ms as the minimum zoom level? What do you think?

ferdnyc:
The `scale` imposed by the zoom slider on the timeline is currently expressed in whole numbers of seconds (1 being the minimum). It _appears_ to be treated as a float, in the code, so in theory setting `scale` to some fraction of a second might be possible. I don't know how it would work in practice.

I also, well... I'm still questioning whether closer zooming actually _**does**_ anything, from a practical/functional standpoint. I completely understand the _perception_ of being able to zoom in closer, but at "1 second" zoom frames (your minimum unit of measure in OpenShot) are already some number of pixels wide. (How many will depend on the frame rate. With OpenShot fullscreened in "Advanced" layout on my 1920×1200 display, that number appears to be approx. 4px for 25fps, only 1px or 2px at 60fps.)

Regardless, because the smallest unit of measure on the timeline is 1 frame, zooming in _closer_ than one frame would have no effect except to widen each of those frames on the timeline. You couldn't be any more _precise_, because you're still limited to 1-frame resolution for all operations. You just have to travel more pixels in distance between each one.

It's kind of like... imagine you have a 640x480 image that you're displaying full-screen on a standard HD monitor. **Each** individual pixel in that image is already being _blown up_ to roughly 2.25 screen-pixels.

You _**COULD**_ zoom in farther on that image, if you really wanted to. But zooming in would make _absolutely_ NO additional information visible. You'd be doing nothing except making the already-blocky pixels larger.

Though it doesn't _feel_ like the same thing (to me, either!), zooming in closer than 1s on OpenShot would pretty much be _exaaaactly_ the same thing: You'd merely be blowing up the information that's already there, and which can't be represented with any more _detail_, it can only be made larger.

ferdnyc:
> Honestly, for a moment, I thought "Whoa! ferdnyc has implemented frame-by-frame view on the timeline!?". Would you like to add that too, perhaps, please? Then I'll be sold, for sure.

Yeah, that won't be happening. But if it helps any, you can consider this PR a _prerequisite_ for that feature. A necessary first step would be uncoupling the zoom levels from the interface scaling factor, which this PR does. So, hey, first step!

(I fear we may have given the wrong impression about how much would be involved in building something like that, though. Just because OpenShot already has a thumbnailing and caching engine doesn't make that feature any less of a massive undertaking.)

Especially when you start to consider the performance impact, not of generating all of those thumbnails, but of **displaying** them in the interface. If someone were to build that mode and have it just blindly shove all of the clip's exploded frames onto the timeline, so that even just a single 10-second clip means you're rendering and interacting with a widget that contains over 200 individual images... well, just the act of zooming to that level / activating that mode would probably take several seconds to complete. Each drag or scroll of the timeline and/or the clip(s) on it, several more.

...It'd be a huge undertaking, to make it actually _work_ usably. (One I'm not currently planning to attempt.)

ferdnyc:
> For me, only having 1 or 2 pixels per frame makes it difficult to sync multiple videos because it's sometimes difficult to click and drag by one pixel. Zooming in further would let the drag be easier because you'd be dragging, say, 5 pixels per frame.

\*nod\* That makes sense. Though I wonder if that could also be (and perhaps would better be) achieved by making the ruler elements wider, rather than going to subsecond zoom? (In effect, redefining every zoom level to be "closer" than it currently is.) The `scale` factor on the timeline (the variable that the zoom setting controls) defines its relationship to the ruler, effectively how many seconds of timeline are compacted into each major ruler division. But the size of those ruler divisions is fixed. If they were wider, there would be more pixels between each second.

(My main worry about sub-second zoom is that the ruler isn't meant to do that. `scale` may be a float, but the major ruler divisions are labeled with "HH:MM:SS" time, so if you zoom closer than 1 second I have no idea what those divisions would even show. Probably "00:00:00 00:00:01 00:00:01 00:00:02 00:00:02" at "500ms zoom". Not an unsolvable problem, but a confusing one if not solved. When simply widening the ruler (altering the `pixelsPerSecond` value that the scaling math works off of) _should_ have the same effect, and allow the "HH:MM:SS" labeling, and the concept of "1 second" as the closest zoom level, to be retained.)

ferdnyc:
> It seems that using git remote add upstream <upstream url>
> and running git fetch upstream
> and git merge upstream/master
> is a lot better way to sync your fork with the upstream.

And actually, now that I thought about that more, I realized that I don't actually ever _use_ the master branch of my fork for anything. So why am I tracking it at all?

```
% git checkout master
% git branch -u upstream/master
% git pull
```
And now my local `master` is tracking `openshot-qt:master` automatically. I have no access to `ferdnyc:master`, which is fine because I should never be using it. If I create a new branch from `master` locally, it's automatically forked from `openshot-qt:master` which is probably what I want anyway.

(I can still `git push -u origin <branchname>` to configure that branch as a tracking branch that pushes into my fork. GitHub has no issues with me pushing changes into `ferdnyc:<branchname>` that aren't in `ferdnyc:master`, because it knows it was forked from `openshot-qt:master` and git doesn't actually _care_ which remote a particular branch interacts with, it's all one giant distributed database as far as it's concerned.)

I'm sure somewhere down the line I'll get screwed over by something that demonstrates why this was all a very bad idea, once I've gotten far enough down the path for whatever "gotchas" leap out to cause maximum frustration with the fewest options for easily cleaning up my own messes. Git's great that way.

jonoomph:
I really like this PR, thanks so much!!! Merged

jonoomph:
After playing with this PR a bit, I kinda wish we had twice as many zoom levels, but I love the exponential nature of them.. just feels like sometimes it jumps too big (too small, and then too large), and needs more gradual changes. So, maybe instead of 14 of 'em, we need more like 28?  Just a thought.

ferdnyc:
> After playing with this PR a bit, I kinda wish we had twice as many zoom levels, but I love the exponential nature of them.. just feels like sometimes it jumps too big (too small, and then too large), and needs more gradual changes. So, maybe instead of 14 of 'em, we need more like 28? Just a thought.

I'll buy that... is it towards the far end of the scale, when zoomed way out, that it starts to feel like it's jumping too far? We could certainly have the exponentiation drop off at a certain point. Or at two points, since the math shifts at both the minute and hour boundaries so they could each shift to a different granularity than the sub-minute zooming.

ferdnyc:
Technically, the defined zoom levels _could_ be a completely arbitrary list, and just impose whatever fixed values we want. That might even be a lot simpler. My admittedly somewhat convoluted implementation avoids taking that step, and ends up performing a bunch of math in the service of two hidden goals: 
1. Continue to internally represent the current zoom in raw seconds, rather than as an indexed "zoom level". This both minimizes disruption of the existing code that expects those units, and allows the available zoom levels to be changed without affecting how that value is interpreted.
1. Remain compatible with any/all zoom parameters stored in existing project files, even though many will be set to some arbitrary number of seconds which doesn't correspond to any of the newly-defined zoom levels.

jonoomph:
Yeah, maybe we just need a list which generally follows a curve, slower in the beginning (when zoomed in), and faster and faster the more you zoom out. For example: 

[
  1,
  3,
  22,
  48,
  72,
  61,
  109,
  130,
  149,
  167,
  184,
  198,
  212,
  224,
  235,
  244,
  253,
  259,
  354,
  269,
  362,
  272,
  361,
  355,
  347,
  417,
  391,
  360,
  439,
  464,
  239
]

jonoomph:
@ferdnyc Also, I would love for you to send me an email: jonathan@openshot.org, since you do so many great PR's. I would like to learn more about you. =)

jonoomph:
Sorry, wrong curve data... updated my previous comment

ferdnyc:
@jonoomph :
> Yeah, maybe we just need a list which generally follows a curve, slower in the beginning (when zoomed in), and faster and faster the more you zoom out.

Well, that's basically what it does now — exponentiation produces a *very* steep curve, probably _too_ steep once you get up into the higher values. Doubling the _y_-value for every _x_ means that you start out at a 45° angle and get precipitously steeper from there. (Here's Wolfram Alpha plotting `2^x`  ...)
screenshot from 2018-04-19 23-00-09

As far as the actual values, in testing I discovered that there are good reasons to snap the zoom levels (whatever they may be) to not just whole seconds, but whole minutes and hours as well. That's why the list isn't just straight exponentiation of the seconds value. "Weird" zooms lead to a correspondingly "weird" timeline ruler.

Since the zoom factor controls the ruler scale, placing the eighth zoom level at 256 seconds (`2^8`) would cause ruler divisions at, and labeled, `00:04:16` `00:08:32` `00:12:48` etc... Which just felt wrong and confusing, when I tried it. So the code special-cases the scale where it crosses both the minute and hour boundaries, which is why the zoom levels are currently computed as:
* 1, 2, 4, 8, 16, 32 seconds
* 1, 2, 4, 8, 16, 32 minutes
* 1, 2 hours 
...but always expressed in units of seconds.

But I'll whip up some sort of adjustment to ease that off, maybe something along the lines of:
* 1, 2, 4, 8, 15, 30, 45 seconds
* 1, 2, 4, 5, 10, 20, 30, 45 minutes
* And then maybe 1 hour, 1h15m, 1h30m, 1h45m, and finally 2 hours. 

That'd be... 20 different levels, on a gentler curve, and they'd mostly produce fairly clean ruler divisions.

(My suspicion is that achieving a comfortable zoom function isn't really going to be about _how many_ levels there are, precisely. The issue with the current set is how distant they become from each other at the far end of the range. I didn't really ever choose any particular _number_ of different levels at all, that was dictated by the set of values produced by my (adjusted) exponentiation function within the range of 1 to 7200 seconds. I discovered that the limiting factor should be 2 hours purely through experimentation: when I took the next step to 14400 seconds (4 hours), the division labels on the far end of the ruler got screwy because their time indexes exceeded 1 day.)

ferdnyc:
I just have to note that, contrary to my worries about it backfiring on me, this:
```console
% git checkout master
% git remote add upstream $CloneURLforParentRepo
% git fetch upstream
% git branch -u upstream/master
% git pull
```
Has actually become my standard setup for **every** new repo I fork. Any new topic branch I create gets forked from the upstream, after which I configure it to push to my fork:
```console
% git checkout master
% git checkout -b topicbranch
% git push -u origin topicbranch
```
And voila. No need to worry about keeping my fork's master branch in sync with the upstream master at all, because it's never used. This is the way Github should work _automatically_. (I may even see about setting up some scripts to do at least the topic-branch linking to origin automatically on every new branch creation.)

PR #1439, opened 2018-03-29
CONTRIBUTING.md: Be more explicit about logfile paths

ferdnyc:
Present and explain full path to necessary logfiles under both Windows and UNIX-like systems.


DylanC:
@ferdnyc - Excellent! This is badly needed.

Issue #1443, opened 2018-03-29
Crashing OVS after playing a moved video fragment


theelectricsoul:
**System Details:**
* Operating System / Distro: Win Server 2106 
* OpenShot Version: 2.4.1 dev1

https://streamable.com/5zwfm

Sry... the keyword "crash" did not return results in help. No idea where the crash log is.

peanutbutterandcrackers:
Here is how you attach log files: [contributing guilelines](https://github.com/OpenShot/openshot-qt/blob/master/.github/CONTRIBUTING.md).

And I suggest you try out the latest daily build. Just head over to the [download](https://www.openshot.org/download/) page and click on the 'daily builds' and grab the latest one and see if that works.

Thanks!

theelectricsoul:
tested 2.4.1-53 as seen in the video of the first post
[libopenshot.log](https://github.com/OpenShot/openshot-qt/files/1863717/libopenshot.log)
[openshot-qt.log](https://github.com/OpenShot/openshot-qt/files/1863718/openshot-qt.log)

peanutbutterandcrackers:
I wonder if it is happening because you are overlapping the left clip over the right, while the normal way to do it is to overlap the right clip over the left? Just a thought. I tried doing the same kind of overlapping here on my machine, and OpenShot did crash after a while....

@DylanC @ferdnyc @N3WWN - I think we've stumbled upon an interesting problem here, gentlemen.

theelectricsoul:
I pinpointed the bug.

https://streamable.com/awnb5

peanutbutterandcrackers:
@theelectricsoul - I see that you put the video on the left over the right. Which is not what us normal users do, normally. Could you try to see if you tried to do the overlapping normally ---- i.e. by putting the video clip on the right side *over* the video on the left one --- and see if the bug can be reproduced then?

Sorry for the late reply. I've not been able to come around to github lately.

If you can replicate the bug even then, I will let the devs know about this right away.

Sorry for the inconvenience.

theelectricsoul:
@peanutbutterandcrackers 

As mentioned in the [last](https://github.com/OpenShot/openshot-qt/issues/1443#event-1550659058) video, it doesn't matter if you overlap from left to right, or right to left. 

--------------
[...]

PR #1491, opened 2018-04-10
Docs: Add Contributing doc

ferdnyc:
Converted from `.github/CONTRIBUTING.md` with pandoc, then cleaned up by hand.

It's a _little_ redundant with the new Developer doc page, but I think it's still useful to have the same contents as `CONTRIBUTING.md` in there directly (and we should endeavor to keep them in sync).

The main thrust for this was @theelectricsoul pointing out in #1443 that the word "crash" doesn't actually appear anywhere in the documentation, when you search for it. Now it does, and the docs provide information on collecting crash logs and submitting them.

### DIFF
--- a/doc/contributing.rst
+++ b/doc/contributing.rst
@@ -1,0 +1,60 @@
+How to contribute to OpenShot Video Editor
+------------------------------------------
+
+**Did you find a bug?**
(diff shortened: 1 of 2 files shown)


jonoomph:
Looks great! Merged

PR #1492, opened 2018-04-10
CONTRIBUTING.md: Where to find daily installers

ferdnyc:
There isn't actually an "Other Downloads" link on the referenced page anymore, so the previous description was confusing.

(The presence of the "Daily Installers" button on that download page is also confusing, as it does nothing and should probably go away. But that's an issue for the website, not the code.) 
**ETA:** Crap, that's not true. You _do_ have to hit the "Daily Installers" button **once** to get the list shown. After that, you can switch operating systems at will, and the button no longer does anything. I'll have to update this slightly and push in another commit.
**Update 2:** Done.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -10,7 +10,7 @@
 
   _Please download our latest daily installer:_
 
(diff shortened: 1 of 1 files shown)


jonoomph:
Thanks! Merged

PR #1525, opened 2018-04-21
Profile cleanup

ferdnyc:
I went down a bit of a rabbit hole tonight, wondering why OpenShot's profile list contained what _appeared_ to be completely identical entries for some profiles.

What I discovered is that four of the profile files _were_ exact duplicates of each other. Since there's no advantage to that whatsoever, I deleted them. I also renamed the file `dv_pal_wide_animorphic` to correct the spelling (it's now `dv_pal_wide_anamorphic`).

1. **How will this affect users loading existing project files?**
    It won't. OpenShot only uses the profile files to populate the Choose Profile dialog list; when a profile is selected that profile's data is copied into the project file. That data will still be there.

1. **How will this affect users editing existing or new projects?**
    They'll see a slightly shorter list of profiles, with fewer duplicates, but will be otherwise unaffected. The file rename will have no effect whatsoever, because the spelling was already correct in the file's `description=` field. That profile will be displayed exactly the same in the Choose Profile dialog.

I still appear to be seeing duplicate profiles at times, independent of this cleanup. (Which means I sometimes saw _three_ of some profiles, when there were file duplicates.) But this should hopefully at least make that easier to chase down, now that the on-disk profile set is unique.

I'll follow up with a comment detailing some further profile-related observations / unresolved questions.


ferdnyc:
So, following this change, the only _fully_ duplicated profile is "VCD PAL" / "CIF PAL", which happen to have exactly the same format. But, since the corresponding "VCD NTSC" and "CIF NTSC" are _not_ the same, we can't change them to "VCD/CIF PAL" and "VCD/CIF NTSC", so I think it makes sense to retain both PAL profiles despite the redundancy.

ferdnyc:
Now we come to the HDV profiles. I removed four of them because they were duplicates of ATSC profiles, _including_ the description field.

Some of the rest are _clearly_ distinct from any other format, as they use video formats that are frankly kind of bizarre. (I believe they're meant to match the recording format from certain HD cameras?) As an example, `hdv_1080_25p` is:
```
width=1440
height=1080
progressive=1
sample_aspect_num=4
sample_aspect_den=3
display_aspect_num=16
display_aspect_den=9
```
Which is nothing at all like `atsc_1080p_25`.

But some of the _other_ `hdv_*` files only differ from the `atsc` counterparts in ways that may or may not even be intentional, and I have no way to really be sure.

For example, the **ONLY** difference between `atsc_1080p_25` and `hdv_1080_25p_1920x1080`, other than the description text, is that `hdv_1080_25p_1920x1080` doesn't contain a `colorspace` parameter. (At all.) The same is true of `atsc_720p_24` and `hdv_720_24p`. 

If those two HDV profiles **did** contain a `colorspace` parameter, they'd be identical to the ATSC profiles. But I don't know what it means when a profie has an _absent_ `colorspace` parameter. Most of the _other_ `hdv_*` files **DO** contain `colorspace=709`, though, same as all of the ATSC profiles. Which makes me think that the HDV files all should have the same `colorspace` definition, which would make `hdv_1080_25p_1920x1080` and `hdv_720_24p` redundant.

DylanC:
@ferdnyc - I'd love to get more feedback from others but it does look good to me at least.

N3WWN:
What happens when existing projects which use the removed profiles are opened?

jonoomph:
LG! Thanks!

ferdnyc:
@N3WWN See PR text

N3WWN:
> @N3WWN See PR text

@ferdnyc I apologize... I see that you did cover that.

Issue #1526, opened 2018-04-21
Drag and drop doesn't work for importing file with Greek characters


Daniel-Mantzakos:
**System Details:**
* Operating System / Distro: Windows 10 Pro 64 bit Version 1709
* OpenShot Version: 2.4.1

**Issue Description and steps to reproduce:**
As the title says, when you drag and drop a file containing Greek characters in its title, it doesn't get imported. There is no crash during the process. The import from the menu though works as expected.

Daniel-Mantzakos:
@ferdnyc I am talking about filename.

This issue happens for any type of file (audio, video and images).

In order to reproduce the issue, you can rename a song or picture or video you already have with the first letter of the Greek alphabet, for example, which is α. You can then drag and drop it into OpenShot and verify that it's not working.

ferdnyc:
Gotcha, thanks.

Hmmm. I renamed a video file to `α.mp4` and an audio file to `ααα.mp3`, and had no problems importing both by drag-and-drop from my Desktop to the OpenShot window.

![screenshot from 2018-04-21 12-48-55](https://user-images.githubusercontent.com/538020/39086632-7faad546-4562-11e8-9148-ad4e95681519.png)

...I'm guessing the reason it works for me is because I'm on Linux. So this looks like it may be a Windows-only issue. I'll try to find some time later on to spin up my Win7 VM, so I can test it there.

Daniel-Mantzakos:
Ok, then it should be a Windows issue. I also found a more general problem. Even if the filename does not include Greek characters, the file won't be imported by drag and drop if the location path of the file contains Greek characters. I did a test by creating a Greek named folder on my desktop and pasting inside it an audio file with only English characters in its filename. The file still wouldn't be imported by drag and drop, because its parent folder was named in Greek.

ferdnyc:
OK, good information, thanks. I'll be sure to test that, too.

DylanC:
@Daniel-Mantzakos - Can you try this in a daily build? @ferdnyc put in a fix for this issue recently.

misterhatoola:
This issue occurred with Hebrew characters as well. Just tried it with the latest daily build, and it works fine - files with Hebrew characters in their filenames, or in the folder name, drag just fine into OpenShot's window.

DylanC:
@misterhatoola - Excellent. Fantastic news to hear. :)

Issue #1288, opened 2018-02-19
Issue : Can not add pictures files on project if picture's title is a number ?


Jehan67:
**System Details:**
* Operating System / Distro: Windows 10 ; version: 1709 ; OS_version : 16.299.248
* OpenShot Version: OpenShot-v2.4.1-14-ge4a5aa1-57-672-x86_64.exe
No logfile

**Issue Description and steps to reproduce:**
Hi there,

I think i've found a bug here. When I start a new project, I add the MOV part and then a lots of pictures. All thoses pictures had a title like 1.jpg 2.jpg 3.jpg, etc...
Each time I tried to load these on the project, there was a runtime error as soon as I accepted to added them.
I tried to load others pictures with letters, it worked. 

**Quickfix** 
My temporary fix is to rename the files you want to add. Worked for me.

peanutbutterandcrackers:
@Jehan67 - Excuse me but are you THE JEHAN??? The one who is working on Zemarmot and fixes bugs in GIMP!??

peanutbutterandcrackers:
P.S: I will test this on my machines and get back to this soon.

Jehan67:
Hi Peanut,

I'm am a Jehan, but only an homonym of the one you think I am :)

Jehan67:
Hi Peanut,

I'm am a Jehan, but only an homonym of the one you think I am :)

peanutbutterandcrackers:
@Jehan67 - You can reopen this issue. I'm pretty sure. I think there's a "comment and reopen" or sth there...

peanutbutterandcrackers:
@Jehan67 - And oh... I see. lol It says you use windows. That shoukd have given it away. The Jehan probably never use windows.... but nice to meet you, anyways. :)

ferdnyc:
Interesting. Here's what happens when I try that in 2.4.0 (which is the version rpmfusion is still distributing for Fedora Linux):

If I import png files `1.png` and `2.png` using the File→Import Files... menu (or Ctrl+F), they're both imported without incident.

If I try to drag-and-drop those two files into the OpenShot window, This dialog pops up:

![screenshot from 2018-02-20 08-15-25](https://user-images.githubusercontent.com/538020/36426048-e87403c0-1616-11e8-82dc-af7cb44a818e.png)

[...]

Issue #1084, opened 2017-11-28
drag and drop not working


primer97:
**System Details:**
* Operating System / Distro: w10
* OpenShot Version: 2.4.1

**Issue Description and steps to reproduce:**
Please allow user to import a video file by drag and drop (on track or on file list)
When trying to do that, nothings happens.
Thank you.

DylanC:
@primer97 - This is already a feature. It should be working.

primer97:
I can drag file from the video list (from inside openshot) 
but when I try to drag from windows explorer it does nothing.

If it should be wrking, it could be a local problem maybe , I don't know. But drag and drop to other solftware is working...
Thank you.

DylanC:
@primer97 - I will test this for you but I suspect it will be working.

misscbeezy:
For me on Windows 7 this is the same for me ONLY when I use the right click "run as administrator" to start 
OpenShot  Launch icon from desktop.  But if I just double click on OpenShot  Launch icon (w/out  right click "run as administrator" ) to start program then I am able to drag and drop with no problem.

DylanC:
@primer97 - Try the solution posted by @misscbeezy. Should work if you start without admin mode.

Issue #923, opened 2017-09-21
drag and drop fail


paco2l:
**System Details:**
* Operating System / Distro: W10?
* OpenShot Version: 2.4.0?

**Issue Description and steps to reproduce:**
if the video clip name contains the spanish's letter "ñ" drag and drop doesn't work

DylanC:
Dup of #677.

DylanC:
@paco2l - This has been resolved in the daily builds now.

PR #1539, opened 2018-04-26
Use QUrl.toLocalFile() to convert dropped URIs

ferdnyc:
This is tested on Linux, still does the right thing. I can't currently test it on Windows (which it targets) because I'm still working on getting an OpenShot build environment working there, but I'm hopeful that it will fix the filename-handing issues that have been plaguing drag-and-drop imports on that platform. The previous method of forcing UTF-8 was clearly causing problems with paths containing non-ASCII characters (on Windows).

There are also some `log.info()` calls that should probably be commented out once this has been tested on Windows and Mac.

Addresses #1526, #880, #677; possibly #1084, #923, #888 (not enough detail in reports); maybe #1288 (but probably not?)

### DIFF
--- a/src/windows/views/files_listview.py
+++ b/src/windows/views/files_listview.py
@@ -273,17 +273,13 @@ class FilesListView(QListView):
         # Reset list of ignored image sequences paths
         self.ignore_image_sequence_paths = []
 
-        # log.info('Dropping file(s) on files tree.')
         for uri in event.mimeData().urls():
-            file_url = urlparse(uri.toString())
-            if file_url.scheme == "file":
-                filepath = file_url.path
-                if sys.platform == "win32":
-                    filepath = filepath[1:] # Remove / at beginning of path (just for Windows)
-                if os.path.exists(filepath.encode('UTF-8')) and os.path.isfile(filepath.encode('UTF-8')):
-                    log.info('Adding file: {}'.format(filepath))
-                    if self.add_file(filepath):
-                        event.accept()
+            log.info('Processing drop event for {}'.format(uri))
+            filepath = uri.toLocalFile()
+            if os.path.exists(filepath) and os.path.isfile(filepath):
+                log.info('Adding file: {}'.format(filepath))
+                if self.add_file(filepath):
+                    event.accept()
 
     def clear_filter(self):
         if self:
--- a/src/windows/views/files_treeview.py
+++ b/src/windows/views/files_treeview.py
@@ -272,17 +272,13 @@ class FilesTreeView(QTreeView):
         # Reset list of ignored image sequences paths
         self.ignore_image_sequence_paths = []
 
-        # log.info('Dropping file(s) on files tree.')
         for uri in event.mimeData().urls():
-            file_url = urlparse(uri.toString())
-            if file_url.scheme == "file":
-                filepath = file_url.path
-                if sys.platform == "win32":
-                    filepath = filepath[1:] # Remove / at beginning of path (just for Windows)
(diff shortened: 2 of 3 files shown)


peanutbutterandcrackers:
@ferdnyc - The hero we need. :)

DylanC:
@ferdnyc - I can test this on Windows and Mac if you like.

ferdnyc:
@DylanC : _Please!_, absolutely. [My testing comment](https://github.com/OpenShot/openshot-qt/issues/1526#issuecomment-383347059) in #1526 has the steps I used to _confirm_ the problem on Windows. So, the ideal here would be that after this patch, those same steps no longer fail to do anything.

N3WWN:
I can test on Mac OS X, too, if it would be helpful when the patch is merged and a daily build is available.

DylanC:
@ferdnyc - This is working fine on my old Macbook (2009 model) with macOS 10.11. Will need to test Windows 10 soon. If all is well I would like to get this in before v2.4.2. However, I will check with the big boss about merging first.

DylanC:
@ferdnyc - Also seems to be working well on Windows 10. I'll see can we get this in.

ferdnyc:
> Also seems to be working well on Windows 10.

Sweet! That's great to hear, since Windows was both the only system where there were any problems with the old code, and my main point of uncertainty as to whether this was the right way to successfully fix them.

Just to belabor the point (in the interest of making sure we get a working fix into the code), did you test using my steps from #1526? If you were able to drag-and-drop import a video file named `C:\Users\<whatever>\Desktop\α.mp4`, where that same file would silently fail to import in 2.4.1 (something that should probably be tested first, on the same system, to ensure that there's an extant problem to be addressed), then I'm prepared to get my hopes up that this actually fixes the issue.

DylanC:
@ferdnyc - Yes, I followed the test in that ticket. Seemed to work on my system. I've asked big boss to merge this but heard nothing back yet. I'm skeptical about hearing back soon, so perhaps I better merge this in so we can get feedback from other testers. (over the weekend) 

As far as I can see this doesn't change anything to do with translation strings or add new ones.

DylanC:
@ferdnyc @peanutbutterandcrackers @N3WWN - Can we all fully test this once it gets into the next build?

N3WWN:
@DylanC Sure thing!  Just let me know which build and when it's ready for testing!

peanutbutterandcrackers:
@DylanC - I was able to drag-and-drop files with Devanagari characters on Linux Mint 18.3. [I hadn't really noticed that there was such a problem before, though, BTW] Also had success with it in Elementary OS 0.4.1 'Loki'. :+1:

DylanC:
@peanutbutterandcrackers - I think Linux was always fine. It was more of an issue on Windows. :)

ferdnyc:
Correct. My only concern with Linux was that this change didn't _break_ anything, and I was already reasonably sure of that. (Same concern for MacOS — which was also fine with the old code — though I was less-certain that it would have no negative repercussions.)

The problem with the old code was that it was manually forcing the paths for dropped files to UTF-8 encoding. That's great for MacOS and Linux, which **use** UTF-8 encoding for their filesystems. (On Linux it's _guaranteed_ that all file paths will be UTF-8, by design and by fiat. On MacOS I'm not sure if that's guaranteed or simply true in practice, but regardless it's proven non-problematic to assume UTF-8 there.

Windows, because it has to be difficult, wants nothing to do with UTF-8 paths. It either uses "plaintext" (probably some default 8-bit encoding, like ISO-8859-1 or CP-1252), as a fallback for non-Unicode-aware code, or else path strings are all encoded as double-wide UTF-**16** (aka UCS-2). Which is why the previous code mangled all filenames containing non-ASCII characters.

PR #1544, opened 2018-04-28
Move CONTRIBUTING.md to project root

ferdnyc:
That way we can reference it from README.md, and include it in source tarballs. Github will [automatically pick it up](https://help.github.com/articles/setting-guidelines-for-repository-contributors/) from this location, same as if it was in the `.github/` directory.


DylanC:
@ferdnyc - I would like to refer to this in my own PR, so it would help if it was in the repo first. Its pretty harmless to merge this imho.

ferdnyc:
> @ferdnyc - I would like to refer to this in my own PR, so it would help if it was in the repo first. Its pretty harmless to merge this imho.

*nod* It should have zero impact, since Github will treat it exactly the same in either location. I think it'd be good to have the file incorporated into the source tarballs and etc., as well, but I'm not sure where those scripts live or if they're even part of the repo. (An update will probably be necessary to replace `README` with `README.md`, as well, so that can get tackled all at once the files themselves are sorted out.)

Issue #1531, opened 2018-04-23
other track names are not preserved by delete track


jpritikin:
**System Details:**
* Operating System / Distro: ? Debian stable/testing on Intel 64
* OpenShot Version: 2.4.1

**Issue Description and steps to reproduce:**

Name all the tracks. Delete a track in the middle. Some of the track names get reset back to generic names "Track 3" or "Track 4"

ferdnyc:
Hmm, this _should_ be an easy one to fix...
(I forgot it was even possible to name tracks, I don't think I've ever tried it!)

ferdnyc:
It was indeed, though not as easy as I'd expected once I got into what was going on. (It wasn't just track names, those were _intentionally_ being cleared. But the track lock status would also get screwed up. It'd actually be "left behind" on the **wrong** tracks, whenever you inserted a new one.

Anyway, I believe it's all fixed now. Or will be as of #1546 being merged.

DylanC:
@jpritikin - Hi there, can you try a daily build? This fix went in and should be available to you.

Issue #454, opened 2016-12-30
Can't remove Track 0


P-Woods:
This is not a bug, merely an issue.

OpenShot-v2.2.0-x86_64.AppImage

When a NEW PROJECT is created, 4 tracks appear by default. 
This is inelegant and irritating but not a bug.

When the surplus tracks are removed, they are renamed from track 0 upward.
This could be confusing.

When Track 0 is removed, Track 1 is renamed Track 0.
When only 1 track remains is is named Track 0.
This is not a BUG, it is the DESIGN!  
This is the sort of design which BUGS users.

Please reconsider this part of the design (which probably "just happened" 
and was never actually designed I hope.)

peanutbutterandcrackers:
@P-Woods - whoa! This one reminds me of Steve Jobs movie portrayals. If you can submit a patch, would you be kind enough to do it, sir? Or, perhaps enable renaming of a track?

@DylanC - what do you think about this, el capitano?

DylanC:
@peanutbutterandcrackers - I think that when you create a new project it should start with one track callled "new track". Then the user should be able to set a track name and then later add more tracks.

I don't like the track 0, track 1 etc. So perhaps this should be raised with the devs.

peanutbutterandcrackers:
@DylanC - Perhaps, let's have the ability to chose the track naming scheme! That'd be neat.

I'm pretty sure Mr. Richard Stallman and the entire free software foundation and every Unix hacker out there who would like to edit their grandchild's kindergarten graduation ceremony would love the fact that the tracks start from 0 and not 1, though. 😄

P-Woods:
The internals of the software will identify "things" in their own way. 
In my experience, using 0 for the first item instead of indicating null caused as many bugs as most common errors.  (I worked on mending some really old software.)
However, PEOPLE need help. Software changing the names of the tracks is not helpful. 
The user should be able to name the track and know the name will not change. 
Starting with "New Track" would be good, but "New Track 1" would be better, with auto indexing for "New Track 2" etc, if the track name had not been changed. (Unless intended to force a naming event, devs - see user interface guides.)

peanutbutterandcrackers:
[...]

Issue #321, opened 2016-09-08
Removing track causes reset of all track titles


PanderMusubi:
Removing track ( in this case Track 0 at the bottom) causes all custom track titles to get reset to Track 0, Track 1, etc. Please fix this so that custom track titles are maintained after deleting a track.

DylanC:
@PanderMusubi - Is this still an issue for you in newer versions of OpenShot?

DylanC:
Confirmed in v2.4.0 and the daily build.
OpenShot-v2.4.0-1-g8a71c3d-55-646-x86_64

DylanC:
@peanutbutterandcrackers - Excellent information! Thanks.

peanutbutterandcrackers:
@DylanC - I hope this will be resolved soon. You're most welcome!

DylanC:
@PanderMusubi - Hi there, can you try a daily build? This fix went in and should be available to you.

PanderMusubi:
@DylanC I'm not using it at the moment.

PR #1546, opened 2018-05-01
Retain track parameters during add/remove track

ferdnyc:
Previously, actionAddTrackAbove()/Below() would always create the new track at the very top of the layer stack (above all existing tracks), then walk the list of those tracks' clips and slide them up one layer.

This caused _track_-level parameters (name, lock status) to be left behind (in the case of lock status) or intentionally discarded (in the case of track-name, which was blanked on all tracks which were moved). Lock status was thus left behind on the "wrong track", relative to the clips that track previously contained.

With this new code, the list of _tracks_ is walked during an add operation, starting from the topmost track, and each track is moved upwards one layer (with all of its properties intact). As each track is moved, all of the clips and transitions on that track are pushed up along with it. Then, a new track is created and slotted into the space vacated by the lowermost moved track.

This method preserves all track parameters, because each track's data in the project file is the _same_ after being moved as it was before. Only the layer numbers change.

All of the label-blanking code was removed from both add methods, as well as from the renumbering (in the other direction) performed during actionRemoveTrack().

Fixes issues #1531 , #321 ; Addresses part of rant #454

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -885,7 +885,7 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         # Get # of tracks
         track_number = len(get_app().project.get(["layers"]))
 
-        # Look for existing Marker
+        # Create new track above existing layer(s)
         track = Track()
         track.data = {"number": track_number, "y": 0, "label": "", "lock": False}
         track.save()
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -910,11 +910,16 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
             existing_track.data["number"] = existing_layer + 1
             existing_track.save()
 
-            # Loop through clips for track, moving up to new layer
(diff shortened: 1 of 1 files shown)


N3WWN:
I think I can sum up my opinion of this PR with a quote...

>  @ferdnyc is a brilliant programmer 

(sorry, I couldn't resist! 🤣)

In all seriousness, this appears to be a very elegant change!

ferdnyc:
Pushed in a second, not-_technically_-related commit to correct a misleading cut-and-paste comment (describing a completely different operation than the one being performed). I happened to notice it while I was in there.

DylanC:
@ferdnyc - Again, this would be nice to get in for the v2.4.2 release. I will need to get testing this soon.

DylanC:
@ferdnyc - Really like this PR, hope to get this change in for v2.4.2 now.

DylanC:
@ferdnyc @N3WWN - Decided to get this change in too. Would be a very worthwhile PR to have in v2.4.2.

Issue #1438, opened 2018-03-29
Openshot preview does not match exported clip (1/4 speed slice not exporting correctly)


jmollo310:
**System Details:**
* Operating System / Distro: Windows 10 1920x1080 NVidia GeForce 1060 3GB, intel i7-700, RAM 16 GB
* OpenShot Version: 2.4.1
**Issue Description and steps to reproduce:**
Using prerecorded nvideo shadowplay clips of PUBG game, I do the following:
1. Enter 1 clip as track 0
2. Enter 2nd (very large clip) as track 1
3. Slice track 1 first keeping right side
4. Align tracks
5. Slice track 1 keeping left side
6. Copy track 1 and place copy in track 2
7. Set track 2 to slow 1/4 speed
8. Export clip

Exported clip shows a stopped stuttered frame where the 1/4 speed track 2 should be and it's about 15 seconds before track 2 slice should be beginning.
Here is a youtube clip showing entire process: https://youtu.be/YdCV-dmZ7GQ

peanutbutterandcrackers:
I am not so sure about the issue itself but could you perhaps please give the latest daily build a shot? Please head over to the [download](https://www.openshot.org/download/) page and click on 'daily builds' button and grab the latest one; uninstall the current OpenShot version and install the latest build, and see if this still persists in the latest version. Thanks!

peanutbutterandcrackers:
P. S: Thank you for the good report.

jmollo310:
Will give it a try later this evening.  Lol, I work as a senior software
engineer, so, used to doing bug reports.

jmollo310:
I installed the latest build and recreated the steps and the issue remains.

jmollo310:
Also, if you have a ftp site, I have the actually files that I can upload.  One is 8 GB though and the other only 100 MB

DylanC:
@jmollo310 - Perhaps can you upload the 100MB zip somewhere and provide us a link?

peanutbutterandcrackers:
@jmollo310 - Dropbox, perhaps? o.O

@DylanC - You sure you are willing to test that big file, Cap'ain? :D

DylanC:
@peanutbutterandcrackers - Of course! :)

jmollo310:
I think the link works as it lets me start the download.

peanutbutterandcrackers:
@DylanC - Cap'ain? We need your help here, good sir!

DylanC:
[...]

Issue #1098, opened 2017-12-04
Black screen in output video FHD 1920x1080 60p


dsieradzki:
**System Details:** i5-7500, 8 GB Ram, GTX 1050Ti
* Operating System / Distro: Kubuntu 17.10 64bit
* OpenShot Version: 2.4.1

**Issue Description and steps to reproduce:**
Black screen in part of output video.
1. Add two clips, on the seconds clip some part from both sides
2. Add Fade in(fast) effect on start slice and Fade out(slow) on the end on the second clip
3. Export as FHD 1920x1080, it's important to export in >60< frames with default codec

When I export in 24 frames is ok, problem is in 60 frames.
Similar problem occurs when I do timelaps(32x) on one slice.

N3WWN:
Same issue as mentioned in at least https://github.com/OpenShot/openshot-qt/issues/1034 and https://github.com/OpenShot/openshot-qt/issues/519 ?

At this time, anything that uses keyframes will not render properly if exported at a different framerate.  If the project was originally set to the export framerate before the fade/animation/etc is applied to the clips, then all will go as expected.

DylanC:
Dup of #519

mikelacsa:
This issue was fixed as mentioned in [519](https://github.com/OpenShot/openshot-qt/issues/519)

NetlistPCB:
I'm seeing the same issue on Windows 10

Issue #1041, opened 2017-11-15
Transition timing doesn't match timeline


N3WWN:
OpenShot 2.4.1

I noticed this on earlier versions and it's still present in 2.4.1, not to mention visible in the videos just published by @jonoomph ...

Transitions take 1/2 as long as they should;  The transition is complete exactly 1/2 way through the duration of the transition.

This is visible here:  https://youtu.be/VE6awGSr22Q?t=6m30s  

If you adjust the YouTube playback speed to 0.25, we enter the transition about 06:18 and the transition is visibly complete about 08:07, but the blue box which represents the transition extends to about 10:03.

To replicate:

- New project with profile of "HD 1080p 60 fps" (my default profile)
- Import a clip
 I used this 25.14s long video: https://videos.pexels.com/videos/aerial-shot-of-mountain-and-ocean-1469
- Drag the clip to the start of the timeline
- Zoom in (to the 2s value)
- Drag the clip to the start of the timeline again, overlapping the first clip by at least 2 sec
 I placed the 2nd clip at about 22s, giving about 3s of overlap.
- Click on the transition and note the duration in the properties
 Mine is 3.20s
- Move the playhead to 22s and step through, frame by frame, until you visibly see that the transition is complete
 For me, it was at 23:49
- Note that the blue box representing the transition continues until about 25:14, but no transition is visible during this 2nd half

I think this may be related to some of the other issues regarding frame markers moving if profiles change, but I'm not sure, so I'm creating a new issue to address it specifically.

Thanks!

DylanC:
Dup of #519

DylanC:
@N3WWN - Can you try v2.4.3 and let us know if this bug is still present?

Issue #640, opened 2017-05-14
Fade-off transition performed 2 times faster than it should.


sknaumov:
**System Details:**
* Operating System / Distro: Windows 10
* OpenShot Version: 2.3.2

**Issue Description and steps to reproduce:**
When I overlap edges of 2 videos blue rectangle is created to make an automatic transition. The problem is that transition is fully performed by the end of the first half of transition rectangle, and in the second half the 2nd video is playing without any transition => transition is 2 times faster than it should be and to create a proper one I need to sacrify some end part of the first video. Say, if I want to have 1 second transition, then currently I need to create 2-second overlap and sacrifice the last second of the first video.

Efenstor:
I double that! Transition is made from "brightness=1.0" to "brightness=-1.0", whereas it must be from 1.0 to 0.0. As far as I know the problem is present there for quite some time, and very bothersome. Please fix it!

DylanC:
@sknaumov - Does this happen in the v2.4.1 release?

sknaumov:
@DylanC - Yes, the problem is still there. Just downloaded and tried v2.4.1 to check.

N3WWN:
Just saw this issue... I think my issue https://github.com/OpenShot/openshot-qt/issues/1041 is a dupe of this one.

DylanC:
Dup of #519

sknaumov:
Looks not like a Dup of #519, because sometimes users say that transitions are incorrectly applied on export, but here it is incorrect even in internal video player. I believe these cases should be treated differently.

DylanC:
@N3WWN - What do you think? Is this a dup of #519 or a separate issue?

N3WWN:
@DylanC Regarding https://github.com/OpenShot/openshot-qt/issues/519 , I believe this to be a separate issue since this issue is talking about the transitions as they behave by default and https://github.com/OpenShot/openshot-qt/issues/519 is talking about keyframes of all types if the export profile doesn't match the project profile.  

I think https://github.com/OpenShot/openshot-qt/issues/1041 may be a dupe of this one, though, as the issues talked about here and in https://github.com/OpenShot/openshot-qt/issues/1041 both pertain the transitions without regard to profile since the symptoms are visible in the internal player.

DylanC:
@N3WWN - I've changed #1041 to be a Dup of this issue. This issue is no longer a Dup of #519.

sknaumov:
Still present in 2.4.2

[...]

PR #1557, opened 2018-05-04
[WIP] Disable incompatible profiles for existing projects

N3WWN:
This PR will grey out the non-compatible profiles in the Profile dialog for existing projects.  

In this context an "existing project" is one that has not been saved (this is the check I'm currently performing).

The reason behind this change is an attempt to prevent users from changing to profiles with different frame rates, which currently break where keyframes are positioned on the timeline.

Issues that mention this include, but may not be limited to: 

#519 
#640 
#1034  
#1041 
#1098 
#1438

So, the only gotcha is someone who builds a project with keyframed items without saving it.

This PR may need to be modified to look for keyframed items in the project and only restrict the profiles based on if they really are incompatible or not.


N3WWN:
Hrm... 😕 

I wish there was some way to view the reasons for build failure.

DylanC:
@N3WWN - Yeah I know,  there really should be a way.

DylanC:
@N3WWN - The build is passing again. 😆

DylanC:
@N3WWN - Can you move your old PR's to the develop branch if possible?

DylanC:
@N3WWN - Closing since this relates to master and not develop. Please re-submit if you plan to get this in. Thanks.

PR #1578, opened 2018-05-10
add appdata

eszlari:
This file is used in software managers like GNOME Software or KDE Discover to display information. It is also used on Flathub: https://flathub.org/apps/details/org.openshot.OpenShot


peanutbutterandcrackers:
Thank you for your contribution to the project. Hope our maintainer will soon see this. And even more so, I hope you keep on contributing to this project in whatever way you can.

Also.... whoa! You're the one maintaining the flatpak for OpenShot! Awesome! Thank you, good sir, for your work. Also, v2.4.2 will be out soon. I hope you support OpenShot through more releases, too. :)

xdg/openshot-qt.appdata.xml
```diff
@@ -0,0 +1,45 @@
...
+		<content_attribute id="drugs-alcohol">none</content_attribute>
+		<content_attribute id="drugs-narcotics">none</content_attribute>
+		<content_attribute id="drugs-tobacco">none</content_attribute>
+		<content_attribute id="sex-nudity">none</content_attribute>
+		<content_attribute id="sex-themes">none</content_attribute>
+		<content_attribute id="language-profanity">none</content_attribute>
+		<content_attribute id="language-humor">none</content_attribute>
+		<content_attribute id="language-discrimination">none</content_attribute>
+		<content_attribute id="social-chat">none</content_attribute>
+		<content_attribute id="social-info">none</content_attribute>
```
ferdnyc: Based on OpenShot's ability to (optionally, but on by default) send tracking metrics back to the "mothership", I believe this should be at _least_ "moderate", possibly even "intense". (The telemetry doesn't _contain_ any personally-identifiable information, but it does include a unique identifier that ties the data to a specific OpenShot instance installed on a specific system, which could theoretically be associated with a particular user. I'm not clear on whether that's enough to raise the rating to "intense".)

ferdnyc: (The "This" in my previous comment being OpenShot's `social-info` content rating, just to be clear.)

xdg/openshot-qt.appdata.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
```
ferdnyc: [Freedesktop says](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html#spec-appdata-filespec) that the `"desktop"` type is legacy, and should be avoided in favor of `type="desktop-application"` for new AppData files.

setup.py
```diff
@@ -65,6 +65,8 @@
 os_files = [
     # XDG application description
     ('share/applications', ['xdg/openshot-qt.desktop']),
+    # AppStream metadata
+    ('share/appdata', ['xdg/openshot-qt.appdata.xml']),
```
ferdnyc: Apparently [these should be stored](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location) in `/usr/share/metainfo/` now; `/usr/share/appdata/` is also read for backwards compatibility but it's deprecated.

xdg/openshot-qt.appdata.xml
```diff
@@ -0,0 +1,49 @@
...
+  <id>openshot-qt.desktop</id>
```
ferdnyc: I guess using `<id/>` to associate the AppData with the application's `.desktop` file [is also deprecated](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html#spec-appdata-filespec). (OpenShot's is `openshot-qt.desktop`, BTW.) According to that spec `<id/>` is now supposed to be a reverse-DNS-style identifier, and `<launchable/>` associates the AppData with a `.desktop` file. So, if I'm interpreting correctly, I'd suggest:
```xml
<id>org.OpenShot.openshot-qt</id>
<launchable type="desktop-id">openshot-qt.desktop</launchable>
```
and while we're at it, also
```xml
<provides>
  <binary>openshot-qt</binary>
</provides>
```

ferdnyc: Oh, and while I'm suggesting fields to add, one more:
```xml
<url type="bugtracker">https://github.com/OpenShot/openshot-qt/issues</url>
```
Thanks!

ferdnyc:
> That's why I left out the renaming at first.

\*nod\* TBH, I'm wondering... does it even have to be done now? AIUI from reading the spec, there's a lot of support for "legacy" paths/names — is there anything, other than it possibly seeming a bit weird, with creating  a `/usr/share/metainfo/org.OpenShot.openshot.xml` AppData file that still references the old `openshot-qt.desktop` file and legacy names? Renaming things like the icon and desktop files might be a bit more disruptive than OpenShot can handle, given all of the automation @jonoomph  has in place. 

(Apologies if it seemed like I was pushing for a complete update to the current standards. I know Gnome's own projects have generally gone that route but it does seem like Freedesktop left an "out" for others who don't necessarily want to disrupt things that much, even when following the latest spec.)

eszlari:
AppStream conforming names are required by flatpak.  At the moment I do this renames in the flatpak build manifest. The plan is to get all these modifications upstream.

ferdnyc:
Ah! I didn't realize. Then, absolutely, I say let's modernize. We'll have loop in @jonoomph when he has a moment to sort out the automated builds, but it sounds like this is something that's long overdue.

jonoomph:
If possible, can you send a new PR (or update this one) to merge into develop (rather than master). I have no problems with merging this in, but I do want to verify this doesn't break the PPA, Debian packages, or Ubuntu packages. And by break, I mean, .desktop file has a missing icon, doesn't launch correctly, or something else related to the shortcut and icon. If it all still works, then sounds good to me. =)

eszlari:
https://github.com/OpenShot/openshot-qt/pull/1967

Issue #1393, opened 2018-03-17
Add "Shut down after processed"


songsammy:
My computer is too slow that I have to wait for ages to get the video exported.
So  I often do the work first , and then export it later at night.
_(I mean , I leave it on **DURING the WHOLE night**. It's a waste of energy)_
Would you please give us a tickbox of "_Shut down after processed_"?

peanutbutterandcrackers:
Yeah! This would be quite neat!

songsammy:
Request new feature

PR #1581, opened 2018-05-12
[WIP] Added an option to the export dialog to shutdown the system 60 seconds after the export is completed.

bnieuwveld:
This should address issue #1393.


DylanC:
@bnieuwveld - Congrats on your first PR! :)

N3WWN:
Congrats on your first PR @bnieuwveld !

How does the code behave if you're running OpenShot as a normal (non-root/non-admin) user?  

- Does the shutdown proceed regardless of user privileges?  
- Is the user prompted to enter a password to elevate privs to allow shutdown?  
- Does the shutdown fail to execute due to improper privs?
- Other?

peanutbutterandcrackers:
I did try executing the main code to shutdown the computer from a non-root user:
```python
import os
os.system("shutdown -h +1")
```
and I was able to shutdown the computer just fine. So, I think this works. I haven't gotten to test this one out, yet, though.

bnieuwveld:
I found the same as @peanutbutterandcrackers - I wasn't running OpenShot or the shutdown command as root, and it still worked. Might require a bit more testing on Windows, although the command didn't have any problems (I might extend it so it displays an actual message on sign-out).

@peanutbutterandcrackers raised the suggestion of making it a variable shutdown delay, rather than it being hardcoded at 60 seconds. The only reason for this currently is that the shutdown command on Linux only allows you to specify the delay in minutes and doesn't allow too much control, i.e. `shutdown -h +1` (Windows allows you to specify seconds). I was hesitant to make it an instant shutdown i.e. `shutdown -h now`, because it seemed like the entire system crashed as soon as the export hit 100%... just didn't seem particularly elegant to me.

Additionally, another suggestion was that the function should schedule the shutdown first, then close the application, and _then_ execute the shutdown command. I think this would be a little more elegant, as currently the application is killed by the shutdown command, which makes OpenShot think it's encountered a crash on next launch. (Doesn't affect anything; just mentions this fact in the logs.)

I'm wondering if it's worth parking this PR and adding the above enhancements...? Any thoughts?

peanutbutterandcrackers:
I would like to now add that we should probably have the main option be something like 'close on complete', and have 'shutdown in x minutes of completion' as another optional check-able thingy.

The reason being some uber nerd might want to have openshot render the whole thing out, and launch a couple of scripts upon successful completion of the renderings. Just might.

peanutbutterandcrackers:
I would really like to hear @ferdnyc's ever-so-constructive suggestion on this one too. :)

ferdnyc:
I don't _particularly_ have any thoughts on this, since I tend to be one of those, "You mean people turn computers **off**!?!" types. Mine only ever powers down completely when I'm physically moving or upgrading it.

I'd suggest looking at how other applications handle this sort of feature, in particular cross-platform ones. Though not necessarily to emulate what they do, since my recollection is that it ends up becoming kind of a Frankenfeature for most of them. But it would at least give us an idea of how this is handled. 

I know a lot end up doing this as an "execute command(s) when completed" type feature, which is a bit more open-ended and would allow the user to choose things like the shutdown delay for themselves. Or use more modern, but less universally-supported, commands, like `systemctl poweroff`. Or even `gnome-session-quit --power-off` which AIUI is the recommended way (this week) to cleanly exit a Gnome Shell session. (It'll also throw up a prompt giving the user 60 seconds to cancel the shutdown, which is nice.)

I do think that OpenShot should fire off the command as part of its own application shutdown, so that it doesn't end up getting forcibly killed. For that reason we may also want to require that the project file be cleanly saved to disk before this can be enabled, to ensure that no work is lost. But I'm not sure of the best way to do this without OpenShot blocking on the command's completion, since most shutdown commands won't return until the shutdown time arrives.

It looks like OpenShot may have to have a separate "helper" program it can trigger to handle the completion command(s). The [python-daemon](https://pypi.org/project/python-daemon/) library might be helpful. I'm thinking something like a `daemon.py` program that detaches from the main program and runs the user commands, allowing OpenShot to continue so that it can close itself gracefully. OpenShot would do something like `subprocess.run(["python3", "/path/to/our/daemon.py", settings_data.get("completion-command")])` which would return immediately and allow OpenShot to close, while the command is launched in the detached process.

N3WWN:
> "You mean people turn computers off!?!"

I don't believe it as I don't turn machines off, either!  😜  

> For that reason we may also want to require that the project file be cleanly saved to disk before this can be enabled, to ensure that no work is lost.

^ +1 for this!

We'd probably have to disable the shutdown option if the project hasn't been saved, or perhaps trigger a project save if shutdown is selected without yet saving.

> "execute command(s) when completed" type feature

I think I like this... it leaves the per-system implementation up to the user.  We could include samples of "shutting down the system in Gnome 3", "shutting down the system in Windows", etc in the docs.

> I do think that OpenShot should fire off the command as part of its own application shutdown, so that it doesn't end up getting forcibly killed.

^ +1 for this, too!


At any rate, as written, I think this feature would require quite a bit of testing.  Not all Linux distros support user-triggered shutdowns the same.  Some use dbus, some use gnome, some use systemctl, some require that you enter a password to confirm shutdown, etc.  And that's just Linux.  We'd have to test similarly with various versions of Mac OSX and Windows.

We don't mean to discourage you... only to help flesh out your contribution. 🤗

bnieuwveld:
> We don't mean to discourage you... only to help flesh out your contribution. 🤗

No discouragement here whatsoever; in fact quite the opposite! I appreciate your thoughts and suggestions on this and am interested to see how best we could implement and extend this feature. This is why the FOSS community is great 😄 

> We'd probably have to disable the shutdown option if the project hasn't been saved, or perhaps trigger a project save if shutdown is selected without yet saving.

For some reason I had it in my head that OpenShot would automatically save a project for you if you tried to export it while unsaved. I even thought I saw some code for it... turns out I was wrong, that's just to give it a placeholder filename of "Untitled Project". Agree that this would be an **essential** addition.

I think this might be one that requires a bit more research and testing. No problem - I'll just start working on the next PR 😉

ferdnyc:
> For some reason I had it in my head that OpenShot would automatically save a project for you if you tried to export it while unsaved.

Yeah, arguably maybe it _should_, but... at the same time, there hasn't really been any particular reason it **needed** to force a save, since the (on-disk) project file doesn't get used for anything during the export process, and crash-recovery is reasonably OK about making sure that work won't be lost even if the export crashes OpenShot or whatever. 

Still, this feature may be the perfect excuse to force pre-export saving (across the board, even if shutdown isn't enabled), which would receive no opposition from me. :angel:

N3WWN:
@bnieuwveld wrote:
> I think this might be one that requires a bit more research and testing. No problem - I'll just start working on the next PR

Don't give up on this PR yet!  😀 

Please just edit the title and add `[WIP] ` to the beginning so we know that this PR is not yet ready to merge.  There's at least one other PR that is a WIP if you're not sure what I'm asking. 

We look forward to your continued contributions!

@ferdnyc wrote:
> Still, this feature may be the perfect excuse to force pre-export saving (across the board, even if shutdown isn't enabled), which would receive no opposition from me.

^This!  ➕1️⃣

peanutbutterandcrackers:
So, on Linux mint's panel properties, I have something like this:
screenshot at 2018-05-22 15-45-19
Note that the indented check-box depends on the previous one. As in:
screenshot at 2018-05-22 15-45-42

I wonder if we could have something similar.
- [ ] Close OpenShot after rendering is complete
        - [ ] Shutdown Computer after completion in ___ minutes

I think it would be clear and simple enough for anyone to understand. And the user would even be able to set the amount of minutes themselves.

Also, could the said external helper program send out wall messages to *all terminals* that the computer is going to shutdown. With the current system, only the tty that did `$ openshot-qt` gets the wall message for some reason. But maybe that's just Python.

ferdnyc:
> I think it would be clear and simple enough for anyone to understand. And the user would even be able to set the amount of minutes themselves.

Well, that's... that all comes down to how the shutdown is even implemented. As I said, if e.g. `gnome-session-quit --power-off` is used, then the timeout is fixed at 60 seconds, but the user does get a nice message popup allowing them to abort the shutdown within that timeframe. But for the most part, I don't imagine the user will _want_ to set a timeout. (They certainly won't want to HAVE to, and as long a there's a sensible default almost everyone will end up just leaving it, because it's kind of the whole point that this is expected to be happening unattended.)

> Also, could the said external helper program send out wall messages to all terminals that the computer is going to shutdown. With the current system, only the tty that did $ openshot-qt gets the wall message for some reason. But maybe that's just Python.

That's just `shutdown`, since all the WIP code does is run that command. It send the message wherever it's able. When executed as a regular user (not as root), it's very possible it doesn't have permission to write to other users' sessions, but in that case there's nothing we can do about that.

At least on my system, `/usr/sbin/shutdown` is nothing but a softlink to `/usr/bin/systemctl`, showing that this is really just a legacy compatibility feature of systemd, and we'd probably be better off using `systemctl poweroff`. The one advantage to using `shutdown` is that it's a single command that's pretty much universally supported on pretty much every platform with identical semantics. 

But there are probably lots of users who might prefer that OpenShot do a `systemctl suspend` or `systemctl hibernate` instead of powering off, which gets us right back to the "run a command at completion" type implementation because it'd be very hard to predict all of the possible post-completion actions and expose all of them via a purpose-built interface. Whereas limiting ourselves to powering off completely after a certain timeout may be providing both too few _and_ too many features at the same time. The configurable timeout, despite being a staple of such functions since time immemorial (for no real reason other than "it's how it's always been done") is of questionable use to the majority of users, but OTOH the lack of other dispositions (something other than a full system poweroff) may prove to be lacking in flexibility.

bnieuwveld:
Having a look through another application's code to determine how it does this, it seems *way* more complicated than this (whether that's by necessity or not, I don't know). It's written in C++ though so makes very little sense to me, but from what I can determine there's a "Qt-ier" way to do this through a `QDBusInterface`. Can send this through to you @ferdnyc if you'd be happy to take a look and/or decipher it for me.

Worth pointing out however that it *only* contains an if/else statement for KDE or Gnome. I don't believe the app runs on Windows (or Mac?) at all.

ferdnyc:
Coincidentally, I was just looking at the `org.freedesktop.login1` DBus service in D-Feet (the d-bus interactive explorer). The service is mentioned in the `systemd.special(7)` man page as an option for "Applications wanting to reboot the system", as they are advised to "execute systemctl reboot (possibly with the --no-block option) or call systemd-logind(8)'s org.freedesktop.login1.Manager.Reboot D-Bus method directly.")

org.freedesktop.login1.Manager contains a **vast** array of methods for session and system management, the screenshot below is actually incomplete but it was all I could fit in one screen. The entire range of functionality possible using tools like `systemctl`, `shutdown`, etc. on the command line appears to be exposed via D-Bus, allowing programmatic management of those same functions without having to resort to launching shell commands.

But the tradeoff for all this flexibility and power? You're quite correct, absolutely anything involving D-Bus is strictly Linux only, and would need to be supplemented by additional code to perform equivalent or alternate actions when running on other platforms.

screenshot from 2018-05-22 20-56-00

bnieuwveld:
> Coincidentally, I was just looking at...

Great minds think alike. Trans-continental telepathy. Brilliant.

> absolutely anything involving D-Bus is strictly Linux only, and would need to be supplemented by additional code to perform equivalent or alternate actions when running on other platforms

This shouldn't be too difficult. The current command for a Windows shutdown (while I haven't tested it in the context of OpenShot) works well, and even allows you to put a custom message below the "You're about to be signed out" popup (on Windows 10 at least - the same one you get when you're about to be forcibly signed out for an update).

That would just leave macOS, which I assume could use a similar command to *nix to shutdown. Or, according to [this page](https://apple.stackexchange.com/questions/103571/using-the-terminal-command-to-shutdown-restart-and-sleep-my-mac), there's a nice AppleScript command to do it gracefully.

And then, if all else fails, we could just have the good old existing `shutdown` command as a fallback?

ferdnyc:
Oh, but re:
- [ ] Close OpenShot after rendering is complete
  - [ ] Shutdown Computer after completion in ___ minutes

> I think it would be clear and simple enough for anyone to understand.

Agreed, but slightly less than _optimally_ simple, since shutting down isn't _dependent_ on closing OpenShot, it **implies** closing OpenShot (or would, once the "make sure to close OpenShot gracefully before shutting down" thing is worked out). 

I will admit without reservation that this is an exceedingly minor point, but: The difference in semantics between this and the "hide buttons" preference of the Mint panel is that the arrows preference _only has meaning_ if you've chosen to enable hide buttons — until that choice is made, the question has no meaningful answer, which is why the user isn't able to answer it. IOW, those nested checkboxes don't represent two choices, but rather a **single** option, which then (iff it's enabled) can be further customized. The two choices are not even _merely_ dependent, they're **inseparable**. If you were to remove the outer option, you'd also **have to** remove the inner one because it would literally have no meaning.

We don't want shutdown to be *dependent* on quitting OpenShot, and it certainly doesn't **only** have meaning once the user has chosen to quit OpenShot. We want it to be an _alternative_ to simply quitting OpenShot. (It will **imply** also quitting OpenShot, yes, but that's because we _want_ it to: We're making the implementation decision that the user _doesn't_ need to explicitly ask for that, it's simply implied.)

So, in my mind, were OpenShot to offer those two possible post-export actions, it would be _in addition to_ an implicit third possibility, the only one it currently offers: doing absolutely nothing. Therefore, adding those two alternative dispositions to the existing (implicit) disposition leads us to a flat list of radio-button selections:

#### After export is complete:
:radio_button: Do Nothing
:white_circle: Quit OpenShot
:white_circle: Shut down computer

(Which, most critically, means that a user wanting to shutdown after export doesn't end up having having to enable _two_ things, when really all they want is _one_ thing.)

### Danger, Will Robinson

...And now this is the _exact_ turning point in any discussion of this sot of feature, where we begin down the road to the inevitable madness I was referring to when I described them as "Frankenfeatures" in my initial comment.

Because now, I'll say "And we may want to extend that list with other options", and give the first example that comes to mind:
:white_circle: Run a command: [___________________]

So _**then**_ someone else suggests, very reasonably, that the user may want to run a command _and_ have OpenShot quit, or the computer shut down, so do we have to restrict it to an either/or?

Makes sense, so now we've got a bunch of _checkboxes_ for various dispositions, letting the user enable some or all of them as they please. (If some are redundant, who cares? Selecting "Quit OpenShot" when they also selected "Shut down" doesn't hurt anything, and if they _don't_ select it exactly the same thing will happen but that still doesn't hurt anything.)

And where things go from there is pretty predictable, as is the end result because it's happened before.

### Where this leads

I present to you the Frankenfeature disposition options of optical media burning tool K3B:

image

Seven different conditions, each with a configurable _matrix_ of responses. Frankensanity.

Or, take MKVtoolnix, the Matroska container format encoder/transcoder/remuxer. They went the other direction, with an open-ended _list_ of actions that can be customized out of five basic Types, and configured to execute on various set conditions. Also, as a tooltip on the checkboxes notes:
> If any of these checkboxes is checked, the action will be executed when the corresponding condition is met. Independent of the checkboxes, every active configuration can be triggered manually from the "job output" tool.

image

So, any action you've configured, in addition to being associated with the set system conditions listed in its "When to execute:" section, can also be enabled on an individual basis on each individual job that's added to the software's processing queue.

Even more flexible, even more confusing (and a bit weird, to be honest, since with the MKVtoolnix system there's nothing to stop you, say, telling the software that when the job queue ends, it should Suspend AND Hibernate AND Power off the computer, and then what does that even mean?), and even more Frankentastic.

DylanC:
@bnieuwveld - Closing since this relates to master and not develop. Please re-submit if you plan to get this in. Thanks.

Issue #1594, opened 2018-05-16
Default audio codec "ac3" not supported on majority/most players


N3WWN:
**Describe the bug**
I just tried uploading a video directly to a new video streaming service and my video did not contain any audio after upload.  Uploading to YouTube and then importing it into the new video streaming service works.  After consulting with the new video service's support engineers, it turns out that ac3 is not supported by their system.  The reason it works when imported from YouTube is that YouTube converts the ac3 stream into aac, which they can then import.

**System Details (please complete the following information):**
 - Operating System / Distro: N/A
 - OpenShot Version: 2.4.1

Additional _open_ issues reported which have a high probability of being due to the use of ac3:

#996
#1047 
#1466 
#1583 

There are other _closed_ issues which are likely due to this, also, but I won't list them here.

N3WWN:
Additional info:
- https://en.wikipedia.org/wiki/Comparison_of_video_container_formats#Audio_formats_supported
- https://en.wikipedia.org/wiki/Advanced_Audio_Coding
- https://en.wikipedia.org/wiki/Dolby_Digital (this is what AC3 actually is, which was patented until recently, which is why fewer devices support AC3 than AAC)
- https://www.macxdvd.com/mac-dvd-video-converter-how-to/aac-vs-ac3-comparison.htm
- https://handbrake.fr/docs/en/latest/technical/audio-codecs.html

The video which I was having issues with on one service and worked fine on YouTube had the audio converted to mp4a.40.2 (Low-Complexity AAC).

N3WWN:
Research indicates:
- Apple TV preset can retain ac3 as default due to ac3 support:  https://support.apple.com/kb/SP724?locale=en_US (checked original Apple TV, 3rd gen, 4th gen and 4K models)
- AVCHD, DVD NTSC, DVD PAL, MP4 XVID presets can retain ac3 default due to this being a standard format for these targets
- Flikr HD and Nokia nHD - Unable to find any recent info on these, so leaving these preset alone
- MP4 x264 preset should change to aac to be supported on as many devices as possible
- MOV x264 preset should change to aac due to lack of ac3 support in older versions of QuickTime
- Xbox 360 - Since this preset is using the AVI container, we can should leave the default as ac3 : https://support.xbox.com/en-US/xbox-360/console/audio-video-playback-faq

DylanC:
@N3WWN - Would you consider this a bug or enhancement?

N3WWN:
[...]

Issue #1583, opened 2018-05-12
Video Audio Unable To Play On Mobile Devices


MisterTeeXD:
**Describe the bug**
When a video is rendered in Openshot and then moved to a Android/iOS Device, then the audio is unrecognised by the device and cannot play

**System Details (please complete the following information):**
 - Operating System / Distro: Windows 10
 - OpenShot Version 2.4.1-90

**To Reproduce**
1. Go to File
2. Click on Export Video
3. Scroll Down to (No Scrolling Needed)
4. See error by moving the video file on a mobile device and play the ideo

**Expected behavior**
The device recognises the audio and plays the video with the audio

**Screenshots**

**Logs**
If you are experiencing a crash, please collect and attach logs of the problem.

**Additional context**
A few builds back, this bug was still there

peanutbutterandcrackers:
Can you please use the VLC app to play the video? I think it's just an encoding/decoding issue. Perhaps we need to have a separate 'mobile devices' profile?

DylanC:
@MisterTeeXD Can you try v2.4.2 and let us know if you have any success?

Issue #1466, opened 2018-04-03
No audio when exporting file for web


dbstlouis:
**System Details:**
* Operating System / Distro: ?
* OpenShot Version: ?
* _Please attach log files if crash_

**Issue Description and steps to reproduce:**
I edited a existing video (interview with audio) and when we attempted to export for web posting there is no audio. What specs are required to ensure audio is preserved?

DylanC:
@dbstlouis - I had this same issue recently. What export profile did you use?

dbstlouis:
I was using HDV 1440i 29.97 fps

DylanC:
@dbstlouis - What did you select as output format on the export dialog?

peanutbutterandcrackers:
Perhaps the audio is there but your media player does not have the ability to decode the audio encoded using the encoder that OpenShot used? I suggest you try playing the video with [VLC Media Player](https://www.videolan.org/vlc/index.html), just to make sure it is not what I imagine it is. Similar issue has happened to me too, before. But it was the media player's fault. Hope that helps!

DylanC:
@dbstlouis - Are you using v2.4.1 or a daily build?

DylanC:
Marking as a bug since I did have this issue too.

DylanC:
@dbstlouis Can you try v2.4.2 and let us know if you have success? A change was put in for the audio issues.

Issue #1047, opened 2017-11-17
Video Export problem


Ben2112:
Ich bin neu bei openshot und mir stellt sich folgendes Problem:
nachdem ich ein Projekt mit mehreren zusammengeschnittenen Videos exportiere, fehlt auf dem exportierten Video der Ton. Video ist ok.
Was mache ich falsch?

DylanC:
EN:

I am new to openshot and I have the following problem:
After I export a project with several edited videos, the audio is missing on the exported video. Video is ok.
What am I doing wrong?

croco3008:
To track down your problem developers will need to know:
1. Your OpenShot version
2. Your operating system
3. Does audio work in preview mode?
4. What export profile you tried? Does some other format will work ok?

peanutbutterandcrackers:
@Ben2112 - Perhaps try setting the 'target' in the export dialogue to something else. I recommend 'OGG(vorbis/theora)'. It just might be that the default audio encoder is not working properly. It could very well also be that the media player that you are using can't decode the audio. Try [VLC](https://www.videolan.org/vlc/) instead. If you do not have the project till now but still have the exported video, you could try re-encoding it into another format using something like [Handbrake](https://handbrake.fr/). Please note that this might not work if the audio encoder didn't do the job while exporting and no audio has been exported at all.

Hope this helps.

Fig: Changing the 'target' during export.
![screenshot from 2018-02-12 07 20 43](https://user-images.githubusercontent.com/26764547/36090895-1ac00616-100a-11e8-8ac5-9e5136476ac1.png)

P. S: Please do make sure that you are also using the latest version of OpenShot (v 2.4.1) and not any older ones.

Good luck!

DylanC:
@Ben2112 Can you try v2.4.2 and let us know if you have success? A change was put in for the audio issues.

PR #1595, opened 2018-05-16
Change default audio codec from ac3 to aac for two presets

N3WWN:
Resolves issue reported in #1594
 
As noted in #1594, there are additional open issues reported which have a high probability of being resolved by this PR:

#996
#1047
#1466
#1583
#1599

### DIFF
--- a/src/presets/format_mov_x264.xml
+++ b/src/presets/format_mov_x264.xml
@@ -5,7 +5,7 @@
 	<title translatable="True">MOV (h.264)</title>
 	<videoformat>mov</videoformat>
 	<videocodec>libx264</videocodec>
-	<audiocodec>ac3</audiocodec>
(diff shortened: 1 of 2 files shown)


ferdnyc:
**+1.** AC3 is the standard audio format for DVD and BluRay media, and should be used there, but when outputting to MP4 file AAC is not only preferable, it's part of the _specification_. In theory it's guaranteed to be supported everywhere MP4 files are. (And especially for 2-channel, there's no reason _not_ to use AAC. Sometimes certain hardware will only do 5.1 or 7.1 if the audio is AC3, but not if it's AAC. But that's never an issue with stereo audio.)

DylanC:
I think this is a minor change to accept before v2.4.2 but a very good one to have in before the next release. Happy to accept responsibility for the merge.

PR #1600, opened 2018-05-17
Fix structure of title SVG

ferdnyc:
`Creative_Commons_1.svg` had an odd "switch" structure in it (marked as an Illustrator extension) that confuses resvg. This fixes up the file by removing the `<switch>`, but leaving the rendering unchanged.


peanutbutterandcrackers:
A non-free software was used in the making of this free software gem!? Richard Stallman just shed a single drop of tear in his pool of GPL3+ Licensed tear.

DylanC:
@ferdnyc - Good find! :)

ferdnyc:
@N3WWN gets the credit, he noticed the rendering issue while testing resvg. I just diffed the SVG source against `Creative_Commons_2.svg` to see where they diverged. Once the Illustrator-tagged section jumped out at me, a little Inkscape object-tree pruning fixed it right up.

ferdnyc:
> A non-free software was used in the making of this free software gem!? Richard Stallman just shed a single drop of tear in his pool of GPL3+ Licensed tear.

Yeah, I'm not _entirely_ clear what the deal with that was — it's possible that Illustrator wasn't involved in creating the title SVG, but rather was used to create the orginal Creative Commons logotype, which was then inserted into the title SVG. Its status as a `<foreignObject>` would seem to favor that interpretation.

peanutbutterandcrackers:
Hmm... I see. That makes sense.

Long live InkScape! Viva La Free Software!

:D

DylanC:
This is a nice change to get in ahead of the possible resvg usage in the future. Even with or without resvg we shouldn't have strange artifacts like this in the svg files. Happy to merge this in.

Issue #293, opened 2016-08-31
Open recently silently fails if path/file don't exist


msbentley:
Using OpenShot 2.1 on ubuntu 16.04 x64, I renamed a directory containing a recent project. When I try to open the project from the "File / Recent Projects" menu, the program doesn't report a file not found message, but just silently ignores the request.

DylanC:
@msbentley - Agreed. We should be outputting a visual error message here.

peanutbutterandcrackers:
@DylanC - Still not fixed in v.2.4.1 Tested today. So, yeah...

peanutbutterandcrackers:
@DylanC - Wonder if we should do prioritizations and milestones too. Like "This enhancement will be solved in v2.4.2" or something...

DylanC:
@peanutbutterandcrackers - True. Something like that would be nice!

peanutbutterandcrackers:
@DylanC - Perhaps you should write to Mr. Thomas regarding that (and also informing him about the pull request that awaits his approval). I would also like to suggest making very small version number bumps. Let's have OpenShot v2.4.2 after v 2.4.1. Let's even have 'unstable' and 'stable' versions, aside of the daily builds. Let's have v2.4.x where x is an even number, be stable - bug fix release and 2.4.x where x is an odd number be an enhancement release --- like GIMP does. And let's have the Major.Minor.x go up to Major.Minor.42 before finally going up to Major.Minor+1.0, or something similar. I'd like to see OpenShot be around even after 25 more years.

PR #1605, opened 2018-05-22
WIP: Attempting to open a recent project which has moved will no longer fail silently.

bnieuwveld:
OpenShot will also offer to remove the missing file from the Recent Projects menu.

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -455,6 +455,23 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
                 self.load_recent_menu()
 
                 log.info("Loaded project {}".format(file_path))
+            else:
+                # If statement is required, as if the user hits "Cancel"
+                # on the load file dialog, it is interepreted as trying
+                # to open a file with a blank name. This could use
+                # some improvement.
+                if file_path != "":
+                    log.info("File not found at {}".format(file_path))
+                    ret = QMessageBox.question(self, 
+                                        _("Error Opening Project"), 
+                                        _("Could not find project {}. Would you like to remove it from the Recent Projects menu?".format(file_path)),
+                                        QMessageBox.No | QMessageBox.Yes)
+                    if ret == QMessageBox.Yes:
+                        self.remove_recent_project(file_path)
+                        log.info("Removing {} from Recent Projects.".format(file_path))
+                        
+                        # Load recent projects again
+                        self.load_recent_menu()
 
         except Exception as ex:
             log.error("Couldn't open project {}".format(file_path))
@@ -1882,6 +1899,15 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         for file_path in reversed(recent_projects):
             new_action = self.recent_menu.addAction(file_path)
             new_action.triggered.connect(functools.partial(self.recent_project_clicked, file_path))
+            
+    # Remove a project from the Recent menu if it cannot be found by OpenShot.
+    def remove_recent_project(self, file_path):
+        s = settings.get_settings()
+        recent_projects = s.get("recent_projects")      
+        if file_path in recent_projects:
+            recent_projects.remove(file_path)
+        s.set("recent_projects", recent_projects)
+        s.save()
 
     def recent_project_clicked(self, file_path):
         """ Load a recent project when clicked """


peanutbutterandcrackers:
Resolves (the long over-due) issue #293. :)

Awesome!

peanutbutterandcrackers:
@ferdnyc: I had the following suggestions regarding this PR, on slack. I thought perhaps you'd have some more input.

>The pop up doesn't give any explanation, which I think might be necessary. "Perhaps it has been moved or deleted?" would have been great. I wonder if you could put in a small check-box (with smaller fonts) the could be checked to remove the project from the recent-files list and only have [okay] button in the pop-up box. I am afraid the current one might be a little bit confusing to many users as often they are only used to clicking 'yes' instead of actually reading the message. Just a thought.

Basically, what I proposed was this:

**Could not find project *%s*. Perhaps it has been moved or deleted.**
- [ ] Remove *%s* from Recent Projects menu
                                                                                               [OKAY]

Sorry, I couldn't markdown it properly. But I hope you understand.

Any suggestions?

I think the dialogue should also offer to browse for the moved project file, though... implied by the issue mentioned above.

ferdnyc:
As always, I think looking at other software is wise.

Here's how Audacity handles a missing Recent file:
image
Blender also displays an error message (not in a dialog, just as a transient popup that doesn't need to be confirmed) and removes the file from the recent list.

GIMP, OTOH, displays a "Could not open file" error in its messages panel, but leaves the file in the list. (GIMP's interface is terrible, though, so we shouldn't be looking to emulate it.)

I'd be most curious how software like Photoshop or Final Cut handles this, unfortunately right now I don't have access to any of those to test.

My instinct says that just removing the file from the list, automatically and without asking the user for permission, is probably the right move. A bad path in the recent list isn't really of any use to anyone. Either the file's been deleted, in which case no need to keep it around, or it's been moved to a new location, and when they open it from that other location it'll get added back to the list. But pruning bad paths from the list leaves more room to keep a list of _good_ paths, since slots on the Recent list are limited.

ferdnyc:
> A bad path in the recent list isn't really of any use to anyone. Either the file's been deleted, in which case no need to keep it around, or it's been moved to a new location, and when they open it from that other location it'll get added back to the list.

I suppose technically a third possibility exists, that the file is merely _temporarily_ offline (like, on removable media) and might return at some later date. But, while I had to mention it for completeness' sake, I really don't feel like that's a common enough situation to design around.

DylanC:
@bnieuwveld Do you want to rebase this on the develop branch? Will probably be needed before we could merge.

peanutbutterandcrackers:
@DylanC - Cap'ain, this is still a work in progress. Work is still to be done regarding the cleanup of recent files and stuff. Perhaps we need to wait a little longer...

bnieuwveld:
@DylanC @peanutbutterandcrackers - Yes, I've been out of action for quite a few weeks, so this definitely needs a look with some fresh eyes. Rebasing the changes would be a good idea though.

ferdnyc:
@peanutbutterandcrackers wrote:
> I think the dialogue should also offer to browse for the moved project file, though... implied by the issue mentioned above.

I would caution against this, only because it makes a lot more assumptions than you might think about the user's intentions and needs.
1. We don't _know_ the file was moved. 
   1. It could've been deleted, in which case there's nothing to browse for.
   1. It could've been renamed, in which case browsing for _that_ filename specifically won't work, so we'd just have to trigger an Open Project...
   1. It could be on an unmounted drive or disconnected network share, in which case they'll need to switch out of OpenShot and make it accessible before they can open it.
   1. They might've just accidentally clicked the wrong entry in the Recent list, and need to click a different one instead.
1. Not all users load project files through the application's browser. The user may prefer to quit OpenShot, use the Explorer/Finder/etc to browse for (or even _search_ for) the project file, and double-click it to relaunch OpenShot. Forcing them into the built-in file browser isn't helpful to those people.
1. If we're just triggering a general Open Project... browse (see 1.2), there's no real benefit to the user by asking if they want to do that. For users who _do_ want to do that, the Open Project toolbar button requires exactly the same 1 clicks as a "Yes" button in a dialog offering to browse. But if they _don't_ want to do that, then we're getting in their way if we make them click "No" on that same dialog to escape our over-"helping".

bnieuwveld:
@peanutbutterandcrackers @ferdnyc - Where did we ever land on this one? My immediate thought would be a message box offering to either browse for the file (though, personally, I'd basically never do this), or remove it from the Recent Projects menu - at least that way it gives the user a choice. However, it might be testing my Qt skills...

speedytechdev:
Any devs on I need help ASAP please

ferdnyc:
> Any devs on I need help ASAP please

Someone else's unrelated PR is not a help forum. If you're having an issue, please [open a new Issue](https://github.com/OpenShot/openshot-qt/issues/new/choose). The project Issues are also not a help forum, so you need to describe your problem completely, not post "I need help". Nobody's going to help with whatever it is you need if they have to drag it out of you first.

speedytechdev:
I am support and this is the only place I could find help. Everyone is reporting that Openshot is crashing and I don't know what to do. Please help.

ferdnyc:
> I am support and this is the only place I could find help. Everyone is reporting that Openshot is crashing and I don't know what to do. Please help.

I told you how to get help. Please do not comment on this PR with unrelated issues again.

peanutbutterandcrackers:
I agree (or rather, can't disagree) with @ferdnyc (after reading his arguments).

bnieuwveld:
> If we wanted to show a message when removing an inaccessible Recents entry

...which I think we do, because a lack of any sort of feedback just doesn't feel right to me. (But I'm happy to hear any disagreements to this!)

> OpenShot's sadly-underused status line would seem like a reasonable place

True, I forgot about the status bar - and agree that it seems like a fitting use for it. Once I make the enhancement (and given the age of this one!), I will open a new PR and close out this one.

peanutbutterandcrackers:
Once your new PR is submitted, we can test it out and give feedbacks. All the very best! :+1: :)

peanutbutterandcrackers:
Well, we can always go back to either [this](https://user-images.githubusercontent.com/538020/40578757-bd0da184-60e7-11e8-8247-2249eff91545.png) or checked-by-default radio button - [this](https://github.com/OpenShot/openshot-qt/pull/1605#issuecomment-391225459) if the status bar idea - alone - doesn't feel quite right to you (both mentioned above).

bnieuwveld:
Superseded by #2202

PR #1627, opened 2018-05-30
Remove ActionGroup from Effects pane

ferdnyc:
The Effects pane currently contains an ActionGroup containing "Show All", "Video", and "Audio" filter buttons.

There are no Audio effects, making this a pointless waste of space (and pushing the clear button for the Filter input into an overflow menu, when the pane is sized to a comfortable width).

This change removes the ActionGroup, and updates the stored Effects pane width in the Advanced view to a narrower size that comfortably holds two columns of Effects with no wasted space.

screenshot from 2018-05-29 22-57-53


peanutbutterandcrackers:
Umm.....

So, OpenShot 1.x did have quite a lot of audio effects. [The plan](https://www.openshot.org/blog/2018/01/01/happy-new-year-2018/) is to port all those effects to 2.x series. Perhaps that is the reason why the ActionGroup is there.

ferdnyc:
That's cool. And as with the Anchor Property, those filters can be added back in _if_ and when that ever happens. But until then, and as things stand right now, it's nothing but useless clutter.

ferdnyc:
I'll need to resubmit this against the `develop` branch.

Issue #1587, opened 2018-05-13
Track actions as always visible buttons (permanent view instead of context menu)


lukashajek78:
If I want to do something with the track (video/audio) then it is very non user friendly to right click and find anything in the context menu. For instance audio fade in/out which is the basic feature.

I would be nicer to have **permanent menu (view) with the most popular actions**.
It could be **shown as buttons or anything which is always visible and does not disappear.**
Now it takes up many clicks to select a desired action on each and every track.

In general I don't like context menu with more than 1 level.
Some options have even 3 levels which is very unclear and slow for using.

ferdnyc:
I _tend_ to agree, and concur that it feels like there's a lot more we could be exposing in the interface instead of relegating to menus.

I did just want to point out that **right**-clicking isn't strictly necessary — the same actions menus can be accessed by left-clicking the sort of V-looking thing at the upper left corner of each clip/transition/track. (Not that this **AT ALL** affects the core issue of being able to access functions.)

The way I see it, there are two issues at play:
1. The need to go into the (context)-menus at all, to perform nearly any operation on the timeline. That tends to be just inconvenient and clunky. It'd be preferable to have toolbar buttons for at least _some_ actions, or even to use system menus (instead of context menus) simply for their repeatability and ease of access. (System menu selections could also be more easily bound to keyboard shortcuts, whereas currently there's really no keyboard access to the context menu functions.)
1. The organization of the many of the current action menu items, which can be annoyingly and uncomfortably deep. The Track and Transition menus are fine, they're pretty spare for the most part. But the Clip menu has turned into a messy catch-all of operations.
   1. Fade and Volume have separate submenus for acting on the "Start of Clip", "End of Clip", or "Entire Clip", and then provide basically the same set of "presets" for each.
   1. Animate adds yet _another_ layer below Start/End/Entire, splitting up the possible animation actions into "Center to Edge", "Edge to Center", and "Edge to Edge", then listing the actual animation actions to be applied below those. Ugh. There _has_ to be a better way.
   1. The Time menu is patently insane. I mean, seriously?
      ![screenshot from 2018-05-29 21-47-00](https://user-images.githubusercontent.com/538020/40694330-10939ec2-638a-11e8-8ee7-4442e48a2bb9.png)
[...]

PR #1643, opened 2018-06-03
New experimental Clip menu layout

ferdnyc:
This change starts something I'd alluded to in a comment on #1587: Flattening the sometimes uncomfortably-deep structure of the Clip context/action menu. (Even though I derided such changes as merely a band-aid, at the time — which is still how I feel, for the record, it's just that sometimes applying band-aids is still useful.)

A new preference is added at the bottom of the General section:

screenshot from 2018-06-03 07-06-07

It defaults to disabled (off), and nothing is different about OpenShot. IFF that preference is turned on, an alternate menu structure will be used for certain submenus of the Clip context menu. So far I've tackled the Fade and Time submenus.

The redesigned Fade menu is completely flat, it looks like this:

openshot-newfademenu

The new Time menu moves the Forward and Backward 1X (normal-speed) options to the top level of the submenu, and chops a level out of the Forward and Backward sub-submenus by combining Slow and Fast speeds as Sections of the resulting sub-submenu. (Both Forward and Backward will have the same submenu, the same way Freeze and Freeze & Zoom do currently.)

openshot-newtimemenu

...And I'm realizing, now, that I've mixed Reverse and Backward terminology in the same submenu, which looks really bad. But, that's why it's a work in progress. The original menu is consistent about always using Backward, but I kind of feel like Reverse works better in the context of the flattened menu.

Things I'm not sure about, or would invite input on (specifically, as in general I invite input on all of it):

1. Whether the Section labels really work, especially for the Fade menu. I'd originally just collapsed it all into a single menu divided by separators, no labeled Sections, and I'd rewritten the option text to incorporate what were previously the submenu labels. But then after I decided to use Sections for the Time menu, I went back and created Sections in the Fade menu so that I could revert the actual options to their original strings, mostly to ease the load on translators.
1. The layout of the two Time sub-submenus. I like the _idea_ of slowest-to-fastest progression, but I'm just not sure about it from an ergonomics perspective. It makes the fast-speed options harder to access, which probably isn't a great thing.
   At one point I had the idea that the sub-submenu would open with a center "Normal" option next to the pointer, with all of the slow speeds located upwards and all of the fast speeds located downwards. But that didn't seem to be easily doable in Qt menus (I'm certain it's _possible_, but it would probably involve overriding the default menu placement code, and I'm not sure that's the best idea, especially where menu placement may have to be adjusted to fit the entire menu onscreen.)
   Given that I've rejected that center-aligned menu layout, though, and allowed it to use the default placement, is there a better arrangement/ordering for the two sets of options?
1. Whether any of the other menus could benefit from similar changes. Most of the others have much larger menus lurking at their lower levels, which don't lend themselves to combining/flattening. But, if anyone has suggestions for how others might be rearranged, please don't hold back.

### TODO for merge
* [x] ~Get rid of 'Experimental' swapping~ (We're gonna leave that in temporarily, during testing)
* [x] Sort out Reverse / Backward terminology (pick one)
* [x] Check that as many strings as possible are unchanged from old menus (AKA "be nice to translators")
* [x] (Optional:) Use Unicode '×' for multipliers, rather than letter X? (e.g. "2×" "1/16×")
   (Yes, in a sense this directly conflicts with the previous entry. But it's likely better for the translators if it's _not_ a letter in the English/Latin alphabet.)


ferdnyc:
Sooooo, this one went over like a lead balloon. It got two thumbs-ups, but _zero_ feedback, and as a result I really have no idea what to do with that. Or it.

DylanC:
@ferdnyc - I guess from my perspective I'm happy with the normal menu. I feel having an option for a different menu style is maybe overkill. It will make users sharing instructions on how to access features more difficult as a user will have to state... "Oh that option is in MenuA>>MenuOptionC but if you have the alternative menu enabled it will be MenuD>>MenuOptionA". etc. 

If you are designing a new menu I would prefer it to just replace the default. Then the question becomes, do users want something new here or are they (like myself) happy with the current default. Really some user feedback would be more useful here.

bobcrikey:
Hello everyone, well there is one added feature to this, maybe a highlight characters (background say red) ink white, to see where you as you go up and down with the mouse pointer you will be able to see
wear the cursor is on your in say fast 1/2 1/4 /1/8 , slow etc

cheers
 bob

MisterTeeXD:
That looks very nice

N3WWN:
I'm in favor of the minimized levels in the menus.  I may just be clunky with the mouse, but I often times end up mousing over something else when trying to get to the next level that just opened up and have to reopen the tree from further up the branch again.

N3WWN:
@DylanC wrote:
> If you are designing a new menu I would prefer it to just replace the default

I agree with this, too.  An action to enable/disable it for a dev/test/experimental/power-user build is good, but I feel that a stable release should have a single presentation for the reasons that @DylanC mentioned.

ferdnyc:
@bobcrikey :
> Hello everyone, well there is one added feature to this, maybe a highlight characters (background say red) ink white, to see where you as you go up and down with the mouse pointer you will be able to see
> wear the cursor is on your in say fast 1/2 1/4 /1/8 , slow etc

Thanks bob, though I'm not sure I _entirely_ understand your suggestion. Do you simply mean a mouse highlight for the current selection? That's already present, the pointer will be followed in blue, same as the previous menu highlights you see — the only reason a highlight isn't showing in that screenshot is that I've got the pointer over one of the section labels ("Slow") which isn't a selectable entry. But if I were to move the mouse down you'd see this:

screenshot from 2018-07-25 10-32-33

The blue highlight would follow the pointer up and down the selectable options.

Apologies if I'm misunderstanding you, though.

ferdnyc:
@N3WWN : (Welcome back, BTW!)

ferdnyc:
Oh, and on this question from @DylanC :
> Then the question becomes, do users want something new here?.. or are they (like myself) happy with the current default menu?

I mean, I'll refer you to #1587 which was the inspiration for this, but... generally, no, I don't get the impression that users are happy with the menus. I don't think it's _merely_ a question of **organization**, and I'm not fooling myself that this change fixes everything for all users and makes OpenShot a joy to use. I outlined what I saw as the larger issues in that bug, and this is merely a quick-fix that doesn't do anything to address those larger issues.

As I said up top, I consider this nothing more than a band-aid, but when you're bleeding all over the floor applying a band-aid is still a useful thing. Rethinking how editing is done in OpenShot completely is a major undertaking and won't happen quickly. In the meantime, anything to make the menus _less_ of a chore to access feels like it can't be a  bad thing. (Unless done wrong, which is why I'm still a bit nervous about those Time menus and would be curious how others find them.... but I don't think anyone can completely judge that by looking at a screenshot, it might **look** fine but then prove uncomfortable in actual use.)

N3WWN:
@ferdnyc wrote:
> (Welcome back, BTW!)

Don't welcome me back too soon 😜   

My opportunities to contribute to OpenShot are based on my $dayjob workload.  Lately, that workload has increased a LOT... it's very ebb and flow but not on anything I'd call a schedule.

Since I watch the OpenShot project, I get copied via email on every update and I'm trying to keep up with the emails, deleting those which I'm not involved with or are outside of my wheelhouse and keeping those that I can/should interact with in a backlog to be addresses ASAP.

When things let up a bit, I'll be able to contribute more.

Sorry for jumping in and out so much, but, sadly, it's the nature of this particular beast. 👹

ferdnyc:
@N3WWN :
> Sorry for jumping in and out so much, but, sadly, it's the nature of this particular beast.

No apologies, we're still fortunate to have you some of the time rather than none of the time!

ferdnyc:
> Since I watch the OpenShot project, I get copied via email on every update and I'm trying to keep up with the emails,

Also, you're a very brave soul (either that, or very foolish) — I only get notifications for the issues I've either  participated in or been tagged in, and even _that_ can be overwhelming.

DylanC:
@ferdnyc - Sounds like most people do like the new menu structure. Do you want to get this in without the experimental checkbox piece in Preferences?

ferdnyc:
> Do you want to get this in without the experimental checkbox piece in Preferences?

That, and I need to fix up a couple of other things (see checklist added to OP) . But sure. I'll clean it up and get it ready.

ferdnyc:
I was also hoping to hear from @bobcrikey about his suggestion above, and whether my reply answered it. Bob?

ferdnyc:
@DylanC 
> Sounds like most people do like the new menu structure.

I mean... we _say_ that, but also that's feedback from people who haven't **used** the new menu structure. They've looked at screenshots. Which I still feel can't really give an accurate impression.

So, let me make an alternate proposal: what do you think about getting it in **with** the Experimental switching? So that people who want to can actually _test_ it via the Daily Builds, and compare it to the old menus using the switch. Then, if they **really do** like it, in practice after kicking the tires, before the next release the switch comes out and the redesign is the new, default/only Clip menu? (Or, if they hate it, we either tweak further or go back to the old menus.)

(The rest of my checklist still applies, so either way it's not quite ready for merge yet. But I'll work on that today.)

ferdnyc:
> people who haven't used the new menu structure. They've looked at screenshots. Which I still feel can't really give an accurate impression.

(I'm still _super_ nervous about swapping Forward/Backward and Slow/Fast in the Time menu progression, and even more nervous about flipping the Slow section upside down, is what this comes down to. I'm not at all sure that's a good idea, or that "it's a slowest-to-fastest progression" is sufficient justification for the new ordering. The former will probably just take some getting used to, the latter may be actively bad ergonomically.)

DylanC:
@ferdnyc - Maybe add it with the option in Preferences then. We can gather feedback on what people prefer.

ferdnyc:
I think this is ready to go now. I did change the speed-change menus — experimental only — to use '×' as the multiplier symbol for speed changes, so the Backward menu looks like this now:

screenshot from 2018-07-27 09-33-31

(And I did standardize on Backward everywhere, because that's how the existing translations have it.)

Also, I am discovering that I have **No. Idea.** how OpenShot's `Time` _Parameter_ actually works. I would've expected it to contain a rate factor for the Clip playback, but it looks like it _actually_ contains... frame numbers? And the playback rate (which can only be _directly_ set in the Clip menu?) is then _implied_ by the rate at which the interpolated frame numbers are coming out, compared to the Timeline clock? Also, because the Clip menu's Time controls reset/override the parameter setting for the entire clip, if you want to do speed-ramping tricks or even just change speeds within a given Clip, you have to do it by manually placing Time parameter keyframes with the corresponding frame numbers? ...Huh.

ferdnyc:

PR #1660, opened 2018-06-07
Use 48kHz as default audio samplerate

ferdnyc:
This PR changes the default project & settings samplerate to 48kHz, and updates every preset to that rate (if it was previously using 44.1kHz) except for two:
* `metacafe.xml`
* `picasa.xml`

I don't even know what to make of those two, so I'm loathe to touch them. But they both use `VGA NTSC` as the default profile and MP3 audio, so 44.1kHz audio is not unreasonable as a default.

For all other presets, either the format standard or the online documentation from the hosting provider recommends 48kHz audio, which will also be the rate used by the majority of input video being edited in OpenShot today. Resampling all of that audio down to (a sometimes-not-officially-supported, e.g. by YouTube) 44.1kHz no longer makes sense.


ferdnyc:
See #1344 and #1594 for some further discussion of samplerate defaults/support.

ferdnyc:
This change is, I would think, g2g post-2.4.2, and I still can't think of any reason **not** to do it. (If anyone can come up with any, please speak up!)

DylanC:
@ferdnyc - I'm happy to put this in. Makes sense as a follow up PR from the AC3 change.

PR #1740, opened 2018-06-28
Adjust clip movement and slicing to the nearest frame position

jonoomph:
Adjust clip movement and slicing to the nearest frame position (based on project framerate)... to prevent placing clips and cuts in between frames, which can have strange results.


sknaumov:
These changes leaded to the bug #2231

PR #1759, opened 2018-07-01
Misc. typos

luzpaz:
Found via `codespell -q 3`

### DIFF
--- a/installer/build-server.py
+++ b/installer/build-server.py
@@ -266,10 +266,10 @@ try:
             github_release = get_release(repo, "daily")
             needs_upload = False
 
-    # Output git desription
+    # Output git description
     output("git description of openshot-qt-git: %s" % openshot_qt_git_desc)
 
-    # Output git desription
+    # Output git description
     output("git description of openshot-qt-git: %s" % openshot_qt_git_desc)
 
     # Detect version number from git description
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -1088,8 +1088,8 @@ $scope.SetTrackLabel = function (label){
  	return transition_size;
  };
  
- // Search through clips and transitions to find the closest element within a given threashold
- $scope.GetNearbyPosition = function(pixel_positions, threashold, ignore_ids){
+ // Search through clips and transitions to find the closest element within a given threshold
+ $scope.GetNearbyPosition = function(pixel_positions, threshold, ignore_ids){
 	// init some vars
 	var smallest_diff = 900.0;
 	var smallest_abs_diff = 900.0;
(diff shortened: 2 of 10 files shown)


src/timeline/js/controllers.js
```diff
@@ -1088,8 +1088,8 @@ $scope.SetTrackLabel = function (label){
  	return transition_size;
  };
  
- // Search through clips and transitions to find the closest element within a given threashold
- $scope.GetNearbyPosition = function(pixel_positions, threashold, ignore_ids){
+ // Search through clips and transitions to find the closest element within a given threshold
+ $scope.GetNearbyPosition = function(pixel_positions, threshold, ignore_ids){
```
luzpaz: source change, please review

src/timeline/js/controllers.js
```diff
@@ -1150,7 +1150,7 @@ $scope.SetTrackLabel = function (label){
 			var abs_diff = Math.abs(diff);
 			
 			// Check if this clip is nearby
-			if (abs_diff < smallest_abs_diff && abs_diff <= threashold) {
+			if (abs_diff < smallest_abs_diff && abs_diff <= threshold) {
```
luzpaz: source change, please review

installer/build-server.py
```diff
@@ -266,10 +266,10 @@ def parse_version_info(version_path):
             github_release = get_release(repo, "daily")
             needs_upload = False
 
-    # Output git desription
+    # Output git description
     output("git description of openshot-qt-git: %s" % openshot_qt_git_desc)
 
-    # Output git desription
+    # Output git description
```
luzpaz: any reason why lines 269 and 272 are exactly the same ?

peanutbutterandcrackers:
Hey there! Thanks for the PR. Hopefully the devs will be able to see and give the Green Light. Hope you do continue to submit some more PRs!

installer/build-server.py
DylanC: Copy and paste? :)

luzpaz: @DylanC I was just wondering if there was a redundancy issue redundancy :wink:

ferdnyc: There's plenty of copy-paste comment duplication (not to mention _wrong_ comments, because the comments weren't edited after being copied, but the code was) in the code.

ferdnyc: The bigger question is why lines 270 and 273 are the same. Because it's not just the comment that appears twice, it's the _code_. Which means it outputs the same line twice for absolutely no reason.

src/timeline/js/controllers.js
ferdnyc: +1, this gets both occurrences of the "threashold" variable, which is only used at these two points in the code.

src/timeline/js/controllers.js
ferdnyc: Corresponding variable invocation to previous.

installer/build-server.py
luzpaz: right, @ferdnyc shall I remove one of them ?

ferdnyc: *shrug* Your call. It's the build script, not part of the OpenShot source, so it really only affects @jonoomph . But I can't see any reason that would be there twice, other than pure accident.

DylanC:
@ferdnyc - LGTM too. Thanks @luzpaz 👍

PR #1773, opened 2018-07-03
Remove ActionGroup from Effects pane

ferdnyc:
The Effects pane currently contains an ActionGroup containing "Show All", "Video", and "Audio" filter buttons.

image

There are no Audio effects, making this a pointless waste of space (and pushing the clear button for the Filter input into an overflow menu, when the pane is sized to a comfortable width).

This change removes the ActionGroup, and updates the stored Effects pane width in the Advanced view to a narrower size that comfortably holds two columns of Effects with no wasted space.

screenshot from 2018-05-29 22-57-53


peanutbutterandcrackers:
Since there is the possibility of needing the action group soon, I wonder if it'd be better if this PR commented out the lines of code rather than delete them? o.O Just a humble opinion.

ferdnyc:
\*shrug\* Obviously, I disagree. There's a _possibility_ of needing lots of code "someday", and git's history is easy enough to work with so the code is never _really_ gone. Commented-out code in your source is never a good sign.

The grouping buttons aren't _necessary_ simply because there are Audio effects added, BTW. Only if the list needs to be filterable down to just audio or just video effects. Given how few effects there are total, right now, it's debatable whether that really **is** necessary.

Also, the previous design is... well... terrible (with the 10-character wide "Filter" box and the clear button that requires opening an overflow menu to access it), so even if audio effects are added I'd hope the removed code would be rethought instead of just restored.

(Heck, if it were me I'd probably drop the text filtering, which is **really** unnecessary, entirely. It could be replaced with the grouping buttons if that's deemed useful.)

peanutbutterandcrackers:
Hmm... I see. Now that you point that out, I can't help but agree with you - like all the other time. lol :+1:

ferdnyc:
Actually, last time I was in the code I noticed that there are more bits of this that I missed (the signal bindings for the controls, which are set up elsewhere in the code) — I should really update the PR to deal with those as well. I'll try to get around to that ASAP. It makes sense to pull it all out in one go, so that it's easier to retrieve if anyone does want to re-enable any parts of this.

DylanC:
@ferdnyc - I'm not sure how I feel about this one. I kinda feel like we are just taking something out that we will end up putting back in later on. (Although later on could be 2 years later 😆)

peanutbutterandcrackers:
@DylanC - Evidently, even the current code could be improved upon (that is why the PR doesn't comment out the code but actually removes it). :)

ferdnyc:
\*shrug\* I'm not married to it. I do think the current UX is bad, with the overly-narrow toolbar atop Effects, but since I'm not putting myself on the hook for rewriting it far be it from me to rip out what technically works but serves no purpose. (This stuff _is_ actual interface clutter.)

(However, the reason the PR doesn't comment out the code is that I was raised[-as-a-programmer] to believe that commented-out code in your released source means that you didn't _finish_ the code, so you shouldn't **BE** releasing it yet. Either the code serves a purpose, in which case it should be executed, or it doesn't and it shouldn't be there. Functions that don't get called "yet", code paths that never get traversed but are "there for later", commented-out code that probably won't work again when it's _uncommented_ because there have been changes in the surrounding structure... those are all bad things.)

DylanC:
@ferdnyc - Maybe a better fix here would be to add an Audio effect, so we can justify having the Action Group tab. 😉

peanutbutterandcrackers:
Or, even better, let's merge this one in (as working on audio effects rather than stability wouldn't be right choice for now)... and let's get all the suggestions from @ferdnyc, document it somewhere (gist?) and then work on it again later on. He is right. He mostly is right.

DylanC:
@peanutbutterandcrackers @ferdnyc - Do any of us even know the requirements for creating an Audio effect? Its hard to know when we don't have any so far.

peanutbutterandcrackers:
Sorry... I don't. I am quite a n00b, actually...

@DylanC - Perhaps we should merge this in, Cap'ain. @ferdnyc has always made good contributions to the project and his suggestion regarding the ActionGroup being a clutter currently is indeed correct. If there will be an audio effect in 2.4.3, pretty sure we'll get some contribution from him regarding that to...
The 'filter' option is still there, so I think this is pretty good to go.

DylanC:
@peanutbutterandcrackers - It does look better for sure. @ferdnyc - I'm happy enough to add this, doesn't look like much to add back in the action group later on. In the meantime maybe lets have a think about the Audio effects, after we focus on crash issues and bugs. 👍

peanutbutterandcrackers:
Awesome, Cap'ain! Thank you! :+1:

ferdnyc:
Oh, crap, you actually merged this huh? Hrm.

When it looked like it was going the other way I put it out of my mind, but I never actually followed up on this:

> Actually, last time I was in the code I noticed that there are more bits of this that I missed (the signal bindings for the controls, which are set up elsewhere in the code) — I should really update the PR to deal with those as well. I'll try to get around to that ASAP. It makes sense to pull it all out in one go, so that it's easier to retrieve if anyone does want to re-enable any parts of this.

So, I went looking to clean that up just now, and accidentally discovered the _perfect_ illustration of the exact argument I was making with this PR!

Even though the filter widgets are removed from the Effects pane, they're still present in the UI definition file, which is good because they're also still used in parts of the code. Once I removed the definitions from `main_window.ui`, I was able to trace the remaining consumers by seeing what that broke.

* **They're used in `windows/models/effects_model.py`**
   
   No surprise there, the code for displaying the effects list has to respond to selections made with the filtering controls. Easy enough to pull that out.

* **They're used in `windows/models/blender_model.ui`**
   
   ### Wait, _**WHAT??!!?**_
   
   It's true. The filters on the Effects pane are used in the Animated Titles window. Why? Well, they're not _supposed_ to be. It appears to have been a complete accident.

   Turns out, the code from `effects_model.py` was cut-and-pasted into `blender_model.py` (Actually, I'm guessing the file was copied in its entirety and then edited.) And when it was edited into `blender_model.py`, the connections to the Effects filters were accidentally left in.

   Which means that, in OpenShot 2.4.2 release, if you set the Effects pane filtering to show Audio effects and then select Title > Animated Title from the menu, the Animated Titles window comes up blank and non-functional. Same thing if you have text typed in to the text-filter field in the Effects pane. Any sort of filtering applied to Effects has the side effect of completely breaking the Animated Titles editor. 

   Not a single one of us ever noticed.

## ...And THAT'S why you don't leave unused code paths hanging around in your source!

(He says, smugly. I honestly didn't _plan_ any of this, but I will take the vindication! :grinning: )

Since I'm guessing I can't still push more changes to this branch, I will create a new branch and submit a new PR to clean this up the rest of the way.

DylanC:
@ferdnyc - I was under the impression that you and @peanutbutterandcrackers really wanted this merged in. :laughing: Oh dear. Well if you submit a follow up PR, I'm sure we can get that in.

ferdnyc:
Well, I _definitely_ want it merged, now that I know the filters will break the Animated Titles editor if they're used! :laughing: Even if the filters weren't removed, we'd have to fix that bug. Simply removing them is the easiest way, and probably what I would've proposed anyway if someone had actually found the bug before I removed them.

I'll have a follow-up PR ready within 10 or 15 minutes.

PR #1776, opened 2018-07-03
Add Makefile to compile .qrc file

ferdnyc:
I added a simple `Makefile` in `src/images/` so that `openshot_rc.py`
can be recompiled from `openshot.qrc` with just a simple `make` command.
Make will also handle recompiling only when the output file is older
than the input `.qrc`.

Also updated `src/images/README` accordingly.


DylanC:
@ferdnyc - Neat change to have in the src. Thanks.

Issue #1777, opened 2018-07-03
AppImage runs spew QFactoryLoader log messages


ferdnyc:
**Describe the bug**
Runs of OpenShot from the AppImage are displaying incredibly verbose `QFactoryLoader` messages on the console. I _believe_ this is because `usr/bin/openshot-qt` in the AppImage is setting `QT_DEBUG_PLUGINS=1`

**System Details:**
 - Operating System / Distro: Fedora 28
 - OpenShot Version 2.4.2 from `OpenShot-v2.4.2-x86_64.AppImage`

**To Reproduce**
1. Execute `OpenShot-v2.4.2-x86_64.AppImage` in a terminal window
2. Respond to the "Create a shortcut?" question
3. Read console output

**Expected behavior**
Only pertinent log messages are output

**Actual behavior**
```console
$ ./OpenShot-v2.4.2-x86_64.AppImage
Loaded modules from current directory: /tmp/.mount_tPV6nB/usr/bin
      launch:INFO ------------------------------------------------
      launch:INFO    OpenShot (version 2.4.2)
      launch:INFO ------------------------------------------------
QFactoryLoader::QFactoryLoader() checking directory path "/tmp/.mount_tPV6nB/usr/bin/platforms" ... 
QFactoryLoader::QFactoryLoader() looking at "/tmp/.mount_tPV6nB/usr/bin/platforms/libqeglfs.so" 
Found metadata in lib /tmp/.mount_tPV6nB/usr/bin/platforms/libqeglfs.so, metadata=
{
    "IID": "org.qt-project.Qt.QPA.QPlatformIntegrationFactoryInterface.5.2",
    "MetaData": {
        "Keys": [
            "eglfs"
        ]
    },
    "className": "QEglFSIntegrationPlugin",
    "debug": false,
    "version": 328193
}


Got keys from plugin meta data ("eglfs") 
QFactoryLoader::QFactoryLoader() looking at "/tmp/.mount_tPV6nB/usr/bin/platforms/libqkms.so" 
Found metadata in lib /tmp/.mount_tPV6nB/usr/bin/platforms/libqkms.so, metadata=
{
    "IID": "org.qt-project.Qt.QPA.QPlatformIntegrationFactoryInterface.5.2",
    "MetaData": {
        "Keys": [
            "kms"
        ]
    },
    "className": "QKmsIntegrationPlugin",
    "debug": false,
    "version": 328193
}


Got keys from plugin meta data ("kms") 
[...]

PR #1778, opened 2018-07-03
launch-linux.sh: Don't set QT_DEBUG_PLUGINS=1

ferdnyc:
Addresses #1777 ?

Haven't tested this since I can't build AppImages, but based on my experiments with the launcher it would solve the problem.


DylanC:
@ferdnyc Thanks. I do think reducing the output is important. Should make it easier to see more important messages.

Issue #1764, opened 2018-07-02
Keyboard shotcut to slice just the selected clip


timlau:
**Describe the bug**
Keyboard shotcuts Ctrl-L & Ctrl-J slice all tracks instead of just the selected one.
(Right click menu -> Slice -> Keep Left etc. does the right thing)

**System Details (please complete the following information):**
 - Fedora 28
 - OpenShot 2.4.2 from flathub.org
https://flathub.org/apps/details/org.openshot.OpenShot

**To Reproduce**
1. Add 2 clips on above each other
2. Select a place in the middle to cut
3. Select on of the clips
4. Press Ctrl-J or Ctrl-L

**Expected behavior**
Only selected clip is sliced

N3WWN:
@timlau - May I ask what tool you used to create this animated GIF showing the issue that you're reporting?  That looks very useful! 😉 

Now, regarding the issue that you've reported, <kbd>CTRL</kbd>+<kbd>J</kbd> and <kbd>CTRL</kbd>+<kbd>L</kbd> are listed in the `Keyboard` section of the `Preferences` as `Slice All: Keep Right Side` and `Slice All: Keep Left Side`, respectively.

There currently are no keyboard shortcuts for `Slice Selected...`

ferdnyc:
As @N3WWN says, <kbd>Ctrl</kbd>-<kbd>L</kbd> and <kbd>Ctrl</kbd>-<kbd>J</kbd> are documented as "Slice All: Keep {Left/Right} Side". Ditto <kbd>Ctrl</kbd>-<kbd>K</kbd> for "Slice All: Keep Both Sides". And they worked that way in 2.4.1 as well, so I think the `2.4.2-issue` label is erroneous. If anything, this is a feature request, as to my knowledge there never was a "Slice Selected" keyboard binding.

Right-clicking on a Clip invokes the Slice function on _that Clip_, same as using the Razor Mode tool. Which is different from "Slice All".

peanutbutterandcrackers:
@N3WWN - Precisely what I was thinking! Pretty neat tool! I'd like to know what it is, too.

@ferdnyc - Fixed it, sir! :)

timlau:
@N3WWN the gif is a screen cast, process with the chop filter in shotcut and rendered to a GIF (1024x720, 10 FPS).(keymon is used to show the keypresses)
I can see that that in the keyboard shotcut in preferences Ctrl-L & J is assigned to slice all, but
[...]

PR #1780, opened 2018-07-03
Move Slice All shortcuts to correct menu

ferdnyc:
The keyboard combos for "Slice All" were displaying in the Slice
submenu on the Clip (under certain conditions), when they should be
on the Playhead "Slice All" submenu.

Addresses #1764


DylanC:
@ferdnyc - Another good change. Thank you! (Good sir)

PR #1836, opened 2018-07-11
Add "Ask a Question" template

ferdnyc:
Doesn't actually change the other two templates, but once again Windows line endings crept in, so this fixes them back to UNIX format.


peanutbutterandcrackers:
Thanks! Merged! :+1:

PR #1849, opened 2018-07-13
Finish cleanup of Effects filtering

ferdnyc:
Also fixes a bug where the Animated Titles window would stop functioning when filtering was used in the Effects pane. (See #1773.)


DylanC:
@ferdnyc - Thank you. (Good sir)

PR #1859, opened 2018-07-16
Correct variable name typo to initialize properly

ferdnyc:
Both `clip.js` and `transition.js` were using the variable `dragLoc` uninitialized, because the top of each file contained the following initialization:

```js
var dragLog = null
```

### DIFF
--- a/src/timeline/js/directives/clip.js
+++ b/src/timeline/js/directives/clip.js
@@ -40,7 +40,7 @@ var track_container_height = -1;
 // 1: can be dragged
 // 2: can be resized
 // 3: class change when hovered over
-var dragLog = null;
+var dragLoc = null;
 
 App.directive('tlClip', function($timeout){
 	return {
--- a/src/timeline/js/directives/transition.js
+++ b/src/timeline/js/directives/transition.js
@@ -40,7 +40,7 @@ var track_container_height = -1;
 // 1: can be dragged
 // 2: can be resized
 // 3: class change when hovered over
-var dragLog = null;
+var dragLoc = null;
 
 App.directive('tlTransition', function(){
 	return {


DylanC:
@ferdnyc - Seems like a pretty harmless fix.

ferdnyc:
@DylanC : Indeed. It was _prooobably_ a harmless typo, as well (certainly doesn't seem like it was harming anything), can't say for sure because TBH I didn't even bother to check whether `dragLoc` was ever being used uninitialized, before the fix... it's possible it wasn't. I just figured, there's no point in initializing a variable that's never used, and if the code bothers to initialize variables at all then it might as well at least initialize the _right_ variables.

DylanC:
@ferdnyc - I'll be looking into this section of code anyway. JS is something I do daily, so I should be helping the parts of the project relevant to my skillset. Still unsure is using JS here a great idea. I had thoughts of a C++ timeline today or Python but it seems like a huge project to do a replacement on what has been implemented already.

ferdnyc:
> Still unsure is using JS here a great idea. I had thoughts of a C++ timeline today or Python but it seems like a huge project to do a replacement on what has been implemented already.

\*nod\* Tho the question then becomes, in C++ or Python with what toolkit? Qt I guess is the obvious answer because it's what the rest of the code is built in, but TBH building the Timeline in traditional, QWidget-based Qt... does sound like kind of a chore. I definitely can see reasons for going down the Angular+jQuery JS path, rather than attempting a QWidget-based implementation. I can't imagine that'd be very enjoyable at all.

One thing that I think might be really interesting to explore, though, is the idea of a Timeline built using a more _modern_ Qt approach. Interface building in Qt today isn't just the QObject/QWidget-based toolkit, they also offer QML for markup-based interface design. Which I imagine might offer a fairly painless migration/conversion from the current Angular-based Timeline... in fact, like everything else this decade, QML _also_ has a JavaScript API to access pretty much everything.

I know you said you're not sure JavaScript is the right approach... but right now the Timeline implementation is external to the entire rest of the interface (and they're totally disconnected from each other, as a result) so both sides wind up jumping through a lot of hoops to achieve even the current (low) 
level of integration. Whereas the impression I've gotten is that a QML Timeline would be a fully integrated component right from the start. (Hopefully that'd mean far less sprawling JavaScript code, and less of it too.)

I know _nothing_ about QML other than what I've read, but Qt seem to treat it as The Future™ most of the time... though how much of that is justified, and how much is just marketing telling people what they think they want to hear, I know too little about QML to even hazard a guess.

[Wikipedia says](https://en.wikipedia.org/wiki/QML#Qt/C++_integration)...
> ### Qt/C++ integration
> QML does not need Qt/C++ knowledge to use and it can be easily extended via Qt.[28][29] Any C++ class derived from QObject can be easily registered as a type which can then be instantiated in QML.
>
> #### Familiar concepts
> QML provides direct access to the following concepts from Qt:
> * QObject signals – can trigger callbacks in JavaScript
> * QObject slots – available as functions to call in JavaScript
> * QObject properties – available as variables in JavaScript, and for bindings
> * QWindow – Window creates a QML scene in a window
> * Q*Model – used directly in data binding (e.g. QAbstractItemModel)[30][31][32]

Issue #1839, opened 2018-07-11
Click on the timeline doesn't move playing position sometimes


singalen:
Seems to mostly work when I click **and drag** the mouse a little.

**System Details:**
 - Operating System / Distro: MacOS 10.13
 - OpenShot Version: 2.4.2
![openshot-timeline](https://user-images.githubusercontent.com/151199/42602629-29985d20-851f-11e8-83f9-e89333b4ef7f.gif)

peanutbutterandcrackers:
Can't reproduce on my [Elementary OS](https://elementary.io) machine. Perhaps it is a mac related issue? o.O

DylanC:
@peanutbutterandcrackers - No problem. There is a bunch of Mac issues I will need to test in a testing session. :)

N3WWN:
It seems to me that the "trick" is clicking on the timeline above the minor graduations (the top of the short lines).  Clicking below the top of the minor graduations results is no playhead movement for me, too.  Even if I click and drag, if the cursor is below the top of the minor graduations, it won't work.

peanutbutterandcrackers:
@singalen - Does @N3WWN's suggestion work for you, sir?

@N3WWN - Is the action a little less responsive on a mac, and thus in need of some work? o.O

N3WWN:
@peanutbutterandcrackers - I don't think it's any different on a Mac.  

I noticed in @singalen 's animated GIF that it was the vertical location of the cursor that determined if the click caused movement or not.  I replicated in Linux, FYI.

If there's no reason to NOT extend the clickable location to move the playhead, I can look at increasing it to include more of the lower portion.

peanutbutterandcrackers:
@N3WWN - I see. I wonder if there is any reason NOT to increase it. But perhaps @DylanC or @ferdnyc  might have some insight into this? I don't really see why not...

ferdnyc:
The strange thing about this is that it's _supposed_ to be clickable all the way down to the bottom of the ruler marks. And it is, in the timeline JS debugger. So, something's being set in Qt that's shrinking the clickable area of the ruler.

DylanC:
@ferdnyc - I don't think something is shrinking the clickable area. I feel the click events are not registering sometimes. It happens at random intervals from my testing. A click on the ruler would just do nothing, clicking the same area again later works.

ferdnyc:
[...]

Issue #1454, opened 2018-03-31
Attempting to extend a video's duration (a strange behavior) produces strange behavior.


Just-Curious:
**System Details:**
* Operating System / Distro: ?  Windows 10 Pro, Version 1709, Build 16299.309 
* OpenShot Version: ? 2.4.1

Originally added to topic https://github.com/OpenShot/openshot-qt/issues/894#issuecomment-377658191.  That topic has been closed.

If the right hand edge of a timeline video is dragged to the right (impossible), the display length shrinks. Doing this repeatedly leaves a "blank" area on the timeline that, in fact, contains the "missing" video. Dragging a video to the timeline and allowing it to "snap in", will result in it snapping to the correct ending of the original video. Adjusting the time scale corrects the display.

https://youtu.be/JcSmhuKoaCA

peanutbutterandcrackers:
Awesome! Thank you for this well-documented report, good sir!

@ferdnyc - I am pretty sure you'll be able to fix this in an afternoon, good sir! :)

Just-Curious:
@peanutbutterandcrackers
Here's a well-documented report that you may also be able to have someone fix some afternoon:
https://github.com/OpenShot/openshot-qt/issues/362

peanutbutterandcrackers:
Well, lets hope @ferdnyc will have some free time soon. OpenShot is a community effort. We are all squireling time away to help out with what we can. And @ferdnyc is a brilliant programmer who has made quite a lot of Pull Requests fixing and enhancing OpenShot in the last couple of months. But it seems that he's been busy of late. Perhaps we'll soon see the return of the super-contributor. :)

ferdnyc:
Yeah, I had noticed this one a while ago, but then lost track of it. Thanks for filing the report, it's good to have it in the system.

As you discovered, it's purely a timeline-rendering issue — the clip still has the same duration, and in reality is still the same length on the timeline, it's just being drawn slightly too short. Dragging the clip to another track, or another position on the same track, will cause it to be drawn at the proper length, as will changing the zoom level. Anything that refreshes the timeline rendering, basically. And all interactions (snapping, slicing, etc.) will continue to function properly over the full length, even the hidden portion.

Because it's purely a visual thing (the issue doesn't affect the software's function in any way, only the UI, and can easily be worked around) it's not the highest-priority issue. But I'll try and take a look, it _might_ be simple to fix and it would be nice to whack it.

[...]

Issue #1425, opened 2018-03-27
Remove timeline clutter (lines, curves, shading, glow, colors)


qubodup:
![Clutter in OpenShot](https://user-images.githubusercontent.com/18312/38000397-9f7ed334-3227-11e8-8bbd-898b5f08e02e.png)

A - Gap between tracks wastes space.
B - Round track and clip corners make less clear where clips begin/start, where to click them, indicate roundness (fades) where there is none.
C - Playhead marker is decorative rather than functional (shaded, not immediately visible on a glance).
D - Track border color grabs attention even though they are the least interactive thing in the timeline.
E - glow (gradient?) behind track name makes it unreadable. Gradient behind tracks makes it less clear where they start/end on first glance.

I suggest to update the timeline with a flat, non-rounded look with solid dark grey background, grey 1px track borders, a solid light blue playhead marker and line.

Samples of less cluttered timelines in video editors:

![Premiere](https://user-images.githubusercontent.com/18312/38000481-0db5c650-3228-11e8-8633-d1025c884164.png)

![Filmora](https://user-images.githubusercontent.com/18312/38000678-0844391c-3229-11e8-96fd-8f8b19212c75.png)

philiplb:
I like this a lot.

peanutbutterandcrackers:
@ferdnyc @DylanC - While I agree with this ticket, I wonder if the rounded-ness is there to be/appear to be more user friendly. I wonder if it is possible to have another theme for OpenShot that will transform it to look a lot more like these uncluttered examples. Or maybe a view mode - 'Professional' or sth.

philiplb:
I see one drawback of the rounded-ness: It renders some UI-Space useless.

ferdnyc:
I think we should start by getting our terminology straight. Nobody could ever accuse the OpenShot timeline of being "cluttered", there's barely anything **ON** it. Especially compared to that first screenshot, which is a _heavily_ cluttered mess of controls. What the OpenShot timeline _could_ be accused of being is "noisy". That's not unfair.

At the same time, the "here's what other programs do" list is a bit of a cherry-pick. 

Here's PowerDirector, complete with rounded timeline elements (nobody thinks they're fades), and a playhead with gradient shading on it (that also obscures the timecodes when it passes them, something I also noticed in the second screenshot above; I just find that choice _bizarre_):

[...]

Issue #1383, opened 2018-03-15
Dragging a marker adds its icon to the timeline as a clip


ferdnyc:
**System Details:**
* Operating System / Distro: Fedora Linux x86_64
* OpenShot Version: 2.4.1

**Issue Description and steps to reproduce:**

OK, this one's actually funny. :laughing: 
1. Add a clip to the timeline.
1. Click "Add Marker".
1. (Move the playhead out of the way of the marker.)
1. Attempt to reposition the marker by grabbing and dragging it.
1. The marker's icon, `blue.png`, is added to the timeline as an image clip.

Before:
![screenshot from 2018-03-14 22-47-12](https://user-images.githubusercontent.com/538020/37441980-54a912ea-27da-11e8-8969-cf950d4dcb1f.png)

After:
![screenshot from 2018-03-14 22-47-21](https://user-images.githubusercontent.com/538020/37441983-57e68e88-27da-11e8-94a4-a3bb2da32560.png)

Since markers _aren't_ repositionable, they probably shouldn't be draggable. (Or, dragging them should reposition them.)

peanutbutterandcrackers:
This one's funny, indeed. I did try dragging other icons to see if those are also draggable. They weren't. Haha

N3WWN:
"That's not a bug!  That's a feature!"  😁 

Now, if we can find some way to promote this easter egg as something useful, there'll be no need to fix it!

DylanC:
@ferdnyc - I don't think we can convince anyone else its not a bug. 😆 Something to be fixed but I would say low priority.

DylanC:
@ferdnyc - Man... I can tell you so far this bug is gonna be painful! :rofl: It doesn't even happen in the browser but only when deployed to Qt.

DylanC:
@ferdnyc - Managed to fix this actually. It was easy enough in the end. :+1:

ferdnyc:
> It doesn't even happen in the browser but only when deployed to Qt.

**RIGHT?** That was one of the first things I noticed. Though later I realized that the mouse pointer did turn into a hand when you _tried_ to drag, which felt like a clue.

> Managed to fix this actually.

`\o/` I knew it had to be something stupid. I'm on my phone now, but I look forward to seeing the changes.

(As a reminder, if you put `Fixes #1383` in a PR text, GitHub will auto close this issue on merge. I've found that quite handy. Not sure how it works with PRs that fix multiple issues, tho.)

[...]

PR #1864, opened 2018-07-16
Timeline cleanup and performance

DylanC:
Attempting to upgrade Angular safely and remove any dud code. Also hope to make some performance improvements.


DylanC:
> Still unsure is using JS here a great idea. I had thoughts of a C++ timeline today or Python but it seems like a huge project to do a replacement on what has been implemented already.

*nod* Tho the question then becomes, in C++ or Python with what toolkit? Qt I guess is the obvious answer because it's what the rest of the code is built in, but TBH building the Timeline in traditional, QWidget-based Qt... does sound like kind of a chore. I definitely can see reasons for going down the Angular+jQuery JS path, rather than attempting a QWidget-based implementation. I can't imagine that'd be very enjoyable at all.
Comment from ferdnyc on #1859 -

One thing that I think might be really interesting to explore, though, is the idea of a Timeline built using a more modern Qt approach. Interface building in Qt today isn't just the QObject/QWidget-based toolkit, they also offer QML for markup-based interface design. Which I imagine might offer a fairly painless migration/conversion from the current Angular-based Timeline... in fact, like everything else this decade, QML also has a JavaScript API to access pretty much everything.

I know you said you're not sure JavaScript is the right approach... but right now the Timeline implementation is external to the entire rest of the interface (and they're totally disconnected from each other, as a result) so both sides wind up jumping through a lot of hoops to achieve even the current (low)
level of integration. Whereas the impression I've gotten is that a QML Timeline would be a fully integrated component right from the start. (Hopefully that'd mean far less sprawling JavaScript code, and less of it too.)

I know nothing about QML other than what I've read, but Qt seem to treat it as The Future™ most of the time... though how much of that is justified, and how much is just marketing telling people what they think they want to hear, I know too little about QML to even hazard a guess.

Wikipedia says...

> Qt/C++ integration
> 
>     QML does not need Qt/C++ knowledge to use and it can be easily extended via Qt.[28][29] Any C++ class derived from QObject can be easily registered as a type which can then be instantiated in QML.
>     Familiar concepts
> 
>     QML provides direct access to the following concepts from Qt:
> 
>         QObject signals – can trigger callbacks in JavaScript
>         QObject slots – available as functions to call in JavaScript
>         QObject properties – available as variables in JavaScript, and for bindings
>         QWindow – Window creates a QML scene in a window
>         Q*Model – used directly in data binding (e.g. QAbstractItemModel)[30][31][32]

DylanC:
@ferdnyc - Makes more sense to continue the conversation here instead.   

> nod Tho the question then becomes, in C++ or Python with what toolkit?

I did mean to say with Qt on this. So it would be C++ with Qt5 or PyQt5.

> TBH building the Timeline in traditional, QWidget-based Qt... does sound like kind of a chore

Yes, I was thinking of a GraphicsView widget and drawing QRects for the clips. The more I was thinking about it the more it sounded like a horrible task and something else would fare better. 

>Whereas the impression I've gotten is that a QML Timeline would be a fully integrated component right from the start. (Hopefully that'd mean far less sprawling JavaScript code, and less of it too.)

Yeah I feel maybe QML is the best fit in terms of easy coding and reducing the code to support the bridge between Qt and JS. At the moment when I try to work on the timeline code it just seems overwhelming and I always feel like starting again from scratch but that would be even more work and probably a harder task. 

If I could draw out and document the requirements of everything the timeline has to support then maybe starting a new timeline implementation would be best.

ferdnyc:
> Yes, I was thinking of a GraphicsView widget and drawing QRects for the clips. The more I was thinking about it the more it sounded like a horrible task and something else would fare better.

Yeah, I think the win there is that Angular+jQuery at least bring a scene graph and nice, highly advanced animation features. Reading more about `QGraphicsView`, it does apparently have its own scene graph (`QGraphicsScene`) and shape primitives like `QGraphicsRectItem` and `QGraphicsPathItem`, as well as `QGraphicsTransform` for animation effects, so maybe it's more capable than I realized. 

But still, as much as [some users may dislike them](/OpenShot/openshot-qt/issues/1425) :wink: ,  I have to wonder if even something as simple as drawing the clips with rounded corners would be nearly as easy in QGraphicsView as it is in the current timeline code, where all it requires is the `border-radius` attribute in the timeline CSS:
```css
.clip_top { 
  display: -webkit-flex; 
  background: -webkit-gradient(linear, left top, left bottom, color-stop(0%,rgba(255,255,255,0.20)), color-stop(100%,rgba(255,255,255,0))); 
  border-radius: 8px; 
  overflow: hidden; 
}
```
There are also a lot of little animations we take for granted. (The snapline when Clips are in alignment, the scrolling of the Playhead when the user clicks the ruler, the scrolling of the snapline from one alignment point to another, etc...)

> At the moment when I try to work on the timeline code it just seems overwhelming

I hear that. I went hunting for #1454 (the clip-resizing display bug), and I came away utterly confused  about everything that was going on in there, but without any fix for that bug. I know I _found_ the code that must contain the bug, I just couldn't find the **bug** in the code, mostly because I didn't understand half of it. But that bug even shows up in the Timeline debug interface (opening `src/timeline/index.html` in Chrome), so I know it's a JS bug.

I also went looking for my own #1383 (about the fact that Markers are draggable, and add their icon as an image clip when dragged to the Timeline), which _should_ be a really elementary thing, but I'll be damned if I could figure out why the things are considered draggable, or how to make it stop. They're _not_ draggable in the debug interface, so that makes it even weirder. Though, I do notice that the cursor changes to a hand when you _attempt_ to drag them, which feels like a clue.

P.S> [QGraphicsView vs. QT Quick / QML](http://blog.qt.io/blog/2017/01/19/should-you-be-using-qgraphicsview/) on the QT Blog. (It's actually titled "Should you still be using QGraphicsView?", but having read it I feel like a more accurate title would be "In defense of QGraphicsView", since that's how it comes off.)

DylanC:
@ferdnyc 

> There are also a lot of little animations we take for granted. (The snapline when Clips are in alignment, the scrolling of the Playhead when the user clicks the ruler, the scrolling of the snapline from one alignment point to another, etc...)

Yeah I think for now on the whole a move to something else would be too much change too soon.

> P.S> QGraphicsView vs. QT Quick / QML on the QT Blog. (It's actually titled "Should you still be using QGraphicsView?", but having read it I feel like a more accurate title would be "In defense of QGraphicsView", since that's how it comes off.)

This makes it a bit more confusing as to which is better in this case. :laughing: 

Soo.. just a quick update. I've updated Angular to the latest version and like before this has broken the dragging behavior. You can drag a clip after selecting it first but not if you just click and start dragging right away. I've seen this in my previous attempt to upgrade Angular. 

However, since we also include Jquery and Jquery UI it could possibly be something in there. I do suspect its more an Angular issue but my (fun) investigations will continue..

DylanC:
@ferdnyc - Fixing dragging wasn't bad and I feel like I've fixed this before. Next up is snapping (oh dear lord). Clip snapping seems to completely not work now but where there is a will there is a way. :rofl:

DylanC:
@ferdnyc - Is there a chance you could test this if you have time? I want someone to test and maybe find something broken before I merge this PR. Seems to work fine inside Openshot and in Chrome from my testing so far.

ferdnyc:
@DylanC Sure, I'll give it a spin in the next hour or two.

ferdnyc:
@DylanC — LGTM! Very cool. I couldn't really detect any differences from 2.4.2 release, which I gather is the goal (at least so far)? All the same functionality, all the same bugs. :rofl: 

(And I just found another one, or at least it's new to me: When you have a locked track and you're dragging clips up and down between tracks, you don't have to try very hard to get a clip to "land" on the locked track and stay there — it's _really_ still on the track it moved from, according to its Properties, it only _appears_ to be on the locked track. But looks like the drag auto-reset on a refused drop sometimes doesn't fire, or can get interrupted by new drag activity before it has a chance to restore the original position.)

But like I said, that's only new _to me_ — exact same behavior as 2.4.2 release, so nothing to do with this PR at all. Everything here looks A-OK from where I'm sitting, in both the debug timeline and the OpenShot interface. :+1:

src/timeline/js/controllers.js
```diff
@@ -1408,19 +1402,6 @@ $scope.SetTrackLabel = function (label){
...
-
-	// Enable all items
-	//$(".clip").draggable("enable")
-
-	// Disable any locked items
-	// for (layer in $scope.project.layers)
-	// {
-	// 	timeline.qt_log(layer);
-	// }
- };
```
ferdnyc: What _will_ we do without this function??

(Which reminds me, one of us should just delete `src/timeline/media/js/debug.js` entirely, as it contains equally critical code.)

src/timeline/js/controllers.js
```diff
@@ -34,9 +34,9 @@ App.controller('TimelineCtrl',function($scope) {
 	$scope.project =
     {
 	  fps: {
-		    num : 24,
-		    den : 1
-		   },
+	    num : 24,
+	    den : 1
+	   },
```
ferdnyc: Die, tabs, die!

ferdnyc: (BTW, this still doesn't quite line up, since lines 36-39 start with a single tab and then switch to spaces, while 35 and 40-44 are indented entirely with spaces. But, honestly, this file's tabs-vs-spaces insanity is... clinical, so maybe that's a whole 'nother PR of its own.)

src/timeline/js/directives/clip.js
```diff
@@ -208,9 +206,7 @@ App.directive('tlClip', function($timeout){
 						// Preview the right side of the clip
 						scope.PreviewClipFrame(scope.clip.id, new_right);
 					}
-
-				},
-
+				}
```
ferdnyc: TBH, I don't hate trailing commas on multi-line (multi-_page_) lists like this. It's not illegal, and it saves you when someone adds another member after the last one and _forgets_ to append the comma. But removing it keeps 'em honest, I suppose.

src/timeline/js/directives/clip.js
```diff
@@ -421,8 +407,9 @@ App.directive('tlMultiSelectable', function(){
 						item = findElement(scope.project.effects, "id", id);
 					}
 					
-					if (scope.Qt)
+					if (scope.Qt) {
```
ferdnyc: Ugh, _THANK YOU_. Conditional blocks without enclosing braces are just bug magnets, they're practically _taunting_ anyone who looks at the code. "Try to add another instruction to me, screw up the program logic!"

src/timeline/js/controllers.js
DylanC: @ferdnyc - Yup, I considered deleting that multiple times already. Might as well just get rid of it.

src/timeline/js/controllers.js
DylanC: @ferdnyc - No worries, I will fix it before merging if possible.

src/timeline/js/directives/clip.js
DylanC: @ferdnyc - I'd only leave it there if its very likely we will be adding something soon. IMHO.

src/timeline/js/directives/clip.js
DylanC: @ferdnyc - Yep, its a personal preference of my own too. I HATE when people do single line if statements. Its just not consistent with the rest of them and yeah it can be a magnet for future bugs.

DylanC:
> LGTM! Very cool. I couldn't really detect any differences from 2.4.2 release, which I gather is the goal (at least so far)? All the same functionality, all the same bugs

Good to hear. Tbh yeah I think that is all I'm aiming for with this one. Its to make important upgrades for the timeline and just remove some cruft. Should give us a more solid foundation for future changes. I'm planning to tackle some individual bugs with separate PR's.

src/timeline/js/controllers.js
DylanC: @ferdnyc - That debug file has been removed btw.

DylanC:
@ferdnyc - Getting this change in. I'm planning to tackle some timeline bugs in my next PR's. Do you know of any issue numbers worth getting attention?

ferdnyc:
The two I mentioned, #1454 and #1383, are both definitely Timeline code bugs, so that's a start. 

There's also #1839, which raised an issue with the ruler that _isn't_ present in the debugger, so that's something going on with OpenShot embedding the Timeline canvas... but it's still related. 

Oh, and also on the "straddling the Timeline JS and OpenShot's Python" front, in the course of discussing one issue [the conversation got around to the Timeline's current keyboard-/multi-selection functionality](https://github.com/OpenShot/openshot-qt/issues/1425#issuecomment-401580034) which is, to put it mildly, a bit lacking and weird, so that could really use sanity-checking and standardizing.

Oh, and the super-ambitious goal, of course: All those various situations where the Timeline should be autoscrolling to follow the Playhead, but instead just lets it wander out of view.

I'd also throw in, though it wasn't the original focus of #1587, the things that could easily be simple clickable controls on the Timeline, but are currently accessible only from the context menu. The obvious low-hanging fruit on this front is the Locked Track control. When you lock a track, a padlock symbol gets added to the track label section, but the only way to actually perform the lock/unlock operation is via the menu. It seems like a no-brainer that the padlock should be clickable to unlock the track, at a minimum. Or, even better, _always_ visible, and clickable to toggle between locked and unlocked states, so you don't need the context menu at all just to perform the simple act of flipping the state of a boolean property.

...That's just off the top of my head. :angel:

PR #1876, opened 2018-07-19
Use track names in "Add To Timeline" window

ferdnyc:
I happened to notice that the track selector in the "Add To Timeline" window (right-click a Project Files entry, select "Add To Timeline...') didn't use the user-defined track labels, it always showed "Track _"
numbered labels even when a custom label was set. There was no reason _not_ to use the user's track names, so now it does.


DylanC:
@ferdnyc - LGTM!

PR #1883, opened 2018-07-20
WIP: Timeline style refresh

ferdnyc:
I decided to at least get started on a timeline style refresh, taking to heart what I felt were some of the better points @qubodup made in #1425. I'm hoping this won't step on @DylanC's toes, with his work in #1864, but if it does then we'll sort that out between us.

So far I've mostly only tackled track styling, but I think it's coming along nicely so I wanted to solicit input. Please, though, let's try not to [bikeshed](https://en.wikipedia.org/wiki/Law_of_triviality#Argument) this to death.

### Observations/Highlights/Caveats
* The intent here is to blend the background of the Timeline area as a whole into the application interface. Note that this only applies to the dark theme, and I don't intent to change that with this PR. Don't ask.
* Selected clips now have their red border go _all_ the way around. Ditto the yellow border when clips intersect with the selection rectangle.
* Though it's not visible here, I also changed the snapline when dragged tracks are aligned from cyan/blue (same basic color as the clip border) to yellow. There's too much of that blue in the interface as it is.
#### Issues
* You'll notice that clips appear to be drawn a pixel too high, relative to the tracks. That's true, and it's true in 2.4.2 release as well. I promise. Look carefully. It's just more noticeable now.
* Ditto the fact that the first-frame keyframe markers (little green tickmarks) seem like they're one pixel too far to the left. Exactly the same as 2.4.2.
* I _still_ cannot find or get rid of that tiny white square at the intersection of the two scrollbar areas (extreme lower-right corner), and it's driving me crazy.
* The marker icon has _got_ to be changed, the current one looks horrible against the application-window gray ruler. It's on my list.

### Side-by-side
Here's the current state of this PR, side-by-side with 2.4.2 release on the left. I'll try to keep this comparison screenshot updated as I push changes into the branch.
screenshot from 2018-07-20 07-51-43


ferdnyc:
P.S> In the same vein as "don't [bikeshed](https://en.wikipedia.org/wiki/Law_of_triviality#Argument)" / "don't ask if the 'light' theme can extend to the Timeline":
#### I have no intention of making this selectable, optional, configurable, etc.
We either agree that the changes are an improvement and it gets merged, or we don't and it doesn't. In the end it's all or nothing. (Though if anyone feels that certain specific changes here are problematic, please do make your case and I'll revisit them.)

ferdnyc:
Poking the usual suspects for feedback: @jonoomph @peanutbutterandcrackers @DylanC @N3WWN

qubodup:
Telling from description/screenshots (thanks!): these changes are great, the app looks less cluttered/ancient and more clean/stylish.

Both the current and the old scrollbar background colors are problematic. Neither solution is better or worse.
I would suggest a lighter grey color for scrollbar backgrounds, as for example seen in the 2nd and 3rd screenshot in this 3-picture premiere pro gallery: https://imgur.com/a/BRwQYTw

ferdnyc:
Thanks @qubodup , I appreciate the feedback.

> I would suggest a lighter grey color for scrollbar backgrounds, as for example seen in the 2nd and 3rd screenshot in this 3-picture premiere pro gallery: https://imgur.com/a/BRwQYTw

You... lost me there. 
1. The second image... actually, I'm not even sure what's a scrollbar vs. what's a something-else bar in that image, but AFAICT:
   1. The timeline canvas has light-colored scrollbars, against a background that's the exact same color as the main canvas background, separated by (actually, outlined by) a dark-colored gutter. (IOW, exactly the same as the new design above, except for the gutter and the color of the scrollbars themselves.) 
   1. The scrollbars on other elements, like the media browser, appear different — and don't seem to give any clear indication of which part is the scrollbar and which part is the background.
1. The third image has *marginally* lighter scrollbars, against a *marginally* darker background, compared to the timeline canvas background. But it's all degrees. (And, honestly, those scrollbars once again have very poor visibility, to my eyes.)

> Both the current and the old scrollbar background colors are problematic. Neither solution is better or worse.

Perhaps we need to settle on some common definitions, again:

```
From WordNet (r) 3.0 (2006) [wn]:
  problem
      n 1: a state of difficulty that needs to be resolved; "she and
           her husband are having problems"; "it is always a job to
           contact him"; "urban problems such as traffic congestion
           and smog" [syn: {problem}, {job}]
      2: a question raised for consideration or solution; "our
         homework consisted of ten problems to solve"
      3: a source of difficulty; "one trouble after another delayed
         the job"; "what's the problem?" [syn: {trouble}, {problem}]
```

When I say "problematic" I mean in the first or third sense of the word "problem". "I think they should be a different color" is not a _problem_, it's bikeshedding. "I think they should look like (some other application) is even less compelling, unless you can explain why it's _better_ in a practical sense (the scrollbars would be more **useful** / more usable) if they look like that. I'm willing to accept that the current and/or new scrollbars pose some sort of _problem_, but you have to explain to me what that is. In its own terms, not in comparison to any other software. (Comparisons are fine for suggested _solutions_ to a problem, but the problem first has to stand on its own.)

(As an example: In 1.2 above I noted that the media browser scrollbars don't clearly indicate which part is the scrollbar vs. which is the background — it seems a little hard to tell, at first glance, whether I'm looking at a small bar scrolled all the way up, or a big bar scrolled all the way down.)

DylanC:
@ferdnyc - Looks great so far! I don't think this will hurt my changes as I probably will be working on functionality only changes and maybe some bug fixes.

ferdnyc:
Thank you, @qubodup, that's _extremely_ helpful.

> I have to scan up to nearly the entire width of OpenShot with my eyes to be able to tell whether or not there is a horizontal scrollbar or not, as I need to see the grip to be able to tell, 

In fairness, that is the reason why the scroll bar grip is bright blue, unlike many apps which seem to favor unobtrusive scrollbars. — Ours is pretty hard to miss, for answering the simple question, "Is there a scroll bar active in this direction (horizontal/vertical)?" But, point taken.

> if I want to be able to click on the scrollbar track to scroll in the clicked direction (standard scrollbar feature), 

**COMPLETELY fair**, and a need the current design utterly fails to meet, agreed. So, that gives us a requirement: the scrollbar track needs to be both..

1. Clearly visible, at all times.
   * (Also, clearly distinguishable from the scrollbar itself, and it should be obvious at a glance whether there is a scrollbar present in the track or not.)
2. Easily targetable for mouse clicks, in the empty track area around the scrollbar grip, in order to reposition the scroll region.

Cool, that gives me a well-defined set of constraints to work within. I'll incorporate those req's into the next changes. Thanks again.

ferdnyc:

peanutbutterandcrackers:
@ferdnyc - Can we bring sexy back?

DylanC:
@ferdnyc - I was gonna ask something similar. 😆

PR #1885, opened 2018-07-20
Some fixes for the timeline debugger

ferdnyc:
Just some minor things, bugs that had crept in because it hasn't been used in so long. It's still useful for testing style changes, though, so this returns it to more complete functionality.


DylanC:
@ferdnyc - I was about to work on fixing the debug functions myself. You got there ahead of me. :+1:

PR #1887, opened 2018-07-20
Update edit-clear icon in Humanity

ferdnyc:
I noticed that the Humanity `edit-clear` icon (the one for clearing the Filter text fields) looked... well... like ass in the dark theme. The original was a semi-transparent outline filled with dark gray gradient, with a light gray shadow. Which doesn't work well against the application theme:

screenshot from 2018-07-20 13-52-00

I reversed the colors to an outline filled with a _light_ gray gradient, with a dark gray shadow. I can't actually show you how it looks in the interface since OpenShot uses my system theme except in the AppImage, and I can't edit the contents of the AppImage to update the icon. But here's a mockup where I placed the icon into a screenshot using the GIMP:

screenshot from 2018-07-20 13-52-17

Also, I made some improvements to `src/images/Makefile`, so that `openshot_rc.py` will be recompiled whenever an icon in the `Humanity/` directory is updated, not only when `openshot.qrc` is edited. And I recompiled the resource file to pick up the new icon.


DylanC:
@ferdnyc - Nice find. I didn't really notice this before! :+1:

PR #1891, opened 2018-07-20
Convert (most) files with DOS line-endings to Unix...

ferdnyc:
A bunch of files in the tree ended up with CRLF line-endings, somehow. Not all editors like this (mine gets pissy about it), and attempting to convert them in the process of making other commits is proving difficult. (For instance, the constant conflicts on my tutorial2 branch.)

So, this PR takes a clean branch off `develop`, and converts most of the files without changing **anything** else.

I didn't convert quite _all_ of the files, as I tried not to touch the ones that I know have active PRs (`main_window.ui` excepted, since converting that will _solve_ my conflicts). Checking in conversions of those files would probably mess those PRs up and cause conflicts in _them_. That may still happen with these, to a minor extent, but it should hopefully be a one-time thing, and I'll be happy to do the work of generating diffs to fix any conflicts that crop up.

(This is also going to completely screw my logspam branch, which touches nearly every source file. Which is why I'd prefer to get this out of the way now, and then I can deal with the fallout.)


DylanC:
@ferdnyc - Done. Maybe keep an eye on the logspam PR just in case.

ferdnyc:
Cool, thanks!

Oh, I fully expect it'll go sideways. I'll probably revert any of the commits that touch affected files, then re-create them by hand in the non-CRLF versions of the files. That's actually the easiest approach, as the changes tend to be scattered but minor.

PR #1772, opened 2018-07-03
WIP: Reduce logspam

ferdnyc:
OpenShot outputs a staggering number of unnecessary, redundant, or generic messages to its log while running, especially on startup. 

(A log message every time a fallback icon is used; every time an interface binding is created; every time an effect is loaded — that one just says "effects_model:INFO category: Video" every time, so _super_ useless; every time the preview is resized; etc...)

The user doesn't care about _any_ of these, and honestly they're of no help to developers either. They're just the equivalent of debug `printf()`s being left in the code. So, this PR aims to wipe them out. 

Along the way, I'm fixing the type of some other messages (lowering some from `log.info()` to `log.debug()`, promoting others from `log.info()` to `log.error()`, etc.) and updating the formatting of any that use Python %-string formatting to instead use `str.format`.

WIP because I still have more files to go through.

PR #1905, opened 2018-07-23
WIP*: Reduce logspam

ferdnyc:
OpenShot outputs a staggering number of unnecessary, redundant, or generic messages to its log while running, especially on startup.

(A log message every time a fallback icon is used; every time an interface binding is created; every time an effect is loaded — that one just says "effects_model:INFO category: Video" every time, so super useless; every time the preview is resized; etc...)

The user doesn't care about any of these, and honestly they're of no help to developers either. They're just the equivalent of debug `printf()`s being left in the code. So, this PR aims to wipe them out.

Along the way, I'm fixing the type of some other messages (lowering some from log.info() to log.debug(), promoting others from `log.info()` to `log.error()`, etc.) and updating the formatting of any that use Python %-string formatting to instead use str.format.

I've also made the logging asymmetrical between the console output and the logfile, and hooked the `debug-mode` preference up to OpenShot's (file-based) logging. So, `openshot-qt.log` will get INFO and higher messages normally, but DEBUG-and-higher when the debugging switch is enabled.

(Replacement for PR #1772)

### IMPORTANT:
\* — This is very much WIP for real, please _do **NOT** merge_ in current state. There are a lot of commits in here I was holding back on my local system, not pushing to my fork / the previous PR, to keep what _was_ in the PR relatively harmless to merge. The changes here are extensive and no longer harmless, and they are not ready to go in yet.


ferdnyc:

PR #1926, opened 2018-07-29
Remove debug.js from index.html

ferdnyc:
The JavaScript file no longer exists


DylanC:
@ferdnyc - Whoops, good catch! :+1:

PR #1944, opened 2018-08-02
Improve zoom scale

### LINKED ISSUE #1943: Zoom scale is unusable
In OpenShot 2.4.2 (https://github.com/OpenShot/openshot-qt/commit/836a0232161950b8da8b673c0ac3acdaf98de548) a new zoom scale was introduced, instead of linear scale. Thanks for working on improving it. Alas, it breaks my processes in these ways:

1. I prefer round scales, to powers of 5: 10, 15, 20, 15, 30, 40, 60, 90, 120. That way you can easily know what the values for intermediate bars are, and generally estimate lengths. The current scale does not have a single "human" value.
2. For synchronizing audio to video, I use scale 3: it is the lowest scale on which the waveform does not disappear. I need this small scale to move tracks by tens of milliseconds. Now I've got a choice between 1 (on which there is no waveform) and 7 (which is too big for synchronizing).

Could you please fix the scale to better values?

Zverik:
Fixes #1943. Zoom scale was not generated programmatically, instead I chose each value by hand, so that the scale bar is immediately readable and tends to show round values.

### DIFF
--- a/src/classes/conversion.py
+++ b/src/classes/conversion.py
@@ -26,37 +26,9 @@
  along with OpenShot Library.  If not, see <http://www.gnu.org/licenses/>.
  """
 
-zoomSeconds = [1,
-               7,
-               14,
-               24,
-               54,
-               94,
-               118,
-               174,
-               205,
-               276,
-               356,
-               445,
-               542,
-               648,
-               761,
-               881,
-               1075,
-               1212,
-               1428,
-               1578,
-               1813,
-               2141,
-               2396,
-               2745,
-               3192,
-               3645,
-               4187,
-               4890,
-               5988,
-               7200
-               ]
+zoomSeconds = [1, 3, 5, 10, 15, 20, 30, 45, 60, 75,
+               90, 120, 150, 180, 240, 300, 360, 480, 600, 720,
+               900, 1200, 1500, 1800, 2400, 2700, 3600, 4800, 6000, 7200]
 
 def zoomToSeconds(zoomValue):
     """ Convert zoom factor (slider position) into scale-seconds """


N3WWN:
@DylanC - Do we have all branches building AppImages at this point?  Or is it just the develop and master branches?  

I ask because, if it's all branches, would it make sense to put this into a new branch so we can play with these values without altering the current ones?  

I will say that I do not like the current zoom values, so it's more than likely that I'll prefer these. 😜

DylanC:
@N3WWN - I think its just the develop and master branches. Btw, its easy to try if you just paste the line of code over the previous. I gave it a shot just now.

I don't have much more feedback but it did **feel** better... if that's even considered an endorsement for it. The increments just seemed a little bit more natural and not so severe.

N3WWN:
@DylanC wrote:
> Btw, its easy to try if you just paste the line of code over the previous. I gave it a shot just now.

I was trying to figure out how to do this... I'm using the AppImage so have to try to intercept the reading of the file between hitting <kbd>Enter</kbd> to execute the AppImage and when OpenShot actually launches.  The /tmp/.mount_* directory that the AppImage creates is read-only, so that's causing issues, too.

Having only spent about 15 minutes on this yesterday, that's where I'm stuck at.

DylanC:
@N3WWN - Oh sorry, I mean't to say in the source code. I have it building on my machine locally. So I put this change in and gave it a whirl. Maybe @ferdnyc can try it when he has the chance. His feedback is usually quite massive and detailed. (almost the size of a little novel) 😆

jonoomph:
This looks great! I just gave it a try and it works nicely! Merged.

PR #1993, opened 2018-08-14
WIP: Timeline actions

DylanC:
Following on from this PR #1914. I've decided to open a new PR for this. The objective is simply to implement some nice track actions such as:

- Toggle Audio
- Toggle Video

I hope to represent each option as a simple icon but I will start with buttons until this works. (firstly) This will aim to solve issue #1587 and other similar issues/duplicates.


DylanC:
@jonoomph - Closing since I don't think I will get around to this one soon. It would require much time and commitment I don't have available at the moment. (However, it still would be a great feature at some point)

PR #2011, opened 2018-08-20
WIP: Rename files according to AppStream specification

eszlari:


eszlari:
@DylanC any update on this?

DylanC:
@eszlari - None yet. I will get onto @jonoomph again to see can we sort this out. 👍

jonoomph:
@eszlari Thanks for this PR and sorry for the delay! That's my fault. This looks really good to me, but I need you to resolve the conflicts before we can continue. Please @ me and let me know when this is ready. I'm going to update this to a WIP until then. Thanks!

eszlari:
@jonoomph I resolved all conflicts.

eszlari:
@jonoomph Any news on this?

DylanC:
@eszlari @jonoomph - Do we know what is preventing this from being merged?
`cx_Freeze.freezer.ConfigError:` cannot find file/directory named `/home/travis/build/OpenShot/openshot-qt/xdg/openshot-qt.svg`

ferdnyc:
@DylanC 

Yeah. I restarted the build, and it failed the same way.

...Ah, I see. Because this PR is _renaming_ that file, without updating the name at all of the places it's referenced. 

```sh
$ grep openshot-qt.svg *(.) .*(.) installer/**/*(.)
setup.py:    ('share/pixmaps', ['xdg/openshot-qt.svg']),
setup.py:    ('share/icons/hicolor/scalable/apps', ['xdg/openshot-qt.svg']),
installer/build-server.py:        shutil.copyfile(os.path.join(PATH, "xdg", "openshot-qt.svg"), os.path.join(app_dir_path, "openshot-qt.svg"))
installer/build-server.py:        shutil.copyfile(os.path.join(PATH, "xdg", "openshot-qt.svg"), os.path.join(app_dir_path, "usr", "share", "pixmaps", "openshot-qt.svg"))
```

I don't see why the files would be renamed in the _source_ tree, anyway. Why should the AppStream specification want to dictate what we call our icon file, and why would we let it? I'd certainly rather refer to it as `openshot-qt.svg`, rather than `org.openshot.OpenShot.svg`.

If the icon has to have that f---ing stupid name in the _destination_ directory, then so be it. But it should be renamed **DURING** install, not renamed in our source tree. `</IMHO>`

ferdnyc:
Yeah, that's what it comes down to, for me. The icon file is a perfect example: We use that same file on Windows and MacOS as well, and its name on all three platforms is `openshot-qt.svg`. Renaming it in our **source** tree breaks things, and is IMHO an unacceptable constraint. The AppStream specification can go pound sand.

I vote we reject this, unless someone can come up with a compelling reason why we need to break our tree to follow external naming conventions.

DylanC:
@ferdnyc 

> I don't see why the files would be renamed in the source tree, anyway. Why should the AppStream specification want to dictate what we call our icon file, and why would we let it? I'd certainly rather refer to it as openshot-qt.svg, rather than org.openshot.OpenShot.svg.

Yeah it does seem a bit extreme to rename a file in source (basically just for Linux) when this is a cross platform project.

> I vote we reject this, unless someone can come up with a compelling reason why we need to break our tree to follow external naming conventions.

I don't have strong enough of a reason here. Maybe @eszlari has the answer?

eszlari:
> If the icon has to have that f---ing stupid name in the _destination_ directory, then so be it. But it should be renamed **DURING** install, not renamed in our source tree. `</IMHO>`

Renaming the icons during installation is ok too.

DylanC:
@ferdnyc - According to the response above we can just replace the file name at installation. Should we do something to implement this for Linux?

ferdnyc:
🤷‍♂️ I mean, I guess we _can_... though I'm not sure why it's necessary.

The authors of the AppStream spec may be married to this reverse-DNS-style naming scheme, but a quick perusal of my Fedora box's `/usr/share/metainfo/` directory tells me that it's a fight they are not winning:

<details>
<summary><code>cd /usr/share/metainfo/; ls</code></summary>

```
blender-fonts.metainfo.xml
ca.desrt.dconf-editor.appdata.xml
cmake-gui.appdata.xml
com.github.gkarsay.parlatype.appdata.xml
com.github.gkarsay.parlatype.libreoffice-helpers.metainfo.xml
com.github.hluk.copyq.appdata.xml
com.obsproject.Studio.appdata.xml
com.ozmartians.VidCutter.appdata.xml
com.twitter.twemoji.metainfo.xml
eog.appdata.xml
evince-comicsdocument.metainfo.xml
evince-djvudocument.metainfo.xml
evince-pdfdocument.metainfo.xml
evince-psdocument.metainfo.xml
evince-tiffdocument.metainfo.xml
evince-xpsdocument.metainfo.xml
firewall-config.appdata.xml
flameshot.appdata.xml
fr.handbrake.ghb.appdata.xml
gaupol.appdata.xml
gdouros-symbola.metainfo.xml
gedit-bookmarks.metainfo.xml
gedit-bracketcompletion.metainfo.xml
gedit-charmap.metainfo.xml
gedit-codecomment.metainfo.xml
gedit-colorpicker.metainfo.xml
gedit-colorschemer.metainfo.xml
gedit-drawspaces.metainfo.xml
gedit-findinfiles.metainfo.xml
gedit-git.metainfo.xml
gedit-joinlines.metainfo.xml
gedit-multiedit.metainfo.xml
gedit-smartspaces.metainfo.xml
gedit-terminal.metainfo.xml
gedit-textsize.metainfo.xml
gedit-wordcompletion.metainfo.xml
gedit-zeitgeist.metainfo.xml
ghostwriter.appdata.xml
gimp-data-extras.metainfo.xml
gimp-dbp.metainfo.xml
gimp-elsamuko.metainfo.xml
gimp-luminosity-masks.metainfo.xml
gimp-wavelet-decompose.metainfo.xml
gitg.appdata.xml
glade.appdata.xml
gnome-control-center.appdata.xml
gnome-subtitles.appdata.xml
gnome-system-monitor.appdata.xml
goldendict.appdata.xml
gparted.appdata.xml
gscan2pdf.appdata.xml
gsmartcontrol.appdata.xml
gstreamer-bad-free.appdata.xml
gstreamer-good.appdata.xml
gucharmap.appdata.xml
inkscape.appdata.xml
julietaula-montserrat.metainfo.xml
libreoffice-base.appdata.xml
libreoffice-calc.appdata.xml
libreoffice-draw.appdata.xml
libreoffice-impress.appdata.xml
libreoffice-writer.appdata.xml
me.mitya57.ReText.appdata.xml
net.poedit.Poedit.appdata.xml
nm-connection-editor.appdata.xml
oflb-dignas-handwriting.metainfo.xml
openshot-qt.appdata.xml
org.avidemux.Avidemux.appdata.xml
org.bunkus.mkvtoolnix-gui.appdata.xml
org.fedoraproject.LangPack-en_GB.metainfo.xml
org.freedesktop.appstream.cli.metainfo.xml
org.freedesktop.fwupd.metainfo.xml
org.freedesktop.GnomeAbrt.appdata.xml
org.gimp.GIMP.appdata.xml
org.gnome.baobab.appdata.xml
org.gnome.Boxes.appdata.xml
org.gnome.Builder.appdata.xml
org.gnome.Calculator.appdata.xml
org.gnome.cantarell.metainfo.xml
org.gnome.Characters.appdata.xml
org.gnome.clocks.appdata.xml
org.gnome.ColorProfileViewer.appdata.xml
org.gnome.Contacts.appdata.xml
org.gnome.Devhelp.appdata.xml
org.gnome.dfeet.appdata.xml
org.gnome.DiskUtility.appdata.xml
org.gnome.Documents.appdata.xml
org.gnome.Epiphany.appdata.xml
org.gnome.Evince.appdata.xml
org.gnome.Evolution.appdata.xml
org.gnome.FileRoller.appdata.xml
org.gnome.font-viewer.appdata.xml
org.gnome.gedit.appdata.xml
org.gnome.gThumb.appdata.xml
org.gnome.Lollypop.appdata.xml
org.gnome.Maps.appdata.xml
org.gnome.meld.appdata.xml
org.gnome.Music.appdata.xml
org.gnome.Nautilus.appdata.xml
org.gnome.Packages.appdata.xml
org.gnome.PackageUpdater.appdata.xml
org.gnome.Photos.appdata.xml
org.gnome.Pomodoro.appdata.xml
org.gnome.Screenshot.metainfo.xml
org.gnome.seahorse.Application.appdata.xml
org.gnome.Shotwell.appdata.xml
org.gnome.Software.appdata.xml
org.gnome.Software.Plugin.Epiphany.metainfo.xml
org.gnome.Software.Plugin.Flatpak.metainfo.xml
org.gnome.Software.Plugin.Fwupd.metainfo.xml
org.gnome.Software.Plugin.Odrs.metainfo.xml
org.gnome.SoundRecorder.appdata.xml
org.gnome.Sysprof2.appdata.xml
org.gnome.Terminal.appdata.xml
org.gnome.Totem.appdata.xml
org.gnome.tweaks.appdata.xml
org.gnome.Usage.appdata.xml
org.gnome.Weather.appdata.xml
org.kde.k3b.appdata.xml
org.kde.kcachegrind.appdata.xml
org.kde.kdiff3.appdata.xml
org.kde.kleopatra.appdata.xml
org.kde.konqueror.appdata.xml
org.kde.konsole.appdata.xml
org.qt-project.qtcreator.appdata.xml
org.shotcut.Shotcut.appdata.xml
org.thonny.Thonny.appdata.xml
PrestoPalette.appdata.xml
projectx.appdata.xml
rclone-browser.appdata.xml
rhythmbox.appdata.xml
simple-scan.appdata.xml
spyder3.appdata.xml
steam.appdata.xml
svg2svgt.appdata.xml
synergy.appdata.xml
system-config-printer.appdata.xml
tagtool.appdata.xml
typing-booster.appdata.xml
vinagre.appdata.xml
vlc.appdata.xml
winetricks.appdata.xml
yelp.appdata.xml
```

</details>

Adoption is decidedly... "mixed". (Though we're strongly in the minority, with our non-reverse-DNS, no-`.desktop` one-word name.)

<details>
<summary>Lots more blather you don't need to bother reading</summary>

...Ah. And I think I see why. According to the AppStream spec's § 2.2.2 [Desktop Applications | File specification](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html):

> #### Note
> In previous AppStream releases, the `<id/>` was used to associate metainfo files with their .desktop files to merge in data from .desktop files into the AppStream generator's final output. In modern metainfo files, the component-ID for `desktop-application` components can be an arbitrary reverse-DNS string (matching the naming rules applying to all AppStream metadata), while the `<launchable/>` tag is used to associate .desktop files with their metainfo files.
>
> Note that even though the component-ID can now be any rDNS name, when updating existing applications, do not change their `<id/>` to drop the .desktop suffix. The rules are relaxed when picking a new component-ID for new applications, but when updating older applications they still need to keep their original `<id/>` (when it's otherwise compliant). The ID is used to uniquely identify applications across distributions and releases and should always remain the same for the same application.

Now, you might think that addresses our situation, and says we _shouldn't_ do this. I certainly did, at first. But reading it more closely, it's actually no help **at all**, and only further confuses the issue.

1. It says that an existing application `<id/>` should not change.
1. But it also says "when it's otherwise compliant".
1. It _also_-also says "do not change their `<id/>` to drop the `.desktop` suffix". 

Our `<id/>` never _had_ a `.desktop` suffix to begin with, it's just `openshot-qt`! (Which I guess means we've been getting it wrong from the start. Probably because this spec is both confusing _and_ keeps changing. A trend that has not abated, I might point out.)

I guess that admonishment about not changing an `<id/>` to drop the `.desktop` is why Gnome-Terminal is `<id>org.gnome.Terminal.desktop</id>` and Totem is `<id>org.gnome.Totem.desktop</id>`, but Gnome Weather is `<id>org.gnome.Weather</id>` and Evince is `<id>org.gnome.Evince</id>`. But then again, Yelp is `<id>yelp.desktop</id>`, Inkscape is `<id>inkscape.desktop</id>`, gsmartcontrol is `<id>gsmartcontrol</id>`, and clearly _everyone_ is confused by this.

In fact, someone apparently got so confused that Evolution's `org.gnome.Evolution.appdata.xml` file ended up like _this_:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<component type="desktop">
  <id>org.gnome.Evolution.desktop</id>
  <metadata_license>CC0-1.0</metadata_license>
  <project_license>GPL-2.0+ and GFDL-1.3</project_license>
  <name>Evolution</name>

[...]

  <launchable type="desktop-id">org.gnome.Evolution.desktop</launchable>
  <provides>
    <id>org.gnome.Evolution.desktop</id>
  </provides>
```

(The `<provides/>` tag is supposed to be for aliases, mostly used when listing the old name after changing to a new one. It makes no sense to be an alias for yourself. Presumably they _meant_ to change the main ID to `<id>org.gnome.Evolution</id>`, but wires got crossed somewhere.)

</details>

Aaanyway, I certainly have absolutely no qualms about doing this for the `.appdata.xml` file, for the `.desktop` file, etc. Even in the source tree, since we really just don't care.

But I'm not sure about renaming the icon files. That doesn't seem to be the norm even among applications following AppStream, and I can't seem to find anything that actually requires it. The AppStream spec and the updated Desktop File Spec from freedesktop both _appear_ silent on the issue, unless I just missed something.

The only place I **can** find where it's addressed is [in the Freedesktop Icon Naming Spec](https://standards.freedesktop.org/icon-naming-spec/latest/ar01s03.html), which admittedly dates back to like 2006, but maybe nothing's bothered to supersede it? It says:

> The “Applications” context should not use this method of falling back to more generic icons. An application must either use a generic application icon name provided by this specification, **or install an icon named the same as the executable for running the application.** 

Still, I suppose it can't hurt to _copy_ the SVG file to the longer name, since it'd be weird to have XML files referencing icon names that don't exist in our tree. It sucks that we'll need to keep two copies around, but we still need (and want!) it to be named `openshot-qt.svg` as well.

However, this change is going to screw things up for all the distribution packagers, even if it is "more correct". That's why I think we should do it when, and only when, we also pull the trigger on renaming the launcher executable (#2796). That's also going to break lots of people, so we might as well do it all at once and let everyone be broken together. I've suggested to @jonoomph that #2796 get merged shortly before the next release, and the change be announced in the release notes so everyone's aware. I still think that's the right approach, and I think this should happen at the same time.

eszlari:
> Adoption is decidedly... "mixed". (Though we're strongly in the minority, with our non-reverse-DNS, no-.desktop one-word name.)

Some of the apps you listed have since moved to the new naming scheme, e.g.:
https://gitlab.com/inkscape/inkscape/blob/master/org.inkscape.Inkscape.appdata.xml.in
https://gitlab.gnome.org/GNOME/gitg/blob/master/data/org.gnome.gitg.appdata.xml.in.in

> But I'm not sure about renaming the icon files.

AFAIK renaming the icons is only needed to make alternative icon themes work.

jonoomph:
I'm going to close this one until conflicts have been resolved, and this is addressed as an install feature (for Linux distros). And honestly, I'm not really sure how to correctly resolve this issue... and maintain as OS-cross platform happy balance.

PR #2029, opened 2018-08-24
Improve waveform display rendering

DerGenaue:
Improve the rendering speed, especially when zoomed very close.
Additionally, when zoomed further out, there are now two plots displayed: the max values and the average values.
There are some ideas to easily extend and improve this further on the library side;
See comments

### DIFF
--- a/src/timeline/js/functions.js
+++ b/src/timeline/js/functions.js
@@ -58,6 +58,7 @@ function drawAudio(scope, clip_id){
 
         // Determine start and stop samples
         var samples_per_second = 20;
+        var block_width = 2; // 2 pixel wide blocks as smallest size
         var start_sample = clip.start * samples_per_second;
         var end_sample = clip.end * samples_per_second;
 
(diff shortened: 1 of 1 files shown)


DerGenaue:
I've created this after I read through some of the other PRs and saw the code that created the rendering of the audio.
This should solve most of the audio rendering performance issues (not addressing the waveform generation delay)

DerGenaue:
I've dug quite a lot through the the code in the QT app (js, py), libopenshot (C++) and libopenshot-audio to see whether I could also improve the generation of the audio samples in their reduced form.

One thing I wanted to address is the `samples_per_second` constant, that is stored in two places (py and js).
I would have that constant transmitted together with the data in the JSON.

Also, the waveform data is based on Max-Values (see [libopenshot-audio: AuidoSampleBuffer/getMagnitude](https://github.com/OpenShot/libopenshot-audio/blob/develop/JuceLibraryCode/modules/juce_audio_basics/buffers/juce_AudioSampleBuffer.cpp#L642))
That library also provides RMS values (Mean) which aren't exposed by the [libopenshot Frame](https://github.com/OpenShot/libopenshot/blob/develop/src/Frame.cpp#L317).
I learned that it is pretty useful to see visually both max peaks and average values, therefore I would propose to also expose this in libopenshot and rewrite the [get_waveform](https://github.com/OpenShot/openshot-qt/blob/develop/src/classes/waveform.py#L54) python to provide both max and average values
While doing all this, I would also suggest to consistently use `waveform_data` as a name and not `audio_data` to make obvious this has a reduced sample rate.

One of the most confusing features imo is that the audio waveform contains the volumes animated on the clip; but *only* if they exist while the rendering is triggered.

All together, I would suggest to change the waveform JSON data to be changed to something like this:
```json
{
    "sample_rate": 30,
    "maxData": [0.8, 0.5, ...],
    "rmsData": [0.6, 0.2, ...],
}
```
Which also isn't influenced by the volume values.
They should instead be rendered separately eg. as a red line on top of the waveform.

I wasn't able to implement these changes because of the dependency on the library and because I do not have a development setup right now (literally just edited the files in the installation folder).
Maybe someone of you guys can implement these changes (they should be pretty simple if you have a dev setup)

DerGenaue:
Also, I would suggest not applying a wait animation on the users cursor while the waveform is generated, but display a wait icon inside the waveform display.
This would make clear that the software is still usable and that only this specific waveform is waiting to be rendered.
This could be especially relevant, when a user triggers waveform generation for multiple clips within a short time

DylanC:
@DerGenaue - Seems like a nice follow up to my PR. I do like your idea around samples_per_second and how we are managing the waveform data. I did also give your code a try:

I guess maybe its just my opinion but it looks a bit blocky with the new code if you have short audio clips zoomed in to 1 second:
screenshot from 2018-08-24 22-36-55

I tried changing that block width but to be honest any value seems to look poor (in my eyes at least). Other zoom levels look much better. Maybe should be be changing the block width if zoomed into 1 second or is that overkill?

DerGenaue:
That is no difference to the old version.
Try changing `samples_per_second` to eg. `50` both in the python and JS code.
That worked fine for me.
(I would actually propose something like that in the future)

DerGenaue:
the `block_width` variable is only relevant when the resolution of the data is higher than the current zoom level (= zoomed out far).
On close zoom, the sample rate of the waveform data makes the blocks bigger.

The blocks looked slightly rounded in the old version because of the `0.5 transform` offset I removed.
This introduces weird artifacts in my version.

DylanC:
> That is no difference to the old version.
Try changing samples_per_second to eg. 50 both in the python and JS code.
That worked fine for me.
(I would actually propose something like that in the future)

Yes, very true. I don't know why I didn't notice the original is the same. :laughing: Its better without the transform, I agree. I'm happy with this change overall.

DylanC:
@DerGenaue - If you take a look at #1864, myself and @ferdnyc discussed a little bit around the idea of rewriting the timeline in QML/Qt5. Although that itself would be a massive project really and may not give the project much benefit. Although in the case of the audio wave the benefit would be huge, so perhaps we should just focus on replacing that JS with C/C++ for more performance. 

> Do you think you could implement the proposed changes right away, as well?

I would implement these changes right away if I was very familiar and confident with the OpenShot code but I am not. You will notice I have only ever made changes to the timeline code which is web based. I'm happy enough working with JS and C/C++ but I've never touched Python really.

DylanC:
@DerGenaue - If you could write an issue describing the issue (copy and paste some of your description if you want) and detail the files that would need to be changed that would be great. Even better would be a PR that I could just test but its up to you. :laughing:

DerGenaue:
Discussion on this is now tracked in #2035

DylanC:
@DerGenaue 

> Do you work on that library, as well?

I don't currently work on the libopenshot. I've my own C++/Qt project I work on in my spare time. 

> Why was the design decision made to make all audio Frame-based?

I wouldn't have an answer to this question. @jonoomph is likely the only person who would know.

Issue #2041, opened 2018-08-28
File conflicts aren't checked properly after opening saved project (ASSETS_PATH is constant)


DerGenaue:
**Describe the bug**

Openshot has an internal constant `info.ASSETS_PATH`, which always points to the users home-directory to a temp location for assets.
This temp location is used for saving eg. titles when they are created.
This temp-location also is used to check for conflicts when saving or creating titles.  
**However**  
After saving, all the asset-files are moved to a local assets-folder directly next to the project save.
This makes all the conflict-checks break after saving, since they only check the static location, but not the local one

**Steps to reproduce**
 1. New Project
 2. Create Title `<name>-1`
 3. Create Title `<name>-2`
 4. Duplicate Title `<name>-1`
   ->The file name is correctly increased to `<name>-3` since `<name>-2` already exists
   ->Trying to save as `<name>-2` will correctly trigger a prompt whether to overwrite the other file
 7. Save and re-open project
 8. Retry 4. : it will not work as intended

**Expected behavior**
As soon as a save has been performed, the `ASSETS_PATH` should point to the local folder so that conflicts are checked correctly and no project files are overridden accidentally.
All newly created files should be created directly in this local folder.

**Attention**
More discussion might be needed; since this does not fix that assets of an already existing project might be overwritten when saving another project in the same location for the first time.

This is one of the underlying issues of #2004

ctsdownloads:
Merged, see above.

Issue #2004, opened 2018-08-19
Multiple Projects in same folder overwrite each others assets


DerGenaue:
**Describe the bug**
When having multiple projects in the same folder, Openshot will "share" the assets and thumbnails-folders
When a title is saved in the application, it is not checked properly, whether this filename already exists and titles from other projects are simply overwritten. Very annoying...

**System Details (please complete the following information):**
 - Windows 10 / 7
 - OpenShot 2.4.2

**Expected behavior**
Separate folders should be used for different projects.
Eg. the project-name should be appended to the folder name
(careful when renaming projects)
Also, titles should not overwrite other files without prompting the user.

**Additional Notes**
Something similar can apparently also happen within one single project:
After restarting OpenShot and opening a old project, try to duplicate a title.
The title-name-counter will be reset (to 1) and an already existing title with this name will be overwritten.

ferdnyc:
#2976 will _partly_ solve this, in that each project's assets will be separated and titles from other projects will no longer be overwritten. **_However_**, I hadn't tested the #2041 case where the Title Editor will trample its _own_ filenames after a project is saved and reloaded, and currently that remains an issue with the #2976 code. (Now that I'm aware of it, though, I'm hopeful I can fix that as well.)

PR #2042, opened 2018-08-28
Improve Title duplication counting

DerGenaue:
Keep the users name when duplicating a title;
Increment number if existing; add if not yet existing

When duplicating, Title names will now be changed in the following way:

* `<name>-1`  
  `<name>-2`
* `<name>`  
  `<name>-1`
* `<name>-`  
  `<name>-1`
* `<name>----`  
  `<name>----1`
* `<name>1`  
  `<name>2`

=
* Increase number if already exists;
* add `-<number>` if it doesn't exist
* except when the title already ends in a `-`; then just add `<number>`

Old behavior:
* `TitleFileName-1`  
  `TitleFileName-2`
* `<name>`  
  `TitleFileName-1`
* `<name>-`  
  `TitleFileName-1`
* `<name>----`  
  `TitleFileName-1`
* `<name>1`  
  `TitleFileName-1`

While doing this PR, I was trying to mitigate #2004;
Which I wasn't able to do.
But at least I figured out the exact reason for it:
See #2041

### DIFF
--- a/src/windows/title_editor.py
+++ b/src/windows/title_editor.py
@@ -28,6 +28,7 @@
 
 import sys
 import os
+import re
 import shutil
 import functools
 import subprocess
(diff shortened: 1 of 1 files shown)


DylanC:
@DerGenaue - So are you saying this partly fixes the issue but not fully?

DylanC:
@DerGenaue - I see. Indeed it does partly fix this problem. This PR is similar to how my audio_wave PR only fixed 1/2 issues but it was still a fix and still worth merging.

PR #2060, opened 2018-09-04
Square the PNG icon file

ferdnyc:
This one's kind of odd. Turns out the dimensions of `openshot-qt.png` are only 64 × 56, which is problematic for some desktop environments. (It's odd, too, because the `.svg` has square page dimensions, and the `.ico` file is a square 64x64. **\*shrug\***)

We've been squaring it during packaging with ImageMagick, using this nonsense:
```sh
convert xdg/openshot-qt.png -virtual-pixel Transparent -set option:distort:viewport \
"%[fx:max(w,h)]x%[fx:max(w,h)]-%[fx:max((h-w)/2,0)]-%[fx:max((w-h)/2,0)]" \
-filter point -distort SRT 0 +repage <outfile>
```
But that's nuts, so I just re-exported from inkscape at 64x64. Fills the frame better, too.


DylanC:
@ferdnyc - Wow, very strange dimensions indeed! Good find! 👍

ferdnyc:
Credit goes to the previous maintainers at rpmfusion. All I did was read through the spec file I inherited, find that `convert` line, and decide it was stupid and I wasn't going to keep doing _that_ if I could help it. :grin:

PR #2103, opened 2018-09-12
Completely remove the old src/locale dir

ferdnyc:
High time I cleaned up my mess. This change eliminates all traces of `src/locale/`.

* Fix translation support code
  * Scripts and `OpenShot.pot` file are moved to `src/language/`
  * `generate_translations.py` and `test_translations.py` are updated for the new tree and filenames
  * Add `show_translations.py`, a script to dump all translated strings for each translation file
* Fix translation packaging
  * `freeze.py` is updated to use the `language` dir, instead of `locale`, for Qt translation files
  * `src/classes/language.py` will look for packaged (frozen) Qt translations in `language/`
  * `freeze.py` will package `qtbase_*.qm` files. (It was only packaging `qt_*.qm` files, even though `src/classes/language.py` looks for `qtbase_` files in the frozen output dir. That appears to have been an oversight.)
* `installer/build-mac-dmg.sh` will no longer include `locale` dir
* Delete `src/locale/README`, as it dated back to 2009 (OpenShot 1.x) and was completely incorrect
* `AUTHORS` file points to `src/language/` for translation credits
* `rm -r src/locale`


DylanC:
"Stupider typo" xD

> Delete src/locale/README, as it dated back to 2009 (OpenShot 1.x) and was completely incorrect

Wow, I think you found an ancient relic here! (Good sir) :laughing: 

LGTM anyway. I might hold off merging this one so quick to see if anyone else will give feedback. :+1:

ferdnyc:
> I might hold off merging this one so quick to see if anyone else will give feedback.

Prudent, it's definitely a big change with the potential to break stuff. If I did everything right, the bits that got removed should already have been unused, but I can only say **for sure** that they're unused _inside_ the source tree. The translation process necessarily involves external resources as well.

> "Stupider typo" xD

Well, I'd changed e.g. this:
```python
{"type": 'QT',
         "prefix": 'qtbase_',
         "path": os.path.join(info.PATH, 'locale', 'QT')},
```
to this:
```python
{"type": 'QT',
         "prefix": 'qtbase_',
         "path": language_path},
```
to this:
```python
{"type": 'QT',
         "prefix": 'qtbase_',
         "path": os.path.join(info.PATH, 'language'},
```
Which sorta looks fine in the editor if you're not paying attention, but the Travis build failed due to the missing close-paren. So I first changed it to _this_:
```python
{"type": 'QT',
         "prefix": 'qtbase_',
         "path": os.path.join(info.PATH, 'language'),
```
Which is just patently moronic. Because the syntax is wrong, yes, but more because I assumed that I'd accidentally written `}` _instead of_ `)`. They're completely different, so you'd think I would've checked whether the `}` was also needed. (Though it do mix them up a lot in Makefiles, with their stupid `$(VARIABLE)` syntax.) Anyway, third time was the charm.

DylanC:
@ferdnyc - Well I guess this is good to go then. I didn't get any other feedback.... soo.. yeah. :rofl:

ferdnyc:
If anything breaks, I'm sure we'll hear about it. :wink:  The important thing was to get it in well before the next release, so there wouldn't be any problems with shipping double translation files. 

(And this way now the path and filename changes will be super obvious to packagers and translators — what with a whole directory missing, a new one taking its place, and everything inside being shuffled around and renamed — so they'll have a chance to deal with the changes.)

jonoomph:
Okay, I've merged these 2 language PRs into the release branch, and updated lots of misc infrastructure things (like LaunchPad and some translation generation scripts), and I've updated the .qm translations and the qrc file. Seems like everything works good on Linux. I'm going to test Windows and Mac next.

PR #2138, opened 2018-09-23
Update appdata

eszlari:
Unfortunately, you did not update this file for the release.

### DIFF
--- a/xdg/openshot-qt.appdata.xml
+++ b/xdg/openshot-qt.appdata.xml
@@ -69,6 +69,7 @@
     <content_attribute id="money-gambling">none</content_attribute>
   </content_rating>
   <releases>
+    <release version="2.4.3" date="2018-09-22"/>
     <release version="2.4.2" date="2018-06-30"/>
     <release version="2.4.1" date="2017-11-12"/>
   </releases>


ferdnyc:
This is something that should also be brought to @jonoomph 's attention, since the best time to do it would obviously be _before_ the release, not after the fact. It'd be good to add a step for updating the AppData file to the final release process.

DylanC:
@ferdnyc - Yeah I'm not sure how but this does need to become part of the official release process.

ferdnyc:
> I'm not sure how

I mean... most release engineers use a checklist. This would be an item on the checklist.

Many maintainers, in fact, publish their Release Procedures, whether it's [in the source tree](https://github.com/electron/electron/blob/master/docs/development/releasing.md), [in the project wiki](https://github.com/snb2013/django-oscar/wiki/Releasing), or what have you. This is so that the procedure is public and agreed upon by all parties involved, and so that anyone involved in the project is able to make releases, even in the unfortunate event that the current maintainer is [hit by a bus](https://en.wikipedia.org/wiki/Bus_factor).

I'm just saying, it's not rocket surgery.

DylanC:
@jonoomph - Can we have this as part of future release?

PR #2171, opened 2018-09-30
title_editor.py: Don't import unused QWebView

ferdnyc:
So, turns out `src/windows/title_editor.py` was importing `QWebView` from `PyQt5.QtWebKitWidgets`. 

"But the title editor doesn't have any web components!" I hear you cry.

Well-spotted. And that's why `src/windows/title_editor.py` shouldn't be importing anything from `PyQt5.QtWebKitWidgets`. 

So, after this change, `PyQt5.QtWebKitWidgets` will only appear at one place in the source: an import statement at the top of `src/windows/views/timeline_webview.py`. As it should be.

### DIFF
--- a/src/windows/title_editor.py
+++ b/src/windows/title_editor.py
@@ -1,27 +1,27 @@
-""" 
+"""
  @file
  @brief This file loads the title editor dialog (i.e SVG creator)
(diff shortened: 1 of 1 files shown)


DylanC:
@ferdnyc - Excellent! Thanks. :+1:

PR #2187, opened 2018-10-04
Title Editor: Font/BG color selection fixes

ferdnyc:
Experimenting with the Title Editor in the course of examining @mcbluefire's title-SVG issue, I noticed a bunch of weird things about how the buttons for Font Color and Background Color worked:

* The color pickers they launched came up set to White every time
* They could only actually be used once, after that they didn't properly refresh themselves
* The text labels could be very hard to see

This PR fixes all of that. 
* The selected colors are now tracked in the (already-defined, but previously unused) `font_color_code` and `bg_color_code` class variables, and the color picker will default to the current value whenever it's launched to choose a new one. 
* The button labels are fully refreshed using `update_*_color_button()` after every color adjustment, ensuring that they restyle themselves to match the updated value. 
* The color buttons now use a luminance-based algorithm to switch between white and black text labels, whichever contrasts better with the selected button color.

### DIFF
--- a/src/windows/title_editor.py
+++ b/src/windows/title_editor.py
@@ -418,9 +418,22 @@ class TitleEditor(QDialog):
                 opacity = 1.0
 
             color = QtGui.QColor(color)
+
+            # Compute perceptive luminance of background color
+            colrgb = color.getRgbF()
+            lum = (0.299 * colrgb[0] + 0.587 * colrgb[1] + 0.114 * colrgb[2])
+            if (lum < 0.5):
+              text_color = QtGui.QColor(Qt.white)
+            else:
+              text_color = QtGui.QColor(Qt.black)
+
             # Convert the opacity into the alpha value
             alpha = int(opacity * 65535.0)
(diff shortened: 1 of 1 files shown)


ferdnyc:
Whoops. Accidentally submitted before I was finished. There will be a description momentarily, I promise. :wink:

ferdnyc:
The one thing this _doesn't_ fix is, the color buttons still come up with the default colors **initially**. They don't pick up the original font and background colors saved to the file. That would require additional parsing in the initial editor-launch phase, though it would be nice to have. 

(But at least now, they correctly _display_ their initial default values, which is an improvement despite the fact that they still don't match the initial state of the title SVG.)

DylanC:
@ferdnyc - Apologies for the late reply. I've been away all weekend on a trip and without my main laptop. (The one that has more than 2Gb of Ram) :rofl: 

Had a look over it tonight and I couldn't spot a problem with it. LGTM!

PR #2196, opened 2018-10-07
Append correct extension to snapshot frames

ferdnyc:
Reading the code for `actionSaveFrame_trigger()`, I happened to notice 
what looks like a minor copy-paste bug. If the save filename chosen by 
the user doesn't have a `.png` extension, instead of adding it, the code 
would've accidentally been appending an `.osp` extension (OpenShot 
project file) instead.

Atom also caught a spurious space character following an `if ... :` 
statement, so I let it fix that as well. That's the other hunk in this 
PR.

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -896,9 +896,9 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         framePath, file_type = QFileDialog.getSaveFileName(self, _("Save Frame..."), framePath, _("Image files (*.png)"))
 
         if framePath:
-            # Append .osp if needed
+            # Append .png if needed
             if ".png" not in framePath:
-                framePath = "%s.osp" % framePath
+                framePath = "%s.png" % framePath
         else:
             # No path specified (save frame cancelled)
             self.statusBar.showMessage(_("Save Frame cancelled..."), 5000);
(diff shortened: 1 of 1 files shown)


peanutbutterandcrackers:
Speaking of fixing bugs, wouldn't it be better if the checks were `if not filePath.endswith('.osp'):` rather than just `'.osp' in filePath`? I am pretty sure I myself would probably create an `osp_portability_test_project` some time in the future once we start working on portable `.osp` files.

Also, "Oh hi, Mark!"

ferdnyc:
> Speaking of fixing bugs, wouldn't it be better if the checks were `if not filePath.endswith('.osp'):` rather than just `'.osp' in filePath`? I am pretty sure I myself would probably create an `osp_portability_test_project` some time in the future once we start working on portable `.osp` files.

Well, that's fine, `osp_portability_test_project` won't match `.osp`, unless you put a dot in front of it.

Only @jonoomph can answer why the test was originally written that way, but my guess would be that it's intended to leave room for something like `myproject.osp.zip` to be a valid and supported `.osp` filename, rather than turning it into `.osp.zip.osp` (lather, rise, repeat). The test _could_ be narrower and more complex, but since AFAIK the current form has never been an issue... `¯\_(ツ)_/¯ `

DylanC:
@ferdnyc - If you click on the "Details" it has a Restart button for that build.

ferdnyc:
> @ferdnyc - If you click on the "Details" it has a Restart button for that build.

Hm. It's gone now so I can't check, but I don't think it did. That may only be available to collabs. Thanks for kicking the hamsters!

PR #2204, opened 2018-10-08
Last CRLF conversions

ferdnyc:
A while back (PR #1891), I converted every Python source file with CRLF (DOS) line endings _except_ `language.py`, because the translation PR was still open at that point. This PR cleans up the tree by converting that file, as well as `presets/chromebook.xml` which also had DOS line-endings.


DylanC:
:+1:

PR #2205, opened 2018-10-08
Implement Properties View

ferdnyc:
An ancient quick-fix that I finally got around to implementing: **Properties View**.

This item (in the **View → Views** menu) is slightly different from both **Simple View** and **Advanced View** in that, rather than completely reconfiguring the OpenShot interface, it merely makes one adjustment: the Properties dock is allowed to occupy the entire left edge of the window, from top to bottom (except the menu/toolbar/statusline). As such, it can be used after _either_ **Simple View** or **Advanced View** (or a custom window layout), to adjust the existing layout. However, both **Simple View** and **Advanced View** will disable the **Properties View** adjustment.

The menu item will use the theme icon `view-list-symbolic`, if such an icon exists. However, since Humanity does not include `view-list-symbolic` (or any usable equivalent), I've also created a simple `properties-view.svg` icon that's used as a fallback.

(A PNG render of the custom icon, though a PNG version is not included in the source tree:) properties-view

#### Initial default OpenShot window state (== Simple View):

image

#### Select View → Views → Properties View from menu

image

#### Resize window and select View → Views → Advanced View

image

#### Select View → Views → Properties View

image

Fixes: #876


peanutbutterandcrackers:
While I was totally fan-girl-ing over this PR (I know no one who'd be as excited as me, except maybe for @N3WWN), I have a few feedbacks... (and I might be asking for a bit too much).

Ideally, it would have been great if properties dock could have covered the entire screen without any "properties view" at all - just drag-n-drop. And I remember us talking about that quite a long time ago. The benefit of that would be that the properties dock could be placed on either side - right or left - of the screen.

However, if that isn't possible, I would like to make this request: If the user has modified the layout (i.e put the properties dock to the right) and then clicks on 'properties view', let the properties dock cover the entire screen on that side. Currently, if I try to do so the dock comes back to the left.

Also, I wonder if there can be some other name than 'properties view' (that's a good one, though) that would be... more elegant? Or maybe just a toggle properties .... sth? Or, what if we had a button on the properties dock itself (next to the close and un-dock buttons) that would do this hulkification?

But thank you, thank you so very much for this!

ferdnyc:
> Ideally, it would have been great if properties dock could have covered the entire screen without any "properties view" at all - just drag-n-drop. And I remember us talking about that quite a long time ago. The benefit of that would be that the properties dock could be placed on either side - right or left - of the screen.

Ideally, sure. If that were possible, this wouldn't be necessary at all. But it's a limitation of Qt's QMainWindow layout that the application is forced to choose which dock area occupies each of the four corners of the window. By default, the top dock area occupies the upper left and right corners, and the bottom dock area occupies the lower left and right corners. In that state, there's no way for any dock widget to occupy the full height of the window unless the top and bottom dock areas are both empty.
 
> However, if that isn't possible, I would like to make this request: If the user has modified the layout (i.e put the properties dock to the right) and then clicks on 'properties view', let the properties dock cover the entire screen on that side. 

Well, that opens up a **huge** can of worms. The thing is, activating Properties View doesn't change anything  about the Properties dock. What it _changes_ is the configuration of the `LeftDockWidgetArea`, giving it preference over the `TopDockWidgetArea` in the upper left corner of the window. The Properties window is then placed into the `LeftDockWidgetArea`, from wherever it was previously (even if it was floating, or not visible), because that's where it has to be in order to benefit from that reconfiguration.

By the same token, the reconfiguration of the `LeftDockWidgetArea` will affect _all_ dock widgets in the interface, not just the Properties dock. If you tried to drag the Effects dock to the left side of the screen after activating Properties View, it too would be full-height. And if you were to activate Properties View and then drag the Properties widget over to the right side, it would (like the Effects dock in its default Advanced View position) only occupy the height of the Timeline (the fixed central area of OpenShot's QMainWindow layout).

Theoretically the Properties View switch could detect if the Properties dock is in the `RightDockWidgetArea`, and make _that_ area full-height instead of the `LeftDockWidgetArea`. But:

1. If you were to place Properties in the `RightDockWidgetArea`, activate Properties View to expand that area to full height, and then drag Properties to the `LeftDockWidgetArea`, Properties would be half-height, the right-hand area would still have upper corner priority even without the Properties dock in it, and you'd have to activate `Properties View` **again** to reconfigure the `LeftDockWidgetArea` for corner priority. At which point _both_ side dock areas would be full-height.
1. So now activating Simple or Advanced View would have to disable upper corner priority on _both_ side `DockWidgetArea`s to fully reclaim their expected layout, instead of (like they do with this change) simply turning off corner priority for the `LeftDockWidgetArea`. 
1. If Properties View is going to detect Properties' position, instead of moving it, then what should it do when Properties is in the `TopDockWidgetArea`? Or floating? If it just defaults to reconfiguring `LeftDockWidgetArea` for corner priority, that would be an invisible change that you'd never know about, until you docked Properties (or something else) there.
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
Which is why:
> Currently, if I try to do so the dock comes back to the left.

> Also, I wonder if there can be some other name than 'properties view' (that's a good one, though) that would be... more elegant? Or maybe just a toggle properties .... sth? Or, what if we had a button on the properties dock itself (next to the close and un-dock buttons) that would do this hulkification?

Two problems there:
1. The dock titlebar is an all-or-nothing thing, the only way to modify it is to reimplement every part of it ourselves.
2. It'd also be misleading because, like I said, _nothing_ about the Properties dock is changed. It's not even resized. Properties View only reconfigures the `LeftDockWidgetArea`. Properties' height will change as a side effect, but only when it's located in that area, which is why the dock is moved there during the reconfiguration. 

I don't know... maybe this is actually a bad idea. It does break the symmetry of the interface, and it breaks the drag-and-drop reconfigurability of the Properties dock's placement in a way that may be unexpected and confusing for users. Worse, the reconfigurations it makes are implicit, and there's no way to turn them off again except via the full reset done by Simple View or Advanced View. I can't just make **both** side dock areas full-height by default, because if the user starts from the default Advanced View then the Effects dock will grow as well, which I think we can agree is undesirable. I'm also loathe to mess with the Preview dock's placement, even if it's just by swapping upper right corner priority from the `TopDockWidgetArea` to the `RightDockWidgetArea`.

OpenShot's interface isn't actually very flexible, by default. The other two View modes completely force the reconfiguration to their parameters, and there's no way to reveal individual docks from the **View** menu — it can only be done implicitly, like by selecting Properties from the a context menu on the Timeline. So, perhaps this kind of interference with the assumptions built into the existing interface, and the user's expectations, is best avoided.

N3WWN:
Thanks for this! :)

ferdnyc:
So, things I just discovered.
1. The Timeline is actually _**completely**_ fine living in a dock widget, instead of occupying the `QMainWindow`'s static center area.
1. With the Timeline docked into the `BottomDockWidgetArea`,  the interface becomes far more reconfigurable.
1. Qt's `QMainWindow` without a center widget (an [unsupported configuration](http://doc.qt.io/qt-5/qmainwindow.html#qt-main-window-framework)) is annoyingly weird.
1. Trying to place a dock into an empty `LeftDockWidgetArea` or `RightDockWidgetArea`, when the window has no center widget and is entirely occupied by the `TopDockWidgetArea` and `BottomDockWidgetArea` contents, is _incredibly_ finicky and frustrating, because the target area is not the entire edge of the window, like you'd expect, but only a very small spot alongside the horizontal splitters.
1. The double splitters created by a missing center widget are likewise quite annoying.

Screencast of the interface in GIF form below. It's tiny, only 5 FPS, and about 10 seconds are cut off at the end, since the max length supported by ezgif.com was 60 seconds @ 5 FPS. So I uploaded the full 1280x800 capture as [an Unlisted YouTube video](https://youtu.be/4zWdewfci9E).

openshot-dockwidgettimeline

It's... really _almost_ usable, except for all the ways in which it isn't usable. (Ignore the missing title on the Timeline dock widget, that's just me not setting it correctly. It was a quick-and-dirty test.) Unfortunately, I don't expect that Qt would make it very easy to get it the **rest** of the way to usable.

But it was an interesting experiment, and shows that a more flexible interface might be posible. I'm not planning to explore it further, but I placed the code that I was running when I recorded this into a branch on my fork: [dockable-timeline](https://github.com/ferdnyc/openshot-qt/tree/dockable-timeline).

ferdnyc:
P.S: 
* You'll note when I first start playing with the interface, I activate Advanced View and the dock placement gets all weird. 
* That's because Advanced View reconfigures the Properties and Effects docks into the `Left`- and `RightDockWidgetArea`, respectively, which are currently shrunk down at the middle of the window.
* Because the Timeline is in the (normally-unused) `BottomDockWidgetArea`, instead of its usual center position, I then have to move them both down to sit alongside it. 
* After that, activating Properties View reconfigures the side docks with corner priority, so that their contents can be full-height.

ferdnyc:
P.P.S: Oh, and kudos to [Open Broadcast Studio](https://obsproject.com), an extremely capable screen recorder. (I still haven't found anything for Linux that will illustrate mouse clicks or provide any other sort of advanced demonstration features, though.)

ferdnyc:
P.P.P.S> OBS Studio, which also has a Qt interface, uses their _preview_ area as the center widget, and places all of the other panels around it as docks. (The **top** dock area is normally unused, in their setup.) That works out well for their needs, but has the same caveat as OpenShot's interface: There's no way to place a dock on the side of the window so that it occupies the full height. The side docks have the same height as the preview area.

An interface screencast from the OBS version 20 release announcement:
obs-interface

peanutbutterandcrackers:
Wow. After seeing that screen-cast, I think *that* should be how Openshot's layout internals should be made to be by default - the kind that it is when the `properties view` is activated. I love it! Perhaps we should get @jonoomph and @DylanC in the loop and consider moving towards that? Highly configurable layout like that is just what we need!

ferdnyc:
> P. S: Just checked out the branch. Nothing seems disadvantageous about going that route to me, though. We wouldn't even need a separate properties view. It would just work.

Oh, there are _definitely_ some very real issues with the center-widget-less interface. Primarily, the two I mentioned in my original comment:

1. The double horizontal splitters, and the potential for a wasteland to be created between them. It's just ugly to have the divide between the top and bottom dock areas not simply take the form of a movable partition, the way you'd expect. Rather, there's a completely empty **chasm** there that's just waiting to be accidentally opened up.
    
    ...Aaaaaand, I just pretty much solved that problem. 

    By imposing a 1-pixel-high, fixed-height widget between the two splitters, as the `QMainWindow` center widget, I've effectively locked the two splitters together. They still show a double handle, but other than that it works perfectly. You can grab anywhere on or between either one, and it acts as just one big splitter that divides the top and bottom dock areas. 

    I know I said I wasn't going to do any more with this, but I had that idea so I had to try it. The code on my branch has been updated with that change, and another to set the title of the Timeline dock and make it non-closeable. (It can still be moved, floated, and resized.)

2. The _other_ big issue with this, though — and one that isn't as easily solved (or at least, nothing's come to mind) is the difficulty in targeting the left or right dock areas when the center widget is basically nonexistent. _**Ideally**_, you'd want the entire right or left edge of the window to be the target drop zone for dropping docks there, but it isn't. The target is actually a tiny spot alongside the horizontal splitter. In practice, I personally find that **intensely** annoying. YMMV, of course. 

    Regardless, Qt doesn't appear to offer us a whole lot of control over it, though there may be ways to play tricks with size or layout constraints that could help. I'm not sure, TBH the `DockWidgetArea`s are one of the very few Qt features that they treat like black magic, providing no real documentation regarding their implementation or functionality. Any other standard Qt widget, you could subclass it and make changes. But I don't get the impression the `DockWidgetAreas` _are_ extensible Qt widgets, they're just magical aspects of the `QMainWindow` class.

> P. P. S: I can already see an added advantage of having timeline as a QdockWidget. If I have many tracks, I can now see all of them at once. If I am doing multi-layered animation, this should be a pretty neat feature.

Well, see... now I'm confused, because _that_ layout ­— the one you have in your screenshot — was easily achievable with the _existing_ OpenShot interface. As long as you move everything out of the top dock area, and use only the left and right dock areas, the timeline grows to full height and so do any docks alongside it.

Here's the exact same arrangement, in OpenShot 2.4.3 release without the dockable Timeline:

image

So, while the dockable/floatable/movable Timeline does indeed enable more an varied window configurations, that actually **isn't** one of them! :laughing:

peanutbutterandcrackers:
Hmm... I see. Once I closed all the top-widgets, that layout was achieved. I didn't know that.

I just checkout out the recent changes and bumped into a few issues I hadn't discovered before.
1. The default `simple view` and `advanced view` are both messed up:
Simple view has too big time-line:
simple view layout
And advanced view is just...:
advanced view
However, if I toggle `properties view`, everything becomes normal. Except for simple view, the time-line is still huge.

2. If I click on a clip and click properties while in `simple view`, I get this mess:
properties meh
However, if I toggle `properties view`, everything becomes neat again.

So, what I think would be the solution for that branch would be to not have a separate `properties view` toggle in it, but just have it as default (except for the actual placement of the properties dock to pillar aspect). [Maybe that is exactly what that toggle does, but what I mean is, if there is some setup before actually placing the properties widget in pillar aspect, have that as default in that branch]

Subsequent layout-changes (using drag-and-drop) while in `properties view` didn't trigger the chasm to appear.

So, the only real problem now, should my shots-in-the-dark suggestions should work, would be problem no. 2. And yes, I find it to be quite annoying too.

ferdnyc:
> 1. The default `simple view` and `advanced view` are both messed up:

I didn't make any changes to Simple or Advanced View, so they simply apply the same stored configurations they always have, ignoring the Timeline completely (as always). They'll definitely mess up the dockable-Timeline interface, because they're set up with the assumption that the Timeline is the center widget.

If you'll recall, I originally introduced that branch with:

* You'll note when I first start playing with the interface, I activate Advanced View and the dock placement gets all weird.
* That's because Advanced View reconfigures the Properties and Effects docks into the Left- and RightDockWidgetArea, respectively, which are currently shrunk down at the middle of the window.
* Because the Timeline is in the (normally-unused) BottomDockWidgetArea, instead of its usual center position, I then have to move them both down to sit alongside it.
* After that, activating Properties View reconfigures the side docks with corner priority, so that their contents can be full-height.

The point was to test the dockable timeline, not to build a whole new, workable interface. The existing stored views would have to be modified to account for having the Timeline in a dock. The only intended-usable View mode is the Properties View, which configures the left and right dock areas for corner priority (if that's not already set, say because loading Simple View or Advanced View removed it).

ferdnyc:
(With the last change I made to configure the center with a `(flex)` × `1px` blank widget, using Simple View or Advanced View is actually the _only_ way to get the central area to open up larger than that. They're stored with the center widget dimensions configured to accomodate the Timeline, and their stored dimensions can override the constraints set for the center widget. But after things are moved around so that the center widget is compressed to `1px` tall, and as long as the left and right docs are configured with corner priority by the Properties View so they won't stretch it out again, it'll stay flattened and keep the double splitters tied together.)

peanutbutterandcrackers:
Hmm... I see. My bad. Sorry.

I really like that branch. I hope we can go that route for the interface.

So, if we make the changes that are triggered by clicking 'properties view' be there by default, both of the problems will be solved! [Excpet, maybe let's not have the properties dock show while the user is in simple view.

And the only problem that we'd be left with will be this issue:
>The other big issue with this, though — and one that isn't as easily solved (or at least, nothing's come to mind) is the difficulty in targeting the left or right dock areas when the center widget is basically nonexistent. Ideally, you'd want the entire right or left edge of the window to be the target drop zone for dropping docks there, but it isn't. The target is actually a tiny spot alongside the horizontal splitter. In practice, I personally find that intensely annoying. YMMV, of course.

...the solution to which, I am sure, will dawn upon you soon enough. :+1:

DylanC:
@ferdnyc @peanutbutterandcrackers - I'm a bit confused here. Do we need this PR or did I read something saying we can achieve the same thing currently (more or less) without this PR?

peanutbutterandcrackers:
@DylanC - It's just that there's [another branch](https://github.com/ferdnyc/openshot-qt/tree/dockable-timeline) that is better than this PR, IMHO, but requires some of the changes in this PR to be made default, in that branch, for now. Only 2 issues remain in said branch and if they are resolved, I am highly in favour of merging that branch into the source rather than just this.

(Kinda' reminds me of the PR where the click event thingy in the tutorial window was being fixed but it turned out to become a complete rewrite of the tutorial window itself...)

ferdnyc:
> (Kinda' reminds me of the PR where the click event thingy in the tutorial window was being fixed but it turned out to become a complete rewrite of the tutorial window itself...)

The difference being, I have no plans to submit that PR. My branch is there for anyone who's interested in fleshing it out. But, as things stand, no I don't think we need this PR. It doesn't seem like a particularly popular way to address the issue.

peanutbutterandcrackers:
@ferdnyc - I think the *only* thing that branch needs is for the changes in this PR to be made the default (without having to manually toggle the `properties view` on) - except for the actual placement of the `properties` dock on the left hand side; if that is done, it'll be all ready to be merged, really. The really small target drop zone for the dock to gain a pillar aspect ratio is something I, and surely a lot of other users, could live with. It would still be a thousand times better than the current thing we've got going on. Had I only known half as much as you I would have gladly fleshed it out. But, as things stand, I can't. And therefore, I would like to most humbly request you, with utmost sincerity, to complete what you have started. Please!!!

ferdnyc:
[Me](https://github.com/OpenShot/openshot-qt/pull/1883#issuecomment-427915655):
> So the changes can get merged, and two months later jonoomph can parachute in and claim they're holding up the release? No thanks.

peanutbutterandcrackers:
Me:

You are a well respected contributor with two accepted major changes: 1. The tutorial window re-write and 2. The locale thingy; the latter one even earning you a [public expression of gratitude from Mr. Thomas himself](https://www.openshot.org/blog/2018/09/22/openshot-243-released-animated-masks-language-improvements-improved-stability-and-more/).
>Huge improvement to language translations (Thanks to Frank Dana - ferdnyc)

Apart from that you have made tons of small changes that have polished OpenShot so much.

So I wouldn't think you need to be that weary about your work going in vain. But I'll go ahead and bring the branch to the attention of everyone at slack (again) just to make sure. I am pretty sure I am not the only one in favour of this. But I'll (try to) get some more devs to chime in. :)

jonoomph:
Thanks again for your work on this @ferdnyc ! Merged into develop now.

PR #2212, opened 2018-10-10
WIP: Stopped Progress Got Stuck- Setup toggle buttons for timeline actions

speedytechdev:
This makes it easier for people to work on video only or only work on audio. (More information coming soon)


speedytechdev:
What would be the best way to toggle the audio and video. For example, set the screen black when the video is toggled. Some help would be great thanks.

peanutbutterandcrackers:
@DylanC - I just realized, this might be a possible duplicate of #1993? It wouldn't be highly beneficial to have two contributors working on the same thing at the same time, differently, would it?

speedytechdev:
@peanutbutterandcrackers I saw the pull request that Dylan made and I saw it was out of date with the main branch so I thought I might work on it. So yes working on it might mean we can get it out sooner.

speedytechdev:
I have a problem and I am not sure how to fix it I was wondering if anyone could help. So I have set up the toggle buttons but the problem is if I turn off audio for one track it will do it on all the tracks. Is there anyone I can fix this?

jonoomph:
This looks like a WIP at the moment, so I updated the title.

DylanC:
@jonoomph - This PR won't be updated any time soon I reckon but it is a much requested feature from the Openshot 1.x series. Hopefully someone with the right knowledge can implement. 😉

Issue #1829, opened 2018-07-10
Media relinking in file properties


fschill:
I'm working on a feature to be able to re-link media files if they have moved (I already fixed the bug to relink them during opening if the project has moved, but would like more flexibility now).

I had the problem that I moved a project to another computer, and OpenShot got stuck in a loop asking where the media files are. I've fixed this problem by auto-linking all files in the same folder, so that it's enough to point to the first file (assuming they are all in the same directory. This works (see my fork), and I'll submit a pull request shortly. 

However, while doing this I thought it would be great to simply change the path in the file properties, and relink media to a new file, e.g. it could also be a recoded version in another video codec (different extension). I seem to have a problem with the MP4 files from my phone, they cause stuttering video output with skipping frames every now and then, both in preview, and in export (unrelated issue). The problem goes away if I recode the videos to MJPEG. So now I'd like to recode videos in existing projects, and relink the media files to the new file.

I have added this feature in file_properties.py, and it seems to work for the media files in the project view. However, all clips that are already in the timeline are not updated for all I can tell. They still show up with the old name (if I drag a renamed file in, it has the new name). Also, when saving and reloading the project, the log shows that it is still loading the old files.

Here is my question:
Can someone explain to me how clips in the timeline are linked to files in the project view? If I delete a file, it is removed from the timeline as well, so clearly there is a listener mechanism to let clips know if something happened to a file. But it does not seem to work for changing file locations.

Your help would be much appreciated. 
PS: my current attemtps are on my fork. Media relinking attempts are in the branch media_relink: https://github.com/fschill/openshot-qt/tree/media_relink

fschill:
Anyone have any hints? I'm stuck on this, need some insight or at least some pointers how OpenShot data structures are managed.

peanutbutterandcrackers:
Wow. You're working on it right now!? Awesome!

Sorry for the tardy reply.

This seems to be related to #1061. However, I would also like to suggest you take a look at #1325.

Pinging @ferdnyc @N3WWN @DylanC @jonoomph here!

fschill:
thanks - yes, I decided that I want to use OpenShot regularly now instead of iMovie (my desktops are all linux, and working on laptops is not as nice). I had a few issues and started looking into the code to fix them. So, time permitting, you might hear more of me in the future... :)

Reading #1325 was useful, it matches what I thought while debugging this. Up to now, my ideas were:

[...]

PR #2276, opened 2018-10-24
Media relinking in file properties dialog

fschill:
(addresses issue "Media relinking in file properties #1829" )
The purpose of this pull request is to allow users to re-link individual media files in the file property dialog. This is useful when moving files to a new location (this could also be done by moving and reloading the project, and relinking while opening). It is also needed if a user wants to re-encode a clip into a different format for editing (e.g. lower bitrate), and link it to the full-quality version at the very end. 

- added a browse button in file properties, to allow the user to re-link individual media files in a project. 
- when changing the source file, the project and timeline are automatically updated with the new file path/name. This is done "brute-force" by looping through all clips and updating as necessary.
- at the moment there are no sanity checks that the user actually selects a compatible file (e.g. the same file in a different location, or the same clip with same length but in a different format). If a completely unrelated file is selected, I didn't detect any serious issues (e.g. crash), but the thumbnails, clip length, etc. are not updated, and it can lead to unexpected behaviour.


peanutbutterandcrackers:
This looks neat too! A great feature to have.

However, this only worked for me once. When I tried to do it again (with not-downloaded-from-9gag-webms but proper footages), OpenShot crashed. Here are the logs:
[libopenshot.log](https://github.com/OpenShot/openshot-qt/files/2523007/libopenshot.log)
[openshot-qt.log](https://github.com/OpenShot/openshot-qt/files/2523008/openshot-qt.log)

But I might try running the tests again. Maybe it's just this computer.

Also, if it could be done, it'd be really great to have the thumbnails, clip-length, etc. all be updated, before the merge.

Neat stuff. Thank you, good sir!

peanutbutterandcrackers:
@ferdnyc and @DylanC - Perhaps we should test this one out a bit more and merge it in, and deal with polishing it up later? Incrementally? How does this PR work for you?

ferdnyc:
Haven't even had a chance to glance at it, sorry.

In general, I don't think there should be any rush to merge new code. If sometthing needs improvement or further development, it can also be improved _before_ it's incorporated into the codebase that developers and testers expect to be able to run from, and which any further releases will be based on. It's one thing if features are built up incrementally, but **testing** shouldn't be a _post_-merge activity!

DylanC:
@peanutbutterandcrackers - I'm looking into #2312 first since it relates to the earlier PR #2275 and I would rather have that PR working close to perfect before moving onto this one. (if you don't mind)

peanutbutterandcrackers:
Hmm... True. Okay then (I guess I was a bit too anxious)

DylanC:
@fschill @peanutbutterandcrackers - I'm happy enough with this. It does work for me at least.

Maybe could we make these lineEdits/text inputs a bit longer to show more of the path?
path_len

Also @fschill would you be able to look at sorting out the conflict? I'm afraid to go at it in case I include the wrong section of code.

fschill:
I'm currently traveling - maybe I'll find some time next week. Not sure 
which conflict you mean, it was up to date when I submitted the PR. Has 
something changed?

ferdnyc:
> I'm currently traveling - maybe I'll find some time next week. Not sure
> which conflict you mean, it was up to date when I submitted the PR. Has
> something changed?

Yes, in `src/classes/project_data.py` — I can't see the details, but my guess is the merge of my #2311 caused the conflict.

DylanC:
@fschill - I could try to fix the conflict here but I would be worried about including the wrong code. I would feel better if someone familiar with this code could resolve the conflict. 

Its probably caused by merging @ferdnyc changes before yours.

src/classes/project_data.py
```diff
@@ -906,42 +906,64 @@ def check_if_paths_are_valid(self):
...
-                    path = os.path.join(starting_folder, file_name_with_ext)
+                # try to find clip with previous starting folder:
+                if starting_folder and os.path.exists(os.path.join(starting_folder, file_name_with_ext)):
+                    # Update clip path
                     file["path"] = path
                     get_app().updates.update(["import_path"], os.path.dirname(path))
-                else:
-                    log.info('Removed missing file: %s' % file_name_with_ext)
-                    self._data["files"].remove(file)
+                    log.info("Auto-updated missing file: %s" % file["path"])
```
ferdnyc: It actually looks like this line might be the _only_ conflict here (it's certainly one of the only) — and it only conflicts because my commit changed it to
```python3
log.info("Auto-updated missing file: %s" % path)
```
whereas @fschill changed it to
```python3
log.info("Auto-updated missing file: %s" % file["path"])
```
...however, those lines are _literally_ equivalent, just two lines earlier we set `file["path"] = path`!

So, it does not matter the _tiniest_ but which version you go with. If that line is the only conflict, then it can be resolved and merged. (If not, I guess let me know what the other conflicts are? I don't appear to have visibility into that, as a non-collaborator non-submitter. ...Or if I do, and I'm just not seeing how, let me know how I can look over the other conflicts.)

ferdnyc:
Oh, wait, nope! Never mind, I found it.

The issue (I _think_) isn't that this branch conflicts with mine, but that it already includes a commit (dddef07c) that was merged in @fschill 's previous PR #2275. I couldn't figure out why I wasn't seeing nearly as many changes as I expected, even after I manually merged the `fschill:file_media_relink` branch with `develop` (and received no conflicts) — it's because it wasn't reapplying `dddef07c`, which is showing as part of this PR but really isn't since it's already merged.

My `ferdnyc:testmerge` branch is a merge of this into develop, which went cleanly. I just opened that as PR #2375, just so we can move past the conflict.

ferdnyc:
> The issue (I _think_) isn't that this branch conflicts with mine, but that it already includes a commit ([dddef07](https://github.com/OpenShot/openshot-qt/commit/dddef07cb5fc058822d6f8127efd58ee01842573)) that was merged in @fschill 's previous PR #2275.

...And that code _was_ subsequently further changed by my #2311, so by including  this PR is effectively wanting to go back to the previous version of `src/classes/project_data.py` **without** my #2311 fixes. 

So, _that's_ the source of the conflict-that-isn't-really-a-conflict. ...I _think_. **#GitIsConfusing**

PR #2312, opened 2018-11-02
[WIP] project_data: Store thumbnail paths aliased

ferdnyc:
Taking a page from the `@transitions` aliasing for OpenShot's internal thumbnail paths, on project file creation/saving this converts `clip["image"]` paths (the thumbnail corresponding to the given media file) to `"@thumbnail/CLIP_ID.png"`, instead of converting to relative file paths anchored at the project file location (e.g. `"../../home/user/.openshot_qt/thumbnail/CLIP_ID.png"`).

On project file _load_, these `@thumbnail` aliases are replaced with `info.THUMBNAIL_PATH`, the absolute path to the active thumbnail cache.

Please kick the tires on this a _bunch_, since I'm very open to the possibility that I may have overlooked some aspect of the project_data management which is not accounted for in this change. But in my testing:
1. It works for all media types (including images and titles)
1. It doesn't interfere with or affect the moving of thumbnails to `$project_file_location/thumbnail/` directories on save
1. And, in concert with @fschill 's file-recovery changes, it restores clip thumbnails for recovered files.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -4,12 +4,14 @@ about: Create a report to help us improve OpenShot
 
 ---
 
+**Please fill out this template** 
+
 **Describe the bug**
 A clear and concise description of what the bug is.
 
 **System Details (please complete the following information):**
  - Operating System / Distro: [e.g. Windows 10, Linux Mint 17.1]
- - OpenShot Version [e.g. 2.4.1]
+ - OpenShot Version [e.g. 2.4.3]
 
 **To Reproduce**
 Steps to reproduce the behavior:
--- a/.github/ISSUE_TEMPLATE/Feature_request.md
+++ b/.github/ISSUE_TEMPLATE/Feature_request.md
@@ -4,6 +4,8 @@ about: Suggest an idea for OpenShot
 
 ---
 
+**Please fill out this template** 
+
 **Is your feature request related to a problem? Please describe.**
 A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
 
--- a/.github/ISSUE_TEMPLATE/i-have-a-question.md
+++ b/.github/ISSUE_TEMPLATE/i-have-a-question.md
@@ -4,6 +4,8 @@ about: Ask a question about obtaining, installing, or using OpenShot
 
 ---
 
+**Please fill out this template** 
+
 <!--Give your question a descriptive title so that others will be able to understand at a glance how they can help. Then, explain in more detail under "My Question" below -->
 
 **My Question:**
@@ -15,7 +17,7 @@ about: Ask a question about obtaining, installing, or using OpenShot
 
 **System Details**
  - Operating System / Distro: [e.g. Windows 10, Linux Mint 17.1]
- - OpenShot Version [e.g. 2.4.1]
+ - OpenShot Version [e.g. 2.4.3]
 
 **Screenshots**
 <!-- If applicable, add screenshots to help illustrate your question. You can include screenshots by copy/pasting them here, dragging-and-dropping into this form, or clicking below and loading images saved to your computer. All images are public, so please don't post screenshots containing personal information. -->
--- a/installer/build-server.py
+++ b/installer/build-server.py
@@ -33,7 +33,6 @@ PATH = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))  # Primary o
 import datetime
 import platform
 import shutil
-from slacker import Slacker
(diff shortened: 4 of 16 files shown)


DylanC:
@ferdnyc - Does this work for only new projects created with this change in place? I'm using the test project located in #2275 and it does not seem to restore the thumbnails for me.

thumbnails_test

ferdnyc:
>> @ferdnyc - Does this work for only new projects created with this change in place? I'm using the test project located in #2275 and it does not seem to restore the thumbnails for me.
>> 
> Yeah, it won't do anything unless the project file was saved with this code in place

Doesn't have to be a new project, though, any save will do. Even an old project file, if it's loaded in with currently-working thumbnail paths, will have them all upgraded to the more resilient form the next time it's saved. 

That works because this change doesn't alter _anything_ about how the project data is managed in-memory by OpenShot — the only modifications are to the (already-complex) conversion process that's performed on the in-memory data before writing it out to disk, and then to the reverse conversion process that's applied to the `.osp` file data when loading it into memory. When in memory, all paths are absolute, same as ever. When saving, previously thumbnail paths were converted from absolute to relative; with this change they're converted to aliases instead.

> However, project files _**written**_ after this change goes into effect will no longer be fully compatible with older versions of OpenShot, since it won't have any idea how to parse the `@thumbnail` part of the path. (In practice I don't know what that would actually affect, whether the files would simply load without thumbnails or whether they'd cause some sort of error/crash in older versions.) So, that's something we should be aware of and keep in mind.

If that's a concern, at the cost of some extra complexity it could be worked around, and the project files written by the new code could still be fully backwards-compatible with older OpenShot versions.

If we wanted to do that, how it would work is:
1. Instead of modifying how the `"image":` property of the clip data is formatted, I'd add a **new**, separate property — let's say `"thumb"`. 
1. The `"@thumbnail/FILENAME.png"` aliased form of the thumbnail path would get written to the `"thumb"` clip property.
1. The full relative path would still get written to `"image":`, same as always
1. The new code would know to look for `"thumb":` properties in project files, and read the `clip["image"]` data from `"thumb":` instead, ignoring the `"image":` path if it is. If a `"thumb":` property isn't found (older project file), then it would fall back setting `clip["image"] from the relative path in the `"image:"` property.
1. _Older_ versions of OpenShot would still be able to read the `"image":` property into `clip["image"]` same as always, since it's still formatted in the same manner it's always been and they just have to do the same relative-to-absolute path conversion. Those versions would completely ignore the `"thumb:"` property simply because they have no idea what it means or what to do with it.

At some point down the line a ways, when nearly all project files contain `"thumb":` properties and all versions of OpenShot in use know how to deal with them, then the `"image":` property could be deprecated from the project file data format, and the legacy code for reading/writing it dropped from the source.

(And at some point long before that, probably, newer versions of OpenShot would store and use the thumbnail path in-memory as `clip["thumb"]` instead of `clip['image"]`, simply because loading `"thumb":` into `clip["image"]` is needlessly-confusing long term.)

peanutbutterandcrackers:
I'd just tested a `.osp` file created using this branch, with v2.4.3 and it looked like this:
screenshot at 2018-11-03 21-01-39

Yeah. Backwards-compatibility would be great. I think we should go that route.

Test Results:
I moved a test project from `~/Videos/Test/project.osp` to '~/Test/project.osp` (with all the footage intact in the same parent directory - `Test`. And it just worked. Without any complaints. The thumnail worked neat too.

However, when I moved the project.osp file, alone, to a separate location, things seemed to work but the thumbnails were still broken. But then again, that is not what this PR addresses (thumbnail regeneration), yet, right? So, so far this works as expected.

Still. Backwards-compatibility for the win.

Thank you, BTW, for this. You're awesome, as always. :+1:

ferdnyc:
Huh, that's actually a lot better than I feared, if it merely loads the file with missing thumbnails, instead of crashing or breaking in some way that makes the file unusable. Still, I agree backwards-compatibility would be preferable. \*sigh*

peanutbutterandcrackers:
@ferdnyc - Any chance of this PR being updated soon, good sir? :)

ferdnyc:
I've had it on hold to focus on #2400, since that was the far bigger issue. Hopefully that can be merged soon and (with luck) will turn out to improve performance for people who've been having issues. 

If all goes well there, I can hopefully circle back around to this one some time next month.

peanutbutterandcrackers:
Ah... I see. It's just that I would have liked share some more OpenShot projects with my friends and that would have been a nice thing. [I'm just asking for selfish reasons :D ]

Anytime you see fit, senor! [But please make it quick if you can :smile: ]

PR #2337, opened 2018-11-07
Hardware support

luloi123:
le logiciel me marque quil ne repond pas quand je fait exporter la video


DylanC:
@luloi123 - This is not where to report an issue - Moved to here #2338.

PR #1828, opened 2018-07-10
Add commandline processing to launch.py

ferdnyc:
This adds standard Python argparse handling for command-line arguments to
launch.py, which now responds to `--help` with:

```console
$ src/launch.py --help
Loaded modules from current directory: /home/ferd/rpmbuild/REPOS/openshot-qt/src
usage: launch.py [-h] [--list-languages] [-V]

OpenShot version 2.4.2

optional arguments:
  -h, --help        show this help message and exit
  --list-languages  List all language codes supported by OpenShot
  -V, --version
```
There is also commented out code for another argument, `--lang`, to set the
interface language on the command line via language code (as listed using
`--list-languages`). That will require the changes from from PR \#1795
before it can be enabled.

PR #2355, opened 2018-11-14
Pass remaining command line args to Qt

### LINKED ISSUE #1872: Opening a previously saved .osp file shows only a blank project [macOS]
Steps to reproduce:
1. Open a new project
2. Import a clip or any other file
3. Select “File > Save project”, assign a project name (without weird characters or blank spaces) and save in whichever folder you wish
4. Close OpenShot
5. Double click on the saved file: a blank project is displayed. The only way I found to reopen it is to select the desired project from “File > Recent projects”.

Software version (for both creating and trying to open the saved file): 2.4.2
Platform: Mac OS X 10.11 El Capitan

### LINKED ISSUE #2162: Unable to open files via command line
Hello,
I installed the latest version (2.4.3), and I noticed the following bug: files, either .osp or any media files, can not be opened in Openshot from any file manager nor from the terminal. It seems to be a regression, as it was working perfectly with the previous version (2.4.2).

System details: Ubuntu 14.04, Openshot 2.4.3.

To Reproduce:
1) Create a project file 'project.osp' with Openshot
2) In a terminal, run: openshot-qt /path/to/project.osp
→ expected result: openshot should open the file
→ what I get: Openshot does not open, with the message: "openshot-qt: error: unrecognized arguments"

It does not work even with a path without spaces or special characters. 
Furthermore, files cannot be opened from the file manager. But I can open the files successfully from within Openshot, using the menu. So that it seems that the problem comes from the command line argument for opening files!

ferdnyc:
Make sure we don't error out when a project file name is passed on the command line, which used to work until I broke it with #1828 .

Fixes #1872 , fixes #2162

### DIFF
--- a/src/launch.py
+++ b/src/launch.py
@@ -41,7 +41,7 @@
  """
 
 import sys
-from argparse import ArgumentParser
+from argparse import ArgumentParser, REMAINDER
 
 try:
     from classes import info
@@ -67,6 +67,7 @@ def main():
                         action='store_true', help='List all language '
                         'codes supported by OpenShot')
     parser.add_argument('-V', '--version', action='store_true')
+    parser.add_argument('remain', nargs=REMAINDER)
 
     args = parser.parse_args()
 
@@ -94,8 +95,11 @@ def main():
     log.info("   OpenShot (version %s)" % info.SETUP['version'])
     log.info("------------------------------------------------")
 
-    # Create Qt application
-    app = OpenShotApp(sys.argv)
+    # Create Qt application, pass any unprocessed arguments
+    argv = [sys.argv[0]]
+    for arg in args.remain:
+        argv.append(arg)
+    app = OpenShotApp(argv)
 
     # Run and return result
     sys.exit(app.run())


DylanC:
@ferdnyc - Thanks, I did feel the issue had something to do with those changes in #1828 but I didn't want to call it out unless I was completely wrong. 😆

ferdnyc:
https://github.com/OpenShot/openshot-qt/blob/f9cbde8513b7e1dd45c2598823a4eae488ec1c2e/src/classes/app.py#L166-L169

P.S: My favorite thing about the commandline processing in `app.py` is that `if (len(args[0]) == 2)` ­— so if you pass _two_ filenames on the commandline, or any other argument in addition to a project filename, it ignores **all** of the arguments.

(...However, that doesn't apply to the arguments recognized by `argparse` in `launch.py`, because it processes the `argv` first. Only the remaining unprocessed arguments (if any) are passed through to the `QApplication` code.)

ferdnyc:
> FYI to any Fedora users, right after I fixed this I deployed new `openshot-2.4.3-2.$arch.rpm` packages to all supported Fedora releases @ RPMfusion, so the fix has already been released for anyone who installed from the `rpmfusion-free` repo.

Whoops, sorry, jumped the gun on that. The new builds are released to rawhide, but they're in <code>rpmfusion-free-updates<b>-testing</b></code> for F27-F29. They haven't been pushed to the main repos yet, but it should only take a few days at most.

PR #2375, opened 2018-11-19
Contents of PR #2276 without duplicated commit

ferdnyc:
This is `fschill:file_media_relink` merged into current `develop`, which git was happy to do with no conflicts on my local system. So, I suppose this would be a candidate for merging instead of #2276.

(If what's here is ready for merge. I feel like we weren't totally clear on that yet in #2276, but then got hung up on the conflict.)


DylanC:
@ferdnyc - Thanks. I think we were happy enough apart from the input could show more of the input field.

PR #2400, opened 2018-11-25
Discard history on project-file save/reload

### LINKED ISSUE #1923: Should we store the history stack in project files?
Bug #1768 deals with an issue loading project data into OpenShot, and while it's not yet _certain_ exactly what caused the issue, one of the things I noticed in examining it was the `"history": { ... }` data included in the project data.

In that particular case, the saved history data alone accounted for more than _half_ of the entire (1.4 MB) project data.

Even examining my own, much simpler test project files, the history sections occupy...
* 285/464 formatted lines (61.5% by line count)
* 5929/9197 formatted lines (64.5% by line count)
* 2020/2586 formatted lines (**78% (!!)** by line count)

Which leads to the obvious question: Should we be flushing the undo/redo history on save? (Or, if not flushing it, then at least omitting it from the save file. It can stay in memory until the end of the session.)

It's worth noting that, currently, the undo/redo history **IS** restored on project load, and can be accessed just as if the new session was a continuation of the old one. So, arguably there's some value to having the history stored in the save file. 

However, it's worth at least asking the question of whether that's really _useful_ in practice, and whether any users really do ever load a project and then immediately Undo or Redo any actions from before saving. I only have my own anecdotal opinion to go on, but I don't think that would ever even have _occurred_ to me. I've certainly never actually done it.

ferdnyc:
On both load and save of project file data, any `"history"` in the JSON will be discarded, replaced with an empty struct:
```python3
self._data["history"] = { "undo": [], "redo": [] }
```

Temporarily, while testing the feature, the total size of the data structure being discarded will first be computed and logged. (I added a `total_size()` function, taken from a python recipe — there's a link in the comment header — to `classes/conversion.py`, which maybe isn't the best spot but oh well.)
```python3
log.info("Discarding history of size {} bytes".format(total_size(self._data["history"])))
# Produces:
# project_data:INFO Discarding history of size 105246892 bytes
```

(That's @peanutbutterandcrackers' 12MB mega-exploded project file — turns out, that history turns into _**One Hundred Five**_ megabytes in memory!)

In my testing, this works seamlessly. History operates normally when _using_ OpenShot, and can be undone and redone as usual. Even after saving a project, the history is still there _in that session_. However, when you load a project, instead of the Undo history being populated, it's now always cleared, so history **only** persists through the session. 

If you load old project files and resave them, the history will be removed. (Actually, that can happen anyway with autosave — I've accidentally lost the long history in more than a few of my test projects already, in testing this code out. Which is making it harder to actually test! But, in a good way. And I could always  turn off autosave.)

Kick the tires, see what you think. Hopefully this gives OpenShot a kick in the ass, performance-wise, especially for people with project files that have ballooned in size. And it should also eliminate any history crashes caused by absolute paths to AppImage mounts.

Fixes #1923


ferdnyc:
P.S> 56fa87a adds logging at the start of project-file save and load operations — previously it only logged at the _completion_ of those operations, which is no help from a debugging standpoint. You want to know if anything untoward happens _during_ the save/load operation.

DylanC:
@ferdnyc - I'm supposed to be on a break from this project **but** I'm very interested and excited to test this change. This is the best PR ever for the project in recent times and could solve a whole lot of bad experiences related to performance issues across the board.

DylanC:
@ferdnyc - I had a brief test of this and it didn't appear to be saving any undo history during my session. I tested both an old project and created a new one.

Should I have deleted the Openshot preferences folder before testing this?

ferdnyc:
> @ferdnyc - I had a brief test of this and it didn't appear to be saving any undo history during my session. I tested both an old project and created a new one.

Meaning, undo didn't work _within_ your session? Oh, it's not supposed to break that! I thought I'd tested that from every angle but I may have missed something.

> Should I have deleted the Openshot preferences folder before testing this?

No, there are no preferences changes or changes to the default project, only save/load of project file data are _supposed_ to be affected. But maybe I missed something in how new-project setup is handled. I'll test through project creation, especially, more thoroughly and carefully when I can... later tonight or tomorrow, hopefully.

ferdnyc:
> I had the project file open in an editor and for some reason expected the undo json to be populated with data.

Yeah, that's specifically what this PR does away with, because of all the problems (performance, path issues, outright crashing) it's caused to be storing history.

I still feel it's a nice feature to provide history persistence across sessions. But _only_ if it can be done safely, reusably, and without destroying OpenShot's performance, which the current implementation falls short of.

ferdnyc:
https://github.com/OpenShot/openshot-qt/blob/e8633bfaa7d2499fb1cd408268a645c805af4c48/src/classes/updates.py#L72-L84

My suspicion, BTW, is that the bug that blows out the history lives right here in the code. You can see where it tests `if "history" in data_dict` right at the end, intending to drop the history — but unless we're in the `only_value` case and `data_dict` is a `deepcopy()` of `self.values`, it'll **never** contain `"history"`. We can SEE all of the keys it contains — they're explicitly set in the lines immediately previous. `"history"` isn't one of them.

I think the test should be something different, at least for the _non_ `only_value` case. (It may be fine, and need to be kept, for the `only_value` case.) I'm just not sure what the correct fix is, because I'm not clear on the intent of that code. (The comment is more narrative than explanatory.) But most likely it's either trying to examine the key, or the value. So, maybe...
```python3
if "history" in data_dict["key"]:
  do_something() # though I'm honestly not sure what
# or
if "history" in data_dict["value"]:
  data_dict["value"].pop("history")
```
(I know that updates with `"key": [ "history" ]` do exist, and they're the ones that blow out the project files. So, the first option may very well be correct. But I still don't know what `do_something()` should be in that case. Quite possibly, it should be to just return from the function completely at that point, bailing on doing the update processing since _history_ updates shouldn't be tracked as `UpdateAction`s — that's where we get into trouble with recursive history stacks.)

ferdnyc:
> Yeah, that's specifically what this PR does away with, because of all the problems (performance, path issues, outright crashing) it's caused to be storing history.

I should also note that I'd prefer to have @jonoomph OK this change before it merges, since history persistence across sessions has been explicitly held up as a desirable feature (see e.g. 86f1d487) and it doesn't feel like it should be my call to do away with that, whatever problems it's causing.

(The recursive history entries is only one of those problems — the other major one being outright crashes, caused by expired AppImage transition paths in the history — but the recursion was thought fixed in #2084, the change that introduced the code I highlighted in the previous comment.)

peanutbutterandcrackers:
Since
1. This PR is going to fix a lot of performance issues
2. Not even GIMP (2.10.8, at least), nor LibreOffice (6.0.6.2, at least), neither InkScape (0.92, at least) saves history across sessions
3. OpenShot is [essentially non-destructive](https://github.com/OpenShot/openshot-qt/issues/1587#issuecomment-399702934), and users can manually change every parameter they've set
4. This PR has [passed all checks](https://travis-ci.org/OpenShot/openshot-qt/builds/459931360?utm_source=github_status&utm_medium=notification)
5. This PR [has been tested](https://github.com/OpenShot/openshot-qt/pull/2400#issuecomment-441785663)

I'd be happy to take responsibility for the merge, if you two are not against it.

ferdnyc:
> I'd prefer to have @jonoomph OK this change before it merges

DylanC:
@ferdnyc - I'm happy to leave it up to @jonoomph. Its more of a fundamental change to the project.

DylanC:
@jonoomph - Can you review this change?

ferdnyc:
@DylanC : Ah, I remember optimism.

My _favorite_ is when he parachutes in to merge his own PRs and still ignores all the rest of them (e.g.: #2450)

peanutbutterandcrackers:
@ferdnyc - We have been notified about this. The plan is to clean up the newer commits to make them translate-able, I guess. Once that is done, plans are to go through each of the PRs. That is why the optimism. You should come join us. It'll be a lot of fun.

ferdnyc:
This PR contains no translatable strings, its only output is to the log.

peanutbutterandcrackers:
Not this PR, sir. The others that have been merged. I guess he's just trying to polish up what is in the main tree already and once that is done (which I hope it is) he'll be in the currently open PRs. Hope that clears up the confusion.

jonoomph:
Heyyo @ferdnyc !  Looking at this now!

jonoomph:
Yeah, originally I thought this would be a great feature, to persist history across sessions. However, as noted above, this has some ugly repercussions... such as bloated file sizes and incorrect paths. One issue I "thought" was resolved (but might not be), is the storing of nested histories. For example, a project has history [], and then someone would save or load a project, which would add the entire project, plus the entire history [] as a nested structure... thus really, really, really ballooning the size. That might still be happening, not sure.

jonoomph:
Merged into develop! Thanks @ferdnyc !

PR #2409, opened 2018-11-29
launch.py: Exit using sys.exit()

ferdnyc:
cx_freeze doesn't like it if we quit OpenShot with `exit()`, triggering a traceback and complaining:
```
  File "/usr/local/lib/python3.4/dist-packages/cx_Freeze-4.3.4-py3.4-linux-x86_64.egg/cx_Freeze/initscripts/Console.py", line 27, in <module>
  File "openshot_qt/launch.py", line 105, in <module>
    main()
  File "openshot_qt/launch.py", line 76, in main
    exit()
NameError: name 'exit' is not defined
```
Using `sys.exit()` is more correct, and should appease cx_freeze.

### DIFF
--- a/src/launch.py
+++ b/src/launch.py
@@ -74,20 +74,20 @@ def main():
     # Display version and exit (if requested)
     if args.version:
         print("OpenShot version %s" % info.SETUP['version'])
-        exit()
+        sys.exit()
 
     if args.list_languages:
         print("Supported Languages:")
         for lang in get_all_languages():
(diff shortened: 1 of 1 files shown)


peanutbutterandcrackers:
Thank you for making OpenShot better, as always.

ferdnyc:
Eh, in this case I was the one who made it worse in the first place. Those were my `exit()` statements that shouldn't have been there in the first place. :laughing:

peanutbutterandcrackers:
Well, nobody's perfect. :D

PR #2474, opened 2018-12-23
Persist undo/redo history correctly across OpenShot sessions

jonoomph:
Lots of fixes to history (undo/redo) persistence, to ultimately prevent HUGE OSP project files. The reason project files would get huge was due to a bug which saved "load" actions (which load an entire OSP project file) into the history. These actions contained a copy of the undo/redo history. So over time, loading a project multiple times, it would create a exponentially larger OSP project file, each containing the previous nested version. This will no longer happen thanks to this PR!

### DIFF
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -339,10 +339,6 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
             # Check if paths are all valid
             self.check_if_paths_are_valid()
 
-            # Discard history
-            log.info("Discarding history")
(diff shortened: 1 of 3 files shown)


ferdnyc:
Does this address the other cross-session history bug from #1923, the problem that absolute (transition) paths in the history data from AppImage sessions turn into crashers for _future_ sessions, if that history is actually used? (Because the referenced paths are no longer mounted.)

jonoomph:
@ferdnyc Nope, that is not addressed. However, I am looking at that now. Thx!

PR #2477, opened 2018-12-26
Rewrite conversion of relative and absolute path support in OpenShot

jonoomph:
I have done a complete rewrite of relative and absolute path support in OpenShot, to work as the JSON is read and write. It uses regex to quickly find all mentions of "path": or "image": , and converts the paths as needed: absolute when loading a project, and relative when saving. Even better, it also catches all paths in the undo/redo history, and fully supports @transitions relative paths. 

I've done quite a bit of testing on Linux, and this seems to work much better than the previous approach. I need to now test on Windows and Mac, to verify this has no OS dependencies.

### DIFF
--- a/src/classes/json_data.py
+++ b/src/classes/json_data.py
@@ -33,8 +33,14 @@ except ImportError:
     import simplejson as json
 
 import copy
+import os
+import re
 
 from classes.logger import log
+from classes import info
+
+# Compiled path regex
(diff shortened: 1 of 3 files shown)


jonoomph:
I also found another case of "bloated" undo/redo history... where we had an updateAction for "history" itself... which contained an entire clone of the history data structure. That was discovered by @peanutbutterandcrackers and also fixed in this branch. :smile:

PR #2490, opened 2018-12-28
Improved Docking (Timeline widget is now dockable and Properties works in pillar mode)

jonoomph:
Thanks to @ferdnyc and @peanutbutterandcrackers for their work on this one! This update changes how OpenShot's UI docks a bit, and make things more flexible. It defaults the "Properties dock widget" to fill the entire vertical space (top to bottom) on the left side of the screen. In "Advanced" view, the properties widget uses the right side of the screen. This gives two primary docking options in our view presets. I've also updated the default window_state and window_geometry for new installs, and updated the "Simple" and "Advanced" views.


jonoomph:
Nice! This is now merged into develop. I will continue testing though on all OSes.

PR #2540, opened 2019-01-19
Integration of resvg SVG library (gitlab and freeze updates)

jonoomph:
Integrating the optional resvg SVG library into openshot-qt's CI, build servers, and freezing.  This ensures that the **libresvg.so|dll|dylib** is included in our builds (if found), and otherwise ignores the file. I've also included a frame refresh when the preview window is resized, to redraw the frame at the new resolution.


jonoomph:
This is related to the libopenshot PR: https://github.com/OpenShot/libopenshot/pull/185#, and the original PR which inspired all of this: https://github.com/OpenShot/libopenshot/pull/143

PR #2543, opened 2019-01-19
Integrate Constant Rate Factor (CRF) into Export Dialog

jonoomph:
Adds a new VP9 preset, and now understands **crf** measurement and **kb/s** in preset quality choices.


src/windows/export.py
```diff
@@ -732,6 +740,7 @@ def accept(self):
             w.Open()
 
             # Notify window of export started
+            export_file_with_path = export_file_path
```
jonoomph: I think we need to remove this line, and the QMessageBox code block below, so we don't introduce a new translatable element this late (fixing to release very soon). I'm also not 100% sure about the idea of a pop-up at the end of an export either. Pehaps this could be brought back as a debug preference later?

jonoomph:
Looks Good! Although we should probably wait until https://github.com/OpenShot/libopenshot/pull/186/ is merged before merging this.

PR #2742, opened 2019-04-17
Hardware Support Integration

jonoomph:
This PR contains new preferences and export dialog changes used by the hardware-support integration PR (https://github.com/OpenShot/libopenshot/pull/220).


jonoomph:
Merged!

SuslikV:
It also makes incompatible it with previous version of libopenshot because the GetAudioDeviceNames() isn't exist there (the _Preferences_ won't open). And this is not mentioned! Just a note. Explicit linking to external lib requires attention when merging.

PR #2763, opened 2019-05-02
Remove simplejson fallback

ferdnyc:
The json module has been standard since Python 2.6, and OpenShot only supports Python 3.0+, so `import json` will never fail. The code to use simplejson as a fallback is outdated and unnecessary.

(My editor also did some blank-line and end-of-file newline cleanup that got sucked in.)

### DIFF
--- a/src/classes/json_data.py
+++ b/src/classes/json_data.py
@@ -1,36 +1,33 @@
-""" 
+"""
  @file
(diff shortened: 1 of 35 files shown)


jonoomph:
LGTM! Thanks @ferdnyc

PR #2769, opened 2019-05-05
Mac installer: Update DMG background, add Retina version

ferdnyc:
These commits:

* Update the copyright date in the `dmg-background.svg` file to 2019 and remove a lot of unused references via Inkscape's "Clean up document" tool
* Replace `dmg-background.png` with a new export
* Add a new `dmg-bacground@2x.png`, the corresponding Retina (double-res) version.

The readme for Node-appdmg [claims](https://github.com/LinusU/node-appdmg#retina-background) that it'll automatically pick up and package a Retina background if one is present alongside the standard-res file. Assuming the version on the builder is new enough, this should automatically do the right thing. If not, it'll just ignore the Retina image, so NBD either way.


jonoomph:
I noticed the font is incorrect on your PNG rasterization. It should use "Hemi Head" (https://www.dafont.com/hemi-head.font). I can also rasterize it if you would like.

ferdnyc:
@jonoomph Whoops, good catch sorry! I didn't have that font installed, you're right. Images updated, I believe they're correct now but please do take another look, you know exactly what they're supposed to look like way better than I do.

jonoomph:
LGTM! Thanks!

Issue #2780, opened 2019-05-11
Tab key order on advanced export options is quirky


jpnance:
**Describe the bug:**
When I'm picking particular frames to export in the advanced export options, I want to be able to press tab to advance from "Start Frame" to "End Frame" but it goes in the reverse order that I would expect.

**Steps to reproduce the behavior:**
1. Go to File > Export Video > Advanced
2. Click on the "Start Frame" field, where it says 1
3. Press the tab key

**Expected behavior:**
I would have expected the "End Frame" field to now be selected but, instead, input field focus has gone away altogether. Pressing shift+tab instead for Step 3 takes the focus to the correct place.

**System Details:**
- OpenShot Version: 2.4.4
- Operating System: Windows 10

Maybe this should be rolled in with https://github.com/OpenShot/openshot-qt/issues/730?

ferdnyc:
I'm going to close this, since tab order in the Daily Builds has already been sorted out, and the fix will be in the next release. The issue was addressed by @SuslikV in #2782 — the PR just wasn't tagged to auto-close.

PR #2782, opened 2019-05-12
Change order of tabbing items in export dialog

SuslikV:
The tabbing order for some fields in export dialog is not correct.
The "Start Frame" (1) and "End Frame" (2) fields is tabbing in reverse order when attempting to tabbing forward between the fields with keyboard in File > Export Video > Advanced dialog window.

Should resolve the https://github.com/OpenShot/openshot-qt/issues/2780

Actually, I copied block of code not just few fields. ~~So, despite it was not tested yet,~~ I think it should work well.

**Edit:** Tested, works OK for me.

### DIFF
--- a/src/windows/ui/export.ui
+++ b/src/windows/ui/export.ui
@@ -499,6 +499,39 @@
              </item>
             </layout>
            </item>
+           <item row="1" column="0">
+            <layout class="QHBoxLayout" name="horizontalLayout_11">
+             <item>
+              <widget class="QLabel" name="lblWidth">
+               <property name="minimumSize">
+                <size>
+                 <width>100</width>
+                 <height>0</height>
+                </size>
+               </property>
+               <property name="text">
+                <string>Width:</string>
+               </property>
+              </widget>
+             </item>
+             <item>
+              <widget class="QSpinBox" name="txtWidth">
+               <property name="sizePolicy">
+                <sizepolicy hsizetype="Expanding" vsizetype="Fixed">
+                 <horstretch>0</horstretch>
+                 <verstretch>0</verstretch>
+                </sizepolicy>
+               </property>
+               <property name="minimum">
+                <number>2</number>
+               </property>
+               <property name="maximum">
+                <number>9999</number>
+               </property>
+              </widget>
+             </item>
+            </layout>
+           </item>
            <item row="2" column="0">
             <layout class="QHBoxLayout" name="horizontalLayout_10">
              <item>
@@ -709,39 +742,6 @@
              </item>
             </layout>
            </item>
-           <item row="1" column="0">
-            <layout class="QHBoxLayout" name="horizontalLayout_11">
-             <item>
-              <widget class="QLabel" name="lblWidth">
-               <property name="minimumSize">
-                <size>
-                 <width>100</width>
-                 <height>0</height>
-                </size>
-               </property>
-               <property name="text">
-                <string>Width:</string>
-               </property>
-              </widget>
-             </item>
-             <item>
-              <widget class="QSpinBox" name="txtWidth">
-               <property name="sizePolicy">
-                <sizepolicy hsizetype="Expanding" vsizetype="Fixed">
-                 <horstretch>0</horstretch>
-                 <verstretch>0</verstretch>
-                </sizepolicy>
-               </property>
-               <property name="minimum">
-                <number>2</number>
-               </property>
-               <property name="maximum">
-                <number>9999</number>
-               </property>
-              </widget>
-             </item>
-            </layout>
-           </item>
           </layout>
          </widget>
          <widget class="QWidget" name="page_4">
@@ -873,6 +873,26 @@
            <string>Audio Settings</string>
           </attribute>
           <layout class="QGridLayout" name="gridLayout_8">
+           <item row="0" column="0">
+            <layout class="QHBoxLayout" name="horizontalLayout_20">
+             <item>
+              <widget class="QLabel" name="lblAudioCodec">
+               <property name="minimumSize">
+                <size>
+                 <width>100</width>
+                 <height>0</height>
+                </size>
+               </property>
+               <property name="text">
+                <string>Audio Codec:</string>
+               </property>
+              </widget>
+             </item>
+             <item>
+              <widget class="QLineEdit" name="txtAudioCodec"/>
+             </item>
+            </layout>
+           </item>
            <item row="1" column="0">
             <layout class="QHBoxLayout" name="horizontalLayout_21">
              <item>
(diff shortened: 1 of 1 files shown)


SuslikV:
The other way to change the order of tabbing is using tabstops, for example: https://github.com/obsproject/obs-studio/pull/1710

ferdnyc:
I pulled `export.ui` into Qt Creator, and turned on the Designer mode's "Edit Tab Order" tool to look at the existing tab order for that dialog.

**It was a _mess_!!!**

Things were scattered all over the place, just crazy. So I used that same tool to manually set the order of _every single control_ in every pane of that dialog, then saved it back again. The result was Designer inserting this lovely list into the UI definition:
```
 <tabstops>
  <tabstop>txtFileName</tabstop>
  <tabstop>txtExportFolder</tabstop>
  <tabstop>btnBrowse</tabstop>
  <tabstop>tabWidget</tabstop>
  <tabstop>cboSimpleProjectType</tabstop>
  <tabstop>cboSimpleTarget</tabstop>
  <tabstop>cboSimpleVideoProfile</tabstop>
  <tabstop>cboSimpleQuality</tabstop>
  <tabstop>cboExportTo</tabstop>
  <tabstop>txtStartFrame</tabstop>
  <tabstop>txtEndFrame</tabstop>
  <tabstop>cboProfile</tabstop>
  <tabstop>txtWidth</tabstop>
  <tabstop>txtAspectRatioDen</tabstop>
  <tabstop>txtPixelRatioNum</tabstop>
  <tabstop>txtHeight</tabstop>
  <tabstop>txtPixelRatioDen</tabstop>
  <tabstop>txtFrameRateNum</tabstop>
  <tabstop>txtFrameRateDen</tabstop>
  <tabstop>cboInterlaced</tabstop>
  <tabstop>txtAspectRatioNum</tabstop>
  <tabstop>txtImageFormat</tabstop>
  <tabstop>txtVideoFormat</tabstop>
  <tabstop>txtVideoCodec</tabstop>
  <tabstop>txtVideoBitRate</tabstop>
  <tabstop>txtAudioCodec</tabstop>
  <tabstop>txtSampleRate</tabstop>
  <tabstop>txtChannels</tabstop>
  <tabstop>cboChannelLayout</tabstop>
  <tabstop>txtAudioBitrate</tabstop>
 </tabstops>
```
It addresses a _lot_ of misordered controls, including some that aren't covered in your PR. (The "Simple" tab's tab order was **Nucking. Futz!**, even with your changes applied.)

...So, I think we should just go the tabstops route. It's obviously Qt's officially-supported method for defining tab order. They know that their tool loves to just scramble the XML for no reason, so this way the ordering of the stanzas in the file won't matter. (Designer is presumably smart enough to know that the `<tabstops />` stanza's ordering needs to be preserved unless tab order is being explicitly altered.) And it's just a pain to have to keep reordering blocks of XML in these files, it always makes the diffs incomprehensible.

Let me know what you think, and if you think I should create a separate PR for the `<tabstops>...</tabstops>` block. (I **can**, but I'm also happy enough to have you update this PR with it instead, if that works for you...)

SuslikV:
I always thought that _tabstops_ only needed when logically it is better to follow (tabbing) columns first (for example), but widgets (for readability of code) better to place in rows (or something similar). Of course, this change is insignificant, but I prefer to implement both solutions - to make clear reading of the code and to add _tabstops_ for convenience and future use. What do you think if I make it both?

Your tabstops list doesn't changes the  _Advanced tab > Export > Profile_'s items tabbing order (at least, I see it like this in my Qt 5.12.3 build of the OpenShot, Windows), so either the list is not complete or I missed something.

SuslikV:
Made same changes to the "Simple" tab's items tabbing order.

ferdnyc:
> I always thought that _tabstops_ only needed when logically it is better to follow (tabbing) columns first (for example), but widgets (for readability of code) better to place in rows (or something similar).

\*shrug\* They're needed whenever they're needed, meaning whenever the order isn't correct. From the Qt Designer manual:
>> Many users expect to be able to navigate between widgets and controls using only the keyboard. Qt lets the user navigate between input widgets with the Tab and Shift+Tab keyboard shortcuts. The default tab order is based on the order in which widgets are constructed. Although this order may be sufficient for many users, it is often better to explicitly specify the tab order to make your application easier to use. 

There doesn't appear to be any "better" or "worse" about it — if a tabstops list is needed, then it's better to have it. Rearranging the widgets has proved an _insufficient_ solution, since they don't stay in order... so it seems like it's needed.

> Of course, this change is insignificant, but I prefer to implement both solutions - to make clear reading of the code and to add _tabstops_ for convenience and future use. What do you think if I make it both?

No harm, of course, I just think committing to keeping the UI files in a certain order is a failing proposition. They tend to get scrambled around by Designer whenever they're edited. But, rearranging them again doesn't hurt anything, so no objections from me.

> Your tabstops list doesn't changes the _Advanced tab > Export > Profile_'s items tabbing order (at least, I see it like this in my Qt 5.12.3 build of the OpenShot, Windows), so either the list is not complete or I missed something.

Hmm. It _should've_. The list it contains:
```xml
  <tabstop>cboProfile</tabstop>
  <tabstop>txtWidth</tabstop>
  <tabstop>txtHeight</tabstop>
  <tabstop>txtAspectRatioNum</tabstop>
  <tabstop>txtAspectRatioDen</tabstop>
  <tabstop>txtPixelRatioNum</tabstop>
  <tabstop>txtPixelRatioDen</tabstop>
  <tabstop>txtFrameRateNum</tabstop>
  <tabstop>txtFrameRateDen</tabstop>
  <tabstop>cboInterlaced</tabstop>
```
...has all the widgets in the correct tabbing order for that section.

Where did you insert it? I notice that, like an **idiot**, I didn't actually _say_ where it belongs in the file, and it might matter. Designer placed the list towards the bottom, after all of the `<widget>...</widget>` tags, right before `<resources/>`. So, it was lines 1035-1066 in my file. (Inserted before the original line 1035, which was `<resources/>`.)

SuslikV:
@ferdnyc Again, I inserted the first list from your post:

```
...
  <tabstop>txtWidth</tabstop>
  <tabstop>txtAspectRatioDen</tabstop>    <<< - this one differs, sure
  <tabstop>txtPixelRatioNum</tabstop>
...
```
and you have changed it in the second:
```
...
  <tabstop>txtWidth</tabstop>
  <tabstop>txtHeight</tabstop>     <<< - this one differs, the next one is different too...
  <tabstop>txtAspectRatioNum</tabstop>
...
```
Copy/Paste typos?

ferdnyc:
> Copy/Paste typos?

Crap... then yes, definitely. I pasted in the wrong version of the tabstops list (in addition to not saying where it goes in the file). Not my best comment.

I've corrected it above. Sorry about that.

ferdnyc:
LGTM, as I noted above the previous tab order of that dialog was a DISASTER. Merging this change as uncontroversial.

PR #2799, opened 2019-05-23
Use actual import source for PYQT_VERSION_STR

SuslikV:
Resolve the PYQT_VERSION_STR import error
The error appear if minimum of PyQt5 is installed (--no-sip-files key etc.).
This changes the source of the string during the import to ensure that
the string is always reachable.

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -35,7 +35,7 @@ from PyQt5.QtWidgets import QApplication, QStyleFactory, QMessageBox
 from PyQt5.QtGui import QPalette, QColor, QFontDatabase, QFont
 from PyQt5.QtCore import Qt
 from PyQt5.QtCore import QT_VERSION_STR
-from PyQt5.Qt import PYQT_VERSION_STR
+from PyQt5.QtCore import PYQT_VERSION_STR
 
 try:
     # Enable High-DPI resolutions
--- a/src/classes/metrics.py
+++ b/src/classes/metrics.py
@@ -39,7 +39,7 @@ from classes import settings
 import openshot
 
 from PyQt5.QtCore import QT_VERSION_STR
-from PyQt5.Qt import PYQT_VERSION_STR
+from PyQt5.QtCore import PYQT_VERSION_STR
 
 
 # Get libopenshot version


src/classes/app.py
```diff
@@ -35,7 +35,7 @@
 from PyQt5.QtGui import QPalette, QColor, QFontDatabase, QFont
 from PyQt5.QtCore import Qt
 from PyQt5.QtCore import QT_VERSION_STR
-from PyQt5.Qt import PYQT_VERSION_STR
+from PyQt5.QtCore import PYQT_VERSION_STR
```
ferdnyc: It _shouldn't_ really matter which module this is pulled from — it's present in both, so the existing code doesn't really cause any problems. I suspect that's backwards-compatibility from all the way back in the PyQt4 days, when the documented sources for those two variables were as above. So, it's not likely to go away.

That being said, it feels _supremely_ weird (in the PyQt5 world) to be pulling `QT_VERSION_STR` from one module, but arbitrarily pulling `PYQT_VERSION_STR` from a **different** module, so purely based on THAT I'm all for this change.

In fact, on that note, my preference would be to instead replace **both** lines 37-38 with just:
```py
from PyQt5.QtCore import QT_VERSION_STR, PYQT_VERSION_STR
```

Mismatching source-module problem solved. (Whether the _previous_ line should also be subsumed into the same one-line import, I leave for better minds to discuss.)

src/classes/metrics.py
```diff
@@ -39,7 +39,7 @@
 import openshot
 
 from PyQt5.QtCore import QT_VERSION_STR
-from PyQt5.Qt import PYQT_VERSION_STR
+from PyQt5.QtCore import PYQT_VERSION_STR
```
ferdnyc: Same.

ferdnyc:
> Resolve the PYQT_VERSION_STR import error
> The error appear if minimum of PyQt5 is installed (--no-sip-files key etc.).

Interesting. I wasn't aware that was an option!

How "minimal" are we talking? Like, is it even still able to be used by OpenShot? What's missing, just the `PyQt5.Qt` module?

...Grepping the codebase, I see that these files actually _are_ the **ONLY** two places where `PyQt5.Qt` is ever imported, so I guess if it's possible to install without it, then we shouldn't need it anyway! Good catch.

SuslikV:
Minimum of PyQt5 I got, all is in this log (second build from sequence): https://ci.appveyor.com/project/SuslikV/openshot-qt/builds/24748659?fullLog=true

Required PyQt modules: ```QtCore QtGui QtMultimedia QtMultimediaWidgets QtNetwork QtOpenGL QtPositioning QtPrintSupport QtQml QtQuick QtQuickWidgets QtSensors QtSvg QtWebChannel QtWebKit QtWebKitWidgets QtWidgets```

Options (also, config file used to specify the modules and link python37 lib):
```python configure.py     --confirm-license     --spec=win32-g++     --sip="C:\Python37-x64\sip.exe"     --no-designer-plugin     --no-docstrings     --no-python-dbus     --no-qml-plugin     --no-qsci-api     --no-sip-files     --no-stubs     --no-tools     --verbose```

The latest build is here (6 days old, so no recent fixes of libopenshot in it), but you may check what functions of OpenShot is missing in the build, if any: https://ci.appveyor.com/project/SuslikV/openshot-qt/builds/24829321/artifacts

The only reason why I did it - it was my wish to get right colors from openshot .mp4 export (you remember it, sure). So, I decided to check if my code is really changes anything (the result was that it only makes right flags).

I have weak HW and connection to build OpenShot myself, but some services allow to use them for free if you are opensource. So, I just made build scripts and got my files as building artifacts. But machine time was limited to 1 hour... Building all stuff takes more time, so I looked for possible optimizations. No one of the pre-build dependencies was work for me (pip, official Qt etc.) - they all fails either function names mismatch or modules missing (webkit). I was thinking that I simply can build the libopenshot and place it instead of existing one - and as it was said - I was wrong, I fail, because I was unable to find exact PyQt files that would match official release of OpenShot... So, I decided to build PyQt from the scratch and get my own "frozen" OpenShot.

I discovered that the most Qt modules was not needed to run the OpenShot (it doesn't use them, many of webkit functions as well). This shortens building time. Finally in 2-3 sequential runs of building I was able to build PyQt with some of the required modules. And then, when I run OpenShot application I got error that PYQT_VERSION_STR cannot be imported... Now you see this PR here.

ferdnyc:
> I have weak HW and connection to build OpenShot myself, but some services allow to use them for free if you are opensource. So, I just made build scripts and got my files as building artifacts. 

\*nod\* I hear that. and the AppVeyor solution definitely makes sense.

As another possibility, that might be more flexible, I'd suggest looking at [Travis CI](https://travis-ci.org/). 
* They _also_ provide free service to open-source efforts
* Their service is _**very**_ well-integrated into GitHub, so much so that enabling it for a fork of an open-source project is very nearly as easy as flipping a switch
* And most importantly, all the work of configuring it for OpenShot's projects is _already done_, since it's what's in use here. (And on the other two trees.) So, the `.travis.yml` file needed to do the basic building is already in the repo, and the result of simply flipping the "On" switch should be a successful build. 
    Starting from there and modifying it to your specific needs is probably less headaches than starting from scratch. Starting from scratch means you get to define everything the way you need, but... I've found that it also means the _existing_ codebase ends up working against you, instead of with you.

I mean, you've already put the effort into AppVeyor, and if it's working out for you then it might not be worth bothering with Travis. But it's another possibility, anyway.

> But machine time was limited to 1 hour... Building all stuff takes more time, so I looked for possible optimizations. No one of the pre-build dependencies was work for me (pip, official Qt etc.) - they all fails either function names mismatch or modules missing (webkit). 

Yup. That's the build tooling and structural stuff I was ranting about in that other Issue. Along with the code age (or, API-support age). It's effectively impossible to build OpenShot from the ground up, because so much of the source tree is 5 years out of touch with the outside world. Its dependencies are all out of date, it's not compatible with _current_ versions of the dependencies (if they even exist anymore) so binary packages are almost completely useless, and it "can't" even be _updated_ to be compatible with these more-current versions because upgrading bloody Ubuntu Twatty Tricycle<sup>1</sup> or whatever to those versions would involve out-of-tree changes that no external project contributor has any ability to make happen.

It's even worse on Windows. I have been attempting to build libopenshot, starting from a clean Windows 7 install, _for the past **year and a half**_.

I have not succeeded.

As far as I can tell, the only way to successfully build libopenshot on Windows is to build it on the OpenShot project GitLab Windows builder, which has all of the dependencies already installed, in the places the build tooling expects/assumes, and in versions that are compatible with the tree.

It could conceivably be impossible to _recreate_ that environment, even, unless you have access to a time machine. (Can't say for sure since I have absolutely no idea how it's set up or what would be necessary to recreate it.)

> (Skipping ahead...)
> So, I decided to build PyQt from the scratch and get my own "frozen" OpenShot.

Right. Noble goal.
* But first you'd need to install Qt.
* And you'll need SIP, which PyQt requires (and is an absolute monster to properly configure/build).
* And FFMpeg.
* And ZeroMQ.
* And JsonCPP. (Though, and outdated version of that is in the source tree.)
* And UnitTest++.
* And ImageMagick. (Which you'll also have to build, since 7.0 isn't supported and 6.x binary downloads are getting hard to come by.)
* And SWIG to build the libopenshot Python/Ruby bindings.
* And Python and Ruby themselves.
* And this comment is probably a more complete list of those dependencies than you've been able to find so far, so good luck sorting them all out. (I'm sure I've forgotten at least one, too.)

> I was thinking that I simply can build the libopenshot and place it instead of existing one - and as it was said - I was wrong, I fail, because I was unable to find exact PyQt files that would match official release of OpenShot... 

That actually shouldn't matter in the slightest. PyQt version, OpenShot is wildly agnostic to.

The real issue is that binary downloads of PyQt are... well, I'm not even sure they're really a _thing_, generally, because they're tied to **a specific Qt version**. Going back to my earlier rant again, as I said PyQt is a generated wrapper around Qt — when you generate that wrapper (build PyQt), the build is specific to the _exact version of Qt_ that it's built for. Change anything about Qt? Gotta rebuild PyQt. (And not just change Qt releases, but literally _anything_. Change the installed module set, or the install location, or whatever: PyQt needs to be rebuilt.) 

So, the only possible way to get a binary download of PyQt is if it's part of a package collection that also includes an exact corresponding binary download (set) for Qt. 

Fedora's dist repo is one such example: Fedora publishes Qt and PyQt packages, along with both WebKit and WebEngine modules for both, and builds/updates are all coordinated together. Whenever something updates, rebuilds all around! The last time Fedora released an updated set of Qt5 builds, it involved [72 individual packages](https://bodhi.fedoraproject.org/updates/FEDORA-2019-74ad6b133a) that all had to be rebuilt together. The amount of effort that goes into coordinating all that is mind-blowing, honestly.

As long as libopenshot is built on _the same Qt_ as PyQt, though, then everything should be fine there.

#### Notes
1. _"upgrading bloody Ubuntu Twatty Tricycle..."_ 
    No, no, that's not fair. Ubuntu _Trusty_ 14.04 support actually **has** finally been dropped. That's what let us finally push the minimum CMake version from 2.8 to 3.1. But it only happened once the release was literally _less than a month_ from reaching its 5-year end-of-life, anyway! And still, [the move to a minimally-capable CMake was a reluctant one](https://github.com/OpenShot/libopenshot-audio/pull/37#issuecomment-480006402). 

    (I actually _cringed_ when they announced that Ubuntu 18.04 [is now a **TEN-year LTS release**](https://itsfoss.com/ubuntu-18-04-ten-year-support/), because I have these terrifying visions of not being able to upgrade anything past the lowest-common-denominator, ca. 2018 standards for another next 9+ years.)

SuslikV:
@ferdnyc Many modern programs has a lot of dependencies. But at least most critical ones should be available for public - it makes life easier ^_^ OpenShot works, it is simple enough, has few bugs - this is general formula of success. You are here, and can help to improve the things. This is open source. Get involved! Long response, low attention to personalities - this is common behavior if you are volunteering.

But if you can - do it, do it better!

Right now I can only make this small PR. I think, that at least in current state (2 strings) it shows where issue was and how it was solved. The next PR can add optimization to listing (1 string, because same module mentioned). And is not a big deal, as for me, if it will be changed in one move.

ferdnyc:
> Right now I can only make this small PR. I think, that at least in current state (2 strings) it shows where issue was and how it was solved. The next PR can add optimization to listing (1 string, because same module mentioned). And is not a big deal, as for me, if it will be changed in one move.

Well, that's _entirely_ up to you — but the point of the PR system, and the ability to request changes, is that PRs can still be developed before merge. If you wanted to make further changes, you can go to the "Files Changed" subtab, click on "Edit file" in the menu for each of the files indicated, and make any additional changes you want to include in the PR — they'll be added as additional commits, and when the PR is merged all of the changes will be incorporated.

In projects where PRs are reviewed seldom and infrequently, having to wait multiple PR cycles to make incremental changes in is kind of impractical. Best to get everything done in a single PR.

PR #2804, opened 2019-05-25
Initial Import and Export support for EDL and XML formats

jonoomph:
Adding the experimental ability to import and export EDL (Edit Decision Lists, CMX 3600) and XML (Final Cut Pro format). This only supports some basic things, like tracks, clips, and a couple keyframes (volume, opacity). Full support of these formats will take more work and likely another contributor to continue this work.


jonoomph:
Merged!

PR #2796, opened 2019-05-21
Rename frozen executable to openshot-qt

ferdnyc:
This renames the executable created by cx_Freeze from `launch`/`launch.exe` to the value of `info.NAME` (currently `openshot-qt`), plus the appropriate extension (if any), so that a running OpenShot will show up in process listings as "openshot-qt" rather than the generic/suspicious-seeming "launch".

It also renames any scripts that call or access `launch.exe`/`launch` to use `openshot-qt.exe`/`openshot-qt` instead.

PR #2810, opened 2019-05-28
WIP: win32: Add metadata to launcher executable

ferdnyc:
This isn't actually a WIP, but it's tagged that way because it has external dependencies (explained below) before it can be merged. 

Inspired by the information @SuslikV provided in #2772 re: attaching Windows metadata to the OpenShot launcher executable, this PR adds code to do so, using an external Win32 binary tool which should be placed on the Windows builder hosts.

**PREREQUISITE:** Before this PR can be merged, a change to the Windows builders @ GitLab is required, to install the `verpatch.exe` executable referenced in this PR. The binary ZIP (attached) was [downloaded from CodeProject](https://www.codeproject.com/KB/install/VerPatch.aspx?msg=3207401). Before extracting, create the directory `C:\Program Files (x86)\Verpatch`, then extract the `verpatch.exe` from the zip file into that directory.

Verpatch binary: 
[verpatch-bin-1.0.10.zip](https://github.com/OpenShot/openshot-qt/files/3230112/verpatch-bin-1.0.10.zip)

The new code will run `c:\program files (x86)\verpatch\verpatch.exe` to add metadata to the Windows `launch.exe` program. This will affect both the Properties dialog, and the Task Manager listing:

Screenshot from 2019-05-28 19-22-49
Screenshot from 2019-05-28 19-23-02

Fixes: #2762

### DIFF
--- a/installer/build-server.py
+++ b/installer/build-server.py
@@ -431,6 +431,23 @@ try:
         if windows_32bit:
             only_64_bit = ""
 
+        # Add version metadata to frozen app launcher
+        launcher_exe = os.path.join(exe_path, "launch.exe")
+        verpatch_success = True
+        verpatch_command = '"C:\Program Files (x86)\Verpatch\verpatch.exe" "{}" /va /high "{}" /pv "{}" /s product "{}" /s company "{}" /s copyright "{}" /s desc "{}"'.format(launcher_exe, info.VERSION, info.VERSION, info.PRODUCT_NAME, info.COMPANY_NAME, info.COPYRIGHT, info.PRODUCT_NAME)
+        verpatch_output = ""
+        # version-stamp executable
+        for line in run_command(verpatch_command):
+            output(line)
+            if line:
+                verpatch_success = False
+                verpatch_output = line
+                
+        # Was the verpatch command successful
+        if not verpatch_success:
+            # Verpatch failed (not fatal)
+            error("Verpatch Error: Had output when none was expected (%s)" % verpatch_output)
+
         # Copy uninstall files into build folder
         for file in os.listdir(os.path.join("c:/", "InnoSetup")):
             shutil.copyfile(os.path.join("c:/", "InnoSetup", file), os.path.join(PATH, "build", file))


jonoomph:
LGTM! Nice. I've installed verpatch at the specified location on the Windows builders, so this should work: :crossed_fingers:

jonoomph:
Oops, looks like an escape issue with the verpatch path... failing on builder. I'm working on a quick fix.

jonoomph:
Or perhaps that is not the only issue... exe_path is undefined. Still working on a fix... but should be quick.

jonoomph:
@ferdnyc Okay, this is now done and merged For Realz :tm:

ferdnyc:
Woo!

I'm actually not _surprised_ that the escaping was a problem — in fact, I'd have been surprised if it **wasn't** — but the thing that confuses me is, I patterned it after the `inno_command` just below. So my **actual** question is, how does _this_ work correctly?? 🤔

https://github.com/OpenShot/openshot-qt/blob/eef69337e077162c589f2580c2503a218875f844/installer/build-server.py#L457-L459

Python's string handling is an unsolvable mystery. And people talk about _Perl_ being write-only code...

ferdnyc:
@jonoomph Any thoughts on the semi-related #2796, BTW? I think it makes even more sense, with this merged, to get rid of the suspicious "`launch.exe`" process name (or "`launch`" on Mac/Linux). 

Makes OpenShot seem like shady malware, at least to those of us scarred by too many years of dealing with shady malware lurking in processes like "`setup.exe`" and "`process.exe`".

jonoomph:
Because \v matches a special character in Python, and \I or \i does not. But I agree, string escaping issues in Python can be a `gotcha` real quick. Yes, I think changing the frozen exe name makes a lot of sense, and should make things seem more legit. No complaints here.

ferdnyc:
:roll_eyes: 

This actually reminds me that I just recently learned (thank you Stack Overflow) of a relevant C++11 feature along these same lines: [Raw string literals](https://en.cppreference.com/w/cpp/language/string_literal). 

You can prepend R to a quoted string, then enclose it in a paired set of delimiters, and it'll disable _all_ escaping between the markers. Sort of like Perl's leveraging of prefixed curly braces:
* `q{'quoted' string}`
* `qq{"double" 'quoted' string}`
* `qw{"word" "quoted" "string"}`
* `s{'standard' s///}{regular! expression!}`

Similarly, in C++11, a line like:
```cpp
std::string raw_text = R"x( ... )x"
```

allows you to place literally _anything_ between the parens.

ferdnyc:
> allows you to safely place literally _anything_ between the parens

(Anything except the exact sequence `)x`, obviously, in my example.)

But that's why they let you choose the delimiters, and that's the _real_ power of the feature. Writing a JSON literal that actually contains the string `)x` for some reason? Just delimit it as `R"json(...)json"`, or whatever else you like.

Issue #1302, opened 2018-02-24
Unable to browse for files on "other locations"


fyrekrig:
Intel® Core™2 Duo CPU E8600 @ 3.33GHz × 2 
NVA8
Kali GNU/Linux Rolling 2018.1
OpenShot 2.4.1
File/ImportFiles does not include "other locations" i.e other partitions on HDD than C:/
How can I navigate to these folders?

ferdnyc:
Mine does. (Fedora 27; currently OpenShot 2.4.1 built by hand, but I don't recall any difference in the rpmfusion openshot-2.4.0 builds.) Sounds like it may be a build issue.

![screenshot from 2018-02-25 01-35-40](https://user-images.githubusercontent.com/538020/36638826-67bf6904-19cc-11e8-8d6d-0f7eaf432701.png)

ferdnyc:
(Also, you're on Linux, you say? You know there's no such thing as `C:/`, right? ...Not even on Windows, there it would be `C:\`.) :wink:

DylanC:
@ferdnyc - I'm assuming its a dual booting PC with Linux and Windows on it. :)

fyrekrig:
> @ferdnyc - I'm assuming its a dual booting PC with Linux and Windows on it. :)
Actually it's not. The first partition used to have a windows install and might have again.
Enclosing screenshots from filemanager and OpenStop "Import files"
![01](https://user-images.githubusercontent.com/29265156/36640648-bbbcc464-1a55-11e8-8853-b078db7bfe7e.jpg)
![02](https://user-images.githubusercontent.com/29265156/36640649-c28e5398-1a55-11e8-9faa-5365127f04d5.jpg)

fyrekrig:
Sorrry made mistake by closing!

peanutbutterandcrackers:
@ferdnyc - Whoa! Neat! Your 'import' dialogue uses the gnome import dialogue for some reason, not OpenShot's own. Which is pretty neat, I think. How did you you do that?

@DylanC - I think we should get ^that going on with OpenShot rather than the default 'import' dialogue. This issue happens with me too. But I have figured out a work-around. :)

[...]

PR #2818, opened 2019-06-01
Lower scroll area minimum height for Preferences

SuslikV:
In preferences.ui the default widget's size of the Preferences
window is set to 350, thus few items at bottom of the main
menu Edit > Preferences > Keyboard_tab is not visible when
scrolled down until Preferences window resized.

This change ensures that scroll area always fits into parent
widget by default, without needs to enlarge the window.

This PR doesn't changes size of the Preferences window itself.

### DIFF
--- a/src/windows/preferences.py
+++ b/src/windows/preferences.py
@@ -144,7 +144,7 @@ class Preferences(QDialog):
                     scroll_area.setWidgetResizable(True)
                     scroll_area.setVerticalScrollBarPolicy(Qt.ScrollBarAsNeeded)
                     scroll_area.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Expanding)
-                    scroll_area.setMinimumSize(675, 350)
+                    scroll_area.setMinimumSize(675, 100)
 
                     # Create tab widget and layout
                     layout = QVBoxLayout()


ferdnyc:
@SuslikV Actually... do you see _any_ difference without that `setMinimumSize()` call in there at all? AFAICT it has absolutely no purpose, and everything is exactly the same without it.

If that's the case, I think I'd prefer to just remove it entirely, and it's one less size constraint to worry about keeping in sync.

SuslikV:
I tested it with Qt 5.12.3 and there it looks completely different - height of the _Preferences_ dialog window is smaller (you may try it). So, yes it changes. I think, it is because parent widget has higher priority of the size constrains if any set (and yes, there is one of the _dlgPreferences_ - the _minimumSize_ in _src/windows/ui/preferences.ui_).

When you update your Qt, you will need this anyway.

jonoomph:
LGTM!

ferdnyc:
> I tested it with Qt 5.12.3 and there it looks completely different - height of the _Preferences_ dialog window is smaller (you may try it). 

Oh, no, I absolutely did try it, and there was no effect whatsoever. That's why I asked. On my system,  `scroll_area.setMinimumSize(675, 350)` , or `, 100)`, or deleted... no change. (Except that I did see the scroll issue in Shortcuts that originally prompted this, so I agree it had to be at least shrunk.)

> So, yes it changes. I think, it is because parent widget has higher priority of the size constrains, if any (and yes, there is one of the _dlgPreferences_ - the _minimumSize_ in _src/windows/ui/preferences.ui_).

Well... _exactly_. That's why I was hoping we could take it out, and just rely on the parent's dimensions. Given that the _parent_ has a fixed size, I'm surprised it would be needed at all: The widgets should just expand to fit the window. Which is what they seem to do, on my system.

But, hey, like I said: That's why I asked. I assume that line was inserted for **some** reason,  just because it's unnecessary on my machine doesn't mean it has no purpose at all.

> When you update your Qt, you will need this anyway.

I'm now running 5.12.4. Nothing has changed. The dialog still looks the same even if I comment out the `setMinimumVersion()` line, on my system. Qt experiences are far from universal, there's quite a bit of customization that goes on across the various installs, distros, etc. We've hit **plenty** of examples of that, just among OpenShot users. (The native file dialog debacle (#1302) comes to mind.)

My problem is actually the opposite. Due to the excessively verbose labels assigned to audio output devices on my system, the Preferences dialog is now _waaay_ too small, and the Preview tab scrolls horizontally. It'd need to be at least `200px` wider, maybe `300px`, for them to fit. I'm not sure what to do with that, since `1000px` seems **excessively** wide as a forced minimum size for the Preferences dialog, and I'm worried it wouldn't fit on some users' screens.

OpenShot-preferences-audio-device-size

I was hoping I could convince Qt to auto-size it to the contents, since it really seems like it should. Fixed pixel dimensions are so 1995. Qt's [extension dialog example project](https://doc.qt.io/qt-5/qtwidgets-dialogs-extension-example.html#), which I looked at for guidance, never even **mentions** size. Nowhere in the code for that project will you find a single width, height, or pixel dimension of any kind. Not only is it presumed that the dialog will be sized automatically, but apparently it even automatically _resizes_ when the contents are changed! That'd be way cooler than futzing with pixel dimensions.

Preferences probably _would_ auto-size, if we were letting Qt build it by creating a data model and connecting it to the dialog. But because we're building and inserting each individual widget "by hand", I guess we're in charge of that. I tried removing the pixel dimensions from the UI file and the dialog failed to even appear, which I assume means it was indeed created with `0x0` dimensions.

ferdnyc:
> Qt's [extension dialog example project](https://doc.qt.io/qt-5/qtwidgets-dialogs-extension-example.html#), which I looked at for guidance, never even **mentions** size. Nowhere in the code for that project will you find a single width, height, or pixel dimension of any kind. Not only is it presumed that the dialog will be sized automatically, but apparently it even automatically _resizes_ when the contents are changed!

Although... the weird thing about that? The weird thing is, the code _really_ doesn't look that different from our own. The dialog-construction code in [finddialog.cpp](https://doc.qt.io/qt-5/qtwidgets-dialogs-extension-finddialog-cpp.html) is fairly similar to what you'd find in OpenShot's building of the Preferences or Export dialog, _except_ that it just pays no attention to size, and trusts Qt to handle it.

I have to wonder if that's where we're going wrong. Would removing _all_ of the forced sizing from the OpenShot code free Qt to do the job it's supposed to do, where it just auto-sizes things appropriately?

Who knows? That's a real chore to properly test out "experimentally", when you can't even be sure whether it'll work. How do you tell the difference between the experiment ending in failure, and just having missed something that should've been removed from the code?

ferdnyc:
> The dialog still looks the same even if I comment out the `setMinimumVersion()` line

I meant `setMinimumSize()`, obviously.

ferdnyc:
One possibility is that the auto-sizing is a feature of only relatively _recent_ Qt versions, and wasn't applicable in earlier releases.

The Qt 5.5 documentation didn't even _have_ a "Dialog examples" section per se, but for instance this [Group Box Example project](https://doc.qt.io/archives/qt-5.5/qtwidgets-widgets-groupbox-example.html) for Qt 5.5 has a [window.cpp](https://doc.qt.io/archives/qt-5.5/qtwidgets-widgets-groupbox-window-cpp.html) file that contains an explicit `resize(480, 320);` at the end of building the dialog contents.

So, if it's the case that auto-sizing only works in recent Qt releases, then we're _really_ in a pickle. OpenShot supports a **HUGE** range of Qt versions. I'm running 5.12.4. You're running 5.12.3. Users of other Linux distros are on anything from 5.11 to 5.9. Last I checked the Windows builds were bundled with Qt 5.6. The AppImages, unless something's changed, are still using Qt 5.2.1 from February 2014!

If OpenShot's code has to target the lowest common denominator of all those Qt versions, then there's basically nothing we can rely on in terms of supported featureset, and everything has to be explicitly forced in the code.

PR #2825, opened 2019-06-07
WIP: Enable Blender GPU rendering

### LINKED ISSUE #2812: Animated titles: Force blender GPU rendering when available
**Describe the new feature:**
As we know, blender can use GPU for rendering if the hardware support it. We can configure Blender thru ctrl+alt+u, "system" tab and in "Cycles compute device" select CUDA (for Nvidia) and then all processes using cycles render engine will use the GPU. However, when called from command line, Blender doesn't use cycles and the GPU acceleration by default, and this is a big improvement to work with OpenShot animated titles. So I suggest to modify the way the server is invoked by OpenShot to force GPU rendering if possible.

Use GPU rendering for animated titles 

One possible an easy solution is to call blender with "- GPU" parameter, but I don't know if OpenShot is doing it internally.
Another way is (as some published in [](https://blender.stackexchange.com/questions/91724/configuring-blend-file-for-remote-machine-gpu-rendering) ), use this python script:
```
# [8] Enable CUDA and activate GPU
# code from https://developer.blender.org/T54099

scene = bpy.context.scene
scene.cycles.device = 'GPU'

prefs = bpy.context.user_preferences
cprefs = prefs.addons['cycles'].preferences   

# Attempt to set GPU device types if available
for compute_device_type in ('CUDA', 'OPENCL', 'NONE'):
    try:
        cprefs.compute_device_type = compute_device_type
        break
    except TypeError:
        pass

# Enable all CPU and GPU devices
for device in cprefs.devices:
    device.use = True
```

ferdnyc:
This is work-in-progress (but functional) code to enable GPU rendering for the Animated Title editor, using the CYCLES renderer add-on that's part of the default Blender distribution, as requested in #2812. 

With assistance and testing from reporter @RafaelLinux , I've managed to condense the configs required to enable and activate CYCLES GPU rendering down to a single chunk of Python code, `src/blender/scripts/gpu_enable.py`, which will be automatically injected into the generated Blender Python script if the corresponding experimental preference is enabled.

The code is _mostly_ bulletproof, in the sense that if the user enables GPU rendering, but their system is not capable of using the GPU renderer, in the majority of cases CYCLES will simply fall back to CPU rendering silently, and still produce the rendered output. 

I say _mostly_, because there are configurations which can cause the render process to fail without generating frames (see #2812 for more detail); those cases can be handled by disabling the GPU rendering code injection, in which case the render pipeline will be completely identical to how it was before this PR. The need to handle those cases better is one of the reasons this PR remains in WIP status.

But, despite needing additional polish, the code is functional and can be enabled and tested today. Whether GPU rendering will provide any _benefit_ when enabled, that's a much larger question. Blender GPU rendering isn't like accelerated video de/encoding — it uses the GPU as a CUDA compute engine, and unless the GPU is sufficiently powerful the speed increase compared to the system CPU can be nonexistent or even negative. (My own low-end Nvidia GT 710 card, a < $50 model, renders between 0% and 20% _slower_ than my system's 4-core Intel Core i5 CPU, making GPU rendering completely pointless for me. But, for some users with high-end or multiple GPUs, CYCLES GPU rendering should offer significant speedups.)

### TODO items, to be ready for merge
- [ ] **Better failure handling:** If the render attempt results in an error, the sensible thing is to disable GPU rendering and try again using the CPU. That can be done one of two ways (our choice)...
  - [ ] Detect the issue automatically, switch off the setting, and try again automatically, with no user intervention.
  - [ ] In the error message dialog that comes up on render failure, if GPU rendering is enabled in the settings, add a button the user can click to turn off the setting and re-attempt the render. If that fails again, they'll get the same dialog _without_ that button, and they'll have to pursue other solutions.
- [ ] **Clean up the Preferences item:** I just stuck it in below the parallel-processing option, since they're both Experimental, but given how much has been added to Performance recently, it looks completely out of place there, and gets kind of lost. Also, the window needs to be taller if it's going to go in that tab.
- [ ] **Add documentation:** GPU rendering setup is complex _enough_ with Blender, that it's worth adding a section to the manual providing what information we can offer regarding proper configuration, along with the caveats I mentioned earlier (like how it might even be _slower_ than CPU rendering in many instances).

Fixes #2812

### DIFF
--- a/src/blender/scripts/gpu_enable.py
+++ b/src/blender/scripts/gpu_enable.py
@@ -1,0 +1,17 @@
+# Enable CYCLES renderer, using GPU (if found)
+bpy.context.scene.render.engine = 'CYCLES'
+bpy.context.scene.cycles.device = 'GPU'
+
+cprefs = bpy.context.user_preferences.addons['cycles'].preferences
+
+# Attempt to set GPU device types if available
+for compute_device_type in ('CUDA', 'OPENCL', 'NONE'):
+    try:
+        cprefs.compute_device_type = compute_device_type
+        break
+    except TypeError:
+        pass
+
+# Enable all CPU and GPU devices
+for device in cprefs.devices:
+    device.use = True
--- a/src/settings/_default.settings
+++ b/src/settings/_default.settings
@@ -456,6 +456,14 @@
     "setting": "omp_threads_enabled"
   },
   {
+    "value": false,
+    "title": "Use Blender GPU rendering for Animated Titles (Experimental)",
+    "type": "bool",
+    "restart": false,
+    "category": "Performance",
+    "setting": "blender_gpu_enabled"
+  },
+  {
     "min": 0,
     "max": 128,
     "value": 12,
--- a/src/settings/_default.settings
+++ b/src/settings/_default.settings
@@ -456,14 +456,6 @@
     "setting": "omp_threads_enabled"
   },
   {
-    "value": false,
-    "title": "Use Blender GPU rendering for Animated Titles (Experimental)",
-    "type": "bool",
-    "restart": false,
-    "category": "Performance",
-    "setting": "blender_gpu_enabled"
-  },
-  {
     "min": 0,
     "max": 128,
     "value": 12,
@@ -502,6 +494,14 @@
     "restart": true,
     "category": "Performance",
     "setting": "decode_hw_max_height"
+  },
+  {
+    "value": false,
+    "title": "Use Blender GPU rendering for Animated Titles (Experimental)",
+    "type": "bool",
+    "restart": false,
+    "category": "Performance",
+    "setting": "blender_gpu_enabled"
   },
   {
     "value": true,
--- a/src/windows/views/blender_listview.py
+++ b/src/windows/views/blender_listview.py
@@ -738,7 +738,7 @@ class Worker(QObject):
                     # change cursor to "default" and stop running blender command
                     self.is_running = False
 
-                    # Wrong version of Blender.  Must be 2.62+:
+                    # Wrong version of Blender.
                     self.blender_version_error.emit(float(self.version[0]))
                     return
 
@@ -751,8 +751,8 @@ class Worker(QObject):
             self.process = subprocess.Popen(command_render, stdout=subprocess.PIPE, stderr=subprocess.PIPE, startupinfo=startupinfo)
 
         except:
-            # Error running command.  Most likely the blender executable path in the settings
-            # is not correct, or is not the correct version of Blender (i.e. 2.62+)
+            # Error running command.  Most likely the blender executable path in
+            # the settings is incorrect, or is not a supported Blender version
             self.is_running = False
             self.blender_error_nodata.emit()
             return
--- a/src/windows/views/blender_listview.py
+++ b/src/windows/views/blender_listview.py
@@ -1,4 +1,4 @@
-""" 
+"""
  @file
  @brief This file contains the blender file listview, used by the 3d animated titles screen
  @author Jonathan Thomas <jonathan@openshot.org>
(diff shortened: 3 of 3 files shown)


ferdnyc:
In the process of developing this feature, I also came up with a way of running OpenShot against a libopenshot build _not_ installed on the system. (Necessary since I had to build the newest version for compatibility with the current Python code, and didn't want to overwrite my _real_ `/usr/lib64/libopenshot.so.17` needed by my installed OpenShot 2.4.4. _Insert standard grumbling about the SOVERSION not being incremented **as soon as** the first breaking change is made to the API._) 

As that would obviously be something hugely useful to other developers, I'd love to find a way to genericize that as a supported feature.

My current method involves both a wrapper shell script (which sets `LD_LIBRARY_PATH`), **plus** edits to `launch.py` which make changes to `sys.path`. (Using the `$PYTHONPATH` envvar simply would not work for me, for reasons I don't really understand.) 

This makes me think that, rather than having to edit `launch.py`, maybe it should support some sort of `user.py` file (installed in `info.USER_PATH`), to be loaded either automatically whenever it exists, or if a new command line flag is passed to the launcher.

SuslikV:
Is this OS independent change? Just wondering.

ferdnyc:
@SuslikV 
> Is this OS independent change? Just wondering.

Good question. In the sense of it _working_ just fine on any OS, it should be, yes. The only changes are to the Python script that drives blender.

Whether the Python code below is sufficient to actually enable GPU rendering on Windows and Mac versions of Blender, or if it needs to be further enhanced for those platforms, I'm not really sure. Hopefully testing will produce those answers. (But this is another area where, should the code _not_ be able to activate GPU rendering on those platforms, the result should just be a harmless fallback to the default CPU rendering.)

Both CUDA and OpenCL are cross-platform technologies (unlike the VDPAU, VA-API, etc. video hardware decoding layers), so there's every chance the same settings that work for Linux will also be applicable on Windows and MacOS.

```py
# Enable CYCLES renderer, using GPU (if found)
bpy.context.scene.render.engine = 'CYCLES'
bpy.context.scene.cycles.device = 'GPU'

cprefs = bpy.context.user_preferences.addons['cycles'].preferences

# Attempt to set GPU device types if available
for compute_device_type in ('CUDA', 'OPENCL', 'NONE'):
    try:
        cprefs.compute_device_type = compute_device_type
        break
    except TypeError:
        pass

# Enable all CPU and GPU devices
for device in cprefs.devices:
    device.use = True
```

If anything, I imagine the `('CUDA', 'OPENCL', 'NONE')` list might need to be extended with additional "flavors" of renderer.

ferdnyc:
In fact, there's at least one other GPU-related technology we should keep an eye on, related to Blender: [AMD's ProRender plugin](https://radeon-pro.github.io/RadeonProRenderDocs/plugins/blender/switching.html). But that's actually a separate _renderer_ entirely. (So, it's an alternative to `'CYCLES'`, rather than to `'CUDA', 'OPENCL'`)

I'm not yet certain from the documentation whether it's a general-purpose accelerator that we'd be able to take advantage of, or if it's only useful with their shader, materials library, etc. which would mean it's not available to us (since the OpenShot title `.blend` files are designed using only the generic built-in featureset provided by Blender).

jonoomph:
This looks really good to me! Since it's flagged by a preference, it would probably be good to merge this before our next release, so we can get some testing on it, right?  @ferdnyc Any big concerns on merging this now, and making future improvements in a different PR?

ferdnyc:
@jonoomph Hmm... because the preference does _completely_ eliminate the changes, and leaves things exactly the way they were in 2.4.4, it's probably pretty safe. Especially since the preference defaults to disabled, currently.

Though I do think we should _move_ the preference (at least vertically within the tab), because I stuck it into the Performance tab before I discovered all of the new Hardware-accel preferences were also added there, and they kind of crowd it. Everything hodge-podged together makes that tab awfully busy.

ferdnyc:
I'm not **crazy** about the idea of an Experimental preference being exposed on the main tab, but I suppose it _could_ go under the Blender path setting. That's probably the next most natural place for it.

jonoomph:
I'm okay with it being buried in the Performance tab for now. Feel free to smash the merge button whenever! Thanks!

ferdnyc:
Sounds good, I'll get it in later tonight.

ferdnyc:
I contented myself with sliding the Blender GPU preference down to the bottom of the Performance tab's settings list. Awaiting Travis' all-clear to merge.

RafaelLinux:
Only if you don't know it ... there are two interesting render engines to use too for blender: Eevee (it's just new) and web4blender (older but curious).

ferdnyc:
@RafaelLinux Thanks, I haven't heard of Eevee yet — is that new with Blender 1.80? We've already received a report (#2901) that the (just-released) 1.80 version doesn't work with OpenShot currently, it's on my list for the weekend to figure out why. I can certainly look into the new renderer options as well (if any) and see if they might be useful to us.

RafaelLinux:
yes, Eevee render engine is just new with the latest Blender release. I'm not sure if it can be used with OpenShot, cause it works in realtime, doesn't not generate a output video.

ferdnyc:
@RafaelLinux 

My further reading eventually led me to understand (hopefully I've got it right) that Eevee is the new _default_ (built-in) renderer for 2.80, to replace the previous "Blender Internal". 

So, it'll definitely be useful to OpenShot (in fact, it'll be used by default), but only when _not_ doing hardware rendering via CYCLES.

RafaelLinux:
In that case, is not interesting from the point of view of a OpenShot user that want a fastest and responsive preview of "Titles" rendered using Blender.

Issue #2816, opened 2019-05-30
RFE: Indicate the track as 'active' at all times, not only at beginning and the end.


udippel:
![Screenshot_20190530_131503](https://user-images.githubusercontent.com/50265752/58632877-39c85f80-82e7-11e9-923e-1b08a64d941f.png)
![Screenshot_20190530_131408](https://user-images.githubusercontent.com/50265752/58632879-39c85f80-82e7-11e9-9df0-6fe74a3d4f95.png)

With a fine timeline, it seems impossible to see which track is set active. Once I scroll to beginning or end, red colour indicates the active track. 
One screenshot show the active track at the start, the other shows no indication when in the center of that same track.

ferdnyc:
Mmmm. This shouldn't be too hard, since it's a change I had already included in the (never-completed) #1883, so the code is already there for the stealing.

ferdnyc:
Fixed in #2827, thanks to @SuslikV

PR #2827, opened 2019-06-08
Make clip selection always visible

SuslikV:
Solves issue https://github.com/OpenShot/openshot-qt/issues/2816

Selection visibility shouldn't relay on zoom level or length of the clip.

The only change from previous attempt (https://github.com/OpenShot/openshot-qt/pull/1883) to improve the UI of OpenShot in this part - is that the `!important` rule was added. Without that change in some environments it is possible to notice slight delay when underlying blue border at top and bottom of the clip is drawn first and only then red (yellow) selection overwrites it.

Before (clip selected)
Selection_clip_before_cropped

After (clip selected)
Selection_clip_after_cropped

Screenshot of the UI for the rectangle selection is under the spoiler:
<details>

Before (rectangle selection)
Selection_rectangle_before_cropped

After (rectangle selection)
Selection_rectangle_after_cropped

</details>

### DIFF
--- a/src/timeline/media/css/main.css
+++ b/src/timeline/media/css/main.css
@@ -83,8 +83,8 @@ img {
 .left_edge_stop { border-left: 2px solid #4b92ad; }
 .right_edge_stop { border-right: 2px solid #4b92ad; }
 .highlight_clip { border-color: #32b7ea; }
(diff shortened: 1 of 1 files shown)


DylanC:
@SuslikV - This change is so simple but so much better!

jonoomph:
LGTM! Thanks again @SuslikV !

PR #2829, opened 2019-06-09
Hardware decoding and preferences improvements

jonoomph:
Adding real-time test to determine valid hardware **decoding** options for the Preferences. This will disable card selection in non-Linux systems, and only display valid options which can decode a test video, and checks the pixel color (for an expected color). Sometimes hardware decode will work without throwing an error, but it will result in a black or green frame (so this check verifies the settings work correctly).


jonoomph:
Merging

SuslikV:
Now change the example .mp4 file to use the file flagged with the used encoding color matrix (to know what decoding color matrix to use). Is this BT.709 colors? Post source image, please.

jonoomph:
Are you getting an error during `make os_test`? Please share what the problem is you are seeing. Everything seems to work fine on my end.

SuslikV:
I know that AMD has own view on how HD files looks like, so BT.601 decoding matrix can be used. Aslo, sampling (0; 0) point was not the best solution - open your test file in OpenShot and _Save Current Frame_, then open saved frame in editor, let's say GIMP.

jonoomph:
Essentially, the new CheckPixel method does the same things as saving a frame in OpenShot and opening that image in Gimp... but only in memory. And then checks the pixel against a known value (with a threshold to account for slightly different decoder values... which can be based on hardware decoders now). The exact color matrix should not be required based on my current testing. Some variation is expected. I hope that makes sense.

SuslikV:
The matrix is not needed if you comparing luma only, and again - video range should be specified. Also, you don't know what version of the ffmpeg will be tested against your test mp4 file (and .mp4 is not .ts) and what decoder should do with the encoded stream of unknown colorspace and range.

PR #2836, opened 2019-06-15
Improve Video Preview background contrast

SuslikV:
**Edit:** screenshots below in this post is no longer actual. Color was changed.

///////////

Before
preview background before

After
preview background after

### DIFF
--- a/src/windows/video_widget.py
+++ b/src/windows/video_widget.py
@@ -607,7 +607,7 @@ class VideoWidget(QWidget, updates.UpdateInterface):
         # Mutex lock
         self.mutex = QMutex()
 
-        # Init Qt widjet's properties (background repainting, etc...)
+        # Init Qt widget's properties (background repainting, etc...)
         super().setAttribute(Qt.WA_OpaquePaintEvent)
         super().setSizePolicy(QSizePolicy.Preferred, QSizePolicy.Preferred)
 
--- a/src/windows/video_widget.py
+++ b/src/windows/video_widget.py
@@ -82,7 +82,7 @@ class VideoWidget(QWidget, updates.UpdateInterface):
         painter = QPainter(self)
         painter.setRenderHints(QPainter.Antialiasing | QPainter.SmoothPixmapTransform | QPainter.TextAntialiasing, True)
 
-        # Fill background with theme color instead of black
+        # Fill the whole widget with the solid color
         painter.fillRect(event.rect(), self.palette().window())
 
         if self.current_image:
--- a/src/windows/video_widget.py
+++ b/src/windows/video_widget.py
@@ -82,7 +82,7 @@ class VideoWidget(QWidget, updates.UpdateInterface):
         painter = QPainter(self)
         painter.setRenderHints(QPainter.Antialiasing | QPainter.SmoothPixmapTransform | QPainter.TextAntialiasing, True)
 
-        # Fill background black
+        # Fill background with theme color instead of black
         painter.fillRect(event.rect(), self.palette().window())
 
         if self.current_image:
@@ -607,10 +607,7 @@ class VideoWidget(QWidget, updates.UpdateInterface):
         # Mutex lock
         self.mutex = QMutex()
 
-        # Init Qt style properties (black background, etc...)
-        p = QPalette()
-        p.setColor(QPalette.Window, QColor("#191919"))
-        super().setPalette(p)
+        # Init Qt widjet's properties (background repainting, etc...)
         super().setAttribute(Qt.WA_OpaquePaintEvent)
         super().setSizePolicy(QSizePolicy.Preferred, QSizePolicy.Preferred)
 


jonoomph:
I'd be interested in feedback on this one from others. This is very subjective, lol, but I'm curious if people think this looks better with increased contrast?

SuslikV:
Anybody who what to test it on their PC right now - just open the `src/windows/video_widget.py` file in suitable text editor (for Windows I'm using `notepad++`) and change the `QColor("#191919")` value at line 616 to  `QColor("#4d4d4d")` , save changes and restart application.

ferdnyc:
Well... OK, first off: I _hate_ it **lighter** colored than the window frame. I don't understand why that would be desirable in the slightest.

But what I don't understand is, what's the advantage to setting that color **AT ALL**? If you remove _all three_ of those lines, and just don't mess with the default palette:

```diff
diff --git a/src/windows/video_widget.py b/src/windows/video_widget.py
index 6bb6816d..5eb23b57 100644
--- a/src/windows/video_widget.py
+++ b/src/windows/video_widget.py
@@ -612,9 +612,6 @@ class VideoWidget(QWidget, updates.UpdateInterface):
         self.mutex = QMutex()
 
         # Init Qt style properties (black background, etc...)
-        p = QPalette()
-        p.setColor(QPalette.Window, QColor("#4d4d4d"))
-        super().setPalette(p)
         super().setAttribute(Qt.WA_OpaquePaintEvent)
         super().setSizePolicy(QSizePolicy.Preferred, QSizePolicy.Preferred)
``` 
Then on my system, you get this, and I cannot for the life of me understand why it wouldn't be the most desirable outcome anyway:
OpenShot-PreviewBorderColor

ferdnyc:
(Or, if you want the edges of the video preview to be set off from the window frame, so they're clearly visible, then I'd say paint the preview with a uniform frame around all FOUR sides. Instead of the current situation, where you can only see the border on two of the four sides.)

SuslikV:
@ferdnyc , In my view,
1) The viewport's preview never should has themed colors (if you need right answer why - because when working with colors, background should be neutral).
2) Viewport - is actually working area that has special meaning - it shows controls and objects that lies outside the output area of the current edit (animated, scaled clips etc.). I think that interactive area can has better contrast with everything else. In other words, I like to know how far I can move my clips at current  preview zoom/pan level (not even implemented yet).

ferdnyc:
> 1. The viewport's preview never should has themed colors (if you need right answer why - because when working with colors, background should be neutral).

And it doesn't. The area in question (the one that shows _around_ the preview area) is part of the preview _widget_, but not part of the video preview. The preview image is drawn _inside_ that area, in a centered rectangle with the aspect ratio locked.

Also, it's still using themed colors. The `QPalette()` function just modifies the widget theme for that particular widget. But Qt doesn't even guarantee that it'll use those exact colors, other factors can affect widget theming.

But that's fine, because the color only affects the widget structure, it's not part of the preview image.

> 2. Viewport - is actually working area that has special meaning - it shows controls and objects that lies outside the output area of the current edit (animated, scaled clips etc.). I think that interactive area can has better contrast with everything else. In other words, I like to know how far I can move my clips at current preview zoom/pan level (not even implemented yet).

That's all well and good, and If someone wants to implement an expanded viewport, that could be done. And I agree, it should not be drawn using the theme colors. By the same token, If panning/zooming/etc. are ever implemented for the Video Preview widget, then presumably its aspect ratio will be unlocked, and the painted area can make use of the entire widget space, at which point it'll be colored however the implementation decides to color it. 

But right now, the Video Preview is a fixed-aspect-ratio rectangle drawn inside a widget that may have larger dimensions in two directions. It just makes sense for that area to be filled with the default window color. In fact, I feel it's a _mistake_ to imply that those border areas are in any way special, by painting them in a different color. 

The hangup here seems to be the fact that OpenShot's Transform tool will paint markers in that border area, when video content extends off the preview area. That fact shouldn't be taken to imply that the border area is **part of** the preview; it's not. The widget paints the markers, but libopenshot renders the preview. So, it's not even _possible_ for any sort of image/video content to extend into those border spaces the same way the markers do, at least with the current implementation, since the renderer doesn't have access to that area. 

That's why the markers are the _only_ things that extend into those spaces. And it's why I say that area is part of the widget structure, not the preview area, and should be colored accordingly to avoid implying otherwise.

jonoomph:
I thought previously we had the viewport black, and you couldn't tell the difference between the black UI and the black rendering from libopenshot... so it would sometimes appear that your margins/borders were larger than reality. So, making it slightly darker was my solution (at the time). However, I do like @ferdnyc 's argument that it should be the same color as the interface itself. That actually makes a lot of sense.

ferdnyc:
It's funny, not long after this conversation I came upon issue 2731, and while it's a bit hard to follow everything going on in there, it _seems_ to me like one issue is that the user was being confused by those dark areas, into thinking the preview was showing the wrong aspect ratio. And, in fact, it sounds like their concern was that the _export_ would therefore have black bars, just because the preview appeared to:

> Openshot will not change the default profile. I'm editing NTSC 640x480 (Preview window is anamorphic at 720x480); when I set the profile to VGA NTSC in Edit > Preferences and restarted the software, the preview screen came in at 854x480, a 16:9 aspect ratio. When I chose the VGA NTSC (640x480) profile in File > Choose Profile, no change. The only time the preview window can be sized with no bars is at the 16:9 aspect ratio. 

I didn't mention it at the time, because I felt I'd already more than sufficiently made my point here. But I really do think those areas of not-obviously-window-border surrounding the preview are a source of confusion for some users, and as a result they're making bad assumptions about the effects of their actions in OpenShot.

SuslikV:
So, what is the decision? Drop?

SuslikV:
@ferdnyc nothing is left from my own PR, but I think that this change is important.

Background color isn't forced anymore. Screenshot and behavior similar to https://github.com/OpenShot/openshot-qt/pull/2836#issuecomment-508961666 (Video Preview area has no other background but window own color).

SuslikV:
Comment in the code was changed to represent actual behavior of the subsequent string.

src/windows/video_widget.py
```diff
@@ -82,7 +82,7 @@ def paintEvent(self, event, *args):
         painter = QPainter(self)
         painter.setRenderHints(QPainter.Antialiasing | QPainter.SmoothPixmapTransform | QPainter.TextAntialiasing, True)
 
-        # Fill background black
+        # Fill background with theme color instead of black
         painter.fillRect(event.rect(), self.palette().window())
```
ferdnyc: With `Qt.WA_OpaquePaintEvent` set, it _may_ not even be necessary to do this manual `fillRect()`. I think the widget should be filled with its background color by default. (And if it is redundant, might as well avoid it, since we're going to be repainting over most of the area with the video preview itself.)

ferdnyc: Eep! Never mind, I just tried it and I am _very_ wrong about that. The painting is a mess without the `fillRect()`,

Still, maybe change the comment to just "Fill widget area with theme color"? ("instead of black" sounds so weirdly specific. It's "instead of white" or "instead of purple", too, but listing all the colors it's _not_ would be tedious.)

ferdnyc:
Looks good. @jonoomph indicated he was on board with this, so merging. Thanks @SuslikV !

PR #2837, opened 2019-06-15
Install icons to hicolor/NxN instead of hicolor/N

Bluewind:
Nearly all other Arch Linux packages also follow this pattern and the Freedesktop specification also only uses this pattern when providing examples.

References:
https://standards.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html
https://bugs.archlinux.org/task/62900

Note: I didn't verify how this behaves in GUI file managers/GUI application launchers since I do not use any of them. I created this patch just based on the bug report and the icon theme spec.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -71,9 +71,9 @@ os_files = [
     ('share/pixmaps', ['xdg/openshot-qt.svg']),
     # XDG Freedesktop icon paths
     ('share/icons/hicolor/scalable/apps', ['xdg/openshot-qt.svg']),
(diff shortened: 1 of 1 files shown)


ferdnyc:
Huh. I never noticed that! While both paths are supported on Fedora, OpenShot is literally the only app on my system using `/usr/share/icons/hicolor/{64,256,512}/apps/`. Everything else is in the `NxN` paths. Thanks!

PR #2846, opened 2019-06-30
SVG hardware-accel icons

ferdnyc:
This started innocently enough an hour or two ago, when I was looking at the "Target:" list in the Export dialog and thought, "Those hardware-accel icons are kind of spindly, and the text is a bit hard to read. Maybe I'll go in and boldface it." So I went poking around.

- Then I discovered they were all PNG files.
- Then I discovered there was an SVG "master" file, but it contained _all_ of the icons, separately, meaning they'd all have to be modified, then re-exported and chopped back up.
- Then I discovered the code was loading them directly from disk, passing `QIcon()` a filesystem path each time.

"Well, this won't do at all!" was my excessively dramatic reaction, especially considering I'm just sitting here in front of my computer, completely alone in my apartment.

So, cut to the aftermath that is this PR. Here, I've:
- Deleted all of the PNG icon images.
- Replaced the source SVG master with a _template_ SVG `hw-template.svg`, stored as both Inkscape-optimized and hand-optimized SVG.
- Included `src/images/gen-hw-icons.sh`, which just runs `sed` a bunch of times to produce individual `hw-accel-$TECH.svg` icon files from the `hw-template.svg` template.
- Created a `hw-accel-none.svg` icon for CPU rendering. (It doesn't get generated from the template, since its design is different. You have to edit it directly.)
- Added all of these new `hw-accel-blah.svg` icons to the Qt resource file template, and compiled a new `openshot_rc.py` from it. (`src/images/Makefile` is also updated so that it'll recompile the QRC if any of the HW accel icons is updated.)
- Modified the code in both `src/windows/preferences.py` and `src/windows/export.py` to load the appropriate SVG icons from the compiled resources, like every other icon.

...Oh, yeah, and along the way I did get around to boldfacing the icons, too. Actually, I sort of totally redesigned them, the outliney thing wasn't working for me. And now there's even more differentiation between accelerated and CPU — though I feel like the CPU icon is a bit too dark, maybe. (Still, it can easily be edited, because SVG!)

Screenshot:

openshot-svg-hw-icons

Oh, yeah, and the DirectX icon's label is in mixed-case, now, because it's supposed to be. (Surprisingly, none of the other technologies use any lowercase letters in their name. I thought at least NVDEC/NVENC would come through with a lowercase 'v', but nope.)


ferdnyc:
It's not your imagination, the templated (green) icons _are_ a little misshapen. I wasn't sure at first, but I happened to browse past one of them zoomed in just now and they're definitely lumpy. It's not so noticeable on its own, but sticks out when compared to the smooth, rounded curves of the "CPU" icon.

I'll get that smoothed out in the template, then generate an updated set of files and update this PR.

jonoomph:
@ferdnyc I love this PR! :rofl:

jonoomph:
Feel free to merge this one once the rounded corners are smoothed to your liking. LGTM!

ferdnyc:
Screenshot from 2019-07-01 19-31-31


Muuuuch better. Travis is done chewing on it, so merging now....

PR #2849, opened 2019-07-01
Fix escaped char in Verpatch path

jonoomph:
This PR is a fix to some issues with https://github.com/OpenShot/openshot-qt/pull/2810. An invalid path and an escaped character are breaking the Windows builds. This fixes them... hopefully.


jonoomph:
Also adding some additional build-server logging, so we get a better sense of progression

jonoomph:
About to merge this one

PR #2853, opened 2019-07-08
project.get(): Accept non-list args (strings)

ferdnyc:
This is another entry in my award-ineligible series, "Don't make programmers write awkward code" (see OpenShot/libopenshot#266). The target today is `project.get()` in `classes.project_data`.

### Don't make programmers write awkward code

All over the source tree, there are requests for project data like this:
https://github.com/OpenShot/openshot-qt/blob/eef69337e077162c589f2580c2503a218875f844/src/classes/timeline.py#L46-L52

Unlike any other lookup method, `ProjectDataStore.get()` _requires_ that it be passed a list, or it will fail with a warning:
https://github.com/OpenShot/openshot-qt/blob/eef69337e077162c589f2580c2503a218875f844/src/classes/project_data.py#L68-L71

But that just makes the calling code awkward. So, with this change, a non-empty, non-list arg is simply promoted to a single-item list by the _code_, instead of by the _coder_:
```python
        if not key:
            log.warning("Cannot get empty key.")
            return None
        if not isinstance(key, list):
            key = [key]
```

That allows `project.get("fps")`, `project.get("width")`, etc. to do what's expected, and makes the code easier to write and to read.

### DIFF
--- a/src/classes/exporters/edl.py
+++ b/src/classes/exporters/edl.py
@@ -46,8 +46,8 @@ def export_edl():
     edl_string = "%03d  %-9s%-6s%-9s%11s %11s %11s %11s\n"
 
     # Get FPS info
-    fps_num = get_app().project.get(["fps"]).get("num", 24)
-    fps_den = get_app().project.get(["fps"]).get("den", 1)
+    fps_num = get_app().project.get("fps").get("num", 24)
+    fps_den = get_app().project.get("fps").get("den", 1)
     fps_float = float(fps_num / fps_den)
 
     # Get EDL path
@@ -67,7 +67,7 @@ def export_edl():
     parent_path, file_name_with_ext = os.path.split(file_path)
     file_name, ext = os.path.splitext(file_name_with_ext)
 
-    all_tracks = get_app().project.get(["layers"])
+    all_tracks = get_app().project.get("layers")
     track_count = len(all_tracks)
     for track in reversed(sorted(all_tracks, key=itemgetter('number'))):
         existing_track = Track.get(number=track.get("number"))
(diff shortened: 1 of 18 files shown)


ferdnyc:
The **BEST** part of this, the great irony of `project.get()` _requiring_ that its argument be a list? The whole reason that was done was so that sub-objects could be looked up by path, using the list arg. So, `project.get(["fps", "num"])`, for example. And, that functionality _is_ made limited use if, by things like the `filter()` methods in the `QueryObject` classes.

But `project.get()` is **never**, _**never-EVER**_ called that way, not even once, in the actual code. Not even where it **COULD** be! Even where sub-keys are looked up, the code doesn't query them using a list. Instead?
https://github.com/OpenShot/openshot-qt/blob/eb3b19cc548c76aaca8ec96c311f8e2d2c9e2b12/src/classes/exporters/edl.py#L49-L50

https://github.com/OpenShot/openshot-qt/blob/eb3b19cc548c76aaca8ec96c311f8e2d2c9e2b12/src/windows/views/timeline_webview.py#L1772-L1775

If anyone was ever curious how you can tell when your API isn't quite in sync with the needs of its consumers, _that's how_.

DylanC:
@ferdnyc - Simple but effective. I like this change.

ferdnyc:
@DylanC Thanks! I find the whole idea of the "series" pretty entertaining, actually. I can imagine it really taking off, right down to an awareness campaign featuring awkward novelty t-shirt designs. :wink: 

This code was what initially inspired it, actually. I ended up tackling libopenshot's logging calls first only because the code they forced was so much _**more**_ awkward and ugly. This is admittedly minor by comparison. We get to type a whole 2 less characters per call, woo.

DylanC:
@ferdnyc 

> This is admittedly minor by comparison. We get to type a whole 2 less characters per call, woo.

Any time saving is better than no time saving. 😄

PR #2856, opened 2019-07-09
Default to three tracks for new projects

ferdnyc:
Five tracks is excessive as a starting configuration. Few users will ever create projects that complex, and with the current default window geometry not even able to fit _two_ tracks, it's just a waste of space. This reduces the default track count to three for both new projects and the debugging interface.

### DIFF
--- a/src/settings/_default.project
+++ b/src/settings/_default.project
@@ -49,20 +49,6 @@
       "number": 3000000,
       "y": 0,
       "lock": false
-    },
-    {
-      "id": "L4",
-      "label" : "",
-      "number": 4000000,
-      "y": 0,
-      "lock": false
-    },
-    {
-      "id": "L5",
-      "label" : "",
-      "number": 5000000,
-      "y": 0,
-      "lock": false
     }
   ],
   "markers": [  ],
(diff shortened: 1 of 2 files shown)


SuslikV:
I'm using 4-5 usually. The bottom one is simple white for "flashing" transitions, the upper one for text (titles), at least 3 in the middle for other transitions (I prefer to see where each track starts and ends, and what is on the top). The scrolling (5 tracks) makes you realize that the application has bigger capabilities - just one of my thoughts.

@ferdnyc How many tracks you are usually using when working with OpenShot?

DylanC:
@SuslikV - I think normally I use only 2 or 3 tracks.

ferdnyc:
> @ferdnyc How many tracks you are usually using when working with OpenShot?

TBH I can't remember the last time I did anything in OpenShot other than "debug OpenShot". But I tend to keep it down to 2 or 3 max simply because it's easier to work with visually. (And I'm on a desktop with a 24" `1920×1200` display, I can't imagine how people with 13" laptop screens manage. But that applies to basically all software.)

There are certainly times when more tracks are useful. And tracks can always be added as needs arise. _Starting_ with 5 just seems excessive to me.

Especially since, now that I think about it, I usually end up working in Track 3 - Track 5, initially (because they're the visible ones, in my 3-track-tall Timeline view) — and if I need more tracks later, I'm probably going to need them _above_ those tracks, not below. So I'll end up adding new ones regardless, and either deleting or **still** not using Track 1 / Track 2.

ferdnyc:
See #2863, for a solution I hate (even though I wrote it) to this impasse.

jonoomph:
Imma close this one, since I'm happy with the current default track layout, and we support custom default project types. This might change in the future though, as we are always reevaluating UI decisions. :+1:

Issue #2841, opened 2019-06-26
dll load failed openshot


nanohtpc:
**Describe the bug:**
After install the latest daily, gives that error..

**Steps to reproduce the behavior:**
Just installed latest daily

**Expected behavior:**
To run the program ...

**System Details:**
- OpenShot Version [e.g. 2.4.4]
- Operating System / Distro: Windows 10 x64

![Screenshot (151)](https://user-images.githubusercontent.com/34278444/60190989-2ed6f100-97f9-11e9-80b9-982ef0b1e70d.png)

SuslikV:
Please, follow the instruction on your screenshot and when it fails second time - upload the log-file named `openshot-qt.log` here (thus it will be short and clean of other errors).

WARNING! By deleting the mentioned folder - you also eliminates all application's preferences (they will resets to their defaults).

nanohtpc:
well guess what, the log is EMPTY, and got the same message... 8(((

pd. how can i backup the settings then?

ferdnyc:
I can reproduce with the x86 version in Wine on Linux. Console output:
```
wine: Call from 0x7bc8eed9 to unimplemented function libopenshot-audio.dll._ZN4juce78this_will_fail_to_link_if_some_of_your_compile_units_are_built_in_release_modeC1Ev, aborting
002e:fixme:ntdll:RtlNtStatusToDosErrorNoTeb no mapping for 80000100
wine: Call from 0x7bc8eed9 to unimplemented function libopenshot-audio.dll._ZN4juce78this_will_fail_to_link_if_some_of_your_compile_units_are_built_in_release_modeC1Ev, aborting
002e:fixme:ntdll:RtlNtStatusToDosErrorNoTeb no mapping for 80000100
002e:fixme:shell:SHGetStockIconInfo Couldn`t find SIID 78, returning default values (IDI_SHELL_FILE)
002e:fixme:shell:SHGetStockIconInfo flags 0x1 not implemented
002e:fixme:shell:SHGetStockIconInfo Couldn`t find SIID 78, returning default values (IDI_SHELL_FILE)
002e:fixme:dwmapi:DwmIsCompositionEnabled 0061D040
```

Looks like an issue caused by the changes to the libopenshot-audio build. I'll hunt down what changed.

ferdnyc:
I believe OpenShot/libopenshot-audio#52 will solve this, sorry about that.

nanohtpc:
Same error on :

[...]

Issue #2748, opened 2019-04-19
error in python script when i start up openshot


AndyIreland:
![openshot error](https://user-images.githubusercontent.com/49172217/56426410-28377500-62b0-11e9-8eba-b963897f6170.jpg)

ThomasBerends:
This looks like a duplicate of #1295 . Could you check if running OpenShot as Administrator solves the issue?

Taurinus:
Exactly what I receive when launching this application.
Neither 32bit or 64bit, stable or latest beta,
always this Python error.
Other Python related programs are doing fine. Python 3.6 and Python3.7 "installed"....

SuslikV:
Also, try to change the default installation path to the folder you have create by yourself "C:\OpenShot" for example.

Taurinus:
So I changed the default installation path to somewhere in the root of my bootdrive, installed with admin rights.
Executed the application with admin rights, trying 32bit & 64bit version, always the same python error pops up.
Next thing I will try is to remove the installed Python 3.7 from windows path or even deinstall it.
And then I'm out of my guts ;-)...

GeoffAdkins:
This error is still happening on Windows 7 32 bit, only Python 3.6 installed by OpenShot, no other Python as far as I can tell. Any fix yet ?

SuslikV:
@GeoffAdkins this means that the openshot for some reason finds the file that itself is not part of the application and thus unable to load it properly. maybe the library was blocked by security software, or you have similar file in Windows system folder. For example, look at the system folder for any file named same as the any dll that is present in the _OpenShot Video Editor_ folder (look for Qt5..., avcodec..., avformat..., avutil..., swscale..., swresample... files first). Is there any of them in C:\Windows\System32\ folder?

peanutbutterandcrackers:
Could you please take a look at this page and see if the work-arounds work for you: https://github.com/OpenShot/openshot-qt/wiki/Windows-Launch-Issues-(cx_Freeze-errors) ? Thank you!

SuslikV:
At line 14 only libopenshot.dll is loads and fails, as far as I understand.

Zipher04:
I have met the same error message on my Windows 10 x86.
I have tried the steps on https://www.openshot.org/issues/new/.
And the daily build(OpenShot-v2.4.4-dev2-1562263849-1b19ae7f-7939e156-x86) gave me a new error message which says "DLL load failed"
[...]

PR #2862, opened 2019-07-10
Permission popup: Don't eat other exceptions

ferdnyc:
The new "Permission Error" popup message is causing major troubleshooting problems, because it eats **ALL** exceptions, and is triggering for lots of problems that have _nothing_ to do with permission errors. This is presenting users with incorrect information (deleting their OpenShot configuration is going to do nothing to solve their problem), and robbing us of access to the useful error information that would be presented if not for the popup.

(See #2748, #2841, probably others. I even had the box come up once during development, due to a bug in my code that had _absolutely nothing_ to do with permissions.)

This change limits the exceptions that can trigger a Permission Error popup box to **only** the actual `PermissionError` exception, which is the only one that should be thrown in the case of an actual permission error.

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -1,28 +1,28 @@
-""" 
+"""
  @file
  @brief This file creates the QApplication, and displays the main window
  @author Noah Figg <eggmunkee@hotmail.com>
  @author Jonathan Thomas <jonathan@openshot.org>
  @author olivier Girard <eolinwen@gmail.com>
- 
+
  @section LICENSE
- 
+
  Copyright (c) 2008-2018 OpenShot Studios, LLC
  (http://www.openshotstudios.com). This file is part of
  OpenShot Video Editor (http://www.openshot.org), an open-source project
  dedicated to delivering high quality video editing and animation solutions
  to the world.
- 
+
  OpenShot Video Editor is free software: you can redistribute it and/or modify
  it under the terms of the GNU General Public License as published by
  the Free Software Foundation, either version 3 of the License, or
  (at your option) any later version.
- 
+
  OpenShot Video Editor is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


jonoomph:
@ferdnyc  LG, I only worry about Windows specific failures to remove a file (i.e. if the file is in-use or locked somehow). Are we sure that PermissionError covers those edge cases? For example, I thought OSError could be raised by Windows in some of those cases.

ferdnyc:
@jonoomph Sure? No. But given the issues, erroring on the side of caution by listing specific exceptions seems the safer approach. I'm happy to add in OSError capture as well, since that's also very likely to be a filesystem error. (Though, I'll turn it back around on you: Are we sure OSError **DOESN'T** also cover things like DLL import failures?)

ferdnyc:
Hmm. OSError encompasses _all_ of this, so I think we'd have to pick and choose a bit. [Ref](https://docs.python.org/3/library/exceptions.html#exception-hierarchy)
```
 +-- OSError
      |    +-- BlockingIOError
      |    +-- ChildProcessError
      |    +-- ConnectionError
      |    |    +-- BrokenPipeError
      |    |    +-- ConnectionAbortedError
      |    |    +-- ConnectionRefusedError
      |    |    +-- ConnectionResetError
      |    +-- FileExistsError
      |    +-- FileNotFoundError
      |    +-- InterruptedError
      |    +-- IsADirectoryError
      |    +-- NotADirectoryError
      |    +-- PermissionError
      |    +-- ProcessLookupError
      |    +-- TimeoutError
```

jonoomph:
Agreed. Let me experiment on Windows real quick to see exactly what is raised.

jonoomph:
Looks like you are correct: `PermissionError` is all I can reproduce when trying to lock a file on Windows. Let's go with it!  Nice!

PR #2863, opened 2019-07-12
 Add support for a user default project file

ferdnyc:
This allows the user to create a file at `info.USER_DEFAULT_PROJECT` (set to `info.USER_PATH/default.project`) which contains their personal default project settings. If this file is found and can be successfully loaded, its contents will be used as the "New Project" initialization data. Otherwise, `info.PATH/settings/_default.project` will be loaded as usual.

Addresses concerns raised by @SuslikV in #2856 

Honestly, I'm not sure about this one... it's got the potential to cause difficulties. It's meant to be a "power user" feature, not a supported normal config, and there's a lot to be said for just telling people, "If you use this, you're on your own for problems it causes you." But, still, it's not a _great_ idea. 

(As far as problems, I'm thinking of things like new fields being added to the system default project file. Those won't automatically be carried over to users' personal defaults, they'll have to manually update them.) 

However, it's also true that the current one-size-fits-all default project may _not_ fit all users equally, and this is at least a solution to that, which doesn't mean that the defaults for _everyone_ have to try and fit the needs of users with more demanding requirements.

### DIFF
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -49,6 +49,10 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
         self.data_type = "project data"  # Used in error messages
         self.default_project_filepath = os.path.join(info.PATH, 'settings', '_default.project')
 
+        # Always set, in case the user creates the file mid-session
+        if info.USER_DEFAULT_PROJECT:
+            self.user_project_filepath = info.USER_DEFAULT_PROJECT
+
         # Set default filepath to user's home folder
         self.current_filepath = None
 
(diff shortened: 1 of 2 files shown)


src/classes/project_data.py
```diff
@@ -252,7 +256,24 @@ def _set(self, key, values=None, add=False, partial_update=False, remove=False):
     def new(self):
         """ Try to load default project settings file, will raise error on failure """
         import openshot
-        self._data = self.read_from_file(self.default_project_filepath)
+
+        # Try to load user default project
+        if os.path.exists(self.user_project_filepath):
+            try:
+                self._data = self.read_from_file(self.user_project_filepath)
+            except (FileNotFoundError, PermissionError) as ex:
```
ferdnyc: The code gets a little weird here, so I should explain:

I originally didn't have the `if` wrapped around this, and was just going to use a `FileNotFoundError` exception raised by `self.read_from_file()` to detect that the file doesn't exist.

But that doesn't work, because there's a bug in the exception code inside `read_from_file()` that loses the exception detail. (I'll file another PR for that.) And even if that bug gets fixed, `read_from_file()` **logs** whenever there's an exception trying to read a file, which I don't want for a file that's _expected_ not to be found most of the time.

So, I couldn't do things that way, and had to add the conditional existence test first.

SuslikV:
Advanced user can save its own template project (to start from it) and re-save it to any new when working on new project. As long as the file save/load compatibility will be preserved across the OpenShot versions - the method will work too.

jonoomph:
@ferdnyc Yeah, I like this change for power users! It matches the basic usage of all these .openshot_qt overrides (such as custom titles, custom transitions, etc...). I would love to work on a Wiki document which lays all of these overrides out nicely (kind of a `How to Customize OpenShot` wiki page)

ferdnyc:
@jonoomph I have been thinking, lately, that all of these great improvements have been made since 2.4.4 released — including a lot of user-facing interface changes... and yet, not one letter of documentation has been updated yet. (I'm talking in terms of the official user manual, mostly.) Seems like _some_ of the changes probably warrant documentation updates.

(I think the power-user wiki page is a great idea, too, and this particular feature definitely isn't anything for the manual. I mean, maybe that wiki page could be _linked to_ from the manual, but nothing more. ...But, that aside, docs for _non_-power users are also good!)

PR #2864, opened 2019-07-12
Check for empty/null path strings

### LINKED ISSUE #2850: Pressing Export Does Nothing (Workaround Included)
**Describe the bug:**
When I press Export on the toolbar (or use the menu or keyboard shortcut), Openshot doesn't open the Export dialog. The program behaves normally afterward, as though I had canceled the dialog (i.e.: it doesn't crash or hang).

1. Open Openshot
2. Click on 'Export'
3. Nothing happens...it's as though the button isn't connected to anything

**Expected behavior:**
I expected to see the Export dialog appear.

**System Details:**
- OpenShot Version [2.4.4]

**Log Files:**
I went through the log and found this exceptional message :)
  File "C:\Program Files\OpenShot Video Editor\windows\main_window.py", line 690, in actionExportVideo_trigger
    win = Export()
  File "C:\Program Files\OpenShot Video Editor\windows\export.py", line 143, in __init__
    if os.path.exists(export_path):
  File "C:\msys64\mingw64\lib\python3.6\genericpath.py", line 19, in exists
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType

**Exception / Stacktrace:**
See above -- my "export_path" project setting was somehow set to null. I was able to work around the problem by replacing null with a valid path, e.g.: "C:\\path\\to\\videos". I would expect the __init__ routine to be able to handle the case where export_path is None.

ferdnyc:
* If os.path.exists() is passed a null argument (say, if it came from a user's JSON data), it will throw a TypeError exception. Protect `export.py` against this. Also correct a typo.
* Refactored some path-related logic in `main_window.py`

Fixes #2850 reported by @dadkins1

### DIFF
--- a/src/windows/export.py
+++ b/src/windows/export.py
@@ -136,12 +136,12 @@ class Export(QDialog):
         self.timeline.Open()
 
         # Default export path
-        recommended_path = recommended_path = os.path.join(info.HOME_PATH)
+        recommended_path = os.path.join(info.HOME_PATH)
         if app.project.current_filepath:
             recommended_path = os.path.dirname(app.project.current_filepath)
 
         export_path = get_app().project.get(["export_path"])
-        if os.path.exists(export_path):
+        if export_path and os.path.exists(export_path):
             # Use last selected export path
             self.txtExportFolder.setText(export_path)
         else:
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -465,6 +465,11 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         app = get_app()
         _ = app._tr  # Get translation function
 
+        # First check for empty file_path (probably user cancellation)
+        if not file_path:
+            # Ignore the request
+            return
+
         # Do we have unsaved changes?
         if get_app().project.needs_save():
             ret = QMessageBox.question(self, _("Unsaved Changes"), _("Save changes to project first?"), QMessageBox.Cancel | QMessageBox.No | QMessageBox.Yes)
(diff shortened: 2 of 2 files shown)


src/windows/export.py
```diff
@@ -136,12 +136,12 @@ def __init__(self):
         self.timeline.Open()
 
         # Default export path
-        recommended_path = recommended_path = os.path.join(info.HOME_PATH)
+        recommended_path = os.path.join(info.HOME_PATH)
```
ferdnyc: This has actually been like that since 2016! None of us noticed. :laughing:

src/windows/main_window.py
```diff
@@ -465,6 +465,11 @@ def open_project(self, file_path, clear_thumbnails=True):
         app = get_app()
         _ = app._tr  # Get translation function
 
+        # First check for empty file_path (probably user cancellation)
+        if not file_path:
+            # Ignore the request
+            return
```
ferdnyc: I moved this check up here, because it felt cleaner than doing that `if file_path != ""` down below. (An empty string evaluates to false in Python, so this `if` is equivalent. Doing it sooner means we can bail on the whole thing, if there's no path to attempt loading anyway.)

src/windows/export.py
DylanC: @ferdnyc - Oh dear. This is funny. 😆

src/windows/main_window.py
DylanC: @ferdnyc - Definitely a good idea to bail on it sooner. 👍

DylanC:
@ferdnyc - Happy to get this one in. Null checking is very sensible for any project.

PR #2867, opened 2019-07-14
Fix import error popup, other exceptions

ferdnyc:
It was pointed out to me that there was a _second_ message popup box in `classes.app` (like the one fixed in #2862, which had even worse sins:
* It was also eating _all_ exceptions, even ones that might have nothing to do with import errors.
* It was displaying the same advice about deleting the user's profile directory, even though there is no situation I can think of where that would be effective at fixing an import error. (Unlike with the other popup, where deleting the directory most likely would solve a permissions error.)
* Also like the other popup, because it was catching the exception and doing nothing useful with it, it was robbing us of important troubleshooting information.

So, while I kept the popup, I removed the bad advice about deleting the user profile, and instead display a full traceback (using the standard `traceback` module) in the popup dialog. I also `raise` the exception, so that the same traceback will be printed to stderr. Any other exception blocks in `classes.app` that end in `sys.exit()` received the same `raise` statements, so that tracebacks are always captured.

Now, if there's an import error (like this one I engineered for myself, by deleting `reroute_output` from `logger.py`), the popup will look something like this:

image

And the same traceback will appear on the console, for debugging purposes.

If we want to make the popup more friendly (though that's hard when it can't be localized to the user's language), I can see including some advice about reinstalling OpenShot as a possible solution to the problem. But deleting their configuration is simply not going to solve an ImportError.

Addresses #2748

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -171,7 +171,6 @@ class OpenShotApp(QApplication):
                 QApplication.setFont(font)
             except Exception as ex:
                 log.error("Error setting Ubuntu-R.ttf QFont: %s" % str(ex))
-                raise
 
         # Set Experimental Dark Theme
         if self.settings.get("theme") == "Humanity: Dark":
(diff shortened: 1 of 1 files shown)


ferdnyc:
Travis passed this, so I'm going to merge it. This issue is still causing real pain for users, and I want to get Daily Builds generated ASAP with the necessary fixes so that we can finally debug their problems.

PR #2868, opened 2019-07-14
Fix path in show_translations.py

ferdnyc:
The `.pot` file is a directory lower than expected in the code.


ferdnyc:
I can't imagine what this could possibly hurt. (Please berate me severely for merging this, if it hurts anything.)

PR #2869, opened 2019-07-14
InnoSetup installer: Add over a dozen langs, retire environment-modification code

ferdnyc:
A pair of InnoSetup config file changes. I'm intentionally burying the lead, here.

### 1. Retire the environment-variable deletion code

Next month will mark the three-year anniversary of the commit (fde27d52d3f3a36f4903eb49f5090384bf60b342) which _removed_ the code in `freeze.py` where `QT_PLUGIN_PATH` and `MAGICK_CONFIGURE_PATH` environment variables.

In honor of that anniversary, I feel it's time to retire OpenShot's code (found in the Windows installer) that arbitrarily _deletes_ those variables from the user's environment during install. If anyone installing OpenShot today has either of those variables set in their environment, it's probable they **didn't** come from OpenShot, and we probably shouldn't be removing them without their say-so (anymore).

### 2. Over a dozen new languages! (And another 6 available with an upgrade.)

I went through the InnoSetup site's [Translation files page](http://www.jrsoftware.org/files/istrans/) and added nearly every available user-contributed translation that _wasn't_ in the OpenShot installer configs, as long as it was listed as being for version 5.5.3+, which corresponds to the version currently in use by OpenShot. (InnoSetup 5.5.3 was released back in 2013. The OpenShot builders are currently using 5.5.7, released at the end of 2015. The most recent release in the 5.x series was 5.6.1, which came out a year ago. The only newer translation file generation is 6.0.0+, which are targeted for the current 6.0.2 version, released in April.)

I also grabbed the 6.0.0+ files, and added those languages to the config file in a commented-out state, since it isn't clear to me that 6.0.0+ files can be used in InnoSetup 5.5.x. (It isn't clear to me that they _can't_, either, since the [differences are pretty minimal](https://github.com/jrsoftware/issrc/commit/1d6d6869b225c73cde4d3e0541b87264c36f3954) between 5.5.3+ and 6.0.0+ translation files. But it just plain isn't clear to me, period.) So, anyway, as I said those languages are commented out, but they're there for later if the InnoSetup version does get upgraded.

[Languages.zip](https://github.com/OpenShot/openshot-qt/files/3389869/Languages.zip)

Here's a zip file of all the added translation files, the files in the zip's `Languages/` directory are the 5.5.3+ files, which would go into whatever directory corresponds to `"compiler:Languages\"` in the setup code. (It's the same place the existing translations are stored in, presumably somewhere within the InnoSetup application directory.)

There's also a subdirectory, `Languages/Inno6.0/`, which contains the 6.0.0+ translation files for the commented-out languages, so we'll have those at the ready as well, should we choose to upgrade InnoSetup. A couple of the files in there are actually newly-official for 6.0, so if the upgrade is done then they _should_ already be installed with the software. A couple of _others_ are languages that **used** to be bundled, but were demoted to "unofficial" status since they weren't being maintained. But those languages still had unofficial 6.0.0+ updates, so I included the updated files since they presumably don't come bundled. There's a note about those in the `Languages/Inno6.0/` directory.

The reason I'm attaching this as a zip rather than committing the files to the repo is, I have _no idea_ what the license status is of these files. I have no idea if we're allowed to redistribute them or not, since the issue is never even remotely addressed by InnoSetup. Someone raised this exact issue once on their user forums, [way back in 2005](http://news.jrsoftware.org/news/innosetup.translations/msg00649.html)... and their post was never answered. So, I figure better to err on the side of caution, and not include them in the repo. I don't feel like trying to explain the licensing situation on that to the Fedora folks, since I have absolutely no idea what the explanation would be.

As I said, I _think_ every one of these languages is supported by OpenShot. I skipped a couple that I was more sure _**aren't**_. But if I'm wrong, and any of the languages here are ones that OpenShot is _not_ translated into, then it'd probably be best to remove them. Just seems cruel, offering your installer in a language that your application doesn't support!


jonoomph:
FYI: Build server is now running 6.0.2 of InnoSetup

jonoomph:
Merged in develop and resolved conflicts... also enabled those disabled languages, since we are on 6.0.2 of InnoSetup (as mentioned previously).

jonoomph:
I have extracted these languages into our build server InnoSetup Languages folder... so, in theory, this should work fine. Here we go...

jonoomph:
A few issues on the build server after I merged this:
- Missing a few referenced language files in the .zip (I found them online and put them on the build server)
- Mismatching extension on one of them (isl vs islu)
- Regarding licensing of the translations, I don't have a great answer for that. I'm going to include all of them. InnoSetup translations are clearing created to be used in other products... and I see no mention of licensing restrictions. So until someone complains... :wink:

ferdnyc:
> A few issues on the build server after I merged this:
> 
> * Missing a few referenced language files in the .zip (I found them online and put them on the build server)

Hm. That's odd. Those may be ones that were either purported to be included with 6.0 (but weren't, or were later dropped again). 

Or — more likely — they were languages that 6.0 lacked, but had been included in 5.3.x (according to their translations page), so (without being able to check) I went on the assumption they were already present on the OpenShot build servers. But since you installed 6.0.2 on a new system, rather than upgrading from that old 5.x installation, there were no old versions to retain.

> * Mismatching extension on one of them (isl vs islu)

Ah, whoops. Well, that should be a moot point anyway, now. My understanding is that, if we're upgraded to InnoSetup 6.x, that distinction goes away and there's no more `.islu`. (Actually, AIUI _**every**_ translation file is "really" ISLU now, since it used to be a special extension for Unicode translations — but now they're all Unicode by default.)

> * Regarding licensing of the translations, I don't have a great answer for that. I'm going to include all of them. InnoSetup translations are clearing created to be used in other products... and I see no mention of licensing restrictions. So until someone complains... wink

Oh, I'm not worried about them being on the OpenShot build servers, that's _**fine**_, and what's anyone going to say? You're the only person with direct access to them! So it's not like they're actually being included in anything we distribute. (The translation strings they contain are compiled into the setup executable, opaquely, but the translation files _themselves_ never leave the build servers. Nor does that compiled executable get distributed anywhere other than the openshot.org site.)

My concern would've been if we were to check them into the _source tree_, which would then require that the relevant applicable license terms be verified by the gatekeepers of the package collections for free-software-only distros like Fedora, in order to have OpenShot packages built from that source tree included in their official repos. So, let's definitely not ever do that, because those people scare me.

PR #2871, opened 2019-07-15
Fix source comment typos

luzpaz:
Found via `codespell`


ferdnyc:
Thanks @luzpaz , all look good from here!

PR #2872, opened 2019-07-16
Windows installation polishing

### LINKED ISSUE #2777: not opening OSP files via file manager
**Describe the bug**
not opening OSP files

1. Go to osp file
2. Click on osp
nothing happens/ OSP does NOT start
**Expected behavior**
I expect OSP to open the osp file

**System Details**
- OpenShot Version: 2.4.4-dev1
- libopenshot Version:  0.2.3
- Platform: Linux-5.0.0-13-generic-x86_64-with-Ubuntu-19.04-disco
- Processor: x86_64
- Machine: x86_64
- Python version: 3.4.3
- Qt5 version: 5.2.1
- PyQt5 version: 5.2.1
- LANG Environment Variable: en_US.UTF-8
- Daily Build Installed: http://github.com/OpenShot/openshot-qt/releases/download/OpenShot-v2.4.4-dev1-1557343973-17a22589-9a1a6093-x86_64.AppImage

**Log Files**
- http://logs.openshot.org/h2ujy8jpq72m0fhaerl78pq0sbcror9f/openshot-qt.zip
- http://logs.openshot.org/h2ujy8jpq72m0fhaerl78pq0sbcror9f/libopenshot.zip

**Exception / Stacktrace**
```No stacktrace found in log files```

video of .osp file NOT opening
https://youtu.be/VTVhUqfotFY

ferdnyc:
I threatened this forever ago, but never got around to submitting it. This PR adds some additional polish to the Windows install of OpenShot, nothing major but a few "little touches" for things that _are_ noticeable when they're **not** corrected. Specifically:

* Optionally (default enabled), file associations for `.osp` project files will be created during the install, so that double-clicking a project file will open it in OpenShot, instead of the user having to choose a program the first time. Those associations will also be cleaned up by the OpenShot uninstaller, when it's used.
* Similar to my MacOS PR #2769, several additional (read: larger) sizes are added to the OpenShot icon file, because `64x64` is a pathetic size in today's high-res world. Support for `256x256` icons was added way back in Windows _Vista_, and Win 10 apparently uses even `512x512`, `768x768`, and `1024x1024` sizes on HiDPI displays. This doesn't go that high, but it does add `128x128`, `256x256`, and `512x512`.
* Instead of setting the _same_ URL three times in a row (which looks **TERRIBLE** in the Win7 Programs & Features control panel, at least), the Subreddit address is used for the "Support" URL, and the "Updates" URL is not set because updates are available at the Publisher URL (which is the same as before).

Fixes #2777

The metadata difference is noticeable even in Win7.

#### Before (ew):

Screenshot from 2019-05-06 06-34-19

#### After (ooh, modern):

image


ferdnyc:
This should be the end result, if I got the URL assignments right. (It's confusing because InnoSetup's mapping is a bit weird — I _believe_ their "AppPublisherURL" is what's mapped to the "URLInfoAbout" registry string, which displays as the "Support link". While their "AppSupportURL" is what's mapped to "HelpLink" in the registry, which displays as "Help Link". It's a very unfortunate clash of terminology.)

image

ferdnyc:
I suppose technically we could insert "https://openshot.org/download/" for the "Updates link". Not sure it's really _necessary_, but it's an option...

ferdnyc:
I was inspired by the fact that I often have multiple versions of OpenShot installed (both Win32 and Win64, at least), and don't necessarily want the .osp file association stolen by the most recent one I install, to make the setting of that association optional.

The commit I just pushed will add a second checkbox underneath "Create a desktop shortcut", default checked, with the label being "Associate OpenShot Video Editor with the .osp file extension" — but localized into the installer language.

image

jonoomph:
Just fixed a conflict with the new **firewall** commit. I'll merge this once Travis finishes.

jonoomph:
Thanks @ferdnyc !!!  This is great.

PR #2873, opened 2019-07-17
Various UI fixes (export dialog, icon resources)

ferdnyc:
This PR combines a few changes, mostly touching on
the UI files in `src/windows/ui/`, or related files.

* For the Export dialog (`export.ui`), I...

   - Named all of the Advanced view toolbox pages (e.g.  `pageVideoSettings`, not `page_2`)
   - Un-rowspanned the tab widget, unnecessary
   - With a combination of stretch settings and size constraints, made the tab widget always expand to fill the window height, with the other elements remaining fixed height
   - Removed the unnecessary vertical spacer UNDER the tab widget, to make the previous possible

* I added more icons to the compiled resources.
   - `openshot.svg` is now at the root prefix (`:/openshot.svg`)
   - New `/about` prefix, containing:
     - `AboutLogo.png`
     - `{bitcoin,kickstarter,patreon,paypal,python,star}-icon.png`
* All of the `.ui` files for windows/dialogs were updated to use `:/openshot.svg` as their window icon, instead of loading it from disk.
* The code in `credits_model.py` was similarly modified to use `:/about/...` resources for the icons it uses.


ferdnyc:
The changes to the Export dialog primarily mean that, now, when the window expands vertically to fit the larger sections of the Advanced toolbox, the expansion will happen _entirely_ within the tab widget, with all of the other widgets staying at a fixed vertical height. Meaning, no more of this:

Screenshot from 2019-07-17 10-32-58

Instead, the tab widget body always reaches all the way down to the progress bar:

Screenshot from 2019-07-17 09-18-06

It's minor, sure, but it just looks better that way.

jonoomph:
Nice, I like the changes to the Export dialog!!!

jonoomph:
Although, I'm not sure I love the "CPU" icon color in that screenshot. Pretty low contrast at the moment. :wink:

ferdnyc:
@jonoomph 

Agreed, I said the same thing in #2846 (the one that changed the color) after looking at it in the interface. But it's been that color since #2846 was merged, this PR didn't change it. :wink: 

Still, easy enough to fix by just editing `src/images/hw-accel-none.svg` and changing both ` stroke="#036980"` on the `<g>` tag and `fill="#036980"` on the `<text>` tag.

...Which I just did, and after fighting with blues that contrasted poorly with the interface gray for a bit, I got fed up and switched to light orange. Much more visible. PR incoming.

PR #2879, opened 2019-07-20
Remove ellipses from "New Project" menu/binding

### LINKED ISSUE #2806: Inconsistent behaviour of 'New Project ...'
This might explain #1234 and #1879.
At first, I also thought Openshot wouldn't work at all. But then I found that 'New Project...' actually works, even removes previous clips and stuff. Though it is inconsistent w.r.t. a basic UI paradigm: a command followed by 3 dots requires some user input. In this case it doesn't. I think that a file selection box (like the one that pops up when you click '+') would sort things out pretty well: offering the user a potential location, an existing or new folder and a project file name. Then the .osp as well as resulting, rendered, file would be stored there. And eventually to be added files ('+') would by default searched in there.

ferdnyc:
As pointed out by @udippel in #2806, our "New Project..." action does not even remotely fit the criteria for [menu items that should be followed by ellipses](https://stackoverflow.com/questions/637683/when-to-use-ellipsis-after-menu-items) as provided by all major systems' Human Interface Guidelines. This change removes the ellipses from both the File menu item, and the keyboard binding label in the settings file (as displayed in the Preferences' Keyboard tab).

Fixes #2806

### DIFF
--- a/src/settings/_default.settings
+++ b/src/settings/_default.settings
@@ -583,7 +583,7 @@
   },
   {
     "category": "Keyboard",
-    "title": "New Project...",
+    "title": "New Project",
     "restart": true,
     "setting": "actionNew",
     "value": "Ctrl+N",
--- a/src/windows/ui/main-window.ui
+++ b/src/windows/ui/main-window.ui
@@ -504,7 +504,7 @@
      <normaloff>:/icons/Humanity/actions/16/document-new.svg</normaloff>:/icons/Humanity/actions/16/document-new.svg</iconset>
    </property>
    <property name="text">
-    <string>New Project...</string>
+    <string>New Project</string>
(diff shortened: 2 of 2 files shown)


SuslikV:
Just merge it and forget until someone implement some dialog to config new project or whatever.

ferdnyc:
@SuslikV At which point, ellipses get added to the menu option. I would certainly hope anyone creating a whole new dialog for the interface would think to update the menu option they're connecting it to, that's pretty basic UI development.

jonoomph:
In old versions of OpenShot, we did have a "New Project" dialog... and thus, the "..." is probably a legacy left over. :smile:

jonoomph:
LGTM!

PR #2880, opened 2019-07-20
Recents list: Never add relative paths

### LINKED ISSUE #2826: Relative project file paths opened on the command line are added to the Recent list as relative paths
**Describe the bug:**
If an `.osp` file is opened via the command line using a relative path, that relative path is added _unchanged_ to the Recent Projects list, making it invalid if OpenShot is launched differently in subsequent runs, even if the actual file is still valid and present

All paths added to Recent Projects should first be made absolute, so that their availability is not dependent on OpenShot's current working directory.

1. Open a terminal
2. Navigate to a directory with a `someproject.osp` file in it
3. Launch OpenShot from the terminal, with the `someproject.osp` file included in the arguments as a relative path, e.g.
    ```sh
    openshot-qt ./someproject.osp
    ```
4. Quit OpenShot
5. Change to a different directory using the `cd` command
6. Launch openshot from that location, _without_ supplying a project file
7. Go to File > Recent Projects in the menu
8. Click on the `./someproject.osp` entry and attempt to reopen the project

**Expected behavior:**
The project file, still located in the same place on the disk, is reopened.

**Actual results:**
A status bar message that the project file is missing, because its path is no longer `./someproject.osp`

**System Details:**
- OpenShot Version: 2.4.4 / 2.4.4-dev2
- Operating System / Distro: Fedora 30

ferdnyc:
Because relative paths on the Recent Files list break whenever you
run OpenShot from a different directory, the submitted path should
always be converted to absolute when inserting into `recent_projects`.

Also, add a defensive check for null/empty `file_path`, before we
start passing it to os library functions.

Fixes #2826

### DIFF
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -851,12 +851,15 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
 
     def add_to_recent_files(self, file_path):
         """ Add this project to the recent files list """
-        if "backup.osp" in file_path:
+        if not file_path or "backup.osp" in file_path:
             # Ignore backup recovery project
             return
 
         s = settings.get_settings()
         recent_projects = s.get("recent_projects")
+
+        # Make sure file_path is absolute
+        file_path = os.path.abspath(file_path)
 
         # Remove existing project
         if file_path in recent_projects:


jonoomph:
LGTM!

PR #2884, opened 2019-07-22
Destroy webview when exiting

ferdnyc:
Since commit aa165c44ec3a40ee1abef9cd7ea17934d81ca706 I've been having issues with OpenShot not exiting cleanly, due to timers in the WebView process that continue trying to access Timeline data after the object's been destroyed. (See comments at aa165c44ec3a40ee1abef9cd7ea17934d81ca706.)

This change sets the TimelineWebView attribute `WA_DeleteOnClose`, which will cause the widget to be destroyed on close. It also calls `timeline.close()` _before_ destroying libopenshot's Timeline object, to stop the WebView timers attempting further data access.

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -136,6 +136,9 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         self.preview_parent.background.exit()
         self.preview_parent.background.wait(5000)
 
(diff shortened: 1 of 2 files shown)


jonoomph:
This looks great! Nice catch

PR #2885, opened 2019-07-22
Migration to new Windows builder (with updated dependencies)

jonoomph:
This includes updated versions of MSYS2, FFmpeg, Resvg, cx_Freeze, Qt5, PyQt5, and much more. This also required some changes to `freeze.py` and `build-server.py`, since so many things were updated. I've also updated the Wiki build instructions for Windows: https://github.com/OpenShot/libopenshot/wiki/Windows-Build-Instructions.

On a side note, now both 32-bit and 64-bit Windows installers include libresvg.

This took a bunch of trial and error, so I will be collapsing the terribly named commits into 1 on merge. :smile:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -83,7 +83,7 @@ windows-builder-x64:
     - Expand-Archive -Path artifacts.zip -DestinationPath .
     - Copy-Item "$CI_PROJECT_DIR/build/install-x64/python/*" -Destination "$CI_PROJECT_DIR"
     - $env:Path = "$CI_PROJECT_DIR\build\install-x64\lib;$CI_PROJECT_DIR\build\install-x64\lib\python3.7\site-packages\;C:\msys64\mingw64\bin;C:\msys64\mingw64\lib;C:\msys64\mingw64\lib\python3.7\;C:\msys64\mingw64\lib\python3.7\site-packages\;C:\msys64\mingw64\lib\python3.7\site-packages\PyQt5;C:\msys64\usr\lib\cmake\UnitTest++;C:\msys64\home\jonathan\depot_tools;C:\msys64\usr;C:\msys64\usr\lib;" + $env:Path;
-    - $env:PYTHONPATH = "C:\msys64\usr\lib;C:\msys64\usr\lib\site-packages\;C:\msys64\mingw64\lib\python3.7\;C:\msys64\mingw64\lib\python3.7\site-packages\;C:\msys64\mingw64\lib\python3.7\site-packages\PyQt5;";
+    - $env:PYTHONPATH = "$CI_PROJECT_DIR\build\install-x64\lib\python3.7\site-packages\;C:\msys64\usr\lib;C:\msys64\usr\lib\site-packages\;C:\msys64\mingw64\lib\python3.7\;C:\msys64\mingw64\lib\python3.7\site-packages\;C:\msys64\mingw64\lib\python3.7\site-packages\PyQt5;";
     - New-Item -path "build/install-x64/share/" -Name "$CI_PROJECT_NAME" -Value "CI_PROJECT_NAME:$CI_PROJECT_NAME`nCI_COMMIT_REF_NAME:$CI_COMMIT_REF_NAME`nCI_COMMIT_SHA:$CI_COMMIT_SHA`nCI_JOB_ID:$CI_JOB_ID" -ItemType file -force
     - $PREV_GIT_LABEL=(git describe --tags --abbrev=0)
     - git log "$PREV_GIT_LABEL..HEAD" --oneline --pretty=format:"%C(auto,yellow)%h%C(auto,magenta)% %C(auto,blue)%>(12,trunc)%ad %C(auto,green)%<(25,trunc)%aN%C(auto,reset)%s%C(auto,red)% gD% D" --date=short > "build/install-x64/share/$CI_PROJECT_NAME.log"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -115,6 +115,7 @@ windows-builder-x86:
     - $PREV_GIT_LABEL=(git describe --tags --abbrev=0)
     - git log "$PREV_GIT_LABEL..HEAD" --oneline --pretty=format:"%C(auto,yellow)%h%C(auto,magenta)% %C(auto,blue)%>(12,trunc)%ad %C(auto,green)%<(25,trunc)%aN%C(auto,reset)%s%C(auto,red)% gD% D" --date=short > "build/install-x86/share/$CI_PROJECT_NAME.log"
     - python3 freeze.py build
+    - editbin /LARGEADDRESSAWARE "$CI_PROJECT_DIR\build\exe.mingw-3.7\launch.exe"
     - python3 installer/build-server.py "$SLACK_TOKEN" "$S3_ACCESS_KEY" "$S3_SECRET_KEY" "$WINDOWS_KEY" "$WINDOWS_PASSWORD" "$GITHUB_USER" "$GITHUB_PASS" "True" "$CI_COMMIT_REF_NAME"
   when: always
   except:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -109,7 +109,8 @@ windows-builder-x86:
     - if (-not (Test-Path "artifacts.zip")) { Invoke-WebRequest -Uri "http://gitlab.openshot.org/OpenShot/libopenshot/-/jobs/artifacts/develop/download?job=windows-builder-x86" -Headers @{"PRIVATE-TOKEN"="$ACCESS_TOKEN"} -OutFile "artifacts.zip" }
     - Expand-Archive -Path artifacts.zip -DestinationPath .
     - Copy-Item "$CI_PROJECT_DIR/build/install-x86/python/*" -Destination "$CI_PROJECT_DIR"
-    - $env:Path = "$CI_PROJECT_DIR\build\install-x86\lib;C:\msys32\mingw32\bin;C:\msys32\mingw32\lib;C:\msys32\usr\lib\cmake\UnitTest++;C:\msys32\home\jonathan\depot_tools;C:\msys32\usr;C:\msys32\usr\lib;" + $env:Path;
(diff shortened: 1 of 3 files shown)


ferdnyc:
Sweet! Awesome that it's finally come together.

Speaking of the libopenshot Windows build instructions: is the full DirectX SDK really required? Because, that download is like half a gigabyte! 

What's actually needed out of it? Maybe there's a smaller redistributable?

jonoomph:
@ferdnyc Not sure about the DirectX SDK... I don't think we need it anymore actually. I don't recall installing it.

jonoomph:
Also, I'm still struggling with the new AWS Windows builder. I have a completely separate 32-bit and 64-bit build environment on the builder. Each one successfully freezes and builds OpenShot. However, the 32-bit version has an error importing the _openshot.pyd file. I'm still troubleshooting this, but it's really kicking my butt. 

It's like a 64-bit DLL has slipped into 32-bit build chain, however, I've done everything I can think of to verify that all DLLs are infact 32-bit, and the correct path. So, I'm still digging, but no success so far on the 32-bit Windows builds.

Oh yeah, I should mention that 32-bit version works on Windows 10... but not older versions of Windows.

jonoomph:
I'm actually configuring an older 32-bit AWS Windows 2008 base, to test this again, without any traces of 64-bit. Ughh.

ferdnyc:
> Each one successfully freezes and builds OpenShot. However, the 32-bit version has an error importing the _openshot.pyd file.

Hmm. Is the 32-bit `_openshot.pyd` compiled against libraries from a 32-bit build of Python?

ferdnyc:
> @ferdnyc Not sure about the DirectX SDK... I don't think we need it anymore actually. I don't recall installing it.

Yeah, I don't remember anything in the `CMakeLists.txt` that would pick it up, either. But I'll double-check.

jonoomph:
Yeah, everything is linked against 32-bit DLLs from MSYS2 (x86 version). But I'm getting:
Screenshot from 2019-07-25 15-05-30

At first I was super excited to get this error. A reproducible version of what many users have reported! However, now I'm not too excited, since it's **incredibly** difficult to debug.

jonoomph:
So, it's failing on our own frozen _openshot.pyd module. And the error "The specified procedure could not be found" seems to indicate perhaps a missing symbol (best guess). And here is another hint: I can get this error on the same Windows server that generated the build. I'm running `launch.exe` from the frozen build directory.

ferdnyc:
Yeah, that's _exactly_ the error Windows 10 32-bit users were getting (#2748) with recent Daily Builds. (And possibly with 2.4.4 Release 32-bit as well.) We never did solve it, AFAIK.

jonoomph:
But, keep in mind this same frozen version works fine on Windows 10. So, perhaps there is some incompatibility with MSYS2 (x86) or one of it's dependencies, which expects a newer version of some DLL. I'm not sure what version of Windows they are building those packages on.

jonoomph:
I'm configured both a windows server 2019 and a windows server 2008 rc2, identically as possible. And both servers generated the same issue with our 32-bit frozen version.

ferdnyc:
Hmmm. Yeah, if you're building with the 32-bit version of MSYS2 itself (as opposed to the MinGW32 "side" of MSYS64) then it's hard to see how a 64-bit DLL could sneak in.

jonoomph:
Correct, I installed the x86 version of MSYS2, and then installed all 32-bit dependencies on that version of MSYS2 (i.e. `C:\msys32'). And none of those paths exist on PATH (those are only set with the gitlab-ci.yml file), so not to accidentally cross the PATHs.

jonoomph:
In my head, I'm imagining a failure when loading _openshot.pyd... and when loading one of its dependencies. Since it only breaks on our module though, it kind of makes me wonder if we're linking against something wrong, on libopenshot-audio, libopenshot, or _openshot.pyd. I just wish I had a technique for bubbling up exactly where the fault is happening.

ferdnyc:
I wonder if the "specified procedure could not be found" error could point to an issue with one of the template functions, perhaps the 32-bit build doesn't generate functions for a particular 64-bit typename.

jonoomph:
`cd C:/Users/Administrator/builds/7546b651/1/OpenShot/openshot-qt/build/exe.mingw-3.7/lib`

Output of `python3 -v`:
```
>>> import PyQt5
# code object from 'C:/Users/Administrator/builds/7546b651/1/OpenShot/openshot-qt/build/exe.mingw-3.7/lib\\PyQt5\\__init__.pyc'

>>> import _openshot
  File "<stdin>", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 670, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 583, in module_from_spec
  File "<frozen importlib._bootstrap_external>", line 1043, in create_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
ImportError: DLL load failed: The specified procedure could not be found.
```

ferdnyc:
Ooh, you could be getting bytecode mismatches! What if you delete all of the `*.pyc` files? (Like, everywhere, basically.)

And maybe even run `python3 -B` to prevent it writing any new ones.

ferdnyc:
(I thought there was a flag for "_ignore_ any .pyc files", but damned if I can find it now.)

jonoomph:
I just realized something... when digging into libopenshot more, the x86 versions of `openshot-example` and `openshot-player` both crash. So, perhaps I'm narrowing in on the real problem. Something wrong how we are compiling libopenshot?

jonoomph:
`openshot-audio-test-sound` works fine... so I don't think it's related to libopenshot-audio

ferdnyc:
Hmmm. Can you pop one of those into Dependency Walker, see what DLLs it's loading? (Or, trying to.)

jonoomph:
x86 libopenshot GitLab build output:
[win-builder-gitlab-output.txt](https://github.com/OpenShot/openshot-qt/files/3433232/win-builder-gitlab-output.txt)

Maybe I need to enable verbose logging for a build, and see all the compile and link commands? That might let us verify things are "as we expect".

jonoomph:
:confused: 

 Screenshot from 2019-07-25 15-54-29

ferdnyc:
I'd enable not only verbose logging, but `-Wall -Wextra -Wpedantic` too, yeah.

jonoomph:
And just to clarify the problem, this works on Windows 10 (32-bit and 64-bit), but fails on older 32-bit versions of Windows (such as Windows 7, Windows Server 2008 RC2, etc...)

ferdnyc:
THAT makes me think it's using some API that didn't exist in earlier Windows.

Another thought: Could it be picking up the wrong imagehlp / dbghelp? (I'm just scouring the CMakeListseses for things we only link in on Windows, that might explain this...)

ferdnyc:
> x86 libopenshot GitLab build output:
> [win-builder-gitlab-output.txt](https://github.com/OpenShot/openshot-qt/files/3433232/win-builder-gitlab-output.txt)

Yeah, there's basically nothing there without `VERBOSE=1`.

jonoomph:
I kind of feel like I'm missing something obvious. But it now appears that libopenshot-audio is not working in 32-bit more on the build server. I have no sound devices on this computer, so this gets tricky to verify things. However, no error appears in the terminal when testing `openshot-audio-test-sound`... but when I launch the command with Windows, it gives me the same error above (`entry point timeBeginPeriod`).

When I profiled `openshot-player.exe` with procmon, up to the point of the error message, the last thing logged is a bunch of lines related to `libopenshot-audio`. So, I'm starting to think perhaps this is related to JUCE, or how we are compiling it... perhaps.  Winmm.dll is the Windows DLL which should contain that symbol... I think.

jonoomph:
Okay, check it out. DependencyWalker says that we are missing those symbols from `c:\windows\syswow64\KERNEL32.DLL`, but those symbols should be from `c:\windows\syswow64\WINMM.DLL`??? I see them listed under WINMM.DLL also. So why is it looking in KERNEL32.DLL for those symbols? Is this is a linking issue?

Screenshot from 2019-07-25 16-31-17

jonoomph:
Here are 3 those symbols in WINMM.dll:
Screenshot from 2019-07-25 16-34-08

jonoomph:
When I inspect the working 64-bit version, it correctly finds the 3 missing entry points in WINMM.dll.

jonoomph:
I noticed that libopenshot-audio x86 version is not being compiled with `-D"CMAKE_CXX_FLAGS=-m32" -D"CMAKE_C_FLAGS=-m32" ` :astonished:

jonoomph:
I've enabled verbose logging and adding the machine flag (-m32) to the libopenshot-audio x86 Windows build: https://github.com/OpenShot/libopenshot-audio/pull/64

jonoomph:
I'm gonna go ahead and merge this in. We'll fix the 32-bit issues in another PR. :+1:

ferdnyc:
D'oh! I even thought to go and double-check that (`-m32`), but only for libopenshot itself, since that's where we were focusing. Glad you tracked it down (hopefully)!

jonoomph:
@ferdnyc This looks promising: https://github.com/msys2/MINGW-packages/issues/4984

jonoomph:
Perhaps the order of -link args is causing kernel32.dll to link instead of winmm.dll

jonoomph:
For example: (thanks to our verbose mode)
```
g++.exe  -m32 -O3 -DNDEBUG  -shared -o libopenshot-audio.dll -Wl,--out-implib,libopenshot-audio.dll.a -Wl,--major-image-version,0,--minor-image-version,1 -Wl,--whole-archive CMakeFiles/openshot-audio.dir/objects.a -Wl,--no-whole-archive /C/msys32/mingw32/lib/libz.dll.a -ladvapi32 -lcomdlg32 -lgdi32 -lGlU32 -lImm32 -lkernel32 -lole32 -lOpenGL32 -lrpcrt4 -lshell32 -lShlwapi -luser32 -lvfw32 -lversion -lwinmm -lwininet -lws2_32 -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 
```
lkernel32 comes before lwinmm

Now... how do we fix this though?

jonoomph:
Fixing this: https://github.com/OpenShot/libopenshot-audio/pull/65

ferdnyc:
That seems like it should do it!

PR #2886, opened 2019-07-22
Adding Missing Images to Cache (improve initial speed of Title dialog)

jonoomph:
When any thumbnail is loaded in the UI, we first check the cache folder. We include a pre-cached image of all assets, in order to speed up initial launch. However, we were missing many Title-related thumbnails, which made the initial load of Titles very slow.


ferdnyc:
If performance is a concern, would these be good candidates for inclusion
in the resource file, even?

SuslikV:
@ferdnyc I never heard that resource storage has any advantages over the regular files for Qt. It just ensures that source always available.

jonoomph:
@SuslikV Same here. Primarily, the issue here is libresvg is very slow rendering SVGs (compared to our previous solution). So, when we have 47 SVGs to render before the UI can display, it is a noticeable lag. By adding a pre-thumbnailed version into the cache folder, this becomes a non-issue, and the UI immediately opens up. 

In fairness, it would also be super fast to use the resource file. But for now, this is easier for me.

Issue #2878, opened 2019-07-19
Freeze View doesn't restores its state on application startup


SuslikV:
**Describe the bug:**
Main menu _View>Views>Freeze View_ remains unchecked when application restarted.

**Steps to reproduce the behavior:**
1. Go to main menu _View>Views_
2. Click _Freeze View_ option. The option changes to _Un-Freeze View_ (corresponding dock widgets not movable now)
3. Restart OpenShot
4. Notice that _View>Views>Freeze View_ was reset by itself and dock widgets are movable again

**Expected behavior:**
_View>Views>Freeze View_ should save its state on change (or exit) and restore saved state on application startup.

**System Details:**
- OpenShot Version 2.4.4
- Win7 x64

PR #2888, opened 2019-07-25
 Restore UI state on startup

SuslikV:
Fixes https://github.com/OpenShot/openshot-qt/issues/2878

Restore frozen state of UI (if was set by user) together with the docks geometry.

### DIFF
--- a/src/settings/_default.settings
+++ b/src/settings/_default.settings
@@ -85,6 +85,13 @@
     "setting": "window_geometry_v2"
   },
   {
+    "value": false,
+    "title": "",
+    "type": "hidden",
+    "category": "Qt",
+    "setting": "docks_frozen"
+  },
+  {
     "value": "",
     "title": "",
     "type": "hidden",
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -95,6 +95,9 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
     OpenProjectSignal = pyqtSignal(str)
     ThumbnailUpdated = pyqtSignal(str)
 
+    # Docks are closable, movable and floatable
+    docks_frozen = False
+
     # Save window settings on close
     def closeEvent(self, event):
 
@@ -1934,12 +1937,14 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         self.freezeDocks()
         self.actionFreeze_View.setVisible(False)
         self.actionUn_Freeze_View.setVisible(True)
+        self.docks_frozen = True
 
     def actionUn_Freeze_View_trigger(self, event):
         """ Un-Freeze all dockable widgets on the main screen """
         self.unFreezeDocks()
         self.actionFreeze_View.setVisible(True)
         self.actionUn_Freeze_View.setVisible(False)
+        self.docks_frozen = False
 
     def actionShow_All_trigger(self, event):
         """ Show all dockable widgets """
@@ -2061,14 +2066,21 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         # Save window state and geometry (saves toolbar and dock locations)
         s.set('window_state_v2', qt_types.bytes_to_str(self.saveState()))
         s.set('window_geometry_v2', qt_types.bytes_to_str(self.saveGeometry()))
+        s.set('docks_frozen', self.docks_frozen)
 
     # Get window settings from setting store
     def load_settings(self):
         s = settings.get_settings()
 
-        # Window state and geometry (also toolbar and dock locations)
+        # Window state and geometry (also toolbar, dock locations and frozen UI state)
         if s.get('window_state_v2'): self.restoreState(qt_types.str_to_bytes(s.get('window_state_v2')))
         if s.get('window_geometry_v2'): self.restoreGeometry(qt_types.str_to_bytes(s.get('window_geometry_v2')))
+        if s.get('docks_frozen'):
+            """ Freeze all dockable widgets on the main screen """
+            self.freezeDocks()
+            self.actionFreeze_View.setVisible(False)
+            self.actionUn_Freeze_View.setVisible(True)
+            self.docks_frozen = True
 
         # Load Recent Projects
         self.load_recent_menu()


src/windows/main_window.py
```diff
@@ -2058,17 +2058,27 @@ def removeSelection(self, item_id, item_type):
...
 
-        # Save window state and geometry (saves toolbar and dock locations)
+        # Save window state and geometry (saves toolbar, dock locations and frozen UI state)
         s.set('window_state_v2', qt_types.bytes_to_str(self.saveState()))
         s.set('window_geometry_v2', qt_types.bytes_to_str(self.saveGeometry()))
+        if self.actionUn_Freeze_View.isVisible():
+            docks_frozen = 'True'
+        else:
+            docks_frozen = 'False'
+        s.set('docks_frozen', docks_frozen)
```
ferdnyc: I'm not crazy about using the menu-item visibility to control the setting variable — why not just update it during `freezeDocks()` and `unfreezeDocks()`? e.g:
```python
    def freezeDocks(self):
        """ Freeze all dockable widgets on the main screen (no float, moving, or closing) """
        s = settings.get_settings()
        s.set('docks_frozen', True)
        for dock in self.getDocks():
            dock.setFeatures(QDockWidget.NoDockWidgetFeatures)
```
and similar for `unFreezeDocks()`. Then nothing special needs to be done for `docks_frozen` during `save_settings()`.

src/settings/_default.settings
```diff
@@ -84,6 +84,13 @@
     "category": "Qt",
     "setting": "window_geometry_v2"
   },
+  {
+    "value": "False",
```
ferdnyc: The value should just be `false`, lowercase with no quotes. That'll store it as a boolean value instead of a string. Otherwise, setting it to `True` (without quotes) later will change the type.

src/windows/main_window.py
```diff
@@ -2058,17 +2058,27 @@ def removeSelection(self, item_id, item_type):
...
 
     # Get window settings from setting store
     def load_settings(self):
         s = settings.get_settings()
 
-        # Window state and geometry (also toolbar and dock locations)
+        # Window state and geometry (also toolbar, dock locations and frozen UI state)
         if s.get('window_state_v2'): self.restoreState(qt_types.str_to_bytes(s.get('window_state_v2')))
         if s.get('window_geometry_v2'): self.restoreGeometry(qt_types.str_to_bytes(s.get('window_geometry_v2')))
+        if s.get('docks_frozen') == 'True':
```
ferdnyc: By the same token, this should just be `if s.get('docks_frozen'):` since it'll be stored as a boolean true/false value and you can just test it directly. (If the setting is missing, that'll still imply false, which is what we'd want.)

src/windows/main_window.py
ferdnyc: (Alternately, the settings update could be in the triggers, instead. There doesn't seem to be any strict pattern in the code for where that's done, and it shouldn't make a difference one way or the other.)

src/windows/main_window.py
SuslikV: Storing non-string values in Settings is usually bad practice. The conversion was made intentionally.

ferdnyc: JSON supports boolean values natively, and the OpenShot convention is to use them:
https://github.com/OpenShot/openshot-qt/blob/e48cc2be1865110a0a8491ff4a5f887994650aa8/src/settings/_default.settings#L464-L470

src/windows/main_window.py
SuslikV: It's not much better. You setting value on startup, me - on close. I have one `if`, you have to read all keys twice instead (on change and on close). Remember, the change triggers on startup.

Anyway, I will add new commit to reflect this changes. Do you have better solution? Global variable or so?

ferdnyc: If you want to avoid the extra settings lookup during load, then like I said the updates can go in the triggers instead of `freezeDocks()` / `unFreezeDocks()`. Then it'll only happen when the user changes the freeze state, regardless whether the saved value is true or false.

src/windows/main_window.py
ferdnyc: > When later you decide to change the UI, for example, if you'll try to replace the boolean checkbox, with the list `0, 1, 2, 3, ... n` you'll be unable to reuse the existing setting (variable) that was saved on users PCs with the previous build of OpenShot.

True, and for all those reasons, we wouldn't want to reuse it. If a situation like that occurs (which it has, in the past), we simply version the setting. `docks_frozen_v2` can be a string, or a list of dock names, or whatever we choose, and (if it makes sense) the old value is migrated from the user's existing settings when there's no value for the new preference. That's already been done for one or two preferences that have changed semantics in the past.

Doing it that way, the user's saved settings remain compatible in both directions (for newer _and_ older versions), since we haven't changed the meaning of the older version's preference, by setting a value that it doesn't know how to interpret.

SuslikV: Really, who cares, user should reconfigure UI because it was programmed in a new way...

ferdnyc: Well... I care, for one. I often run multiple different versions of OpenShot for debugging and troubleshooting purposes. So I don't want to have to worry about my preferences file breaking an older version.

src/settings/_default.settings
jonoomph: Agree, value should be `false`, for JSON sake

src/windows/main_window.py
jonoomph: Yeah, if we are correctly storing the boolean value, then `if s.get('docks_frozen')` is pretty elegant. :+1: If for some reason this changes in the future to a more complex type, we would certainly change the "key" and handle the upgrade path from old setting to new setting. But this is a pretty minor issue, so let's just agree to disagree, and stick with the boolean value for now. That gets my vote, haha.

jonoomph: Otherwise, I like this! LG!

SuslikV: > ...So I don't want to have to worry about my preferences file breaking an older version.

@ferdnyc It never would, if you always reading and processing strings. This is the advantage of the method - forward and backward compatibility, unlike the different variable types.

ferdnyc: @SuslikV We seem to have reached a point where you can either win this argument, _or_ you can get your PR merged. Your choice.

SuslikV: The `QDockWidget::AllDockWidgetFeatures` or _QDockWidget.AllDockWidgetFeatures_ was deprecated, in favor of using individual flags. Thus I strongly recommend to use strings here. I'll try to rewrite the code one more time.

SuslikV:
Some suggestions were taken into account with the new commit. Is it better now? I can re-base on top of the current _develop_ and remove part of the history before the last commit if needed.

SuslikV:
Rebased on top of the current _develop_ branch.
Co-authors was added because original implementation was changed (string variable vs boolean).

jonoomph:
This looks good to me!! Nice and clean booleans. :wink: Thanks @SuslikV

Issue #2889, opened 2019-07-25
Custom Export Profile 20170x1080 resolution at 30 FPS


Jarli01:
**Question:**
I'm trying to create a custom MP4 with a resolution of 20170x1080 at 29.97 FPS, when I create a custom profile and use these details, relaunch OSVE I can see the profile but when I click export I'm prompted with "Sorry, there was an error exporting your video: Could not open codec"

**System Details:**
- OpenShot Version 2.4.4
- Operating System / Distro: Windows 10

**Screenshots:** 
The custom profile on the left and the source file details on the right.
https://i.imgur.com/QLrumEd.png

Thank you in advance

ferdnyc:
@Jarli01 
I'll be honest, I don't know that anyone's ever tried to use dimensions even _close_ to those with OpenShot. I hope you have a disturbing amount of RAM in your system!

My first question would be whether FFmpeg can handle those dimensions, just internally. It might be worth installing whatever native Windows build of FFmpeg is available and trying to play the source file with `ffplay` (their reference player), to see if it balks as well.

Also, if the pixel aspect is 1:1, then I'd think you'd need to set your profile's display aspect ratio to something that matches the video dimensions, e.g.
```
display_aspect_num=18676
display_aspect_den=1000
```

But I'd be pretty surprised if that alone was preventing it from working.

Jarli01:
So I was able to convert to WEBM and the aspect ratio came out at 9999x1080, obviously way smaller than what this person is looking for. But it does look good on my tiny monitor compared to what this wants to be put on. 

Going to do some more testing

Jarli01:
Here's the output from ffplay

	ffplay.exe HALP.mov
	ffplay version N-94387-g923d5c489f Copyright (c) 2003-2019 the FFmpeg developers
	  built with gcc 9.1.1 (GCC) 20190716
[...]

PR #2892, opened 2019-07-26
Raise max video dimensions to 40000x40000

ferdnyc:
Believe it or not, 9999 pixels to a side isn't enough for some users' projects, and the only thing preventing them from working in larger dimensions was the max value the export dialog would accept.

(Also, 9999 is a terrible value to stop at because most codecs require dimensions in multiples of 2, 4, 8... 40,000 is still just as arbitrary, but at least it's a multiple of 2, 4, _and_ 8.)

Addresses #2889

### DIFF
--- a/src/windows/ui/export.ui
+++ b/src/windows/ui/export.ui
@@ -526,7 +526,7 @@
                 <number>2</number>
                </property>
                <property name="maximum">
-                <number>9999</number>
+                <number>40000</number>
                </property>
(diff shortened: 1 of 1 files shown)


jonoomph:
Who am I to argue with this. Same logic though... is 40,000 enough? Did you test 40,000 and see if your computer catches on fire?

jonoomph:
LG though

PR #2893, opened 2019-07-26
Add SVG to image-sequence extensions

### LINKED ISSUE #675: Feature request: Import Image Sequence for SVG files
**System Details:**
* Operating System / Distro: Ubuntu 16.04.2
* OpenShot Version: 2.3.3
* _Please attach log files if crash_

Can the drag and drop import image sequence apply to SVG files as well as PNG?  I have a process to overlay GPS data onto videos I take with my action camera.  Part of the process is creating SVG with the GPS data on that and merging that sequence to the video using OpenShot.  Previously this could be done in OpenShot Version 1.4.3, but there was a bug where it could only import a certain number of SVG files at one time (http://openshotusers.com/forum/viewtopic.php?f=11&t=355).  

I was hoping this bug was fixed in the latest version, but when I drag and drop an SVG I am not prompted to import as a sequence.  For my latest video, I have over 18,000 SVG files, I would prefer not to convert these to PNG and using the old system, I have to load the files in batches.

ferdnyc:
We occasionally get requests to support SVG as an image sequence filetype, and AFAICT there's no  reason we can't. Simply adding `svg` to the list of recognized filename extensions in the image sequence code made it Just Work™, at least in my testing.

This change allows importing of image sequences that comprise groups of sequentially-numbered SVG files, in addition to the other supported image formats.

(Note that I'm not currently able to test this with ReSVG, and the caveat should be noted that QtSVG-rendered SVG image sequences will likely not produce accurate results.)

Fixes #675 and possibly others

### DIFF
--- a/src/windows/views/files_treeview.py
+++ b/src/windows/views/files_treeview.py
@@ -205,7 +205,7 @@ class FilesTreeView(QTreeView):
 
         # Get just the file name
         (dirName, fileName) = os.path.split(file_path)
-        extensions = ["png", "jpg", "jpeg", "gif", "tif"]
+        extensions = ["png", "jpg", "jpeg", "gif", "tif", "svg"]
         match = re.findall(r"(.*[^\d])?(0*)(\d+)\.(%s)" % "|".join(extensions), fileName, re.I)
 
         if not match:


jonoomph:
LG!

Issue #2752, opened 2019-04-23
Make log less verbose


SuslikV:
**New feature:**
I want to see new feature as commandline key or option in settings to allow user to log only critical errors not INFO level. Really, about 25MB logs... For what purpose this? Just to catch the application's bugs? It is not the way how application should work. Also, the log is on the SSD drive, which lifetime is short, and application just writes a lot of unneeded info to this log.

**The solution I'd like to see:**
Add either new commandline key "verbose_level=0..3" or add similar option in the settings of the application.

**Alternatives:**
Log only ERRORS by default, and add special option to enable logging of INFO, DEBUG etc. level of the application messages, but only if asked for.
OR
Allow, to fail silently on write to log operations, thus I can restrict access of the application to the log file and it isn't harm anything for me.

FinixFighter:
I agree with you!

PR #2895, opened 2019-07-27
WIP: Make OpenShot less verbose

SuslikV:
Attempt to implement the: https://github.com/OpenShot/openshot-qt/issues/2752

I believe that end-user will able set _Debug_ checkbox in the application's _Preferences_ to get old-style verbose log back...

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -125,6 +125,9 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
                 event.ignore()
                 return
 
+        # On exit always write INFO messages to the log
+        log.setLevel(logging.INFO)
+
         # Log the exit routine
         log.info('---------------- Shutting down -----------------')
 
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -32,6 +32,7 @@ import sys
 import platform
 import shutil
 import webbrowser
+import logging
 from operator import itemgetter
 from uuid import uuid4
 from copy import deepcopy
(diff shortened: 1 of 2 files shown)


jonoomph:
@SuslikV Perhaps we should add more options around debug mode. For example, allowing the user to choose different log levels for the UI, and be able to toggle the libopenshot debug mode separately. My only concern about this PR is the fact that "debug" mode in libopenshot is crazy, crazy verbose... since it outputs many things per Frame... and all normal log statements from the UI are pretty much lost.

SuslikV:
What you consider normal to have in the log for regular edit task?

From the logs posted by users in the last months - only once it was helpful to resolve the issue (during import of mp3 with image bug). But in all other cases - it is easier to ask user to write all steps that needed to reproduce the issue on developer's end. I personally think that application is mature enough to remove excessive info from the log for now. It can take years to implement perfect solution.

SuslikV:

Also, it is still unknown in what direction this changes should move.
I don't think that "Showing dialog blah-blah-blah", "Position is N", "Updating key to value blah-blah-blah" are  useful in the logs in normal (not debug) mode. The operations as: load new sources, export settings (not menus!), errors, brief environment or system info - should be logged, sure. But logging user action in UI?! I'm strongly against this.

If some user action changes important part of the program the trace in the log may be added (for example, Preferences window running test for HW compatibility silently - this is no good).

jonoomph:
Closing PR due to conflicts and lack of updates.

PR #2897, opened 2019-07-28
Add Qt5Core and Qt5Svg DLLs when missing from frozen folder

jonoomph:
Add Qt5Core and Qt5Svg DLLs when missing from frozen folder (this fixes missing icons in daily builds). Not sure why cx_Freeze doesn't include these already... but it's probably related to using Python3.7, which is isn't perfectly supported yet.

### DIFF
--- a/installer/build-server.py
+++ b/installer/build-server.py
@@ -433,12 +433,16 @@ try:
             else:
                 shutil.copytree(os.path.join('C:\\msys64\\mingw64\\share\\qt5\\plugins', replace_name), os.path.join(exe_dir, replace_name))
 
-        # Copy Qt5Core.dll to root of frozen directory
-        qt5_core_path = os.path.join(exe_dir, "lib", "Qt5Core.dll")
-        new_qt5_core_path = os.path.join(exe_dir, "Qt5Core.dll")
-        if os.path.exists(qt5_core_path) and not os.path.exists(new_qt5_core_path):
-            output("Copying %s to %s" % (qt5_core_path, new_qt5_core_path))
-            shutil.copy(qt5_core_path, new_qt5_core_path)
+        # Copy Qt5Core.dll, Qt5Svg.dll to root of frozen directory
+        paths_to_copy = [("Qt5Core.dll", "C:\\msys64\\mingw64\\bin\\"), ("Qt5Svg.dll", "C:\\msys64\\mingw64\\bin\\")]
+        if windows_32bit:
+            paths_to_copy = [("Qt5Core.dll", "C:\\msys32\\mingw32\\bin\\"), ("Qt5Svg.dll", "C:\\msys32\\mingw32\\bin\\")]
+        for qt_file_name, qt_parent_path in paths_to_copy:
+            qt5_path = os.path.join(qt_parent_path, qt_file_name)
+            new_qt5_path = os.path.join(exe_dir, qt_file_name)
+            if os.path.exists(qt5_path) and not os.path.exists(new_qt5_path):
+                output("Copying %s to %s" % (qt5_path, new_qt5_path))
+                shutil.copy(qt5_path, new_qt5_path)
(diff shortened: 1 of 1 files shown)


ferdnyc:
@jonoomph 

> Not sure why cx_Freeze doesn't include these already...

I subsequently discovered that it _is_ including them, along with all of the other Qt DLLs — they're just ALL going into the `lib\PyQt5\` directory now, for some reason.

ferdnyc:
(Actually, I'm not entirely sure why there even is a `lib\PyQt5\` directory, since lib dirs are very "un-Windows"...ey. Normally any DLLs would be placed right alongside the executables that require them.)

jonoomph:
I agree. If we can determine the ones that are unneeded, we can drop them in another PR. This seems to fix the missing icon issue, and + your prune changes, I can no longer break this version of OpenShot on any of my Windows computers! Nice!

ferdnyc:
I would assume the failure to place the Qt DLLs in the root folder is because it knows it's already included them, and automatically avoids duplicate files when it's building the frozen tree. Which is a good thing, for the most part. We just need to convince it that they don't belong in the PyQt5 module directory... though I remember reading that cx_Freeze had _specific_ support for packaging PyQt5 applications, so I'd think it would get this stuff right.

Oooh, unless that PyQt5 support is what's _placing_ the Qt DLLs in the PyQt5 directory, because normally that works if you're packaging a PyQt5 application exclusively. Maybe it's libopenshot's own dependency on Qt, separate from and in addition to PyQt5's, that's fouling the works here? cx_Freeze might not account for **that**.

ferdnyc:
**BINGO!**

From the [release notes](https://cx-freeze.readthedocs.io/en/latest/releasenotes.html) for cx_Freeze 5.1.1:
> Copy dependent files to the same directory as the file it depends on, not the root directory; also add a sample for PyQt5 to demonstrate its correct use (Issue 328).

ferdnyc:
And it appears that's causing similar problems with other uses of cx_Freeze (anthony-tuininga/cx_Freeze#372):

> when using the option 'include_msvcr': True , VCRUNTIME140.dll gets copied to the lib folder instead of its parent folder, i.e. the one that will contain the produced .exe:
> 
> [screenshot]
> 
> causing the following error:
> 
> [screenshot]
> 
> Moving the VCRUNTIME140.dll from /lib to its parent directory fixes it.
> 
>

jonoomph:
Yeah, we have `msvcrt.dll` in our root. We no longer have a dependency on `VCRUNTIME*.DLL`

jonoomph:
I'm really curious to hear feedback from people who have never been successful running OpenShot on Windows. I hope we've solved more of the general issues.

ferdnyc:
We did actually get one such report already, from @Zipher04 in #2748 ([comment](https://github.com/OpenShot/openshot-qt/issues/2748#issuecomment-515703246)), who saw the Windows network-access challenge (firewall permissions) dialog for the first time yesterday! That made me laugh<sup>1</sup>... but was also sort of sad. In a good way, though. :laughing: 

#### Notes
1. (Mostly because it just sounds like such a _weird_ "first encounter with OpenShot" tale... at least until you realize that, actually, it's _every_ Windows user's first encounter with OpenShot.)

jonoomph:
Yeah, that's good news! haha

jonoomph:
It is probably possible, using InnoSetup to write a registry key to allow us firewall permission.. FYI, if anyone is interested into figuring that out. Since the installer is running with Admin permission, it should be able to write those keys, etc...

jonoomph:
Perhaps, something like this: http://www.gasov.com/content/view/124/45/

jonoomph:
@ferdnyc I'm gonna take a quick shot at a Firewall exception in InnoSetup... I'll have a PR shortly

PR #2898, opened 2019-07-28
New windows firewall rule during install

jonoomph:


jonoomph:
This seems to work fine during my tests. You must delete your previous rules before testing though... and if it works, you should get no "firewall" popup when installing/launching this version of OpenShot in Windows. I believe this syntax works for Windows 7 and newer versions of Windows.

My only question is, should the code be moved to some "Post Install" hook vs where I put it. @ferdnyc Any thoughts on that?

ferdnyc:
Hmmm. See my PR #2872, where (a slight ways down in the comments) I eventually included a commit to add `.osp` file associations during the install.

Even though the firewall rule's not _technically_ an "Additional shortcut" (neither is the association), I think it might be better to stick it in that section as well, as it's still an optional "thing" created on the local system.

("Additional shortcuts" is a standard InnoSetup string with translations into every available language, so despite it not being a great fit it's a safe choice for the group label.)

ferdnyc:
(And it'd be great to get #2872 merged as well, because I was really surprised that we're _not_ associating `.osp` files with OpenShot during the install! ...I rezzed up the icons a bunch, too, for make more prettier.)

ferdnyc:
The installer page in question (the one with the "Additional shortcuts" group) has the overall heading "Additional tasks", which is _certainly_ a good description for the firewall-rule insertion. A separate group other than "Additional shortcuts" can certainly be created for that particular task, on the "Additional tasks" page.

ferdnyc:
(The specific commit from that PR which adds file-association creation to the InnoSetup configs: 3ed1cc9f6fcb9774934469638b3904a035288f14)

ferdnyc:
OK, from a little InnoSetup doc reading, the _command_ to do the firewall rule installation itself **HAS** to be in the `[Run]` section, as that's the only place where you can launch external programs from the installer.

What _could_ be done, though, is to add an entry in `[Tasks]` making the rule insertion _optional_, so the user has some control over whether that's done. Appending `; Tasks: (taskname)` to the line in the `[Run]` section would run the command only if that task is enabled, controlled by a checkbox on the Additional Tasks page. (The task can be default-enabled, as the one for creating the file association would be. So, the user would have to explicitly uncheck it if they wanted to opt out of the firewall rule creation.)

One other thing I'd suggest, separate from the Task question, is adding a corresponding `[UninstallRun]` section with a command to _delete_ the firewall rule. Always polite to clean up after ourselves. (If you do create a Task for the `[Run]` line and associate the `[UninstallRun]` line with the same task, the uninstaller will only run that command if the task was enabled on install.)

jonoomph:
Thanks @ferdnyc! I've added an optional Task for this, and checked it by default. I've done some testing, and this seems to work nicely. Uninstall hook to remove the rule works nice as well!

ferdnyc:
Neato, I'll be sure to grab the next daily build on my Win7 VM, so I can poke the tires.

ferdnyc:
Well! I certainly could've saved us _both_ some time if I'd just read that link you posted in [this comment](https://github.com/OpenShot/openshot-qt/pull/2897#issuecomment-515802538) over at #2897 **_before_** commenting here, since literally Every. Single. Thing. I suggested was already covered on that page. 😖

ferdnyc:
@jonoomph Wooo! `\o/`

image

**WOOOOOOOOOO!**

image

PR #2902, opened 2019-07-30
Don't use deprecated "AllDockWidgetFeatures", don't make Timeline closable

ferdnyc:
As @SuslikV pointed out, `QDockWidget::AllDockWidgetFeatures` is [deprecated in Qt](https://doc.qt.io/qt-5/qdockwidget.html). Their recommendation is to use the individual features it expands to (Closable, Floatable, Movable), a recommendation this PR implements.

It also corrects a bug where the Timeline widget would become closable on unfreeze.


jonoomph:
Awesome! Thanks

PR #2903, opened 2019-07-30
Change "CPU" hardware-acceleration icon to light orange

ferdnyc:
The medium blue/gray didn't stand out well against the interface.

New color:

image


jonoomph:
Nice, no complaints! Looks much better.

PR #2904, opened 2019-07-30
Fixing cx_Freeze issues with renamed EXE (openshot-qt.exe)

jonoomph:
Fixing regression from https://github.com/OpenShot/openshot-qt/pull/2796/, where an ".exe" needs to be appended in Windows, and the `copyright` attribute is invalid.


jonoomph:
Looks like the AppImage breaks with these changes. The app image executes this file (inside itself):
`/usr/bin/openshot-qt`

That file has the following contents:
```
# Launch application
HERE="$(dirname "$(readlink -f "${0}")")"
exec "${HERE}"/openshot-qt "$@"
```

Which I believe is just calling the same file (circular loop... hangs forever)

jonoomph:
So, looks like we have a conflict on the naming of a file somewhere... I'm gonna dig around.

jonoomph:
Okay, got everything fixed I think.

- AppImage Fixed and Verified
- Win32/64 versions fixed and verified
- Mac version verified (although I noticed an error with signing the DMG file... probably unrelated)

PR #2906, opened 2019-07-31
Make Preferences window taller & wider

ferdnyc:
#### Increase width by `100px` to `800px` (for "Playback Audio Device")

Turns out Linux isn't the only system with long audio device names:

image

#### Increase height by `100px` to `450px` (for Performance tab)

All the new options don't quite fit in the window anymore.


jonoomph:
Noice!! LGTM

PR #2908, opened 2019-07-31
Properties: Enable clear button on filter input  field

ferdnyc:
I noticed that the filter text input area for the Properties dock didn't have a clear button, like the other ones do. 

So I went into Qt Creator, enabled the default one for the QLineEdit widget, and to my surprise it JFW™ without my having to do anything else or write a single line of code.

PropertiesFilterClear_optim

In fact, I think it works _better_ than our existing clear buttons (which are outside the input field and always visible, taking up space). Now I wish they all worked like this.

But, for now, at least Properties' filter has a clear button.

### DIFF
--- a/src/windows/ui/main-window.ui
+++ b/src/windows/ui/main-window.ui
@@ -376,6 +376,9 @@
        <property name="placeholderText">
         <string>Filter</string>
        </property>
+       <property name="clearButtonEnabled">
+        <bool>true</bool>
+       </property>
       </widget>
      </item>
     </layout>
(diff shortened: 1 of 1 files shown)


jonoomph:
LGTM! This is really cool! :100: I vote for updating all the "clear" buttons to this same style!

ferdnyc:
Aack! Darn, too slow. I was just finishing another commit to fix the rest of the filter fields. :laughing: 

I'll just submit another PR. :grin:

PR #2909, opened 2019-07-31
Remove the rest of the external filter-field clear buttons

ferdnyc:
We can just use the Qt magic built-in ones, with less code.

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -2152,9 +2152,8 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         self.filesFilter = QLineEdit()
         self.filesFilter.setObjectName("filesFilter")
         self.filesFilter.setPlaceholderText(_("Filter"))
+        self.filesFilter.setClearButtonEnabled(True)
         self.filesToolbar.addWidget(self.filesFilter)
-        self.actionFilesClear.setEnabled(False)
-        self.filesToolbar.addAction(self.actionFilesClear)
         self.tabFiles.layout().addWidget(self.filesToolbar)
 
         # Add transitions toolbar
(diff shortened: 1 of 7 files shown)


ferdnyc:
(The clear button property for QLineEdit was introduced in Qt 5.2, according to the docs, so we have just left the vast Qt 5.1 userbase out in the cold with these changes. :angel: )

jonoomph:
I love it! Nice work!

DylanC:
@ferdnyc - Might as well. Qt is very deep into the 5 series and v6 will be on the way next year.

ferdnyc:
@DylanC 

> Might as well.

Oh, that was very much tongue-in-cheek. If there **are** any Qt 5.1 users left out there (and I can't imagine that there actually _are_, at least not ones who are also looking to build and run OpenShot on their Qt 5.1 dinosaurs), then we're doing them a _favor_ by no longer enabling them!

> Qt is very deep into the 5 series and v6 will be on the way next year.

That could be a difficult time for OpenShot, actually, and one worth planning for in advance. I can almost guarantee there won't be any hope of a Qt6WebKit, even Qt6WebEngine might not be in the cards...

DylanC:
> That could be a difficult time for OpenShot, actually, and one worth planning for in advance. I can almost guarantee there won't be any hope of a Qt6WebKit, even Qt6WebEngine might not be in the cards...

Indeed. The team would have until November 2020 most likely but it could be a huge project to rewrite the timeline to something else. Maybe Qml and PyQt would be the way forward. Hopefully by then I can make time to be available for some dev work on this new timeline but its hard to tell atm.

ferdnyc:
@DylanC 

One happy surprise from the coming [Qt 6 apocalypse](https://blog.qt.io/blog/2019/08/07/technical-vision-qt-6/): They're dumping QMake for CMake. (Actually they're _switching to_ CMake as the build system for Qt itself, and halting development of QMake. It'll still be supported, if anyone's really attached to it for some reason.) So, if nothing else building libopenshot's Qt bits should get easier.

PR #2913, opened 2019-08-02
Fix bugs related to "New Project"

jonoomph:
Fix bugs related to "New Project" and a missing user project default file (this is a regression). @ferdnyc When the ~/.openshot_qt/default.project file was missing, clicking "New Project" would not correctly clear and reload the default project data... leaving a half broken project loaded. This simplifies and fixes it.

### DIFF
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -49,10 +49,6 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
         self.data_type = "project data"  # Used in error messages
         self.default_project_filepath = os.path.join(info.PATH, 'settings', '_default.project')
 
-        # Always set, in case the user creates the file mid-session
-        if info.USER_DEFAULT_PROJECT:
-            self.user_project_filepath = info.USER_DEFAULT_PROJECT
-
         # Set default filepath to user's home folder
         self.current_filepath = None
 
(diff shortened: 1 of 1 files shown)


src/classes/project_data.py
```diff
@@ -49,10 +49,6 @@ def __init__(self):
         self.data_type = "project data"  # Used in error messages
         self.default_project_filepath = os.path.join(info.PATH, 'settings', '_default.project')
 
-        # Always set, in case the user creates the file mid-session
-        if info.USER_DEFAULT_PROJECT:
```
jonoomph: This would always be true

src/classes/project_data.py
```diff
@@ -257,20 +253,17 @@ def new(self):
...
-                    new_project_file, ex))
+                log.warning("Unable to load user project defaults from {}: {}".format(info.USER_DEFAULT_PROJECT, ex))
             except Exception:
                 raise
             else:
-                log.info("Loaded user project defaults from {}".format(
-                    self.user_project_filepath))
-
-        # Fall back to OpenShot defaults, if user defaults didn't load
-        if not self._data:
```
jonoomph: if not self._data would always be true for an existing project loaded... and thus, it would never clear your existing project when clicking "New Project"

src/classes/project_data.py
ferdnyc: Yeah, originally I had the `if os.path.exists()` up here, but then I realized that was a bad idea for the reason given in that comment. But without the test, the rest of what was here became pointless, so I should've removed it completely.

src/classes/project_data.py
ferdnyc: I really cocked this one up, huh? (And you didn't even mention that the exception log message tried to output `new_project_file`, despite no such variable existing.)

I think the _idea_ with the `if not self._data:` was that I'd test to see whether `self.read_from_file()` **successfully** loaded the user's default project data, so that if it failed due to bad syntax or whatever I'd just roll with it. Which is a nice enough idea, I suppose, if only I'd remembered to set `self._data = None` at the top of the method!

PR #2914, opened 2019-08-02
Show waiting cursor when launching Preferences

jonoomph:
Show waiting cursor when launching Preferences, and normal cursor once preferences has loaded fully. This is needed due to Preferences running a bunch of hardware acceleration checks now, and thus is much slower to load.


ferdnyc:
OOC, are those checks run _every_ time, or is the window only slower the first time it's opened?

ferdnyc:
Those double Travis runs look weird, and seem redundant until you discover they happened an hour or two apart. The first was when the branch was created (well... when it was pushed), the second happened when the PR was opened. (GitHub could do a better job of presenting that info instead of glomming them together with no timestamps or anything, but meh.)

PR #2915, opened 2019-08-02
WIP: Refactoring blender animations for 2.80 support

### LINKED ISSUE #2901: Blender 2.80 doesn't work with OpenShot 2.4.4
win10 x64 :

test with last stable / last daily

blender 2.80 released today...

jonoomph:
- alpha_mode no longer exists (replaced with bpy.context.scene.render.film_transparent)
 - horizon_color no longer exists (replaced with bpy.data.worlds[0].color)
 - material_object.diffuse_color now accepts 4 values (including alpha)
 - material_object.alpha is no longer used (and deleted.... inherits from diffuse color i think)
 - material_object.use_shadeless is no longer used (disabled for now)
 - bpy.ops.object.text_add has new parameters (refactored it to work)

@ferdnyc I know you mentioned you were going to take a look at this, but hopefully you haven't started yet. :wink: 

Fixes #2901


jonoomph:
This doesn't fix some of the more complex blender animations, but fixes lots of the simple ones.

jonoomph:
Blender 2.8 changed quite a bit

ferdnyc:
Blender *2.80 (comes after 2.79) — I've updated the PR title. :grin:

ferdnyc:
I hadn't started looking at it yet, so this is great as it saves me a bunch of digging. Especially since the changes are apparently so extensive.

@rafaellinux mentioned a new renderer option as well, for GPU rendering... oh, I notice you saw that. Well, I'll still poke my head in and see what I can find out about Eevee.

ferdnyc:
Huh. Interestingly, it seems the Blender folks themselves suffer from a bit of `2.80` / `2.8` [confusion](https://www.blender.org/download/releases/2-80/):

> #### Special Thanks
> The Blender developer community is being supported by the organizational powers of Blender Foundation and its spin-off Blender Institute. The people who work for the Foundation and Institute did a tremendous job to bring Blender is where is it nowadays.
> 
> Special thanks goes to Tangent Animation and Aleph Objects, who funded 4 additional developers to work full-time on Blender 2.8 during the crucial 2017 period. This enabled us to work on the viewport, Eevee, collections/layers, UI and tools redesign.
> 
> Thanks goes to everyone who contributed to the Code Quest, the massively successful 3 month workshop in Blender Institute during spring 2018.
> 
> And we thank everyone who joined the Development Fund in 2nd half of 2018 and 2019. This helped us to keep the core of Blender contributors together to work on 2.8.
> 
> And last but not least: special thanks to the blender.org community – the developers, documenters, bug reporters and reviewers – it is thanks to them that we can start this wonderful new era of Blender 2.8x!

(There's also a `releases/2-8` redirect to `releases/2-80` on their site. Guess some members of the team didn't get the memo! :wink: )

src/windows/views/blender_listview.py
```diff
@@ -174,7 +174,10 @@ def currentChanged(self, selected, deselected):
             elif param["type"] == "color":
                 # add value to dictionary
                 color = QColor(param["default"])
-                self.params[param["name"]] = [color.redF(), color.greenF(), color.blueF()]
+                if param.get("name") == "diffuse_color":
+                    self.params[param["name"]] = [color.redF(), color.greenF(), color.blueF(), color.alphaF()]
+                else:
+                    self.params[param["name"]] = [color.redF(), color.greenF(), color.blueF()]
```
ferdnyc: Would it be easier to just set _all_ colors in RGBA format, rather than special-casing? I mean, the [QColor docs say](https://doc.qt.io/qt-5/qcolor.html#alpha-blended-drawing) that fully opaque is the default, so `alphaF()` should just be `1.0` for any solid, non-transparent colors. Couldn't hurt to just always pass it in, and it might become useful later if we decide to add alpha to any other color params.

src/blender/scripts/colors.py
```diff
@@ -123,32 +123,30 @@ def load_font(font_path):
...
 # Shadeless Background
-if params["shadeless"] == "Yes":
-	material_object4.use_shadeless = True
-else:
-	material_object4.use_shadeless = False
+# TODO: Unsupported in Blender 2.8 (not sure of workaround yet)
+#if params["shadeless"] == "Yes":
+#	material_object4.use_shadeless = True
+#else:
+#	material_object4.use_shadeless = False
```
ferdnyc: The [workarounds](https://blender.stackexchange.com/questions/131015/where-is-shadeless-material-option-for-blender-2-8?noredirect=1&lq=1) here seem a bit... complex, though maybe they'll turn out to be pretty easy to simplify. The linked answer to [Node for setting render colour, regardless of lighting](https://blender.stackexchange.com/a/8110) seems especially useful, though it's presented in terms of Blender UI screenshots so that'd have to be translated into script code. And on the plus side, the new version will work the same with or without the GPU renderer. (The old option, "use shadeless", was a feature of the previous internal renderer, so presumably it would've either not worked or even crashed when using the GPU renderer.)

src/blender/scripts/colors.py
```diff
@@ -252,8 +250,8 @@ def load_font(font_path):
 except:
 	bpy.context.scene.render.image_settings.file_format = params["file_format"]
 	bpy.context.scene.render.image_settings.color_mode = params["color_mode"]
```
ferdnyc: The `try:` part of all these blocks:

```py
bpy.context.scene.render.file_format = params["file_format"]
bpy.context.scene.render.color_mode = params["color_mode"]
```
should almost definitely just be eliminated, especially if we're moving to 2.80 as minimum version. Neither of those properties is present in the [RenderSettings API docs](https://docs.blender.org/api/blender2.8/bpy.types.RenderSettings.html#bpy.types.RenderSettings).

ferdnyc:
I pushed a commit onto the PR branch to fix the GPU injectable, since `user_preferences` was renamed.

src/windows/views/blender_listview.py
jonoomph: This was mainly for the preview image... so it has more contrast (instead of the background color of the widget... or whatever). But yes, we could probably always default to RGBA and then adjust the widget color in Qt to handle that (I think). It would need some experimenting.

src/blender/scripts/colors.py
jonoomph: Yeah, I'm a bit worried that a few of these *.blend files will need to be extensively redone, especially the node-based ones... which have various "glare" and "lighting" nodes + transparency rendering as the background. Still lots of stuff broken on them in Blender 2.80.

src/blender/scripts/colors.py
jonoomph: Agreed

ferdnyc:
@jonoomph 

I have a commit to clean up the logging in `blender_listview.py`, see 2f2a5541fc136b2bfff27ae66a9ebac806b8b00d
I can push it onto this branch if you like. Mostly it:
* Removes all the "We entered this method" logging, as simply not useful to have always enabled.
* Consolidates all of the "Updating parameter: foo" / "New value is: N" message pairs into single "Set parameter foo to N" lines
* _Comments out_ the parameter update logging for text fields, since _**every**_ keypress causes another log message, which is just pure noise
* **Adds** logging (at Error level) for the errors that are displayed via QMessageBox, since those actually _weren't_ being logged.

jonoomph:
@ferdnyc Sure, go ahead and push it to this branch. Thx

ferdnyc:
@jonoomph  Done and done.

src/blender/scripts/colors.py
ferdnyc: This one actually isn't that bad after all, wrapping my head around the new interface was the trickiest part, but once I worked that out replicating the setup of Background.Material to use an Emission shader for the camera (as described in that "Node for setting render colour..." issue above was pretty straightforward. Unfortunately, I have to start over and redo it cleanly, and can't right now because it's about to rain and our roof leaks, so I have more pressing issues to deal with. :laughing:

ferdnyc: I think I've got it, and even managed to make it configurable by adding a second material definition with the Light Path - gated Emission shader mixed in, called "Background.Material.Shadeless", with a fake user so it stays in the Blend file even though it's unassigned. 

The Python code just assigns one or the other material, depending on the value of `use_shadeless`.

..Honestly, having done that, I can't see that it makes _any_ difference. The background appear pretty much shadeless even if that option's disabled. But, meh, now it's there. (And there is some _vaguely_-detectable difference in the color tone, at times.)

ferdnyc: Pushed that commit onto the branch, and a second one to save/restore keyboard focus when the preview updates. It was driving me nuts that I couldn't focus the preview frame slider and just use the arrow keys to switch frames, because every time the preview refreshed it would lose focus. So, now I can!

src/windows/views/blender_listview.py
ferdnyc: I tried this, turns out it doesn't work. Blender is very picky about what color parameters include an alpha value and which don't. Get it wrong and it's either:
```
ValueError: bpy_struct: item.attr = val: sequences of dimension 0 should contain 4 items, not 3
```
or
```
ValueError: bpy_struct: item.attr = val: sequences of dimension 0 should contain 3 items, not 4
```

src/blender/scripts/colors.py
ferdnyc: Arrrgh. I just realized why the param doesn't seem to make a difference, the code doesn't actually _change_ the material assignment for the background right now. So, the param still has no effect, in terms of the visible render. (And I apparently have a vivid imagination, thinking I saw any difference at all.)  Oops.

ferdnyc: And now that I look at it, this wouldn't be easy at all because the color needs to be a keyframable parameter, but which parameter that is would be totally different in the node-shaded material case. I'm going to revert the changes I made to `colors.py`/`colors.blend`, since they don't work. I suspect the best option might just be to dump "use_shadeless". The animation looks fine without it.

RafaelLinux: I just downloaded Blender 2.8 and latest (at this hour) OpenShot appimage and updated the executable path in OpenShot. However, I got the typical window with error "You need blender ....". I even closed an reopen OpenShot, but nothing changes ...

ferdnyc:
Speaking of double-loads:

The other day I started toying with the idea of adding validation to the Preferences window's `browse` type (simple `if os.path.exists($pref_string):` checking of the provided path string), in response to another "I have the Blender path set in the Preferences but when I hit Render it still doesn't work" type bug report.

And then I asked myself, "Well, why not do the _version checking_ there as well? Instead of waiting until the user's actually trying to use the program, to tell them the string they entered in a completely different window is no good."

So that evolved into this plan where I'd move validation for the Blender and Inkscape executable dependencies into their own separate functions in a new `src/classes/external_app.py` module, which Preferences would trigger after changes to the corresponding path string, or the animated titles dialog could fire off if the path wasn't flagged as already having been verified in the current session. But it'd get triggered on window load, and only _once_ per session unless the path changed (after which a positive result is cached in memory for the remainder of the session), rather than doing it at the start of **every** single Blender execution.

So that's kind of where I'm at with that, and I have (not-quite-even-)half-baked code coming together to implement it. Figured I'd get your take on it and ask if you had any concerns or suggestions before I continue too much farther down that path. So, this is me, asking. :wink:

jonoomph:
@ferdnyc Yeah, I like the idea of validating the path (and version) dependencies when they are entered on the Preferences window. However, I'm not sure if caching the results makes much of a difference (performance wise), but I'm not opposed to it either.

ferdnyc:
> However, I'm not sure if caching the results makes much of a difference (performance wise), but I'm not opposed to it either.

\*shrug\* Probably not a huge one, but `blender -v` takes between 0.35 and 0.6 seconds to run on my SSD-backed system. Right now it's run, and the output parsed and checked, before every single execution of the actual render process. Skip through a few preview frames, that feels like it would add up pretty quickly.

Especially when there's almost no danger the answer will change, once it comes back as "valid'. (For "invalid", sure, we should keep checking every time, in case the user just installed/upgraded the application after the previous try.)

jonoomph:
True, yeah... go with your plan! :100:

src/windows/views/blender_listview.py
```diff
@@ -708,10 +703,14 @@ class Worker(QObject):
     blender_error_with_data = pyqtSignal(str)  # 1007
     enable_interface = pyqtSignal()  # 1008
 
+    @pyqtSlot()
+    def Cancel(self):
+        """Cancel worker render"""
+        self.is_running = False
+
```
ferdnyc: Let's try that again:

Would this let blender run until it finishes its current job? Could it actually do a self.process.terminate() as well, to pull the worker out early?

jonoomph: Yes, good idea. Done.

jonoomph:
@ferdnyc Did you have anything else to add to this PR before I merge things?

ferdnyc:
> @ferdnyc Did you have anything else to add to this PR before I merge things?

Nope, still working on the revamped version-checking code, but that can be a separate PR. If the new animations are ready I'd say it's g2g!

ferdnyc:
(Speaking of the version-checking, fun fact: Did you know that Inkscape on Windows installs _both_ an `inkscape.exe` AND `inkscape.com` in its install directory? And the `inkscape.exe` executable doesn't respond to any of the standard command-line switches? So, to do things like query the version installed, you have to _replace_ `inkscape.exe` with `inkscape.com` when launching the executable, or you'll get an Inkscape window instead of a console response to your query? ...That's pretty f*d up, right there.

Also not actually important for anything, since we don't have any Inkscape version criteria at the moment, but I figured if I'm going to check Blender, might as well do it for Inkscape as well. And this was my discovery of the week.)

jonoomph:
Blender 2.80 support has arrived!

jonoomph:
On a side note, I really love the new version of Blender. It is very inspiring, the entire Blender story leading up to this release. It will continue to be an inspiration for OpenShot and myself. :+1: Good job Blender devs!!!!!

ferdnyc:
> On a side note, I really love the new version of Blender.

It's quite impressive, a huge change. It's surprising that they managed to revamp their entire node interface into a set of dynamic sidebar menus, so you almost never have to look at the actual node graphs unless you want to — but they're still _there_, if you do, so really now it has two entire, almost separate interfaces for a lot of its functionality. Something to fit every user, I guess!

PR #2925, opened 2019-08-07
Rename Linux metadata v2

eszlari:
After PR https://github.com/OpenShot/openshot-qt/pull/2011 was closed, this PR only renames:

* `xdg/openshot-qt.xml`
* `xdg/openshot-qt.appdata.xml`
* `xdg/openshot-qt.desktop`

So it shouldn't interfere with non-Linux platforms.


ferdnyc:
@eszlari 

I added a commit to fix a couple of instances where these files were used in `installer/build-server.py`, to avoid breaking the AppImage builder. With that in place, I have no objections to this. \*shrug\*.

ferdnyc:
I pushed this to the branch `pr-2925` here in the repo so that the project build servers could get at it. They all completed the build successfully, and the resulting AppImage runs just fine on my system. It seems all the bases are covered here, so I'm going to merge this. Thanks, @eszlari !

PR #2930, opened 2019-08-10
Razor tool: Disable right-click

### LINKED ISSUE #1044: Right-click cuts when razor tool enabled
I would expect a right-lick to just open the context menu always. Left click would engage a selected action like razor tool. This is Win7, v. 2.4.1.

ferdnyc:
Previously it would _both_ perform the cut _and_ open the element's context menu, which was confusing. Since menu access is disabled in razor mode (as indicated by the removal of the elements' menu icon when the mode is activated), this simply carries through that intention.

Fixes #1044

### DIFF
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -725,62 +725,62 @@ App.controller('TimelineCtrl',function($scope) {
 	}
  };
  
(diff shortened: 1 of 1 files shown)


jonoomph:
LG

PR #2931, opened 2019-08-10
Preferences: Avoid code duplication for browse type

ferdnyc:
A _really_ minor thing, but I happened to be in `preferences.py` and noticed we could avoid some code duplication for the "browse"-type prefs.

### DIFF
--- a/src/windows/preferences.py
+++ b/src/windows/preferences.py
@@ -221,19 +221,16 @@ class Preferences(QDialog):
                     widget.setToolTip(param["title"])
                     widget.valueChanged.connect(functools.partial(self.spinner_value_changed, param))
 
-                elif param["type"] == "text":
+                elif param["type"] == "text" or param["type"] == "browse":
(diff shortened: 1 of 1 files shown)


SuslikV:
3 checks instead of 2?

ferdnyc:
> 3 checks instead of 2?

A small price to pay, to avoid writing the same three widget-creation lines twice (and dealing with subtle bugs down the road, when future commits alter one or the other so that they're no longer identical).

Compilers optimize away simple boolean tests, anyway. (Usually by inlining the different branches, so effectively that code will end up unspooled to look exactly the way it did originally. The difference is, because it's the _compiler_ (or, bytecompiler) duplicating the code, not the coders, the duplicated sections will always be identical.)

ferdnyc:
Another trivial change that's received no objections, and which I can confidently say has no impact on the code's functionality. Merging.

PR #2932, opened 2019-08-12
Preferences: Raise max samplerate to 192kHz.

ferdnyc:
I thought I'd fixed this forever ago (I know I _noticed_ it forever ago), but I guess not since it's still there. Oops.

The range-max for the field was set **lower** than the largest selectable value.

### DIFF
--- a/src/settings/_default.settings
+++ b/src/settings/_default.settings
@@ -152,7 +152,7 @@
     "restart": true
   },
   {
-    "max": 128000,
+    "max": 192000,
     "title": "Default Audio Sample Rate",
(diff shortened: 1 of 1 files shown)


SuslikV:
What are you waiting for? It should be merged and forgotten.

ferdnyc:
@SuslikV 

I'm generally reluctant to merge my own changes, unless it's something inconsequential like a formatting, comment, or documentation change. No matter how small it is, I prefer a second pair of eyes.

But, this **is** a very small change, and I suppose I can count your comment.

Issue #2933, opened 2019-08-12
Application crashes when Debug Mode is enabled and main window closed


SuslikV:
**Describe the bug:**
Application crashes when Debug Mode is enabled Preview thread is still running and main window was closed.

**Steps to reproduce the crash:**
The crash happens if you:
1) Enable _Debug Mode (Verbose)_ in application _Preferences_
2) Import media file (video) to OpenShot
3) Right-click over the added in p2 file and click _Preview File_ option
4) Click playback button inside new _Preview_ window
5) While playback of the media file is running close OpenShot's main window (the Preview should stay on screen)
6) New dialog will appear that prompts to save the changes - click "No" button to not save the changes
7) Application will crash in `libopenshot.dll` module

**Expected behavior:**
Debug Mode itself shouldn't introduce new errors in the code. Application should exit without the errors.

**System Details:**
- OpenShot v2.4.4 (OpenShot-v2.4.4-dev2-1565453395-4bb54b08-f17a7ed4-x86_64.exe)
- Windows 7, x64

PR #2934, opened 2019-08-12
Prevent application crash when Debug Mode is enabled

SuslikV:
Fixes https://github.com/OpenShot/openshot-qt/issues/2933

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -147,6 +147,7 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         self.timeline_sync.timeline = None
 
         # Close & Stop libopenshot logger
+        openshot.ZmqLogger.Instance().Enable(False)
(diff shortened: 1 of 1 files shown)


ferdnyc:
@SuslikV 

Hmmm. It sounds to me like the real problem is that `ZmqLogger::Close()` doesn't _implicitly_ also disable the logger. I think it would be better if it did, so that the caller doesn't have to jump through that extra hoop. 

What do you think about adding an `enabled = false;` to `ZmqLogger::Close()` in libopenshot, instead of this change? 

Assuming that has the same effect (seems like it would), I think it's a bit cleaner since we never _don't_ want the logger to be disabled when it's closed, and this way we won't have to worry about the callers remembering to do it themselves, when they shouldn't have to.

Good catch, though, either way.

SuslikV:
Closing this in favor of https://github.com/OpenShot/libopenshot/pull/308

PR #2936, opened 2019-08-13
Log the start and end of the session

SuslikV:
Make the exit event more noticeable in the log. Threads may leave entries below the last line. But the overall look of the log-file is much better now (as for me).

Example of the new log-file end with the Preview thread still running in background (debug mode enabled):
```
...
logger_libopenshot:INFO VideoPlaybackThread::run (before render) (frame->number=359.0000, need_render=1.0000)
logger_libopenshot:INFO PlayerPrivate::run (determine sleep) (video_frame_diff=-10.0000, video_position=359.0000, audio_position=369.0000, speed=1.0000, render_time=0.0000, sleep_time=33.0000)
     cutting:INFO btnPlay_clicked
     cutting:INFO pause (icon to play)
preview_thread:INFO onModeChanged
 main_window:INFO ================ Shutting down =================
 main_window:INFO Prompt user to save project
preview_thread:INFO onModeChanged
preview_thread:INFO exiting thread
logger_libopenshot:INFO Timeline::Close ()
 main_window:INFO ================================================

     cutting:INFO closeEvent
preview_thread:INFO exiting thread
```

Just minor improvement for beauty of the log-file.

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -37,6 +37,7 @@ from PyQt5.QtGui import QPalette, QColor, QFontDatabase, QFont
 from PyQt5.QtCore import Qt
 from PyQt5.QtCore import QT_VERSION_STR
 from PyQt5.QtCore import PYQT_VERSION_STR
+from PyQt5.QtCore import pyqtSlot
 
 try:
     # Enable High-DPI resolutions
@@ -61,6 +62,13 @@ class OpenShotApp(QApplication):
             # Import modules
             from classes import info
             from classes.logger import log, reroute_output
+
+            # Log the session's start
+            import time
+            log.info("------------------------------------------------")
+            log.info(time.asctime().center(48))
+            log.info('Starting new session'.center(48))
+
             from classes import settings, project_data, updates, language, ui_util, logger_libopenshot
             import openshot
 
@@ -80,7 +88,7 @@ class OpenShotApp(QApplication):
         # Log some basic system info
         try:
             log.info("------------------------------------------------")
-            log.info("   OpenShot (version %s)" % info.SETUP['version'])
+            log.info(("OpenShot (version %s)" % info.SETUP['version']).center(48))
             log.info("------------------------------------------------")
 
             v = openshot.GetVersion()
@@ -94,6 +102,9 @@ class OpenShotApp(QApplication):
             log.info("pyqt5 version: %s" % PYQT_VERSION_STR)
         except:
             pass
+			
+        # Connect QCoreApplication::aboutToQuit() signal to log end of the session
+        self.aboutToQuit.connect(self.onLogTheEnd)
 
         # Setup application
         self.setApplicationName('openshot')
@@ -233,3 +244,20 @@ class OpenShotApp(QApplication):
 
         # return exit result
         return res
+
+    # Log the session's end
+    @pyqtSlot()
+    def onLogTheEnd(self):
+        """ Log when the primary Qt event loop ends """
+
+        try:
+            from classes.logger import log
+            import time
+            log.info('OpenShot\'s session ended'.center(48))
+            log.info(time.asctime().center(48))
+            log.info("================================================")
+        except Exception:
+            pass
+
+        # return 0 on success
+        return 0
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -100,7 +100,7 @@ class OpenShotApp(QApplication):
             log.info("python version: %s" % platform.python_version())
             log.info("qt5 version: %s" % QT_VERSION_STR)
             log.info("pyqt5 version: %s" % PYQT_VERSION_STR)
-        except:
+        except Exception:
             pass
 			
         # Connect QCoreApplication::aboutToQuit() signal to log end of the session
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -101,9 +101,6 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
     # Save window settings on close
     def closeEvent(self, event):
 
-        # Close any tutorial dialogs
-        self.tutorial_manager.exit_manager()
-
         # Prompt user to save (if needed)
         if get_app().project.needs_save() and not self.mode == "unittest":
             log.info('Prompt user to save project')
@@ -120,6 +117,12 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
                 # User canceled prompt - don't quit
                 event.ignore()
                 return
+
+        # Log the exit routine
+        log.info('---------------- Shutting down -----------------')
+
+        # Close any tutorial dialogs
+        self.tutorial_manager.exit_manager()
 
         # Save settings
         self.save_settings()
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -101,6 +101,10 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
     # Save window settings on close
     def closeEvent(self, event):
 
+        # Some window managers handels dragging of the modal messages incorrectly if other windows are open
+        # Hide tutorial window first
+        self.tutorial_manager.hide_dialog()
+
         # Prompt user to save (if needed)
         if get_app().project.needs_save() and not self.mode == "unittest":
             log.info('Prompt user to save project')
@@ -114,6 +118,8 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
                 self.actionSave_trigger(event)
                 event.accept()
             elif ret == QMessageBox.Cancel:
+                # Show tutorial again, if any
+                self.tutorial_manager.re_show_dialog()
                 # User canceled prompt - don't quit
                 event.ignore()
                 return
--- a/src/windows/views/tutorial.py
+++ b/src/windows/views/tutorial.py
@@ -287,6 +287,11 @@ class TutorialManager(object):
             self.dock.raise_()
             self.dock.show()
 
+    def hide_dialog(self):
+        """ Hide an active dialog """
+        if self.current_dialog:
+            self.dock.hide()
+
     def re_position_dialog(self):
         """ Reposition a tutorial dialog next to another widget """
         if self.current_dialog:


src/windows/main_window.py
```diff
@@ -153,6 +156,9 @@ def closeEvent(self, event):
         # Destroy lock file
         self.destroy_lock_file()
 
+        # Main window widget closed
+        log.info('================================================\n')
```
ferdnyc: Personally, I'm good with this _except_ for the newline at the end of this line. I just don't think a blank line in the logfile is very helpful, especially since (as you say) other threads from the session can still log messages **after** that line.

If we're going to insert newline separators to signify session boundaries in the logfile, I'd rather insert one **BEFORE** the _first_ log message from a NEW session. That way, we can unequivocally guarantee that the blank line is the exact division between two sessions.

As far as the end-of-shutdown marker, either logging a line of `====` twice in a row, or logging something like `======= Main thread terminating =======` would be my preference, over the line with an extra newline appended. 

That way, if any further lines do follow that marker, it'll be clear that they come from background threads but still belong to the preceding session. (Especially if we do the newline-before-the-first-message thing.) And its absence will be a clear indicator of abnormal shutdowns, should something crash inside the `closeEvent()` handler.

ferdnyc: (In fact, while you're in there...)

Since the logfiles don't have timestamps, I've often thought that logging just the start and end time of each session would be a useful thing. So, as part of that session-boundary logging, maybe the first and last-but-for-stragglers log messages could be something like this:

```py
# Start
import time
now = time.asctime()
log.info('\n================================================')
log.info('=========' + now.center(30) + '=========')

#End
import time
now = time.asctime()
log.info('=========' + now.center(30) + '=========')
log.info('================================================')
```

SuslikV:
Meanwhile I found another small bug in the logging system. I'll report it in another thread to not mix it up with these enhancements.

SuslikV:
Updated. New logging code.

Now it looks like this (in between of two sessions, the previous one had thread that was terminated only after main application closed):

```
...
preview_thread:INFO self.player.Position(): 1
     cutting:INFO btnPlay_clicked
     cutting:INFO play (icon to pause)
preview_thread:INFO onModeChanged
 main_window:INFO ---------------- Shutting down -----------------
 main_window:INFO Prompt user to save project
preview_thread:INFO onModeChanged
preview_thread:INFO exiting thread
     cutting:INFO closeEvent
preview_thread:INFO exiting thread
         app:INFO             OpenShot's session ended   
         app:INFO             Thu Aug 15 14:58:26 2019   
         app:INFO ================================================
         app:INFO ------------------------------------------------
         app:INFO             Thu Aug 15 14:59:05 2019   
         app:INFO               Starting new session     
         app:INFO ------------------------------------------------
         app:INFO          OpenShot (version 2.4.4-dev2)    
         app:INFO ------------------------------------------------
         app:INFO openshot-qt version: 2.4.4-dev2
         app:INFO libopenshot version: 0.2.3
...

```

SuslikV:
More over, the automation build log shows that it exits the application without calling `quit()` - there is no entry that session was ended.

It seems that the `main_window.py` missing next part inside the _closeEvent_ method (at the end):
```
        get_app().quit()
```
But this is another minor issue...

ferdnyc:
> More over, the automation build log shows that it exits the application without calling `quit()` - there is no entry that session was ended.
> 
> It seems that the `main_window.py` missing next part inside the _closeEvent_ method (at the end):
> 
> ```
>         get_app().quit()
> ```
> 
> But this is another minor issue...

Hmm, I don't believe the close event should be calling `quit()` itself, that doesn't sound right... (but I could be wrong).

The special-casing of `mode == "unittest"` in `MainWindow.closeEvent()` complicates things a bit. It probably wasn't intended to affect the shutdown process but, now that I look at it, in that mode the `event` will be neither `accept()`ed **nor** `ignore()`d. I think that's probably the issue.

src/classes/app.py
```diff
@@ -232,3 +246,22 @@ def run(self):
...
+        try:
+            from classes.logger import log
+            import time
+            now = time.asctime()
+            new_session = 'OpenShot\'s session ended'
+            log.info("         " + new_session.center(30))
+            log.info("         " + now.center(30))
+            log.info("================================================")
+        except:
+            pass
```
ferdnyc: [No (more) bare `except:`s](https://moserei.de/2017/09/20/python-bare-except.html), please. The code is riddled with enough that I've been meaning to get rid of.

`except Exception as ex:`, or `except Exception:` if you don't need access to `ex` (like here).

src/classes/app.py
```diff
@@ -94,6 +105,9 @@ def __init__(self, *args, mode=None):
             log.info("pyqt5 version: %s" % PYQT_VERSION_STR)
         except:
             pass
+			
+        # Connect QCoreApplication::aboutToQuit() signal to log end of the session
+        self.aboutToQuit.connect(self.onLogTheEnd)
```
ferdnyc: Smart. I wonder if more of the stuff in our `closeEvent()` handler shouldn't _actually_ be in a handler for the `aboutToQuit` signal? But, that's a side question.

SuslikV:
I made the changes (variables number reduced, concatenation number reduced, types for exceptions set).

About the `quite()` usage inside the _exitEvent_ - I've seen it as part of adaptation: https://github.com/obsproject/obs-studio/pull/742#issuecomment-301389877 in OBS Studio.

ferdnyc:
> About the `quit()` usage inside the _exitEvent_ - I've seen it as part of adaptation: [obsproject/obs-studio#742 (comment)](https://github.com/obsproject/obs-studio/pull/742#issuecomment-301389877) in OBS Studio.

Yeah, but they're unsetting [`QApplication::quitOnLastWindowClosed()`](https://doc.qt.io/qt-5/qguiapplication.html#quitOnLastWindowClosed-prop), because they have a system tray icon. We're not / we don't, so it _should_ be automatic. However, [`Qt::WA_QuitOnClose`](https://doc.qt.io/qt-5/qt.html#WidgetAttribute-enum) "Makes Qt quit the application when the last widget with the attribute set has accepted `closeEvent()`.", so I'm still leaning towards the missing `accept()` as the culprit.

ferdnyc:
Hmm. I just tried it, and _neither_ accepting the event, nor adding `get_app().quit()` makes the `onLogTheEnd()` handler run from the unit test code.

SuslikV:
> Hmm. I just tried it, and _neither_ accepting the event, nor adding `get_app().quit()` makes the `onLogTheEnd()` handler run from the unit test code.

In both local and travis logs?

ferdnyc:
> > Hmm. I just tried it, and _neither_ accepting the event, nor adding `get_app().quit()` makes the `onLogTheEnd()` handler run from the unit test code.
> 
> In both local and travis logs?

I only tried it locally (`PYTHONPATH=/path/to/my/libopenshot/build/src/bindings python3 src/tests/query_tests.py`), but I can't imagine it being any different on Travis if it doesn't even work locally.

SuslikV:
And if you force `self.show()` inside `__init__` of `main_window.py`?

src/classes/app.py
```diff
@@ -80,7 +88,7 @@ def __init__(self, *args, mode=None):
         # Log some basic system info
         try:
             log.info("------------------------------------------------")
-            log.info("   OpenShot (version %s)" % info.SETUP['version'])
+            log.info(("OpenShot (version %s)" % info.SETUP['version']).center(48))
```
ferdnyc: Hmm. While that _does_ work, in this one particular case only, I'm sort of inclined to use an intermediate variable. Because boy is that unreadable, compared to a modified version of your previous edit...
```py
app_name_str = "OpenShot (version %s)" % info.SETUP['version']
...
log.info(app_name_str.center(48))
```
Personally I hate the `%`-based (old-style) format strings, and this kind of situation is exactly why. In modern-er python this would be,
```py
log.info("OpenShot (version {})".format(info.SETUP['version']).center(48))
```
which is at least _less_ unreadable (IMHO) — though I'd still probably use an intermediate variable to hold the formatted string.

ferdnyc:
> And if you force `self.show()` inside `__init__` of `main_window.py`?

Hm. You may be right about that being the reason, but currently when I run the unit tests no GUI appears at all. And since that's the way they're _supposed_ to run, anything that would change it is a bad idea. So, even if that would work, better if they just not process that signal. (Not like they _need_ to.)

ferdnyc:
@SuslikV 

Heh. Nope, I'd already tried `Qt::QueuedConnection`, _and_ I tried connecting the `aboutToQuit` signal as a `Qt.QueuedConnection` as well. No dice.

I think you're right. I think it's because, since the main window doesn't ever actually _show_, nothing ever fires off the `aboutToQuit` signal. There probably isn't a clean way to get around that without either forcing or refactoring a lot of things, which isn't worth it really. 

Like I said, not running `onLogTheEnd` is fine for the unit tests. They work as they're intended to. They're just not a testbed for the full application lifecycle. But that's what actual user testing is for.

SuslikV:
I updated the logging order of the MainWindow close event and moved tutorial dialog exit after the confirmation that can appear if project was changed during tutorial.

This aims at two targets (after project was modified and not saved yet):
1) Tutorial will stay open on accidental main window close and next _Cancel_.
2) On _Cancel_ during main window close there will be no entry in the log that shutdown started.

In both these cases the close event is ignored, so no real termination is happened - nothing to close or log.

ferdnyc:
> 2. On _Cancel_ during main window close there will be no entry in the log that shutdown started.

That makes total sense...
 
> 1. Tutorial will stay open on accidental main window close and next _Cancel_.

But this one's an issue, unfortunately. (I was just about to merge it, then I saw that change.) 

The tutorial being closed before the save dialog pops up is intentional. If the tutorial is open, it causes very weird interactions with the save dialog. (Like, at least on Linux, trying to move the save dialog — which normally works, and moves the main application with it, so it stays centered in the application window — will instead drag the main window, _**without** moving the save dialog_! But that only happens when the tutorial is left visible. Something to do with the weird window properties those docks have, I'm guessing.)

If the user cancels the save and wants the tutorial back, they can use the Help menu to launch it again.

src/windows/main_window.py
```diff
@@ -101,9 +101,6 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
     # Save window settings on close
     def closeEvent(self, event):
 
-        # Close any tutorial dialogs
-        self.tutorial_manager.exit_manager()
-
```
ferdnyc: Please remove, as discussed in comments.

SuslikV:
@ferdnyc about `quit()` - you said that you didn't tested it on travis, so I just used this PR to did it ;) Yeah, it didn't changes much for "unittest" mode and `show()` isn't case as far as I see.

SuslikV:
I can hide tutorial window and restore it if cancelled to workaround this strange behavior of some window managers.

SuslikV:
@ferdnyc is this new workaround with hiding the tutorial window works for you?

SuslikV:
I need to fix some indentation, so will rebase it one more time if the workaround works, of course.

ferdnyc:
@SuslikV That sounds good to me, I'll take a look in a couple of hours after I sort out dinner.

ferdnyc:
@SuslikV The tutorial hiding/restoration works nicely, so I'd say this is good to go. Merging, thanks!

PR #2938, opened 2019-08-14
Zoom centered on the playhead, if visible

ferdnyc:
Reading the git history for `src/timeline/js/controllers.js`, I noticed commit 492f68c0 which added centering around the mouse cursor for <kbd>Ctrl</kbd>+mousewheel zooming. 

I'll be honest, I never even knew about that feature! But, inspired by it, I added logic to compute the playhead position and use **that** location to center the zoom, if the cursor isn't over the canvas and the playhead is currently visible on-screen. So, scale adjustments using the zoom slider / buttons will try to keep the playhead visible while zooming.

If the playhead is scrolled offscreen in either direction, the code falls back to the usual centering at the left edge of the canvas.

My editor also cleaned up a whole _bunch_ of non-blank blank lines, sorry. Someday I'm just going to stop apologizing for that.


jonoomph:
Looking good! Some conflicts though...

jonoomph:
I think I resolved everything correctly

Issue #2939, opened 2019-08-15
Two exceptions in the log on application close (settings file not writable)


SuslikV:
**Describe the bug:**
Instead of safely logging the error that save file cannot be written on exit - application runs into another exception.

**Steps to reproduce the behavior:**
1. Run OpenShot
2. Make the `openshot.settings` file not accessible for writing (set read-only attribute)
3. Close OpenShot
4. See 2 errors in the log instead of single one.
One error is about the permissions, another exception is about the unknown name `log`.

**Expected behavior:**
Application should log that the error happened instead of spawning more errors.

**System Details:**
- OpenShot 2.4.4: OpenShot-v2.4.4-dev2-1565798384-1cdddba0-f17a7ed4-x86_64.exe
- Windows 7, x64

**Log Files:**

<details>
  <summary>Click to expand</summary>

```
...
timeline_webview:INFO UpdateLayerIndex
     version:INFO Found current version: {"openshot_version": "2.4.4"}
 main_window:INFO foundCurrentVersion: Found the latest version: 2.4.4
preview_thread:INFO onModeChanged
preview_thread:INFO exiting thread
   json_data:ERROR Couldn't save user settings file:
C:\Users\***\.openshot_qt\openshot.settings
[Errno 13] Permission denied: 'C:\\Users\\***\\.openshot_qt\\openshot.settings'
  exceptions:ERROR Unhandled Exception
Traceback (most recent call last):
  File "I:\Temp\OpenShot-dev\01\{app}\classes\json_data.py", line 151, in write_to_file
    with open(file_path, 'w') as f:
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\***\\.openshot_qt\\openshot.settings'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "I:\Temp\OpenShot-dev\01\{app}\classes\app.py", line 229, in run
    self.settings.save()
  File "I:\Temp\OpenShot-dev\01\{app}\classes\settings.py", line 118, in save
    self.write_to_file(file_path, self._data)
  File "I:\Temp\OpenShot-dev\01\{app}\classes\json_data.py", line 156, in write_to_file
    raise Exception(msg)
Exception: Couldn't save user settings file:
C:\Users\***\.openshot_qt\openshot.settings
[Errno 13] Permission denied: 'C:\\Users\\***\\.openshot_qt\\openshot.settings'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #2940, opened 2019-08-15
Import missing variable

SuslikV:
Fixes: https://github.com/OpenShot/openshot-qt/issues/2939

**Edit:**
Log entries in case of error now will look like this:
```
preview_thread:INFO exiting thread
   json_data:ERROR Couldn't save user settings file:
C:\Users\***\.openshot_qt\openshot.settings
[Errno 13] Permission denied: 'C:\\Users\\***\\.openshot_qt\\openshot.settings'
         app:ERROR Couldn't save user settings on exit.
Couldn't save user settings file:
C:\Users\***\.openshot_qt\openshot.settings
[Errno 13] Permission denied: 'C:\\Users\\***\\.openshot_qt\\openshot.settings'
```

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -226,6 +226,7 @@ class OpenShotApp(QApplication):
         res = self.exec_()
 
         try:
+            from classes.logger import log
             self.settings.save()
         except Exception as ex:
             log.error("Couldn't save user settings on exit.\n{}".format(ex))


ferdnyc:
Hm. Good catch.

With `settings.save()` (or actually `json_data.write_to_file()`) already logging the failure, I wonder whether `app.run()` even really _needs_ to log the same condition again... but since it **is** logging, it needs access to the log.

PR #2946, opened 2019-08-25
Preferences: Start "Browse" at location of current setting

ferdnyc:
When the user clicks the "Browse" button next to a browse-type preference item, instead of always starting at `QDir.rootPath()`, take the old value for that setting and open at its parent directory. 

If the parent directory doesn't exist, walk up the path and start at the first extant directory encountered. (Or the root of the filesystem, our old starting point, if every path element fails.)


jonoomph:
LG!

PR #2948, opened 2019-08-25
Title Editor: Reuse previous selection in Font selector

### LINKED ISSUE #2945: Title Editor window's "Change Font" button always opens with Ubuntu 10.5pt selected
Migrated from the conversation at #2069: Whenever the user clicks "Change Font" in the Title Editor window, no matter what font is actually being used in the title, the dialog _always_ opens with the same font selected: The application-default Ubuntu Regular, at 10.5pt size.

This both makes it impossible to _check_ what font is currently in use (the name isn't displayed anywhere in the interface), and it causes confusion for the user by giving them the impression that their selections in the dialog are ignored.

1. Open Title Editor, choose a template, click "Change Font"
2. Select a visibly different font from Ubuntu, such as Comic Sans or Times New Roman
3. Confirm the selection, and note the font change in the title preview
4. Click "Change Font" again, and notice that the window once again comes up with Ubuntu Regular selected.

**Expected behavior:**
The initially-selected font in the dialog would match the current selection in the interface

**System Details:**
- OpenShot Version [e.g. 2.4.3]: 2.4.4 and Git HEAD
- Operating System / Distro: Linux and Windows, at least (most likely All)

ferdnyc:
Open the font selection dialog with the previous selection as the default, instead of always resetting to the application font (Ubuntu).

When the window is created or a new file is loaded, we construct a default `QFont()` from the "Bitstream Vera Sans" typeface to use as the initial selection. 

Just as with the color picker, because we are not actually parsing the information from the data file, **this initial selection may be incorrect**... especially if the title file is a user-customized copy of a template. (However, it has a much better chance of being correct on the first try than defaulting to Ubuntu, which none of our templates use as their default font.)

In addition:
* Some code-style fixes suggested by the linter
* If the user cancels a font/color dialog or accepts it without changing the selected item, don't force a save of the file data or a refresh of the preview image as there's nothing to update.

Fixes #2945

### DIFF
--- a/src/windows/title_editor.py
+++ b/src/windows/title_editor.py
@@ -346,11 +346,11 @@ class TitleEditor(QDialog):
             self.update_font_color_button()
             self.font_color_code = col
 
-        # Something changed, so update temp SVG
-        self.writeToFile(self.xmldoc)
+            # Something changed, so update temp SVG
+            self.writeToFile(self.xmldoc)
 
-        # Display SVG again
-        self.display_svg()
+            # Display SVG again
+            self.display_svg()
 
     def btnBackgroundColor_clicked(self):
         app = get_app()
@@ -366,11 +366,11 @@ class TitleEditor(QDialog):
             self.update_background_color_button()
             self.bg_color_code = col
 
-        # Something changed, so update temp SVG
-        self.writeToFile(self.xmldoc)
+            # Something changed, so update temp SVG
+            self.writeToFile(self.xmldoc)
 
-        # Display SVG again
-        self.display_svg()
+            # Display SVG again
+            self.display_svg()
 
     def btnFont_clicked(self):
         app = get_app()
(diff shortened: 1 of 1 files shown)


ferdnyc:
I also looked at setting a filter on the font selection dialog so that it doesn't display bitmap fonts (which are not compatible with SVG files, so they can't be used in the Title Editor), but unfortunately font filtering only works with the Qt built-in (non-native) font dialog, and sometimes (MacOS) not even then, apparently.

Which would mean, on Linux, giving up this:

Screenshot from 2019-08-25 07-29-27

for this:

Screenshot from 2019-08-25 07-28-32

Too drastic and not worth the tradeoff, IMHO.

SuslikV:
Are you sure that the _Bitstream Vera Sans_ present on all systems?

jonoomph:
LG

ferdnyc:
> `QFont()` is smart, anyway — I just re-tested in my Win7 VM (which does not have the B.V.S. font installed), and it just falls back to the application default font (Ubuntu Regular) automatically. That one _is_ guaranteed to be available, as it's distributed as part of OpenShot.

That statement changes somewhat in the presence of the Resvg renderer, since a bundled font that exists only in the OpenShot application directory still isn't usable _by Resvg_. So, even Ubuntu isn't guaranteed to be available for SVG rendering, unfortunately. (In fact, it's far less likely to be installed on an average system than a common default like B-V-S, although that's probably still not that common either outside of Linux.)

Issue #221, opened 2016-06-22
Rejecting valid image files for image sequence import


crablab:
When trying to import multiple .jpg files as an image sequence, OpenShot gives an error saying that they are not valid image, sound or video files (which they are). 

Image here: 
![image](https://cloud.githubusercontent.com/assets/5859718/16252725/9663b922-3828-11e6-98ac-099c313d28b3.png)

opensourcegrappler:
I also have this error.

Changing the filename of the image seems to fix it (in my case image000.png to foo.png) allows the image to get imported. But if the image is part of an image sequence (it is in my case) the name change prevents the rest of the image sequence from being imported.

DylanC:
@crablab - Do you get this error in v2.3.4 or in the daily build?

opensourcegrappler:
I haven't used openshot since I commented on this issue 1 year ago. So I don't know, sorry.

DylanC:
@crablab - Image sequence is working much better now. You should try v2.4.

zerekw:
I'm still getting this error in 2.4 on OSX 10.12.6. It fails the first time and then on trying to reimport the file it successfully imports the single image but not the sequence.

zerekw:
Must be something with the name. Ran an automator script to modify my filenames from "G0011602.JPG" -> "gopro1.JPG". Now everything seems to be working.

zerekw:
I suspect its something with the number. When I tried to start from a large number, ie. "gopro1234". It choked up again. When I changed the sequence to start from a file named "gopro1.JPG" it worked.

DylanC:
@zerekw - Yes, sounds like it was designed to work with simple numbers for now.

tahteche:
In my case an underscore in the file name seems to have the same effect.

dougisfunny:
From what I can see, it's not so much 'simple' numbers. They just have to start with 0s.
This works:

> Sequence01.jpg
> Sequence02.jpg
> Sequence03.jpg
> Sequence04.jpg
> 
As does this
> Sequence04.jpg
> Sequence05.jpg
> Sequence06.jpg
> Sequence07.jpg
>
This does not

> Sequence11.jpg
> Sequence12.jpg
> Sequence13.jpg
> Sequence14.jpg
>
 Nor Does this
> Sequence05.jpg
> Sequence06.jpg
> Sequence07.jpg
> Sequence08.jpg
>
Looking through the code I tracked it to files_treeview.py line ~200

```
[...]

PR #2949, opened 2019-08-26
Image sequence import improvements

ferdnyc:
Several fixes to the process of importing image sequences:

- Also log (at warning level) the exception that caused an import failure QMessageBox
- Add paths, filenames to logging of image sequence discovery
- Reset `ignored_image_sequence_paths` after a successful import

### DIFF
--- a/src/windows/views/files_listview.py
+++ b/src/windows/views/files_listview.py
@@ -194,8 +194,10 @@ class FilesListView(QListView):
             file.save()
             return True
 
-        except:
-            # Handle exception
+        except Exception as ex:
+            # Log exception
+            log.warning("Failed to import file: {}".format(str(ex)))
+            # Show message to user
             msg = QMessageBox()
             msg.setText(_("{} is not a valid video, audio, or image file.".format(filename)))
             msg.exec_()
--- a/src/windows/views/files_listview.py
+++ b/src/windows/views/files_listview.py
@@ -192,6 +192,9 @@ class FilesListView(QListView):
 
             # Save file
             file.save()
+            # Reset list of ignored paths
+            self.ignore_image_sequence_paths = []
+
             return True
 
         except Exception as ex:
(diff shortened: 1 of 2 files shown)


ferdnyc:
The last one bears some explaining:

The current logic around `ignored_image_sequence_paths` is broken and weird. It's tied to the current _view_ object for the Project Files, so literally just switching between Thumbnail and Details view will clear out the ignored paths. The ignored paths are _also_ cleared whenever a drag-and-drop import is done. So, they're not really stored very persistently.

But, _within_ the current view, the list can be annoyingly persistent. For instance:
* If I import an image sequence, then _remove_ it from Project Files and try to reimport it, that sequence won't be recognized the second time because the directory STILL on the ignored paths list. 
* Importing two different image sequences from the same directory _also_ fails, even if they have completely distinct names. But if I switch the view from Thumbnail to Details or vice versa, then it imports fine. There's no sense to that.

So, as a result, now the list is just cleared after every successful import. Which means the user can theoretically import the same image sequence twice, if they choose to... and so what? They could always do that anyway, by switching the view or using drag-and-drop between the two import attempts.

TBH I'm tempted to just remove the ignore list entirely, as I'm pretty sure this change means that it has no actual purpose anymore. I just haven't looked at whether it might provide some protections _during_ the sequence discovery process, that would make it still useful.

jonoomph:
:+1: Yeah, image sequences are a bit wonky. I'm tempted to re-implement them from the ground up, and use our QtImageReader to fully support them, including looping support (bounce or reset), etc... Also, support for missing files is broken when checking for image sequences.

ferdnyc:
@jonoomph See the discussion at #221, on that front. (Wherein I try to trick a user into doing that rewrite, after they discovered that it's possible to set a `start_number` for FFmpeg so that the sequence isn't forced to be numbered from 0. Which is actually really great, and would improve the image sequence support greatly, but would require a complete reimplementation of the discovery/import process along the lines of what I sketch out in the last comment.)

PR #2951, opened 2019-08-26
Use openshot.OPENSHOT_VERSION_FULL for version checking

ferdnyc:
Now that libopenshot defines `OPENSHOT_VERSION_FULL` (which includes the `-dev#` extension, for development builds), use that instead of `GetVersion().ToString()` to record and compare libopenshot versions in application logic, logs, and project data.


ferdnyc:
I'm going to merge this, because it would be nice to get the `OPENSHOT_VERSION_FULL` string from `libopenshot` in the metrics-driven Issue reports, when they're opened.

PR #2956, opened 2019-08-28
classes.updates: Fix history-cleaning; logging tweaks

ferdnyc:
- Only check for history if the update is a complex type (not `only_value`)
- Check that value/old_value is a dict before trying to use `in`
- Log the exception caught, when there is one

This fixes a bug introduced in 33c0c219a310b814b7656227c2938fc64c80604e where a message would be logged on every `only_value` update (e.g. zoom change):
```
project_data:INFO _set key: ['scale'] values: 20 add: False partial: False remove: False
     updates:INFO Warning: failed to clear history attribute from undo/redo data.
```
This was due to a TypeError exception "argument of type 'int' is not iterable" (which was not logged). With these changes, every `except` block in `classes.updates` captures and logs the text of the exception caught (a personal crusade).

Additionally:
- Log the listener/watcher name, when rejecting duplicate adds

This last change means that every `log.LEVEL()` call in `classes.updates` now contains at least _**some**_ specific context information (another personal crusade).

### DIFF
--- a/src/classes/updates.py
+++ b/src/classes/updates.py
@@ -76,15 +76,15 @@ class UpdateAction:
                          "partial": self.partial_update,
                          "old_values": copy.deepcopy(self.old_values)}
 
-        # Always remove 'history' key (if found). This prevents nested "history"
-        # attributes when a project dict is loaded.
-        try:
-            if data_dict.get("value") and "history" in data_dict.get("value"):
-                data_dict.get("value").pop("history", None)
-            if data_dict.get("old_values") and "history" in data_dict.get("old_values"):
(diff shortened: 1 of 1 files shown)


ferdnyc:
I just caught on to two of these calls:
```python
log.info('Warning: failed to clear history attribute from undo/redo data.')
```
which even in my updated code were still using `log.info*()`. Changed them to (natch):
```python
log.warning('Failed to clear history attribute from undo/redo data. {}'.format(ex))
```

ferdnyc:
And with that tweak, and a month having passed on this small change, I'm going to merge it because I'd really like to get those false "Warning:"s out of people's logs.

PR #2960, opened 2019-08-31
Prev/Next Marker: Auto-scroll timeline to follow playhead when it moves offscreen

kartchnb:
…Currently, this can be used by the user to automatically jump to the current playhead position rather than manually searching for it.  The timeline also recenters itself when jumping to a marker or to the beginning or end of the timeline.

Modified "main-window.ui" to add the "actionCenterOnPlayhead" action.
Modified "_default.settings" to add a new hotkey (Ctrl+Up) for the "ationCenterOnPlayhead" action.
Modified "controller.js" to add the "centerOnTime" function to perform the actual work of scrolling the timeline.
Modified "timeline_webview.py" to add the "centerOnTime" function as an intermediary for the function in "controller.js".
Modified "main_window.py" to add the "centerOnPlayhead_triggered" function to implement the action behavior.  Also added code to automatically recenter when jumping to the beginning or end of the timeline or moving to a marker.


ferdnyc:
(Copying relevant responses from #2959. Because for reasons that escape even me, I wrote a bunch of stuff _there_ instead of _here_...)

@kartchnb 

Wow, this sounds amazing! I will definitely give it a look over the weekend. At first glance it looks awesome, though, the style looks good and the code seems generally sound. 

I'll leave a couple of comments now, about minor things I noticed on a read-through — one of them I see you already addressed, the unnecessary `scale` param to `centerOnTime()` — and I'll follow up with a more complete review once I've had time to check out the PR branch and give it a spin.

This will be really great to have in OpenShot, though! In fact, it strikes me that my PR #2938 already adds code to determiner whether or not the playhead is visible in the window, which could be abstracted out into a boolean function. (It's not currently abstracted out of the zooming code, but it could and probably should be anyway.) Combined with this, and pretty much all the building blocks are in place to even take a stab at general playhead follow-scrolling, too!

<hr>

> I actually attempted to do playhead follow-scrolling, but found that the performance seemed to suffer (at least on my machine), so I abandoned it for now. If it can be done efficiently, I think it would be nice to at least provide as an option. 

Oh, yeah, as things _currently_ stand I don't think it's feasible. I'd have to look at the code again, but IIRC we're updating the playhead position way too often as it is. Then you add repositioning the scroll into every one of those updates, and the drag on the system just gets immense.<sup>1</sup>

But that was the part that got me thinking: We don't really _need_ to reposition on _**every**_ update. We only need to scroll — whether it's for a Next/Prev Marker movement, _or_ while playing — **IFF** the playhead actually moves out of view. As long as it stays IN view, there's really no need to update the scroll position at all. 

Without having _tried_ it, still, my suspicion on reading the PR was that re-centering on _every_ move might feel "twitchy". (If you're at the start of the timeline and on the left half of the screen, it won't matter since the center can't move any farther left from where it is, but otherwise that sounds like an awful lot of scrolling.)

When a clip fits entirely within the timeline at the current zoom level, I actually _like_ that I can skip around it from one end to the other without anything moving except the playhead.

Combine (what could be) my `isPlayheadVisible()` chocolate with your `centerOnTime()` peanut butter, and we can have the best of both worlds: Move only the playhead as long as it stays within the current view, and scroll only when necessary to avoid losing track of it.

And maybe, eventually, doing the same thing for playback.


#### Notes
1. A digression: 

    Traditional media players don't have nearly as much going on as OpenShot does, during playback, but they also have interfaces that can be decoupled from the player itself, and even displayed remotely. So, those interfaces have to be lightweight, and protocols for remote control/display (like MPRIS2) are designed with that in mind. The communication between the player and the frontend during playback is surprisingly minimal. 

    When playback starts, the player sends a message to the frontend: "Play from $TIME index at $SPEED"... and _**that's it**_. They aren't expected to communicate again until one of them has a message for the other that changes that condition: Stop, Pause, Track Change, Play from $TIME (a seek), or Play at $SPEED (a rate change). 

    The playback position (time) indicator displayed while playback in underway, ticking away once a second (if not once a frame) or whatever? After playback is started, the time index itself doesn't actually come from the player: It's computed locally by the frontend, and updated on a timer. The frontend knows the start time and it knows the playback speed, that's all it needs to compute and update the elapsed time as play continues. There's no reason the backend needs to provide further information unless something happens to delay the actual output. (Exceptional conditions, e.g. buffering, lag, read errors, etc.)

    A different project I'm involved with (the [GSConect](https://github.com/andyholmes/gnome-shell-extension-gsconnect/) Gnome Shell extension) has a feature that mirrors media player controls to a paired Android device running the KDE Connect app. So, for example, when VLC is playing a video, I get a playback notification on my phone with standard transport controls, and I can effectively use my phone as a remote control. It's pretty neat.

    Anyway, we had a problem with a particular media player that was emitting MPRIS2 position updates just _four_ times per second — it doesn't sound like a lot, but because each of those had to be transformed into a network packet and sent to KDE Connect, the drain on system resources got _immense_. The only solution was to simply turn off updates from that software. The KDE Connect architecture just isn't designed to handle 250ms-interval updates, it's just too much.

    When OpenShot is playing, forget four times a second. IIRC it's updating the playhead position every _frame_ — 24, 30, even 60 times a second. You do the math.

src/windows/main_window.py
```diff
@@ -1311,6 +1323,25 @@ def actionNextMarker_trigger(self, event):
...
+        
+        # Calculate frames per second
+        fps = get_app().project.get("fps")
+        fps_float = float(fps["num"]) / float(fps["den"])
+        
+        # Calculate the playhead position in seconds
+        playheadPosition = player.Position() / fps_float
+        
+        # Center the timeline on the playhead position
+        self.timeline.centerOnTime(playheadPosition)
```
ferdnyc: It seems like there's too much math being done on the Python side here, when the JavaScript side already has the necessary information. 

It would be far more efficient to simply have a JavaScript function `$scope.centerOnPlayhead()`, and let it do all of those lookups. They're _easy_, from that side, as all of the data is basically right there. Then, the Python side just calls that with no arguments when necessary.

See `$scope.MovePlayhead` and `$scope.MovePlayheadToFrame` for examples of the necessary lookups.

src/windows/main_window.py
```diff
@@ -1311,6 +1323,25 @@ def actionNextMarker_trigger(self, event):
             get_app().window.refreshFrameSignal.emit()
             get_app().window.propertyTableView.select_frame(frame_to_seek)
 
+            # Center the timeline on the marker
+            self.timeline.centerOnTime(closest_position)
+            
+    def actionCenterOnPlayhead_trigger(self, event):
+        log.info('actionCenterOnPlayhead_trigger')
```
ferdnyc: These are basically debug `printf()`s that shouldn't even be in the _existing_ code, it's fine for testing but I'd really rather not commit any new ones.

src/windows/main_window.py
ferdnyc: (Basically, on the JavaScript side, `$scope.centerOnPlayhead()` can just be...):
```js
$scope.centerOnPlayhead = function() {
  return $scope.centerOnTime($scope.project.playhead_position);
);
```

src/timeline/js/controllers.js
```diff
@@ -413,6 +413,19 @@ App.controller('TimelineCtrl',function($scope) {
...
 	 $("#scrolling_tracks").scrollLeft(new_cursor_x);
  };
+ 
+ // Center the timeline on a given time position
+ $scope.centerOnTime = function(centerTime) {
+    // Get the width of the timeline
+    var scrollingTracksWidth = $("#scrolling_tracks").width();
+    
+    // Calculate the position to scroll the timeline to to center on the requested time
+    var pixelToCenterOn = centerTime * $scope.pixelsPerSecond;
```
ferdnyc: This should be `parseFloat(centerTime)`, just to protect against any weird string values passed in.

ferdnyc:
So, having tried this out, while it's _awesome_ when moving the playhead off-screen, I really prefer the old behavior of **not** scrolling when the playhead is moving between two locations that are both _on_-screen. Right now, it feels disconcerting (to me, anyway) how, even if two markers are just a few pixels apart, the playhead appears to be _stationary_ while everything _else_ moves underneath it.

Plus, since the playhead is being **animated** between point A and point B, it creates these weird flashes where you'll catch glimpses of the playhead still in motion as the scroll position updates. (The fix for that would be to actually do the scroll update **first**, _then_ animate the playhead to its new location, as unintuitive as that seems.)

#### But as far as animating only-when-necessary:
My PR #2938 branch is now updated with JavaScript utility functions for _both_ `$scope.isTimeVisible()` and `$scope.isPlayheadVisible()`, which should make it easy to wire up conditional, automatic scroll updates as _part of_ the JavaScript, rather than the Python having to make additional separate calls for both actions.

IOW, rather than the Python emitting a `SeekSignal(time)`, then separately calling `centerOnTime(time)` (which I think is like 5 context switches, minimum), the `SeekSignal()` handler on the JS side (which is `MovePlayheadToFrame()`) can gain something like:
```js
// It'll already bail if $scope.playhead_animating is true,
// so after it converts frame# -> position_seconds
if (! $scope.isTimeVisible(position_seconds)) {
  $scope.centerOnTime(position_seconds); 
}
// And _then_ do the
$scope.MovePlayhead(position_seconds);
```
That way, you see the viewport shift, and a split second later the playhead animates into position, same as when you **don't** have to scroll and it just moves across the screen.

If there needs to be an additional state variable put in place, so that it won't update the scroll during playback, that's easy enough to add.

kartchnb:
Thank you again, Frank.  I appreciate the guidance and advice, especially on how to divide the code between JavaScript and Python.  

I’ll work on these changes and commit when they’re done.

Sent from Mail for Windows 10

ferdnyc:
I'm not 100% happy with it yet, but this is my best stab so far at a center-on-playhead icon, in context:

image

(The green arrows started as the ones inside the window outline in Humanity's "leave fullscreen" icon, which I'm now thinking were a mistake.)

ferdnyc:
> especially on how to divide the code between JavaScript and Python.

It's one of the squirrellier things about working in the OpenShot codebase, for sure, and the existing code isn't a perfect template for _how_ to do it either. (In fact, there are plenty of examples scattered around of what **not** to do.) 

It's tempting to, and the existing code often does, treat the two sides as just different sections of the same codebase, with program flow meandering between the two like it's just another function call. But the Timeline is an entire WebKit instance running alongside Qt — it's pretty smooth as long as it can operate mostly-asynchronously, so confining as much as possible to either one side or the other can help keep things responsive.

src/windows/main_window.py
```diff
@@ -2246,6 +2279,7 @@ def setup_toolbars(self):
         self.timelineToolbar.addSeparator()
         self.timelineToolbar.addAction(self.actionAddMarker)
         self.timelineToolbar.addAction(self.actionPreviousMarker)
+        self.timelineToolbar.addAction(self.actionCenterOnPlayhead)
```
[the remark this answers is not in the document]
kartchnb: I agree.  I'll move the "center on playhead" control after the prev and next marker controls.

src/windows/main_window.py
kartchnb: I'm embarrassed to say that I spent 15 minutes figuring out how to implement centerOnPlayhead function because I didn't read your last comment...

I've never worked with JavaScript before.  Should every function have a return statement?  It works without it, but is it good practice to include one?

src/windows/main_window.py
kartchnb: Right.  Being unfamiliar with the code, I was mimicking what I saw elsewhere, but agree there's no point in keeping them in.

src/timeline/js/controllers.js
kartchnb: Thank you.  Done.

kartchnb:
> I'm not 100% happy with it yet, but this is my best stab so far at a center-on-playhead icon, in context:
I like it.  Is there a way for me to get a copy so that I can incorporate it?  If not, I can recreate it.

src/windows/main_window.py
ferdnyc: Only if necessary. TBH the only reason I wrote that with a `return` was that I was sketching it out off the top of my head, and wasn't sure whether `centerOnTime()` itself returned anything. (Obvious in retrospect that it doesn't, but in the moment I think I just shrugged and went, "Well, if it does, then we should pass it along.") So I tossed it in there on the theory that it was safer than possibly losing a return value. :laughing:

src/windows/main_window.py
ferdnyc: \*nod\* That's what gets us. The code is fine in many ways, and passes the pragmatist's litmus test of, "well, it _works_." (And, really, that **is** the ultimate test of quality.) But it also contains a bunch of antipatterns that just keep getting reinforced by (bad) example.

(The bare `except:` clauses all over the Python code fall into that same category. Those aren't just bad, they're **#EEEVIL**. But there's still plenty of them in the code.)

ferdnyc:
> > I'm not 100% happy with it yet, but this is my best stab so far at a center-on-playhead icon, in context:
> 
> I like it. Is there a way for me to get a copy so that I can incorporate it? If not, I can recreate it.

Sure, I just pushed a commit to the branch that adds the SVG file, incorporates it into the resource file, commits an update to the _compiled_ resource file, and makes the necessary changes to `src/windows/ui/main_window.ui`. (IOW, it should now be in place as the toolbar icon.)

ferdnyc:
And, I also just cherry-picked the commit from #2938 that adds the visibility utility functions and pushed that onto this branch as well, so those are available to you.

ferdnyc:
(Just as a note about that, because there are other commits on the branch, when doing updates you're going to want to avoid incorporating them by force-pushing modifications to your existing PR commit. (Which is a common approach used by people with hangups about PRs containing just a single so-called "clean" commit. Even though the committer always has the _option_ to squash the PR branch into a single commit during the merge, anyway, so really all that fretting and history-rewriting is just mountains out of molehills.)

The **ideal** approach would be to pull from GitHub first so you get those changes, then add your own as further new commit(s). But since you probably already _have_ existing local changes that aren't on GitHub yet, better to finish and commit those first, then pull the branch and hope you don't end up having to resolve too much in the way of merge conflicts. The cherry-pick only resulted in a single inconsequential conflict that was easily resolved, so hopefully the same would be true incorporating it into your local tree as well.

You could also `git stash` your local changes instead of checking them in, then do the `git pull` to cleanly incorporate the additional commits. Then a `git stash pop` will restore your own edits to the code, which is the point at which you'd still end up resolving merge conflicts, if there are any.)

src/windows/main_window.py
kartchnb: I believe I've removed all of my debug statements.  I didn't go after any others, though.  Even though I wanted to.

kartchnb:
> My PR #2938 branch is now updated with JavaScript utility functions for _both_ `$scope.isTimeVisible()` and `$scope.isPlayheadVisible()`, which should make it easy to wire up conditional, automatic scroll updates as _part of_ the JavaScript, rather than the Python having to make additional separate calls for both actions.

I think I've got this implemented.  The addition of checking if the playhead is out of the view before recentering is great.  When navigating between markers that are both in the same view, the timeline doesn't shift.  The timeline also doesn't shift when the playhead moves out of view due to the video playing normally.  

To get this behavior, I had to add a variable to main_window.py called centerTimelineAfterSeek that should be set to True before calling SeekSignal.emit() if the timeline should follow the playhead.  It's reset to False in movePlayhead.  This feels like a hack, but I'm not smart enough to figure out how to crack into the SeekSignal functionality to add a second parameter to indicate if the timeline should be moved.  I'm willing to tackle it again if you have any hints for me.

I'm always interested in tips and pointers.  Thanks for all the advice so far!

src/windows/views/timeline_webview.py
```diff
@@ -2537,13 +2537,19 @@ def PlayheadMoved(self, position_frames):
             self.window.previewFrame(position_frames)
 
     @pyqtSlot(int)
-    def movePlayhead(self, position_frames):
+    def movePlayhead(self, position_frames, followPlayhead = False):
```
ferdnyc: This should be `@pyqtSlot(int, bool)` to match the function.

Actually, given that `followPlayhead` is optional, it'd be appropriate to decorate it both ways [("chain" the decorators)](https://www.riverbankcomputing.com/static/Docs/PyQt5/signals_slots.html#the-pyqtslot-decorator):
```python
@pyqtSlot(int)
@pyqtSlot(int, bool)
def movePlayhead(...):
```

(Though, here again, I'm not sure the decoration actually has any purpose.)

src/windows/views/timeline_webview.py
```diff
@@ -2537,13 +2537,19 @@ def PlayheadMoved(self, position_frames):
...
-        code = JS_SCOPE_SELECTOR + ".MovePlayheadToFrame(" + str(position_frames) + ");"
+        code = JS_SCOPE_SELECTOR + ".MovePlayheadToFrame(%s, %s);" % (str(position_frames), "true" if followPlayhead else "false")
         self.eval_js(code)
 
+    @pyqtSlot(float)
+    def centerOnPlayhead(self):
+        """ Center the timeline on the current playhead position """
+        # Execute JavaScript to center the timeline
+        cmd = JS_SCOPE_SELECTOR + '.centerOnPlayhead();';
+        self.page().mainFrame().evaluateJavaScript(cmd)
```
ferdnyc: Curious, is there a reason you don't just use `self.eval_js(cmd)` here? (I know the existing code is a weird mix of using it and not using it, so there's really no right — or wrong — answer to that question.)

ferdnyc:
> PS: I marked a couple of your comments as "resolved", but now I'm thinking I shouldn't have done that. Sorry!

No worries, GitHub would've marked them "Outdated" anyway, when you pushed more changes. But they're really more for your use than mine.

> I think I've got this implemented. The addition of checking if the playhead is out of the view before recentering is great. When navigating between markers that are both in the same view, the timeline doesn't shift. The timeline also doesn't shift when the playhead moves out of view due to the video playing normally.
> 
> To get this behavior, I had to add a variable to main_window.py called centerTimelineAfterSeek that should be set to True before calling SeekSignal.emit() if the timeline should follow the playhead. It's reset to False in movePlayhead. This feels like a hack, but I'm not smart enough to figure out how to crack into the SeekSignal functionality to add a second parameter to indicate if the timeline should be moved. I'm willing to tackle it again if you have any hints for me.

Yeah... that is a _teeny_ bit hackish.

Here's how I'd do it: c09de9d8f56841c2e0175bf76a82b66b499ad22a

Basically, add a state variable in the JS, default _true_, for whether the playhead should be followed. (Because, the only time It _shouldn't_ be is when the previewer is in Play mode). A setter function lets that variable be toggled from the Python, and `movePlayheadToTime()` checks that state variable, instead of an extra arg.

Then, `onModeChange()` in the preview thread — the function that previously did _nothing_ other than log its own name like a simpleton — gets wired up to set the variable `False` when Play mode is activated, and set it `True` otherwise.

The rest is just passing values around from context to context.

Doing it that way, the state is kept where it's actually needed (in the JavaScript), and the signals don't have to be changed (which would be a pain). 

And it'd even be possible to add additional things to `onModeChange()`, like maybe when `openshot.PLAYBACK_STOP` is activated, after it toggles follow mode back on, it also triggers a `centerOnPlayhead()` to find the wayward red line. (I'm not saying add that in, even... just pointing out how wiring it in to `onModeChange()` creates options.)

Though I do have to say, as I was testing that out, at one point I'd accidentally flipped the state so that follow was _enabled_ during playback... and I kind of liked it! Recentering only once per half-width, the impact on performance really seemed pretty negligible. It might even be something to consider making possible, like maybe as an option in the preferences. ...Still, that's totally out of scope here.

src/timeline/js/controllers.js
```diff
@@ -413,6 +438,25 @@ App.controller('TimelineCtrl',function($scope) {
...
+    // Get the width of the timeline
+    var scrollingTracksWidth = $("#scrolling_tracks").width();
+    
+    // Calculate the position to scroll the timeline to to center on the requested time
+    var pixelToCenterOn = parseFloat(centerTime) * $scope.pixelsPerSecond;
+    var scrollPosition = Math.max(pixelToCenterOn - (scrollingTracksWidth / 2.0), 0);
+    
+    // Scroll the timeline using JQuery
+    $("#scrolling_tracks").scrollLeft(Math.round(scrollPosition + 0.5));
+ };
```
ferdnyc: So, I do seem to have encountered one bug, and I _think_ this is the source.

If you have a clip on the timeline that starts right at frame 1 (time index `00:00:00:01`), and you use Prev Marker to jump to the left edge of the clip, the playhead line isn't actually visible, because it's scrolled just _one_ pixel off the left edge of the canvas.

I think that `Math.round(scrollPosition + 0.5)` is causing it to round up to 1, when we want the value to be 0.

ferdnyc: JS has a [`Math.floor()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/floor) as well — if the goal is just to offset the rounding for half-values, that should work. When you add `0.5`, any values >= `x.5` will get pushed over to `x+1`, but everything else will round down to `x`.

src/windows/views/timeline_webview.py
```diff
@@ -2537,13 +2537,19 @@ def PlayheadMoved(self, position_frames):
...
+    def movePlayhead(self, position_frames, followPlayhead = False):
         """ Move the playhead since the position has changed inside OpenShot (probably due to the video player) """
-
         # Get access to timeline scope and set scale to zoom slider value (passed in)
-        code = JS_SCOPE_SELECTOR + ".MovePlayheadToFrame(" + str(position_frames) + ");"
+        code = JS_SCOPE_SELECTOR + ".MovePlayheadToFrame(%s, %s);" % (str(position_frames), "true" if followPlayhead else "false")
         self.eval_js(code)
 
+    @pyqtSlot(float)
+    def centerOnPlayhead(self):
```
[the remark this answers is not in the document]
kartchnb: Yeesh.  That's embarrassing.  Should have caught that.

src/timeline/js/controllers.js
kartchnb: Yes, that's what I *meant* to do.  Math.floor is definitely what I wanted.  Thank you.

Thank you for the patience and explanations.  It's not lost on me that it would have taken less time and effort for you to make these fixes yourself rather than pointing them out and explaining them to me.  Hopefully, this means my next attempt will be better.

src/windows/views/timeline_webview.py
kartchnb: No reason.  I think I originally wrote this function in a different place in the source file and the surrounding code used self.page().mainFrame().evaluateJavaScript and I didn't know any better.  self.eval_js is more concise, so I'll use it.

src/windows/views/timeline_webview.py
kartchnb: Thanks for the tip about chaining decorators.  Although this suggestion is overcome by moving the followPlayhead functionality to controller.js, I appreciate the education.

src/timeline/js/controllers.js
ferdnyc: > It's not lost on me that it would have taken less time and effort for you to make these fixes yourself rather than pointing them out and explaining them to me.

Oh, not at all. Sure, maybe I _could_ write it, but the fact is that I didn't. You did! And we can always use (I'm trying not to say "desperately need", but... yeah, that) more people familiar with and working in the OpenShot code, so getting contributors up to speed is hardly an imposition.

Plus, honestly I can easily see myself making that same round-vs-floor mistake in my own code, and _not_ catching it in testing. It's always easier to take a step back and see things like that when it's someone else's code.

And that's kind of the whole idea behind code review — a new thing for me, but something I'm trying to embrace — really: that a second pair of eyes is helpful no matter who you are or how well you know the code. In fact, it only really works if all of the participants _could_ write the code.

(TBH, I probably wouldn't have thought to abstract out the visibility-detection logic from #2938 into a reusable `isTimeVisible()`, without this PR. And even if I had, I probably wouldn't have _bothered_. The code was **there** in my original PR, sure, but it's definitely better organized as a result of our discussions and brainstorming here.)

ferdnyc:
With the latest changes, I think this is good to go. It looks great to me, anyway, and there are no remaining issues that I've been able to spot.

CC'ing @jonoomph and bouncing him a review request, to get his take.

ferdnyc:
Oh, one final suggestion @kartchnb : This PR needs a better title. The current one _really_ buries the lead, in terms of the playhead-follow / center-on-playhead functionality it **doesn't** at all describe. :laughing:

ferdnyc:
Nevermind, I couldn't handle the severe awesomeness imbalance between this PR and how it was described in its own title, so I took matters into my own hands. :wink:

kartchnb:
> Nevermind, I couldn't handle the severe awesomeness imbalance between this PR and how it was described in its own title, so I took matters into my own hands.

I like it!  Thank you!

src/windows/views/timeline_webview.py
ferdnyc: The key difference, really (as I'm sure you noticed) is that `eval_js()` takes care of setting up a  delay-and-retry queue if the WebView hasn't finished loading yet. So, anything that _might_ be called during setup, before the timeline has finished loading, should always use `eval_js()`.

But that doesn't generally describe an entirely manual action such as `CenterOnPlayhead()` — unless the user is excessively twitchy and impatient, I suppose — so it's definitely a dealer's-choice sort of thing. But, as you say, `eval_js()` is _way_ more concise, and there certainly doesn't seem to be any _harm_ in using it anyway, especially for a user-triggered, infrequently-called action.

ferdnyc:
@kartchnb Apologies. I know it's been over a month since you submitted this, and more than 3 weeks since it was _completed_, and there's still been no action from our end.

@jonoomph is clearly busy with other things currently, and I _really_ want to get this change into OpenShot as it's a huge improvement over the current functionality, so I'm going to exercise my discretion and merge it. There's been ample opportunity for testing and review, and ultimately the most effective testing will be to just get it in front of users via the daily builds. So, let's do that now.

Thanks again for putting this together!

PR #2964, opened 2019-09-01
Fix render progress display in Animated Titles editor

ferdnyc:
Blender 2.80 doesn't output "Part" messages anymore, so the frame regexp was always failing, and the progress bar never updated.

This switches to detecting the final "Sce:" message for the scene, and using it to update the progress bar.

It also cleans up an unused method, and eliminates unnecessary slot methods like
```py
def onRenderFinish(self):
    self.render_finished()
```
in favor of decorating `render_finished` as a slot and connecting the signal directly.

### DIFF
--- a/src/windows/views/blender_listview.py
+++ b/src/windows/views/blender_listview.py
@@ -314,10 +314,6 @@ class BlenderListView(QListView):
         # Enable the Render button again
         self.win.close()
 
-    def close_window(self):
-        # Close window
-        self.close()
-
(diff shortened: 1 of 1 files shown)


ferdnyc:
I'm going to merge this, because I really want to get it into the Daily Builds at least (and as I'm preparing to roll a new Fedora version for RPM Fusion, now that Blender 2.80 is in the Fedora repos).

Issue #2962, opened 2019-09-01
MP4 compression like/with ffmpeg


shadowvzs:
![compare](https://user-images.githubusercontent.com/17709504/64074549-580e7480-ccb5-11e9-8764-582ab22ffb67.png)

Hi, i recorded a video and edited with OpenShot (cut/fade etc) and i **exported (_mp4, 720p, h.264_)** it then i used ffmpeg for compress and the size difference is very huge:

- exported output size: **16.2Gb**
- after ffmpeg: **1.2Gb**

ffmpeg command: `ffmpeg -i vagatlan.mp4 -vcodec h264 -acodec aac output.mp4`
quality loss: _i didn't noticed any quality difference in video, look like the quality is **SAME**_ :)

possible integret a ffmpeg conversion option after the export or compress video like ffmpeg?
since size difference is really huge at 2h+ videos :)

SuslikV:
Look at the _Advanced_ tab of the exported video quality settings. Type `18 crf` or `23 crf` instead of the bitrate.

georg-d:
> Look at the _Advanced_ tab of the exported video quality settings. Type `18 crf` or `23 crf` instead of the bitrate.

I did try with two very different values (e.t. 10 and 35) many spellings (e.g. "crf 35" and "35crf" and "10 crf" and just "35" etc) but file size and visual quality is always the same. At least on my machine, crf setting is not working for all h.264 encoders, see https://github.com/OpenShot/openshot-qt/issues/4002

PR #2967, opened 2019-09-02
Add tooltips for Bit Rate control of the video

SuslikV:
Aimed at issues similar to the: https://github.com/OpenShot/openshot-qt/issues/2962
From UI it is unknown to the end user (since feature was implemented) that other control is available.

Option for the _Advanced_ tab of the video export settings now has tooltip to aware user that he can try to use CRF like setting:

openshot export tooltip bitrate

### DIFF
--- a/src/windows/export.py
+++ b/src/windows/export.py
@@ -265,6 +265,11 @@ class Export(QDialog):
         # Determine the length of the timeline (in frames)
         self.updateFrameRate()
 
+        # Tooltips
+        bitrateTT = _("Type <b>N crf</b><br> where N = 0..63, to use Constant Rate Factor control for quality setting instead of the bitrate")
+        self.lblVideoBitrate.setToolTip(bitrateTT)
+        self.txtVideoBitRate.setToolTip(bitrateTT)
+
     def delayed_fps_callback(self):
         """Callback for fps/profile changed event timer (to delay the timeline mapping so we don't spam libopenshot)"""
         # Stop timer


src/windows/export.py
```diff
@@ -265,6 +265,10 @@ def __init__(self):
         # Determine the length of the timeline (in frames)
         self.updateFrameRate()
 
+        # Tooltips
+        self.lblVideoBitrate.setToolTip(_("Type <b>N crf</b><br> where N = 0..63, to use Constant Rate Factor control for quality setting instead of the bitrate"))
+        self.txtVideoBitRate.setToolTip(_("Type <b>N crf</b><br> where N = 0..63, to use Constant Rate Factor control for quality setting instead of the bitrate"))
```
ferdnyc: If _nothing_ else, please change this to:
```python
msg = _("Incredibly long string I'd be crazy to write twice.")
self.lblVideoBitrate.SetToolTip(msg)
self.txtVideoBitrate.SetToolTip(msg)
```
...Why would you _do_ that to yourself?? I'm sure it was a copy-paste, but still, what if it had to be edited? :laughing:

SuslikV: > ...what if it had to be edited?

Same copy/paste rule would apply.

But OK for convenience I'll reformat this.

SuslikV:
Why it takes so long to be merged?

jonoomph:
Closing PR due to conflicts and lack of updates. Also, I don't think translations will go well for this.

PR #2973, opened 2019-09-03
Don't copy permissions when dup'ing template files (animated titles, still titles)

### LINKED ISSUE #2972: Make sure copied files actually have write permissions in ~/.openshot-qt
openshot-qt copies all kinds of Python or SVG files into a temporary location, i.e.`~/.openshot_qt/blender/8f9c347a-ce4e-11e9-b006-a4c3f0b51071/lens_flare.py`. Now, on the NixOS those files are installed with `0444` permissions. The originals have no `+w` bit -- and consequently the copy lacks that permission and cannot be used properly by the edit. This trivial problem breaks the title generator on NixOS, among other things (https://github.com/NixOS/nixpkgs/issues/48591).

Would you consider adding code into oneshot-qt that simply performs a `chmod u+w` on the newly created temporary file after copying it?

ferdnyc:
A user reported that the `.py` scripts in the blender output directory and the `.svg` title files from the title editor are being created read-only on their system, due to the original templates being installed read-only. This change switches from `shutil.copy()` to `shutil.copyfile()`, which will duplicate the templates _without_ copying their original permissions.

Fixes #2972

### DIFF
--- a/src/windows/title_editor.py
+++ b/src/windows/title_editor.py
@@ -177,7 +177,7 @@ class TitleEditor(QDialog):
         self.filename = os.path.join(info.TITLE_PATH, "temp.svg")
 
         # Copy template to temp file
-        shutil.copy(template_path, self.filename)
+        shutil.copyfile(template_path, self.filename)
 
         # return temp path
         return self.filename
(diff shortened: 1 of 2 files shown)


ferdnyc:
This fix worked for NixOS (where it was originally reported), and it's a trivial change so I'm going to go ahead and merge it.

Issue #2969, opened 2019-09-02
Openshot 2.4.4 incorrectly exports vertical video


sknaumov:
**Describe the bug:**
Vertical video looks very small, with big black frames around all edges. Original video has black frames only from left and right of the video and looks about 2 times bigger.

**Steps to reproduce the behavior:**
1. Create custom profile with the following content:
```
description=Vertical 1080p 30fps
frame_rate_num=30000
frame_rate_den=1001
width=1080
height=1920
progressive=1
sample_aspect_num=1
sample_aspect_den=1
display_aspect_num=9
display_aspect_den=16
```
2. Change OpenShot profile to the custom one, restart OpenShot.
3. Import original video into new project.
4. Export video with default settings.

**Expected behavior:**
Get the same video as original one.

**System Details:**
- OpenShot Version 2.4.4
- Windows 10

**Screenshots:**

Original:
![vlcsnap-2019-09-03-01h03m47s966](https://user-images.githubusercontent.com/1166661/64134464-c9017980-cde6-11e9-8eaf-a7a0374bf124.png)

Exported from OpenShot with default HD 1080p 1920x1080 profile.
![vlcsnap-2019-09-03-01h12m46s894](https://user-images.githubusercontent.com/1166661/64134611-1205fd80-cde8-11e9-88e5-e64bae2acb0e.png)

Exported from OpenShot with custom vertical profile:
![vlcsnap-2019-09-03-01h02m56s763](https://user-images.githubusercontent.com/1166661/64134468-d0c11e00-cde6-11e9-856a-3c134ab3bad5.png)

SuslikV:
Try to save your project file (.osp) and then load it again (use new empty folder for the project's file).

sknaumov:
@SuslikV Just tried it. Unfortunately, it doesn't work, even if empty directory is used to save the project file.

sknaumov:
It looks like I found the root cause: video file that came from my Huawei P30 has 1920x1080 dimensions even though it is oriented vertically. Somehow VLC and other media players still open it properly, but OpenShot can't detect this irregularity. I reencoded the file using `ffmpeg.exe -i Orig.mp4 -vf scale=1080:1920 Out.mp4` and now I see proper video preview and proper exported video.

Could OpenShot automatically detect and apply a workaround for such cases?

SuslikV:
Short file sample is needed to test it out (about 2 seconds is well enough).

sknaumov:
[...]

PR #2975, opened 2019-09-03
Update data from the default project too

SuslikV:
**Edit:**
Inspired by: https://github.com/OpenShot/openshot-qt/issues/2969
It doesn't fixes it completely but applies profile as it was intended during startup of the application.

With the https://github.com/OpenShot/openshot-qt/commit/eae7170494118af6dca5b30d972d80f3f54e1c6b was introduced new `changed(self, action)` method for `VideoWidget` that applies default values of some parameters if data not exist.
This PR ensures that required data will be filled on project load during application startup.

Issues connected to the incorrect aspect ratios of the videos loaded in _Preview File_ (cutting.py) isn't affected by this PR.

### DIFF
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -288,6 +288,8 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
                     self._data["width"] = profile.info.width
                     self._data["height"] = profile.info.height
                     self._data["fps"] = {"num" : profile.info.fps.num, "den" : profile.info.fps.den}
+                    self._data["display_ratio"] = {"num": profile.info.display_ratio.num, "den": profile.info.display_ratio.den}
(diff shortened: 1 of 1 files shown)


src/classes/project_data.py
```diff
@@ -288,6 +288,8 @@ def new(self):
                     self._data["width"] = profile.info.width
                     self._data["height"] = profile.info.height
                     self._data["fps"] = {"num" : profile.info.fps.num, "den" : profile.info.fps.den}
+                    self._data["display_ratio"] = {"num": profile.info.display_ratio.num, "den": profile.info.display_ratio.den}
+                    self._data["pixel_ratio"] = {"num": profile.info.pixel_ratio.num, "den": profile.info.pixel_ratio.den}
```
ferdnyc: This nonsense (manually constructing the dict) actually sent me off on _quite_ the tangent today, and resulted in OpenShot/libopenshot#321. Which may hopefully eliminate the need to do that, eventually. But, for now, it is indeed necessary.

Issue #2247, opened 2018-10-17
v2.4.3: default title name overwrites old ones...


valterfukuoka:
**Describe the bug**
Creating a new title, the default title auto naming is not incrementing from the last number...

**System Details (please complete the following information):**
 - Operating System / Distro: Debian 9 Stretch, x86/64
 - OpenShot Version [e.g. 2.4.3 app image for linux]

**To Reproduce**
1. open new project
2. create 2 new titles
3. close openshot
4. run openshot again, open the project
5. try to add some more titles...

**Expected behavior**
if the last created auto name title is "2", then the new one is expected by be "3"...

**Screenshots**

![overwrite-a](https://user-images.githubusercontent.com/5293980/47060958-805ec300-d20a-11e8-8bf5-34f294505ecb.png)

![overwrite-b](https://user-images.githubusercontent.com/5293980/47060963-86ed3a80-d20a-11e8-8488-0407e731bf89.png)

Thanks,
Valter Fukuoka

DylanC:
@DerGenaue - Can you have a look at this? In #2042 did you cater for this scenario?

DerGenaue:
Nope;
the underlying issue is the one I have described in #2041 :
OpenShot has a temporary working directory for all the assets and another directory right next to the project file.
Upon saving, all relevant assets are moved from the temp directory to the project dir.
This, unfortunately also means that creating new assets will ONLY CHECK IN THE TEMP dir and thus not consider any files in the project dir.

PR #2976, opened 2019-09-05
WIP: Store project data in separate per-project asset folders

### LINKED ISSUE #2961: When you close your project, the default titlename can be overwrite existing files from the project before you closed it
When you save, close your project, the default titlename can be overwrite existing files from the project before you closed it.

I added the timestamp to the default title name to avoid overwriting it by accident.

openshot_qt/windows/title_editor.py

import time

Change Line 214:

self.txtFileName.setText(_("TitleFileName-%d") % i)

to

self.txtFileName.setText(str(int(time.time())) + "-" +
_("TitleFileName-%d") % i)

ferdnyc:
This PR represents a _**MASSIVE**_ overhaul of OpenShot's management of the asset files it creates, to avoid a severe data loss exposure pointed out by user @rohezal in #2961.

The problem in a nutshell (from https://github.com/OpenShot/openshot-qt/issues/2961#issuecomment-528186783):

> Uggggh. OK, now I think I'm beginning to understand. You're talking about the case where you save _multiple_ project files to the same directory, say `/tmp/all_my_OSPs`. 
> 
> If those projects both contain title files with the same filename, e.g. `TitleFileName-1.svg`, then they will both save that title file into the same `/tmp/all_my_OSPs/assets/` directory, and when the second project is saved, **\*bam\***, there goes the title file for your first project. Silently and without warning, no less.
> 
> Yeah, that's bad. That's **very bad**.

The solution proposed and, here, implemented:

> I think, rather than timestamping the SVG filenames, what I'm going to do is take a page from Audacity with its `filename.aup` + `filename_data/...` storage model — except, the OpenShot data dir for `filename.osp` will be called `filename_assets`, to avoid any possible clashes with same-named Audacity projects.
> 
> So, in the new model, two OpenShot projects saved to `/tmp/all_my_OSPs/` would be stored like this:
> 
> * `/tmp/all_my_OSPs/project1.osp`
> * `/tmp/all_my_OSPs/project1_assets/`
>     * `/tmp/all_my_OSPs/project1_assets/TitleFileName1.svg`
>     * `/tmp/all_my_OSPs/project1_assets/thumbnail/`
>         * `/tmp/all_my_OSPs/project1_assets/thumbnail/T455452.png`
> * `/tmp/all_my_OSPs/project2.osp`
> * `/tmp/all_my_OSPs/project2_assets/`
>     * `/tmp/all_my_OSPs/project2_assets/TitleFileName1.svg`
>     * `/tmp/all_my_OSPs/project2_assets/thumbnail/`
>         * `/tmp/all_my_OSPs/project2_assets/thumbnail/TAF78ED.png`
>     * `/tmp/all_my_OSPs/project2_assets/$SOME_BLENDER_UUID/`
>         * `/tmp/all_my_OSPs/project2_assets/$SOME_BLENDER_UUID/animated_title_name.py`
> 
> ...and so on.
> 
> No more separate `thumbnail` and `assets` dirs with no obvious relationship to the project files that generated them, no more commingling and potential collisions for different projects' asset files. I think it'll just be cleaner all around. And loading old projects with the old layout will be no problem, they'll just get stored into the new directory layout the next time they're saved.

This PR (which I've created as a branch in the project repo, so we can get test builds) moves to storing project data in the structure outlined above, based on Audacity's model. A new project file key, `"asset_folder_name"`, is also added and set to the name of the project file's "sidecar" directory.<sup>1</sup> 

The new code will load project files in the older format, and seamlessly convert them on save. (Any asset files stored in the old "sidecar" dirs will be  copied into the new structure, along with the assets currently located in the `$HOME/.openshot_qt` space as usual.)

Project files which have been converted into the _new_ format will load without problems on older OpenShot versions, they don't really care exactly where their asset files are saved to as long as the paths in the project file are correct, and they still are after this change.

The one exception to the above is the disconnected-path recovery for image sequence Clip objects (but not the imported File objects they reference), because as it turns out OpenShot 2.4.4+ have a bug that was preventing Clip path recovery from working properly for any project files, old or new. This PR also contains a fix for that issue.

In the process of reworking the asset storage model, I also massively overhauled the `move_temp_paths_to_project_folder()` function's logic, eliminating a lot of "blind" copying that  was preserving unnecessary files that are never referenced in the project data. (In particular, the thumbnail directory was being copied wholesale with the project file, even if it contained thumbnails for deleted clips or from other projects.) Now, only files which are explicitly referenced by the project data at save time are copied into the project's assets dir.

In addition, rather than effectively going through all of the path-update checks _twice_ over (once for the project files, then again for the clip data), `move_temp_paths...()` now keeps a record of all paths updated when going through the Files data, and uses that to update any Clips which reference the same paths. This makes the process of updating the list of Clips (typically much larger than the list of Files) quicker and far less complex than previously.

**PLEASE** test this code as extensively as you're able, kick the tires hard and hit it with as many non-standard or imperfect project file configurations as you can think of. The whole point of this is to **REDUCE** the risk of data loss vs. the previous code, so it's no good if the new code just breaks in different ways.

To facilitate testing, the entirety of `move_temp_paths_to_project_folder()` is currently **RIDDLED** with debug logging at nearly every step of its execution. Those logs WILL be coming out in a future commit<sup>2</sup> (before the PR is merged, or at least that's my intention). Given the nature of this change and the need for careful testing, I decided not to remove them before submitting the PR as I normally would.

#### Notes & TODOs
* [x] 1: (**NOTE TO SELF**) I just realized I need to make sure there isn't a bug with the name failing to update if a new-style project is resaved under a different name, because I might not have accounted for that.
* [ ] 2: Get rid of all the debug logging before merge.
* [x] 3: Handle asset duplication when new-style projects are loaded and saved under a different name (see first comment below).
* [x] 4: Change Blender output directory names from GUID strings to short IDs, since they no longer really need to be globally unique.
* [ ] 5: Enhance the Title Editor's automated file naming so that, if a project is loaded with asset files in a `projectname_assets` dir, the filenames in that directory are also checked for conflicts when generating new filenames for Title Editor output.

Fixes #2961
Does **NOT** fully address Title Editor naming aspects of #2041 #2004 #2247


ferdnyc:
I just realized the new code has another... well, I'd say _bug_, personally, but at a minimum it's another situation that isn't currently accounted for in the code, and therefore won't be handled optimally:

If you save a project in new-style format, say as `MyProject.osp`, and that project has newly-created assets, they'll go into `MyProject_assets` in whatever directory `MyProject.osp` is saved to.

Say that you then come back to that project at some later date, load it again, create some more assets (generate new titles or another animated title sequence), and then decide to save it somewhere else under the name `UpdatedStuff.osp`. Because the _old_ generated assets are being loaded out of the `MyProject_assets` directory rather than the `$HOME/.openshot_qt/` space, they're **not** considered temp files that need to be migrated.

So, when the new `UpdatedStuff_assets` directory is created, the _newly_-generated assets will be saved into it, but `UpdatedStuff.osp` will still be referencing the contents of `MyProject_assets` for the **OLD** generated assets. That'll _work_ as long as those files are still there, but especially with this new storage model I think it would be preferable to have each project's assets be self-contained.

So, I think I need to add the old `asset_folder_name` as a **third** source of files that can be copied<sup>1</sup> into the project's assets directory at save time, in addition to the application temp space (`$HOME/.openshot_qt/`) and the old-style asset directories.

1. Copied, instead of moved, because when saving a project under a new filename, the _old_ project file still exists, and the user may intend to still use that original project, as well as their new revised version. So, we don't want to do any sort of moving or renaming from the old asset directory to the new one, because that would break the references in the original project file. The new project should take its own copies of any files they share.

jonoomph:
Quick thought: it scares me a bit to use the Filename as part of the directory name (due to length, non-allowed characters in dir names, etc..., renaming projects manually, etc...). But, what about using the Project's "id", which should be a short GUID? 
 - My Unique Crazy Project Name.osp
   - thumbnails/YUV2SXZ6BP/

or 

- /YUV2SXZ6BP/
     - /thumbnails/
     - /blender/
     - /etc/...

jonoomph:
I'll look through this PR with more detail today!

ferdnyc:
Well, part of the goal here is to make things _clearer_ for the user, and give them more control over their own data. So, I guess, let me respond first with an exercise:

Say I have a project file "`Wedding Video.osp`" that I want to copy onto another drive/computer. Which files do I need to take with it, from each of these (abbreviated) directory listings?

This one?
```
$ tree
├── 09fdc32c-cfb2-11e9-85a7-2c27d72bd51f
│   ├── fly_by_1.py
│   ├── TitleFileName0001.png
│   ├── TitleFileName0002.png
|  ...
│   └── TitleFileName0080.png
├── 3fdfc442-cfc9-11e9-9d79-2c27d72bd51f
│   ├── fly_by_1.py
│   ├── TitleFileName0001.png
│   ├── TitleFileName0002.png
|  ...
│   └── TitleFileName0080.png
├── 490ca23a-ccc5-11e9-815d-2c27d72bd51f
│   ├── AnimatedTest0001.png
│   ├── AnimatedTest0002.png
│   ├── AnimatedTest0003.png
|  ...
│   ├── AnimatedTest0080.png
│   ├── fly_by_1.py
│   └── TitleFileName0040.png
├── assets
│   └── TitleFileName-1.svg
├── thumbnail
│   ├── 02AYQQ9L9A.png
│   ├── 65O572FSBX.png
│   ├── AB0FHT4X0Z.png
│   ├── AG7DBAXG5C.png
│   ├── DNT3EIVHZN.png
│   ├── EBCNGJHQJ5.png
│   ├── FVCZHKMZJU.png
│   ├── HFUELRTCR8.png
│   ├── K5HM1J88XT.png
│   └── WBH20FJAQU.png
├── VolumeMixingTests.osp
└── Wedding Video.osp

5 directories, 257 files
```

And, this one?
```
$ tree
├── VolumeMixingTests.osp
├── Wedding Video.osp
├── ZVYZ9U25JP_assets
│   ├── 09fdc32c-cfb2-11e9-85a7-2c27d72bd51f
│   │   ├── fly_by_1.py
│   │   ├── TitleFileName0001.png
│   │   ├── TitleFileName0002.png
|   |  ...
│   │   └── TitleFileName0080.png
│   ├── 490ca23a-ccc5-11e9-815d-2c27d72bd51f
│   │   ├── AnimatedTest0001.png
│   │   ├── AnimatedTest0002.png
|   |  ...
│   │   ├── AnimatedTest0080.png
│   │   ├── fly_by_1.py
│   │   └── TitleFileName0040.png
│   ├── thumbnail
│   │   ├── 65O572FSBX.png
│   │   ├── AB0FHT4X0Z.png
│   │   ├── DNT3EIVHZN.png
│   │   ├── FVCZHKMZJU.png
│   │   ├── HFUELRTCR8.png
│   │   ├── K5HM1J88XT.png
│   │   └── WBH20FJAQU.png
│   └── TitleFileName-1.svg
└── ZYEM3VUG2L_assets
    ├── 3fdfc442-cfc9-11e9-9d79-2c27d72bd51f
    │   ├── fly_by_1.py
    │   ├── TitleFileName0001.png
    │   ├── TitleFileName0002.png
    |  ...
    │   └── TitleFileName0080.png
    ├── thumbnail
    │   ├── 02AYQQ9L9A.png
    │   ├── AG7DBAXG5C.png
    │   └── EBCNGJHQJ5.png
    └── TitleFileName-1.svg

7 directories, 258 files
```

And last, how 'bout this one?
```
$ tree 
.
├── VolumeMixingTests_assets
│   ├── 09fdc32c-cfb2-11e9-85a7-2c27d72bd51f
│   │   ├── fly_by_1.py
│   │   ├── TitleFileName0001.png
│   │   ├── TitleFileName0002.png
|   |  ...
│   │   └── TitleFileName0080.png
│   ├── 490ca23a-ccc5-11e9-815d-2c27d72bd51f
│   │   ├── AnimatedTest0001.png
│   │   ├── AnimatedTest0002.png
|   |  ...
│   │   ├── AnimatedTest0080.png
│   │   ├── fly_by_1.py
│   │   └── TitleFileName0040.png
│   ├── thumbnail
│   │   ├── 65O572FSBX.png
│   │   ├── AB0FHT4X0Z.png
│   │   ├── DNT3EIVHZN.png
│   │   ├── FVCZHKMZJU.png
│   │   ├── HFUELRTCR8.png
│   │   ├── K5HM1J88XT.png
│   │   └── WBH20FJAQU.png
│   └── TitleFileName-1.svg
├── VolumeMixingTests.osp
├── Wedding Video_assets
│   ├── 3fdfc442-cfc9-11e9-9d79-2c27d72bd51f
│   │   ├── fly_by_1.py
│   │   ├── TitleFileName0001.png
│   │   ├── TitleFileName0002.png
|   |   ...
│   │   └── TitleFileName0080.png
│   ├── thumbnail
│   │   ├── 02AYQQ9L9A.png
│   │   ├── AG7DBAXG5C.png
│   │   └── EBCNGJHQJ5.png
│   └── TitleFileName-1.svg
└── Wedding Video.osp

7 directories, 258 files
```

In fact, let's make the second and third even easier.

Number two:
```
$ ls
VolumeMixingTests.osp  Wedding Video.osp  ZVYZ9U25JP_assets  ZYEM3VUG2L_assets
```

Number three:
```
$ ls
VolumeMixingTests_assets  Wedding Video_assets
VolumeMixingTests.osp     Wedding Video.osp
```

Addressing specific concerns:
> Quick thought: it scares me a bit to use the Filename as part of the directory name

I could see that if I was proposing to use the project's free-form Unicode _title_ string, or something. But that's specifically why I started with _filename_, a string that we already know is guaranteed to meet the constraints of the filesystem we're saving to.

> (due to length,

The only length limit of consequence is Windows' 260-character cap in the Win32 API. (And on Windows 10 you can even raise that, now, tho it's still in place by default.)

We're already storing Blender dirs named with GUIDs next to the user's filename, so if they're not hitting it with _that_, then the only way they could possibly hit the cap in **this** setup would be the combination of `filename_assets` directory path, _plus_ Blender GUID dir-name, _plus_ blender filenames.

But with this setup, there's really _absolutely_ no reason why the Blender paths need to be globally unique anymore. So, if we're worried about length, I could see switching the _Blender_ dir names to short IDs, and that'd reclaim enough characters to make this scheme safe when paired with OSP filenames up to 20 characters long, even if they happen to be stored at the very _extremes_ of the 260-character limit.

> non-allowed characters in dir names, etc...,

Are there any characters allowed in filenames, but _not_ in dir names? On any of the three OSes? I'm not personally aware of any. If any do turn up, it should be easy enough to sanitize for them, as it will be a _very_ short list.

> renaming projects manually, 

Well, that's why the asset dir name gets stored in a key in the project file, so it's not tied to the filename past the point where we generate it. (Also means we can modify the scheme at any point in the future, if we need to tweak it.) The project file itself can actually be renamed any way they please, and it won't affect anything. 

They can't rename the assets _directory_ manually, but... why would anyone do that, and if they _do_ then why isn't the correct outcome that it screws up their project file loading and they learn not to do it again? (I mean, they could manually rename the `ZYEM3VUG2L_assets` directory, too, and it would break things just as much.)

I'm all for defensive programming, but I don't think we should unduly inconvenience the average user by engineering on training wheels for saboteurs. They're going to find a way to screw it up no matter what we do, anyway.

Plus, this scheme has been working for Audacity for _decades_ now. I love not reinventing the wheel.

I don't know, I just look at _this_:
```
$ ls
VolumeMixingTests_assets  Wedding Video_assets
VolumeMixingTests.osp     Wedding Video.osp
```

and from a _user_ perspective, in particular, it **feels** really hard to come up with any arguments that trump that. And it feels worth tweaking in whatever ways we have to, so that we can preserve as much of the basic simplicity and bloody obviousness as possible. IMHO.

ferdnyc:
> > (due to length,
> 
> The only length limit of consequence is Windows' 260-character cap in the Win32 API. (And on Windows 10 you can even raise that, now, tho it's still in place by default.)
> 
> We're already storing Blender dirs named with GUIDs next to the user's filename, so if they're not hitting it with _that_, then the only way they could possibly hit the cap in **this** setup would be the combination of `filename_assets` directory path, _plus_ Blender GUID dir-name, _plus_ blender filenames.
> 
> But with this setup, there's really _absolutely_ no reason why the Blender paths need to be globally unique anymore. So, if we're worried about length, I could see switching the _Blender_ dir names to short IDs, and that'd reclaim enough characters to make this scheme safe when paired with OSP filenames up to 20 characters long, even if they happen to be stored at the very _extremes_ of the 260-character limit.

(We _could_ also set a cap on the number of characters taken from the filename, when generating the asset directory name. So, if someone really does name a file `this is my video project made with OpenShot on Thursday September 5 in the year of our lord two thousand and nineteen.osp`, we trim that shit down to just `this is my video pro_assets/`.)

jonoomph:
`this is my video project made with OpenShot on Thursday September 5 in the year of our lord two thousand and nineteen.osp` hehe

jonoomph:
Okay, I like this approach! "key"ing the folder in the OSP project solves most of my concerns. And your argument about filename vs directory name characters is compelling (I'm also not 100% sure, but it's probably okay). I also like the idea of shortening the assets folder slightly (if the filename is crazy long), and I like the idea of shorter Blender folders (possibly even with the name of the blender file at the start, for more clarity). Nice!

jonoomph:
I fixed the Mac builder which had lost WIFI (this now passes all builders). I'm gonna try and test this out locally this evening. Everything looks good to me though.

SuslikV:
While assets at the end looks nice, from the machine's management point of view it is better to place it to the beginning of the filename and make it shorter (only imagine how 30 projects will mess up the folders names, if each will start from different letter and user will have more folders under the same directory). I think, that the `a_` is short and simple, especially if unique id is used.

jonoomph:
We could always rearrange the folder name to:  
 - assets_MyProjectA
 - assets_AnotherProject
 - etc...

ferdnyc:
> While assets at the end looks nice, from the machine's management point of view it is better to place it to the beginning of the filename and make it shorter 

...Why? (On the "place it to the beginning" part, not the length.)

> (only imagine how 30 projects will mess up the folders names, if each will start from different letter and user will have more folders under the same directory). I think, that the `a_` is short and simple, especially if unique id is used.

Well... See, I'm confused because, these dirs aren't really meant to be kept _together_. Nobody ever needs to see "all of my OpenShot asset folders in this directory", all clumped together. That just doesn't seem like a realistic need.

What a user _does_ need to see, realistically, is which asset folders go with which projects. _Starting_ them with the same name makes that much more automatic for the user, and allows easy matching of `.osp` files and their asset folders by visual inspection.

(I admit I'm also biased towards Linux, where directories and files are intermingled both in the shell, and by default in the file manager. None of this "sort folders first" nonsense. In a setup like that, `Wedding Video_assets` and `Wedding Video.osp` end up sorted _right next to each other_. It's also possible to copy them both, with `cp -R Wedding* /somewhere/else/`. Both of those conveniences are lost if the asset folder name is `a_Wedding Video/`.)

ferdnyc:
On this point, BTW:

> Say I have a project file "`Wedding Video.osp`" that I want to copy onto another drive/computer. Which files do I need to take with it, from each of these (abbreviated) directory listings?

FTR, yes I absolutely think that a natural evolution of this change would be an "asset consolidation" feature, which copies **ALL** project assets into the asset folder from wherever they live, making the project completely self-contained and portable. Adding code to do that becomes a relatively trivial thing, after this is in place, and it would provide a solution for **countless** problems users have reported when trying to move/copy project files to other locations or computers. 

(And here again, it's a fairly standard feature: Audacity, Blender, Pitivi<sup>1</sup>, Cinelerra... though weirdly enough not Shotcut, at least not that I can find.)

#### Notes
1. (Pitivi actually doesn't have a _consolidation_ feature, but it has a feature to create a `.tar` archive containing the project file and all assets, for migration purposes.)

ferdnyc:
OK, I believe I've now tackled both 1 and 3 on my PR TODO list: assets will also be migrated from _new_-style asset folders, if they're not in the same location as the one we're currently saving to. So, all project file references should either point inside the current asset folder, or to external disk locations. No more refs to older asset folder locations.

ferdnyc:
Oh, forgot to mention: I also added a length cap, the asset folder name will only use the first 30 characters max from the filename (plus `_assets`). Haven't changed the Blender dirs over to using short IDs instead of GUIDs yet, though — I should add that to the checklist.

SuslikV:
If you stop at names instead of IDs for folders names, then it is OK to have `_a` at the end of the filename. And really, who needs that long word `_assets`? For  what purpose? If you clever enough to remove the folder - you will know what this folder means and that it was made by the program.

jonoomph:
Sorry for the delay on testing this. Looking now, and I'm trying to fix an issue I found with the undo/redo history, where thumbnails break on Clip objects.

jonoomph:
@ferdnyc I'm starting to question the logic of having a thumbnail path included in the JSON. Perhaps, a much better solution would be to always look for thumbnails in the /.openshot_qt/thumbnail/ folder, using the following naming: FILE_ID-FRAME# (i.e. ABD234-1.png, ABD234-200.png). Then, when a thumbnail is requested, we can always check the local /.openshot_qt/thumbnail/ path first, and create a new one if it's missing. On a related note, I'm thinking perhaps we need a local python web server on some port, serving these thumbnail images, or a placeholder if a thumbnail is not yet available. That would automatically queue up missing thumbnails, return placeholder or cached images, and generally make these more resilient, even if thumbnails were deleted or lost.

Just thinking out loud.

jonoomph:
My only concern with this PR is the breakages it causes with undo/redo. I've been distracted on a few other tickets this past week, and haven't had a chance to revisit this one. I still think we need to handle thumbnails differently, and probably using a local HTTP server, which can handle things like "thumbnail doesn't exist", or "thumbnail is already pending", or "return a placeholder image in the meantime", or "server an image from /.openshot_qt/thumbnails/ or /project_folder/assets/ folder", etc... 

Then, with that change, we wouldn't need to include the thumbnail "path" in the JSON at all... since our thumbnail server could handle it more dynamically... that is my current thought. :smile:

ferdnyc:
Apologies, I've been neglecting this myself.

And while I was hoping to look at it today, I'm currently replying on my
phone because my cable is down, so that's out.

Latest text from my provider says "by 6", but the first text at noon said
"by 3" and it's only 2pm now. So, we'll see. But forces willing, I should
be able to get back to this tonight.

ferdnyc:
> Latest text from my provider says "by 6", but the first text at noon said "by 3" and it's only 2pm now. So, we'll see. 

At least they kept their word on that.



> My only concern with this PR is the breakages it causes with undo/redo. 

Mmm, because the elimination of the thumbnail "kitchen-sink" approach means they won't be retained for any items in the history?

Honestly, I _still_ feel it's a bad idea to preserve that history across sessions, and the issue isn't limited to thumbnails. Try this in OpenShot 2.4.4:
1. Create a new project
2. Import two video files, throw them both on the Timeline as Clips
3. _Remove_ one media file from the project, causing its associated clip to be deleted as well.
4. Save the project, and quit OpenShot
5. **Delete** (or rename) one of the two files you'd previously imported into the project
6. Open the project in OpenShot again
7. Hit Undo a couple of times

Congratulations, now you have a video clip on your Timeline that points to a nonexistent file, and OpenShot has _no idea_ there's anything wrong.

ferdnyc:
@jonoomph Looks awesome! Caused some conflicts, though, do you want to resolve or should I give it a try?

jonoomph:
I'm working on it... fixing to merge develop back into this

jonoomph:
@ferdnyc I'm going to make some pretty big changes to move_temp_paths_to_project_folder(), to ignore any thumbnail paths stored in the JSON, and just copy thumbnail files from the temp to new folder. Thumbnail path has become more dynamic, and now includes frame # in the filename (FILEID-FRAMENUMBER.png). I see no reason to store this anymore with our new HTTP approach.  Thus, this method needs some updating to continue working. :smile: Should be done shortly.

ferdnyc:
> @ferdnyc I'm going to make some pretty big changes to move_temp_paths_to_project_folder(), to ignore any thumbnail paths stored in the JSON, and just copy thumbnail files from the temp to new folder. Thumbnail path has become more dynamic, and now includes frame # in the filename (FILEID-FRAMENUMBER.png). I see no reason to store this anymore with our new HTTP approach. Thus, this method needs some updating to continue working.  Should be done shortly.

There is still one reason to store it, possibly: Backwards compatibility. With the code so far, new projects could still be loaded in 2.4.4, old projects could still be loaded by this code. Are we sure we want to throw that away?

ferdnyc:
(If we ARE going to make a breaking change to the file format, I'd love to make some other changes. Like storing file IDs instead of paths in the Clip.Reader, so that we don't have to check/recover the same file path a dozen times when Clips are re-used on the Timeline. That'd probably require some changes to Export, to copy the file paths for libopenshot. ....Or we could change libopenshot instead.)

jonoomph:
@ferdnyc As I'm testing and refactoring some things related to assets folder, I have an idea which might simplify things greatly. It's possible we already discussed and dismissed this idea, but let's quickly chat about it again. :wink: 

**Proposal:** What if we dynamically set the asset path: `{{ProjectName}}_assets`. If we find the path, great, then we use it. If we don't find the path, no prob, we create the folder, and start generating new thumbnails there.

This would allow a user to externally rename a project file and rename the assets folder (`Template` and `Template_assets` could be renamed to `MyNewProject` and `MyNewProject_assets` for example). Then the user would open the renamed project, and all is well. If they only renamed the *.osp, that is fine too, although, thumbnails would be re-generated, and blender files would be missing.  But even that, as an edge case, is easy to rectify. Just copy the contents of your old assets folder into the new one, and restart OpenShot. Nothing permanently broken.

**Problem with current assets implementation**
Currently, we are storing the assets path in the project data. That creates a bunch of strange edge cases (discovered while testing), such as, externally renaming a project, and having a mis-matching assets folder name, or having to add code to detect such a change, and copy assets to a newly named folder, etc... Or, if we don't update the asset path for externally renamed projects, it could result int many projects (i.e. `Template.osp` renamed to `XYZ.osp`) using the original asset folder (`Template_assets`).

**Proposal 2:** We scrap the assets folder approach, and require each *.osp file to be saved in an empty folder, or we auto-create a Project folder, and then save inside that, and leave all the inside folders (thumbnail, blender, etc...) alone... since only the current project would use them. But this is also a bit quirky, and forces the user to create more folders than they probably want to.

Just trying to make sure our end solution is predictable for users, and easy to move projects and assets to other folders, thumb drives, etc... and not end up with shared asset folders, and/or broken paths to assets folders. Not trying to change this for the sake of changing it... that's for sure. But currently, it's a bit quirky and complex to debug... so just trying to simplify a bit. :smile:

jonoomph:
Oh, and I forgot to mention, one other quirk of using an assets folder, is that it changes our relative file paths used in JSON.  From `~/.openshot_qt/thumbnails` (`./thumbnails/...) to `~/Project1_assets/thumbnails/` (`./Project1_assets/thumbnails/`)... which requires changing and resaving all paths in the JSON.

jonoomph:
@ferdnyc Okay, I'm making some really good progress on this. A few more ideas that seem to work great:

1) Stop copying assets back to ~.openshot_qt on project load (this is silly and unneeded)
2) Update info.THUMBNAIL_PATH, info.BLENDER_PATH, and info.TITLE_PATH on project load to point at our project's asset folders... pretty much tells OpenShot to use those folders for an existing project... vs our temp ~.openshot_qt folders.

ferdnyc:
I'm on my phone now, so this won't be very informative, but all of that
sounds great because we should really try to get rid of $HOME/.openshot_qt
completely. It's an outdated path, and I think trying to use it is the
source of our permission problems on Windows 10.

Qt has a set of built-in variables for things like settings and cache
paths, which are automatically set to the approved paths on the various
OSes. On Windows they use the AppData dirs, on Linux everything goes under
$HOME/.config and $HOME/.cache, etc. It'd be a really good idea to start
using those, and would probably solve a lot of our ongoing problems.

ferdnyc:
> Qt has a set of built-in variables for things like settings and cache paths, which are automatically set to the approved paths on the various OSes.

Those locations are part of the [QStandardPaths enum](https://doc.qt.io/qt-5/qstandardpaths.html), and specifically the ones we should be using are (roughly):

|                 Qt path                  |              instead of             |
|:------------------------------------:|:---------------------------------|
| `QStandardPaths::CacheLocation` | `~/.openshot_qt/thumbnails` |
| ⋮ | `~/.openshot_qt/cache` |
| ⋮ | `~/.openshot_qt/preview` |
| ⋮ | `~/.openshot_qt/preview-cache` |
| ————————————— | ———————————— |
| `QStandardPaths::ConfigLocation` | `~/.openshot_qt/openshot.settings` |
| or (Qt 5.5+ only) | `~/.openshot_qt/presets` |
| `QStandardPaths::AppConfigLocation` | `~/.openshot_qt/profiles` |
| ⋮ | `~/.openshot_qt/transitions` |
| ————————————— | ————————————— |
| `QStandardPaths::DataLocation` | `~/.openshot_qt/backup.osp` |
| or (Qt 5.4+ only) | `~/.openshot_qt/openshot-qt.log*` |
| `QStandardPaths::AppLocalDataLocation` | `~/.openshot_qt/libopenshot.log*` |

jonoomph:
@ferdnyc Please take a quick look and test of this code. I've done quite a bit of testing on it, and so far, seems like it's working really well. As mentioned in the commit (which is quite large, sorry about that), here is a quick summary:

`"path": "@assets/title/TitleFileName-3.svg"`
`"path": "@assets/blender/CTOE15R9GN/TitleFileName%04d.png",`

`@assets` is now replaced at runtime, based on `get_assets_path()` static function calculated path. It's always based on the project's filename. Things that this PR currently supports well:

1) New projects backup.osp recovery of unsaved changes (initial project before saving) works well
2) Saving project creates assets folder, copies asset files (thumbnails, titles, blender), and replaces assets folder path with `@assets` placeholder. Also, info.TITLE_PATH, info.BLENDER_PATH, info.THUMBNAIL_PATH are pointed to the project's assets folder. This means all future titles, blender, and thumbnails are created directly in the assets folder (not in ~.openshot_qt)
3) Saving 'Project As' changes the project filename, and creates a copy of the assets folder, to match the new project name. The original project can be deleted, and everything will still work.
4) Loading a project is similar to saving one, the info.TITLE_PATH, info.BLENDER_PATH, info.THUMBNAIL_PATH are pointed to the project's assets folder. No assets are copied. All new assets are created in the assets folder (not in ~.openshot_qt). Any `@assets` placeholders are replaced with the absolute path, based on `get_assets_path()`
5) Externally renaming project.osp & project_assets folder. As long as the name matches, everything will work when the renamed project is opened in OpenShot.
6) Before every auto-save (i.e. before we've actually saved anything), a copy of the *.osp project file is copied into ~/.openshot_qt/recovery/. A new preference exists to set 'how many recovery copies to keep'. Defaults to 30 (so a few MB of files, but should allow for the last 30 auto-saves to be recovered, in case something gets clobbered). File names are UNIXTIMESTAMP-ProjectFilename (i.e. 1573858367-Cloud123.osp). Older files are automatically deleted. To actually use a recovery file though, the user would need to copy it back to the original location, and rename it to the original name. So, a little quirky, but will prevent massive data loss hopefully.

jonoomph:
Looks like I have a thumbnail breakage in our Python unit tests... oops. Looking into it.

jonoomph:
@ferdnyc I think this is about ready to merge into develop. Once this is merged, I'm going to try and merge other `openshot_qt` PR's which don't look too scary, and then branch for release. That is the plan at least. :wink:

jonoomph:
Merging this now. If we find any issues, we can modify in a new PR. :crossed_fingers:

PR #2986, opened 2019-09-09
A bunch of filepath-related fixes

ferdnyc:
* Clean up & organize the paths list in `classes.info`
  - Group the members into app paths, user paths, and files
  - Only attempt to create user paths, not app paths
  - Eliminate `EXPORT_TESTS`
* Create `info.BACKUP_FILE`, instead of generating the filename in 4-5 different places
* Replace _all_ remaining instances of `path.encode('UTF-8')` with `os.fsencode(path)`<sup>1</sup>
* Also replace `json.loads('"%s"' % path, encoding="utf-8")` in json_data.py with `os.fsencode(path)`, since according to the pydoc for `json.loads()`:
   > The ``encoding`` argument is ignored and deprecated.
  
  So, that wasn't actually doing anything useful anyway.

#### Notes
1. **Windows. _Does. Not. Use!_** UTF-8 encoding for its filesystem paths.

   Even if it works 99% of the time (Python protecting us from ourselves), `open(path.encode('UTF-8'))` or `os.path.exists(path.encode('UTF-8'))` is WRONG.
    
    If necessary, (and it probably isn't), `os.fsencode(path)` is the correct cross-platform way to encode a file path in the native filesystem encoding. That will generate the proper encoding on all systems.

   As of [PEP-529](https://www.python.org/dev/peps/pep-0529/) (which only arrived with Python 3.6), Python _will_ use UTF-8 for path _strings_ even on Windows. This is done as a convenience for code authors, since Python `str` strings already use UTF-8 encoding. (That's how Python protects us from ourselves, and why `.encode('UTF-8')` really shouldn't be necessary anyway.) 

    But even PEP-529 recommends `os.fsencode()`, rather than an explicit `.encode('UTF-8')`, as the proper way to explicitly encode path strings for the local OS.


jonoomph:
Looking good! conflicts...

ferdnyc:
Resolved! Travis likes it, so merging...

Issue #2703, opened 2019-03-23
Weird new Advanced view


ferdnyc:
Just asking, is this... _meant_ to be the layout of Advanced View in 2.4.4?

![image](https://user-images.githubusercontent.com/538020/54859729-3ad98100-4ce7-11e9-8a1a-64ce1fcd1c4d.png)

ThomasBerends:
Next to 'Show all' on the left there is a small > button. You can click that to have the project files window back on the regular position.

ferdnyc:
@ThomasBerends 

All that button does ­— all it's _supposed_ to do — is open this menu:

![image](https://user-images.githubusercontent.com/538020/55820026-20b1e800-5ac8-11e9-86ff-dabfd2279049.png)

ferdnyc:
(Regardless, while it's true the user can rearrange the interface however they like, my question was more about the stored default layout.)

DylanC:
> Just asking, is this... meant to be the layout of Advanced View in 2.4.4?

Apparently so. 🤣

ferdnyc:
@DylanC 

Yeah... I think I may submit a PR to try and change that.

What do people think of this? Obviously it's a little comical at the default window size, because the default window size is FAR too small for the OpenShot interface in its current form, especially if 2/3 of it is going to be dominated by the video preview. But with the window made larger, I think it'd be fairly usable.

(I **really** don't like Properties on the right. It puts the value on the _outside_, farther away from the Timeline, with the label in between them. That's just bad ergonomics.)

![Screenshot from 2019-07-18 22-03-24](https://user-images.githubusercontent.com/538020/61504166-67a55880-a9a8-11e9-963f-c1c8343a406c.png)

ferdnyc:
Closing as we're currently at 2.5.1, which has its own, all-new weird layout.

PR #2987, opened 2019-09-09
New Advanced layout

ferdnyc:
See #2703. The 2.4.4 layout is just nonsense.

This PR applies the layout proposed in my comment there:

> What do people think of this? Obviously it's a little comical at the default window size, because the default window size is FAR too small for the OpenShot interface in its current form, especially if 2/3 of it is going to be dominated by the video preview. But with the window made larger, I think it'd be fairly usable.
> 
> (I **really** don't like Properties on the right. It puts the value on the _outside_, farther away from the Timeline, with the label in between them. That's just bad ergonomics.)
> 
> Screenshot from 2019-07-18 22-03-24
> 

It also removes `window_splitter` _and_ `window_splitter2` from the settings file, as neither of them is used for anything.


jonoomph:
Well, I like it! We will soon be revisiting the layouts in the next version of OpenShot, so I like the idea of improving ergonomics in the meantime.

PR #2988, opened 2019-09-10
WIP: Rework Preview File optimizations

SuslikV:
**To do**

- [ ] separate regular _Video Preview_ and _Preview File_ optimizations. Currently (and in this PR too), when _Preview_ (file) window scaled up/down the regular _Video Preview_ content inherits same optimizations (scale). Thus if _Preview_ (file) window resized to almost minimum - any file in _Video Preview_ will be shown at bad quality too.
- [ ] see if aspect ratio/canvas size of the preview file can be separated from the current profile., to render video correctly and without black borders on the sides.

**What is done**
The logic to initialize the Timeline object now uses correct geometry data that itself updates only after the _show()_ method in Qt.

**Side effects**
* _Preview_ (file) window now closes together with the main window (parent->child widget relationship; no new entry in taskbar; when application minimized - all opened previews minimizes too)
* _Split Clip_ (project files menu) is modeless window now (can be changed back to the modal by simple `win.setModal(True)` before `win.show()` if needed).

**Behavior**
Currently it (PR) looks like total regression: quality of _Preview_ (file) drops, aspect ratio and scaling are wrong.

### DIFF
--- a/src/windows/cutting.py
+++ b/src/windows/cutting.py
@@ -59,11 +59,11 @@ class Cutting(QDialog):
     SpeedSignal = pyqtSignal(float)
     StopSignal = pyqtSignal()
 
-    def __init__(self, file=None, preview=False):
+    def __init__(self, parent=None, file=None, preview=False):
         _ = get_app()._tr
 
         # Create dialog class
-        QDialog.__init__(self)
+        QDialog.__init__(self, parent)
 
         # Load UI from designer
         ui_util.load_ui(self, self.ui_path)
--- a/src/windows/cutting.py
+++ b/src/windows/cutting.py
@@ -74,6 +74,15 @@ class Cutting(QDialog):
         # Track metrics
         track_metric_screen("cutting-screen")
 
+        # Disable user controls until file ready
+        self.sliderVideo.setEnabled(False)
+        self.btnPlay.setEnabled(False)
+        self.lblVideoTime.setEnabled(False)
+        self.lblInstructions.setEnabled(False)
+        self.widgetControls.setEnabled(False)
+
+        self.preview = preview
+
         # If preview, hide cutting controls
         if preview:
             self.lblInstructions.setVisible(False)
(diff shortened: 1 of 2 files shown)


ferdnyc:
> * separate regular _Video Preview_ and _Preview File_ optimizations. Currently (and in this PR too), when _Preview_ (file) window scaled up/down the regular _Video Preview_ content inherits same optimizations (scale). Thus if _Preview_ (file) window resized to almost minimum - any file in _Video Preview_ will be shown at bad quality too.

smiley-grinning

You know, as bugs go that's one of the rare ones that's actually more _entertaining_ than it is annoying, at least. **#IAintEvenMad** (Much like #1383, "Dragging a marker adds its icon to the timeline as a clip", until @DylanC had to go and ruin the fun by _fixing_ it.)

SuslikV:
It wasn't to be annoying if you didn't try to read extended logs where timeline updates its max size values at least couple of times and set them at default (upon creation of the widget). I yet don't know how deep the bug was integrated into the code. At least, frame screening works OK and uses same techniques.

ferdnyc:
> It wasn't to be annoying if you didn't try to read extended logs where timeline updates its max size values at least couple of times and set them at default (upon creation of the widget). 

Oh, it's _definitely_ quite annoying. Which makes it all the more impressive that it's **even _more_** entertaining than that! (At leas, it entertains me. I'm a sucker for weird side-effect bugs.)

> I yet don't know how deep the bug was integrated into the code.

I imagine the answer is "deeply", since the cutting/preview window and the preview dock are using the same video widget. (Different _instances_, but same implementation.) Which means they both connect themselves to all of the same signals.

That will manifest as _other_ weird side-effect bugs, too. (It's like christmas!)

In fact, here's one I thought of that would probably be in there as well, and sure enough I just confirmed it with a 2.4.4-dev2 build from last month:
1. Load a project with media in the Project Files and clips on the Timeline
1. Open the Preview File window on one of the Project Files
1.With the preview window still open, right-click on one of the Timeline clips and choose Transform
1. Hit Play in the preview window.

You probably saw this coming, but yes: the transform interface will appear in the preview window. 

(It doesn't appear until you hit Play only because activating Transform doesn't automatically refresh the video widget in the preview window, only the one in the dock. But as soon as the window copy **does** refresh, you can see that **both** video widgets predictably caught the signal to activate Transform.)

Which I guess presents an opportunity for outside-of-the-box thinking about the _role_ of the cutting/preview window. Like, does it really make **sense** to manage this second, separate window, with its own copy of the video widget, at all?

What if the Preview dock (with its _existing_ video widget) were to absorb the functionality and UI of Split Clip / Preview File in their entirety?

1. Preview File could load the selected media into the Preview dock, and hide the normal transport controls in favor of the preview display's slider and time index widgets.
1. Split Clip could _maybe_ do the same thing, though inserting the (fairly large) cutting-functionality UI into the Preview dock probably wouldn't be all that comfortable. However, there might be ways to work around that:
    1. Insert the cutting UI into the Project Files dock instead?
    1. Open a floating window containing _just_ the cutting UI controls, automatically placed directly above/below the Preview dock (above only if it's at the bottom of the screen), but still reuse the Preview dock's video widget
1. Or, maybe when Split Clip gets activated, instead of creating a new window to show the video widget in, the Preview dock gets automatically _floated_, and the cutting UI inserted into it. Then, when the user's done with that mode, it automatically re-docks itself and shifts back to the normal UI.

I'm just brainstorming here, tossing out possibilities. Most of these wouldn't work, or at least would need to be improved before they'd be realistic options to consider. But my point is mostly that we don't have to limit ourselves to "fixing" the cutting/preview window's functionality, as the only way to address its issues. Getting rid of it entirely or replacing it with something better might have even more benefits, in the long run.

Thanks for tackling it, though, I have no doubt there's plenty of room _for_ improvement.

SuslikV:
To be honest, I see no sense in _Preview File_ at all. From few video edits I made with the OpenShot, the only reason I found the option and the window itself - is the issues tracker of the github (where was error described and I didn't knew where people found this, because I do all edits in Timeline and preview my files before import. Also, I'm import clips one by one usually: one import - one Timeline object, then next file, next Timeline object. I simply don't need the _Preview File_ or _Split Clip_ as standalone feature. I'm, personally, prefer timeline's slice and custom player - it is easier. More over the right-click menu over the project files window is useless for me too - I'm prefer to drag files, effects and transitions onto the timeline. Effects tab is last from all is in use for me).

SuslikV:
More over, as soon as this PR changes a lot of code strings, even in case of full success, I don't think that anybody will be brave enough to accept these changes to their's code. Feature works right now, and small improvements wouldn't be game changer. I don't use vertical videos (this was original purpose of the PR), so probably will drop the development in this area.

Is this feature needed?

SuslikV:
I'm going to drop this. _Preview File_ is useless for me.

PR #2994, opened 2019-09-12
Blender Python scripts: Code formatting

ferdnyc:
I noticed that the Python files in `src/blender/scripts` were indented with _tabs_ instead of spaces, and had other weird formatting issues, so I went through each of them and used Atom's Beautify tool (which mostly runs `autopep8` on the file) to fix the formatting, then did a quick visual scan to make sure it didn't do anything too weird.

### DIFF
--- a/src/blender/scripts/blinds.py
+++ b/src/blender/scripts/blinds.py
@@ -1,20 +1,20 @@
-#	OpenShot Video Editor is a program that creates, modifies, and edits video files.
+# OpenShot Video Editor is a program that creates, modifies, and edits video files.
 #   Copyright (C) 2009  Jonathan Thomas
 #
-#	This file is part of OpenShot Video Editor (http://launchpad.net/openshot/).
+# This file is part of OpenShot Video Editor (http://launchpad.net/openshot/).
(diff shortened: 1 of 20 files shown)


jonoomph:
Nice catch! LG

src/blender/scripts/blinds.py
```diff
@@ -1,84 +1,85 @@
...
+    'horizon_color': [0.57, 0.57, 0.57],
+    'resolution_x': 1920,
+    'resolution_y': 1080,
+    'resolution_percentage': 100,
+    'start_frame': 20,
+    'end_frame': 25,
+    'animation': True,
+}
+
+# INJECT_PARAMS_HERE
```
ferdnyc: Whoops! Well, messed this one up. I didn't catch that my reformatting inserted a space here, which breaks this line of the Blender code:

https://github.com/OpenShot/openshot-qt/blob/211eed62a24fd4cc3b5fa55dd3ba02fbed697416/src/windows/views/blender_listview.py#L532-L533

I'll merge a fix in a moment.

PR #2997, opened 2019-09-14
Properties: Ignore overlapping context menu events

### LINKED ISSUE #2996: Properties context menu isn't protected against overlapping requests
@SuslikV provided the following screencast, showing an issue with the context menu for the Source property of the Mask effect. (This is a "reader" type property that has to build a list of all project media _and_ all transitions before it can be displayed, which can take some time.)

_Originally posted by @SuslikV in https://github.com/OpenShot/openshot-qt/issues/2968#issuecomment-531306949_

ferdnyc:
Because "reader" type properties require a fairly lengthy context menu that can take some time to display, this PR adds a lock variable to protect against repeated invocations of the event handler from being queued up during menu construction, and subsequently trampling all over each other. 

All menu requests that come in while the lock is set (i.e. while the menu is being assembled) will be ignored until the menu is displayed, at which point the lock is released. The release mechanism involves the event handler re-invoking itself, which should ensure that any earlier queued invocations occur first. (If not, I may have to turn that into an `emit()` instead of a direct invocation, to place the release call behind any others on the slot's queue.)

A short `QTimer()` is also set whenever this protection is actually triggered, to ensure that the lock is always reset no matter what state the interface gets into.

@SuslikV, could you please test this? And confirm whether it genuinely...
Fixes #2996

#### Update
As I mentioned in the comments below, while I was in the file I did a bunch of housekeeping:

1. The name `property` is a builtin in Python 3, and while it's not a **reserved** word (luckily, as this code wouldn't work at all if it was), using a variable named `property` is still asking for trouble. So, this renames all instances of just `property` to instead be `cur_property` (the currently property being acted upon). Other variables like `property_name`, `property_min`, etc. were left untouched.

2. There were a _ton_ of unused variables in the code, Codacy's static analysis caught them. I made them go away.

### DIFF
--- a/src/windows/views/properties_tableview.py
+++ b/src/windows/views/properties_tableview.py
@@ -63,17 +63,15 @@ class PropertyDelegate(QItemDelegate):
         row = model.itemFromIndex(index).row()
         selected_label = model.item(row, 0)
         selected_value = model.item(row, 1)
-        property = selected_label.data()
+        cur_property = selected_label.data()
 
         # Get min/max values for this property
-        property_name = property[1]["name"]
-        property_type = property[1]["type"]
-        property_max = property[1]["max"]
-        property_min = property[1]["min"]
-        readonly = property[1]["readonly"]
-        keyframe = property[1]["keyframe"]
-        points = property[1]["points"]
-        interpolation = property[1]["interpolation"]
+        property_type = cur_property[1]["type"]
+        property_max = cur_property[1]["max"]
+        property_min = cur_property[1]["min"]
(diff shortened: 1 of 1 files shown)


ferdnyc:
Also added  bunch of "while I was in there" cleanup.

SuslikV:
I can confirm this change solves the issue for me.

ferdnyc:
\o/ Thank you kindly!

ferdnyc:
Going to merge this, since @SuslikV has tested it and reported that it solves the issue it was intended to address.

Issue #2882, opened 2019-07-21
Disabled video isn't really disabled unless Scale is set to "None"


ferdnyc:
While working through some of the older open issues, I came across a feature request for a way to disable video for clips on the timeline and though, "Well... we already have that, don't we?"

It turns out we **_do_**, but there's a bit of a bug in that when I disabled the video on the top clip of two that I had on different tracks of the Timeline, I ended up with this:

![image](https://user-images.githubusercontent.com/538020/61589583-1ea4fe00-ab7a-11e9-879a-e1d3f960ebab.png)

...The workaround there was to **also** set the scaling to "None", at which point the black rectangle disappeared and I was able to see the bottom clip without any interference from the video-disabled upper one.

Now, obviously _disabling_ a clip's video entirely should override all other processing of the clip's video content, so the fact that scaling is even being performed in any mode appears to be a sign that an enable check is missing from some of the video processing code. This may well prove to be a libopenshot bug, but I'm filing it here for now as it can always be transferred later.

_Created (with significant editing) from a comment originally posted by @ferdnyc in https://github.com/OpenShot/openshot-qt/issues/1260#issuecomment-513539243_

NecroWizard666:
@ferdnyc your workaround doesn't fix my issue. I'm using an image overlay on a video track. Switching the scale to "None" on the image clip doesn't make the black center go away ("Enable Video" is set to "-1", setting it to "0" makes the image entirely black).

This issue only started happening after I tried to export my project to another drive, manually reconnecting the project files.
![Screenshot from 2019-09-16 16-17-13](https://user-images.githubusercontent.com/55415486/64999669-812a2a00-d89d-11e9-8a70-c66cefd46536.png)

ferdnyc:
@NecroWizard666 Apologies that it took so long to respond to this.

> @ferdnyc your workaround doesn't fix my issue. I'm using an image overlay on a video track. Switching the scale to "None" on the image clip doesn't make the black center go away ("Enable Video" is set to "-1", setting it to "0" makes the image entirely black).

From what you describe there, the black square _isn't_ that image clip — it's another clip _covering_ the image clip. That's why, when you enable video for the image clip, you can see it covered by the black square, and when you disable its video (the image clip's), the frame is entirely black — the black square is still there, but covering nothing because it's the _only_ thing still there.

[...]

Issue #1060, opened 2017-11-21
Why is the centre blacked out?


CodyMol:
**System Details:**
* Operating System / Distro: ?
* OpenShot Version: ?
* _Please attach log files if crash_

**Issue Description and steps to reproduce:**

Why is the centre of the picture blacked out on certain clips? Aargh!!

![snip](https://user-images.githubusercontent.com/33879755/33096872-6417cc40-cf00-11e7-92a3-616b7c84174f.JPG)

N3WWN:
Can you also post a screenshot of your OpenShot window showing your timeline?  

Preferably, zoomed out to show the entire clip that this is occurring in with the playhead showing the blacked out portion in the preview frame.

CodyMol:
Thanks for asking. Can't do that as I keep binning the clips when it happens. I will save he next one to show you.  Enjoy.  CodyMol.

CodyMol:
Found out it happens when you split the sound, as you play the clips it blacks out the centre?  Trying to paste the screen shot but not having much luck there either.  Enjoy. CodyMol

CodyMol:
Managed a link to the screen shot.  The screen blacks out when you split the sound/pics. Helps!


![image](https://user-images.githubusercontent.com/33879755/33207468-c0500c5a-d105-11e7-9df5-899eef2ac797.png)

DylanC:
@CodyMol - Thanks.

N3WWN:
If you swap the order on the timeline, does the box go away?  If so, that may be a workaround while the bug is being addressed.

Currently you have:
* Track 3 => `GOPR0014.MP4 (all channels)`
* Track 2 => `GOPR0014.MP4`

Does the box remain if you swap them like this:
* Track 3 => `GOPR0014.MP4`
* Track 2 => `GOPR0014.MP4 (all channels)`

Or leave the video where it is on Track 2 and move the audio from Track 3 to Track 1 like this:
* Track 2 => `GOPR0014.MP4`
* Track 1 => `GOPR0014.MP4 (all channels)`

DylanC:
@N3WWN - This workaround is good, thanks. @CodyMol - Marking as a bug. Hopefully we can get this resolved for you.

N3WWN:
Just a guess, but I think the first thing to check is to see if the audio-only clip has the "has video" flag set to true.  If that flag is true, but the reader cannot find a valid frame (because there is none in an audio-only clip), I believe the reader returns a black frame.

[...]

PR #3004, opened 2019-09-17
Set scale=none on audio clips when splitting audio

ferdnyc:
Works around issue #2882, where `has_video = 0` clips show as black square when scale is set to 'best fit'.

(Note that this is only a partial workaround, as it will only affect new split-audio actions, not clips that have already been split, or any other type of `has_video = 0` clips.

Addresses https://github.com/OpenShot/openshot-qt/issues/1060#issuecomment-531983099


ferdnyc:
I'm going to merge this since it's pretty simple, I tested it extensively, and I want to get at least this partial workaround into the daily builds.

PR #3005, opened 2019-09-17
Animated Titles: Don't add files if canceled

ferdnyc:
This fixes a bug where a canceled animated title render would still add its files to the Project Files list.


ferdnyc:
Merging my own code (which I hate), because as with the previous blender fix, I want to get this in for the next Fedora package build.

PR #3006, opened 2019-09-17
Animated Titles: Use QPlainTextEdit for multi-line

ferdnyc:
Qt offers [QPlainTextEdit](https://doc.qt.io/qt-5/qplaintextedit.html) which is supposed to be slightly faster in some situations than QTextEdit, and since we're _always_ explicitly accessing the content as plain text, it just makes sense.

(Also, another "while I was in there": The `QBlenderEvent` class was entirely unused, so I yanked it.)

### DIFF
--- a/src/windows/views/blender_listview.py
+++ b/src/windows/views/blender_listview.py
@@ -128,8 +128,8 @@ class BlenderListView(QListView):
                 self.params[param["name"]] = _(param["default"])
 
                 # create spinner
-                widget = QTextEdit()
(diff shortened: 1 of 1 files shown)


jonoomph:
I could have sworn we use `QBlenderEvent`, lol. It was tricky to get our own thread to sent UI events back to the Blender dialog... which is what this relates to. But perhaps we implemented that a different way in the end, and this became a left over?

jonoomph:
:+1:

PR #3009, opened 2019-09-22
Profile dialog: Remove skip_first, fix loading

ferdnyc:
Since eae7170494118af6dca5b30d972d80f3f54e1c6b the profile dialog hasn't displayed the selected profile information on initial load, as `skip_first` was preventing the initial refresh.

This PR breaks out the update logic into two separate slots, connected to different signals:
- `dropdown_index_changed` takes the `currentIndexChanged()` signal and just refreshes the dialog display, including on initial load
- `dropdown_activated` will receive the `activated()` signal, which is sent only on user interaction (not when the code updates the combo-box), and will perform the actual project data update if the selection has actually changed

Also:
- Removed the unused local Settings handle
- Removed the unused `widget` argument to the slot(s)
- Removed the bare debug-printf logging of the profile-file's full path every time `dropdown_index_changed` is called

### DIFF
--- a/src/windows/profile.py
+++ b/src/windows/profile.py
@@ -27,7 +27,6 @@
 
 import os
 import sys
-import functools
 
 from PyQt5.QtCore import *
 from PyQt5.QtGui import QIcon, QStandardItemModel, QStandardItem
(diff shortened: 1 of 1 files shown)


ferdnyc:
Merging this, because it corrects a really disconcerting issue.

PR #3012, opened 2019-09-22
AppImage builds: Exclude more system libs

### LINKED ISSUE #1775: AppImage build is bundled with outdated libfontconfig
**System Details:**
 - Operating System / Distro: Fedora 28, any recent Linux distro
 - OpenShot Version 2.4.2 from `OpenShot-v2.4.2-x86_64.AppImage`

Adding to the list of libraries that desperately need to be updated in the official builds, running OpenShot from the AppImage results in fontconfig log spew when it picks up the `Ubuntu-R.ttf` font:
```console
         app:INFO Setting font to /tmp/.mount_iw4BV5/usr/bin/images/fonts/Ubuntu-R.ttf
Fontconfig warning: line 5: unknown element "its:rules"
Fontconfig warning: line 6: unknown element "its:translateRule"
Fontconfig error: line 6: invalid attribute 'translate'
Fontconfig error: line 6: invalid attribute 'selector'
Fontconfig error: line 7: invalid attribute 'xmlns:its'
Fontconfig error: line 7: invalid attribute 'version'
Fontconfig warning: line 9: unknown element "description"
Fontconfig warning: "/etc/fonts/conf.d/10-hinting-slight.conf", line 4: unknown element "its:rules"
Fontconfig warning: "/etc/fonts/conf.d/10-hinting-slight.conf", line 5: unknown element "its:translateRule"
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 5: invalid attribute 'translate'
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 5: invalid attribute 'selector'
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 6: invalid attribute 'xmlns:its'
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 6: invalid attribute 'version'
Fontconfig warning: "/etc/fonts/conf.d/10-hinting-slight.conf", line 8: unknown element "description"
Fontconfig warning: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 4: unknown element "its:rules"
Fontconfig warning: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 5: unknown element "its:translateRule"
Fontconfig error: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 5: invalid attribute 'translate'
Fontconfig error: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 5: invalid attribute 'selector'
Fontconfig error: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 6: invalid attribute 'xmlns:its'
```
...and so on.

This is an [extremely](https://bugzilla.redhat.com/show_bug.cgi?id=1566762) well [known](https://bbs.archlinux.org/viewtopic.php?id=235643) fontconfig [issue](https://bugs.chromium.org/p/chromium/issues/detail?id=829890) with apps that bundle their own version, when running on systems that use a _newer_ version of fontconfig. The `/etc/fonts/` files from the latest fontconfig contain XML properties for new features, which are not supported by older versions of the library.

I can't seem to determine what version of fontconfig is bundled with the AppImage (it doesn't appear to have _its own_ version string embedded anywhere within the file!), but based on the output we know it's older than the current 2.13.0 release, which is when the config file format was extended.

Please update the AppImage builds to bundle fontconfig-2.13.0. Systems using older versions will not be impacted, as `libfontconfig.so.1` from fontconfig-2.13.0 can read `/etc/fonts/` files from older versions with no problems.

### LINKED ISSUE #3010: Segfault in _IceWrite
**Describe the bug**
Segfault on Linux, stack trace ending with `_IceWrite`.

Try to start AppImage.

**System Details**
- OpenShot Version: 2.4.4-dev2
- libopenshot Version:  0.2.3
- Platform: Linux-5.0.0-29-generic-x86_64-with-Ubuntu-19.04-disco
- Processor: x86_64
- Machine: x86_64
- Python version: 3.4.3
- Qt5 version: 5.2.1
- PyQt5 version: 5.2.1
- Qt Detected Languages: ['en-US', 'en']
- LANG Environment Variable: en_US.UTF-8
- Daily Build Installed: http://github.com/OpenShot/openshot-qt/releases/download/OpenShot-v2.4.4-dev2-1568687222-f8c180c1-f17a7ed4-x86_64.AppImage

**Log Files**
- http://logs.openshot.org/7vbvcz6wi2fdva9m36u2cnf9ah3py3dn/openshot-qt.zip
- http://logs.openshot.org/7vbvcz6wi2fdva9m36u2cnf9ah3py3dn/libopenshot.zip

**Exception / Stacktrace**
```/usr/lib/x86_64-linux-gnu/libICE.so.6                                          0x7cd8  [0x7f00a60d4cd8]
/usr/lib/x86_64-linux-gnu/libICE.so.6 _IceWrite                                0x48  [0x7f00a60d9558]
/usr/lib/x86_64-linux-gnu/libICE.so.6 IceFlush                                 0x14  [0x7f00a60d9644]
/tmp/.mount_rhaUsz/usr/bin/platforms/libqxcb.so                                          0x4b054  [0x7f00a6998054]
/tmp/.mount_rhaUsz/usr/bin/platforms/libqxcb.so                                          0x4b1ff  [0x7f00a69981ff]
/tmp/.mount_rhaUsz/usr/bin/platforms/libqxcb.so                                          0x4bf21  [0x7f00a6998f21]
/tmp/.mount_rhaUsz/usr/bin/platforms/libqxcb.so                                          0x4cbb5  [0x7f00a6999bb5]
/usr/lib/x86_64-linux-gnu/libSM.so.6 _SmcProcessMessage                       0x8dc  [0x7f00ada5141c]
/usr/lib/x86_64-linux-gnu/libICE.so.6 IceProcessMessages                       0x368  [0x7f00a60dd978]
/tmp/.mount_rhaUsz/usr/bin/libQt5Core.so.5 QMetaObject::activate(QObject*, int, int, void**) 0x7e6  [0x7f00ad0482a6]
/tmp/.mount_rhaUsz/usr/bin/libQt5Core.so.5 QSocketNotifier::activated(int, QSocketNotifier::QPrivateSignal) 0x2e  [0x7f00ad0b457e]
/tmp/.mount_rhaUsz/usr/bin/libQt5Core.so.5 QSocketNotifier::event(QEvent*)          0x59  [0x7f00ad0542a9]
/tmp/.mount_rhaUsz/usr/bin/libQt5Widgets.so.5 QApplicationPrivate::notify_helper(QObject*, QEvent*) 0x8c  [0x7f00a89f6c8c]
/tmp/.mount_rhaUsz/usr/bin/libQt5Widgets.so.5 QApplication::notify(QObject*, QEvent*)  0x3d6  [0x7f00a89fbe56]
/tmp/.mount_rhaUsz/usr/bin/PyQt5.QtWidgets.so sipQApplication::notify(QObject*, QEvent*) 0x86  [0x7f00a93cc426]
/tmp/.mount_rhaUsz/usr/bin/libQt5Core.so.5 QCoreApplication::notifyInternal(QObject*, QEvent*) 0x6d  [0x7f00ad020c2d]
/tmp/.mount_rhaUsz/usr/bin/libQt5Core.so.5                                          0x2b542f  [0x7f00ad06e42f]
/usr/lib/x86_64-linux-gnu/libglib-2.0.so.0 g_main_context_dispatch                  0x2ae  [0x7f00ac31b9ee]
/usr/lib/x86_64-linux-gnu/libglib-2.0.so.0                                          0x4ec88  [0x7f00ac31bc88]
/usr/lib/x86_64-linux-gnu/l...

### PR DESCRIPTION
**Pull request by ferdnyc**
In #3010 a user reported crashing on startup, trying to run the latest Daily Builds under Ubuntu 19.04. 

#### TL;DR

I had the GitLab builders build a standard AppImage from this PR branch, meaning it excluded the `libxcb-*` plugin libs and the font libs (see below). That build fixed the launch crashes in Ubuntu 19.04, for the user in #3010, and also runs fine on my Fedora 30 system. So that's 2 in the win column.

If anyone has a different distro/release they can test it on, though, you can download `OpenShot-v2.4.4-dev2-1569159864-f8c180c1-f17a7ed4-x86_64.AppImage`, which was built from this PR branch, here: https://drive.google.com/open?id=1bqSudb-MITqqNJoJrKZub1WHuN-eoicN

#### Bundled-vs-system library conflicts?

The traceback pointed to a segfault in `libICE.so.6`, which is a system (non-bundled) lib, but it was immediately after calls into `libqxcb.so` which is a _bundled_ lib. My theory was that the system and bundled lib versions were separated by too much time and too many updates to still be fully compatible.

#### `libxcb-` plugin exclusions
I picked apart the dependency tree some, and noticed that while we're bundling `libqxcb.so` (which is a Qt plugin, so that makes sense), and we're excluding `libSM`, `libICE`, and `libxcb` (which are common system libs, so that also makes sense), we're _bundling_ all of the other parts of `libxcb` — that's where I suspect we get into trouble.

For example, a Daily Build from around two months ago:
```console
$ ls /tmp/.mount_1cXWx1/usr/bin/ -R|grep -i xcb
libX11-xcb.so.1
libxcb-dri2.so.0
libxcb-dri3.so.0
libxcb-glx.so.0
libxcb-icccm.so.4
libxcb-image.so.0
libxcb-keysyms.so.1
libxcb-present.so.0
libxcb-randr.so.0
libxcb-render.so.0
libxcb-render-util.so.0
libxcb-shape.so.0
libxcb-shm.so.0
libxcb-sync.so.1
libxcb-util.so.0
libxcb-xfixes.so.0
libxcb-xkb.so.1
libqxcb.so
```

So, I added any library filename that `startswith("libxcb-")` to the exclusion list, since those are all plugins that should really be kept in version sync with the main `libxcb.so.1`.

I didn't exclude `libqxcb.so` of course, and I didn't exclude `libX11-xcb.so.1`. (Though maybe I should've, since we _do_ exclude `libX11.so.6` and I believe they're also distributed together. I just wasn't entirely 100% sure whether Wayland systems would all be guaranteed to have `libX11-xcb.so.1` installed.)

#### font library suite exclusions

I also added these additional libs to the exclusion list:
```
                            "libfreetype.so.6",
                            "libfontconfig.so.1",
                            "libcairo.so.2",
                            "libpango-1.0.so.0",
                            "libpangocairo-1.0.so.0",
                            "libpangoft2-1.0.so.0",
                            "libharfbuzz.so.0",
                            "libthai.so.0",
```
Every library there except the main `libcairo.so.2` is part of an interdependent suite that makes up the Linux desktop font rendering system. (`libthai.so.0` is part of Pango. `libpangoft2` is Pango's FreeType2 interface. Harfbuzz is an OpenType layout implementation. Other related candidates for exclusion, if we were bundling them in the first place: FriBiDi (a bi-directional script handling library), `libpangoxft`, `libpangomm`. But we're already not bundling them.

We've been bundling all of the ones listed above for a long while... I _believe_ unnecessarily. Most of them are on [the pkg2appimage exclusion list](https://github.com/AppImage/pkg2appimage/blob/master/excludelist) (or they're dependencies of libs on that list), and AFAIK should be present on any current linux distro. As long as they're **all** excluded together, shouldn't be any problems, and it'll prevent issues like the startup logspam from that outdated copy of libfontconfig we've been bundling.

`libcairo.so.2` is the only questionable one, but as the comments in that pkg2appimage list note, it's closely tied in to pango and bundling cairo without pango can cause problems. Plus, it should be present on basically any Linux system. `libcairo.so.1` hasn't existed since 2006.

Fixes #3010, fixes #1775


ferdnyc:
The more I think about it, the more I think that if we're excluding _both_ `libX11.so.6` **AND** `libxcb.so.1`, we _should_ also exclude `libX11-xcb.so.1`, which the package on my Fedora system describes as the "libX11/libxcb interoperability library". It just doesn't make sense that we'd bundle any specific version of that, _without_ bundling either of the libraries it makes interoperable. There's way too much danger of version mismatch problems, and it's just nonsensical that a system which contains both `libX11.so.6` and `libxcb.so.1` would be lacking `libX11-xcb.so.1`. I'm going to push in another commit adding that file to the exclusion list.

wilx:
Not that I understand this fully but isn't it possible to take the other extreme approach and bundle in everything except glibc? X11 has to be on wire compatible AFAIK, so that should work, IMHO.

ferdnyc:
> Not that I understand this fully but isn't it possible to take the other extreme approach and bundle in everything except glibc? X11 has to be on wire compatible AFAIK, so that should work, IMHO.

In theory, yes, and in th Flatpak / Snap / etc. model that's basically what's done, packages are pratically run in fully-isolated containers. In _practice_ with AppImages, though, that doesn't work out particularly well, because the executable is still running inside the host system. That implies a certain generation of libraries being used by other processes it has to communicate with — like the display server, window manager, etc. — as well as a system tuned to those libraries.

(The fontconfig thing is a perfect example. Fontconfig released a new version a year or two ago which expanded its config file grammar. When an older version of `libfontconfig.so` is loaded by an application running on a system that was installed with the _newer_ version, the older `libfontconfig` complains mightily about all of the `/etc/fonts.d/` files that contain XML tags/attributes it doesn't know how to parse. ...Fortunately, _all_ it does is complain, and not crash, but you see my point. Unless you're truly running completely isolated from the host system, you're better off using the libraries installed on it, at least the ones you can depend on being available pretty much universally.)

jonoomph:
This looks good to me, but I'm not 100% sure if this is more or less compatible that what we currently use. Have we verified this PR works with the latest version of Distros (at least the few big ones)? If so, I'll happily merge it. :wink:

ferdnyc:
Conflict resolved, so that's the first step.

> This looks good to me, but I'm not 100% sure if this is more or less compatible that what we currently use.

Basically, the way I've come to understand the situation is: Bundling _fewer_ libs is generally MORE compatible (counterintuitively), _**if and only if**_ those libs are "standard equipment" that can be assumed to be present on pretty much any Linux system. (Or any Linux system capable of running the app, at least, since we require a GUI and a desktop session as part of the existing environment, so e.g. a headless server install doesn't count.)

So, the question becomes: Are the libs being excluded really all "standard" libs?

IMHO the answer is a resounding "yes" in most cases, and a fairly safe gamble worth taking on the rest. We've seen the issues that a bundled `libfontconfig` can cause, for instance, and I challenge anyone to find a desktop Linux build anywhere that doesn't already provide its own `libfontconfig`. No-brainer, excluding that one.

The ones I admit to considering questionable, myself, are the edge cases like the `pango` and `cairo` components. I _believe_ those can be treated as "universally available", at this point, but I won't deny there's a certain amount of finger-crossing involved on that.

> Have we verified this PR works with the latest version of Distros (at least the few big ones)?

Ubuntu yes, Fedora yes, but so far that's all. I'd love to have a wider sample, but that requires users of other distros to step up and do the testing. Results from Arch, Mint, Manjaro, MX, and Solus would be great, and they'd really round out the "popular kids" category nicely.

Results from some older CentOS/RHEL, Ubuntu LTS, and Debian versions would be helpful as well, since the intention here is to _remain_ compatible with even older releases/installs, while not messing up newer ones by bundling outdated versions of system libs we could be pulling from the host environment. But I'm less worried about latest-version compatibility (in the sense that I consider it a near-certainty this will improve things on _that_ front) than I am about possibly breaking things for users in less bleeding-edge scenarios.

But, ultimately "Not _extensively_" would be the short answer to this question:
> Have we verified this PR works with the latest version of Distros (at least the few big ones)?

No, this hasn't been tested as widely or as much as I'd _like_ it to be tested. However, anecdotal evidence so far is positive, and it's been tested to the full extent of my own _abilty_ to test it, current.

jonoomph:
Okay, I'm convinced! :stuck_out_tongue_winking_eye: I like your thought on "less is more". If this has widespread breakages, I suppose we could always make a tweak, and do a quick bug-fix release. I might also have time this week to test on a few non-Ubuntu distros in your list above.

ferdnyc:
> Basically, the way I've come to understand the situation is: Bundling _fewer_ libs is generally MORE compatible (counterintuitively), _**if and only if**_ those libs are "standard equipment" that can be assumed to be present on pretty much any Linux system.

I guess an even better way to phrase/conceptualize that, really, is: With an AppImage, the packaged executable is still running in the host environment (not containerized/virtualized). So whenever a library is bundled that _also_ will be present and in use on/by the host system (think `libwayland`, `libfreetype`, or `libglib`, which are going to be loaded from the host system _even if_ they're bundled, because the `Xorg` and `mutter` and `gnome-shell` and etc. processes have already loaded them), the bundling adds another dimension to the matrix of versions/dependencies that all have to be at least nominally compatible between the package and the host system(s). So, each added "system" library creates a set of additional opportunities to encounter some _incompatibility_ with a host environment.

(That's not true for things like, say, the Qt libs. Even if Qt is installed on the host system, it'll just be ignored as long as the necessary components are all correctly bundled. So, it's safe, and desirable,  and the _point_ of AppImages, to bundle that kind of stuff.)

So, what you want to bundle are any and all application libraries, plus any system libraries that _need_ to be included because at least some of the target platforms the app is expected to run in will lack that library. And then you want to stop right there, bundling **only** the minimal set of those libraries, and nothing more. (Because of the whole bundling more == deeper compatibility matrix with more opportunities for breakage thing.)

PR #3015, opened 2019-09-25
Update "bug" label for bug report issue template

ferdnyc:


ferdnyc:
@jonoomph Yep, that's got it.

So, the easiest approach is to just match whatever string GitHub auto-fills into the search field when you click on a label, as the label name for the templates. (In this case, '`:beetle: bug`'.)

PR #3017, opened 2019-09-26
Simplify Unity Launcher logic

### LINKED ISSUE #2992: Ubuntu users: Are Unity Launchers still a thing?
I happened to notice that there's still code in OpenShot to interface with the Unity Launcher system on Ubuntu, for reporting export progress:

https://github.com/OpenShot/openshot-qt/blob/483b719694da7d5d9413f7dab40535f1c505abd5/src/windows/main_window.py#L2393-L2406

As a Fedora user, I'm not really up on the state of Canonical's distro-specific featureset, so I'm just curious: is that still a thing?

I know that Ubuntu moved away from their Unity desktop and made the switch to GNOME a few releases back, so does that mean the code here is outdated / useless, or is there still relevant functionality in the GNOME-based Ubuntu desktop?

ferdnyc:
Not _removing_ the Unity Launcher logic, since #2992 produced no clear info on whether it's still useful, but **greatly** streamlining things (especially in the 99% of cases where it's _not_ used):

- Only try to connect to launcher one time, during `__init__()` of the main window
- Don't even connect slots if there's no launcher (removes the need to receive `ExportFrame` once per frame written, just to ignore it)
- No need for `self.has_launcher`, since we'll never even call the slots if we _don't_ haz.
- We only have one desktop ID now (AppImage uses same), so no checking multiple possibilities
- Move the ID to `info.DESKTOP_ID`, rather than hardcoding it/them

Fixes #2992


ferdnyc:
I'll be honest: I have _no idea_ if this one works.

I failed in my efforts to make both the old code and the new code work, testing in an Ubuntu 18.10 VM. So, they seemed equally non-functional to me! But I'd feel better if someone who actually had a working Launcher progress meter could confirm that it _still _works with this code.

jonoomph:
I think this stopped working when Unity was killed and replaced with GNOME Shell. I believe we can rip it all out, and the only people effected are running old version of Ubuntu with Unity launcher (i.e. a small number of people I believe).

ferdnyc:
@jonoomph I suspect you're right. Though, Gnome-Shell-era Ubuntu _does_ still seem to support display of progress meters on the dock icons (can be observed e.g. when downloading a file in Firefox, at least with my Ubuntu 18.10 test VM's desktop session), it'd be cool for OpenShot to make use of that... however it needs to make use of that.

jonoomph:
@ferdnyc Do you know where we can find documentation on Gnome-Shell "progress bars on icons" support? I did some googling, but failed. :wink:. The only articles I see are related to Unity and Feature requests. I'm probably just not searching the "right" term.

ferdnyc:
I _believe_ it's still 

> The only articles I see are related to Unity and Feature requests. I'm probably just not searching the "right" term.

I'm not sure that _isn't_ still the right stuff. [This article](https://www.omgubuntu.co.uk/2017/09/ubuntu-dock-now-supports-progress-bars-badge-counts) — admittedly more than two years old, now, so (as it disclaims right at the top) it's possible the info is no longer relevant — is about them _implementing_ support for the Unity Launcher API in the post-Unity, Gnome-Shell-driven Ubuntu desktop dock.

It's kind of sounding like nobody ever really developed a replacement, so they're still using the venerable old [Unity Launcher API](https://wiki.ubuntu.com/Unity/LauncherAPI).

I'd love to find recent, positive confirmation of that, but haven't, and it's frustrating because we're in something of a "can't prove a negative" situation. I too have been _trying_ to find documentation on what we should be using _**instead of**_ the outdated Unity Launcher stuff... and I too have been failing... but, it's _possible_ the reason we can't find anything like that is because no such thing exists, and Unity Launcher **is** actually still the supported API.

jonoomph:
@ferdnyc Do you feel this one is ready for a merge (after we resolve conflicts), or should we just close this one? I don't have a strong feeling on it. :smile:

jonoomph:
Merging this one! I had to resolve a few conflicts, and I think I successfully did that! I'll test this one after the merge again!

jonoomph:
@ferdnyc NOTE: I know I'm running an old version of Ubuntu (so maybe ignore this error I get):
```
      logger:ERROR /home/jonathan/apps/openshot-qt-git/src/windows/main_window.py:2790: PyGIWarning: Unity was imported without specifying a version first. Use gi.require_version('Unity', '7.0') before import to ensure that the right version gets loaded.
      logger:ERROR   from gi.repository import Unity
```

ferdnyc:
Aha! I can put that back easily enough. I'm not on a Debian-based system, so I haven't been able to test this at all. (Which was always the concern.)

ferdnyc:
I just pushed a commit to this branch, reverting to the old import code. (But keeping the rest of the changes.) I'll open a new PR.

ferdnyc:
Just merged #3493, should fix that.

PR #3019, opened 2019-09-27
classes.project_data: Exception fixes

ferdnyc:
- Don't translate exception messages (it's [not encouraged](https://stackoverflow.com/a/16106706/200794))
- Avoid slicing of caught exceptions / loss of context:
  - `raise Exception("message" + str(ex))` becomes `raise Exception("message") from ex`
  - `raise Exception(ex)` becomes `raise`
- Raise RuntimeError instead of Exception, for generic failures


jonoomph:
LG

PR #3020, opened 2019-09-27
Fixes to the browser-launching methods in `main_window`

ferdnyc:
- Don't log successes
- Log failures at error level, include exception message text
- Don't log about-window accept/reject
- Correct error message referring to doc package

- Add `new=1` to `webbrowser.open()` calls.
  The irony of the slots that use `webbrowser.open()` (formerly) logging their successes is, it wasn't necessarily telling the truth anyway. I discovered that on **my** Linux system, at least, calling `webbrowser.open(url)` returned `True`, but did absolutely _**nothing**_. Even though I have a browser window open. It just never managed to open the page. Still claimed it did, though!
  Adding `new=1` to the calls made them work for me, and should do no harm on other systems.

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -803,7 +803,7 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
 
     def actionHelpContents_trigger(self, event):
         try:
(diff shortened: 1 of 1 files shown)


jonoomph:
LGTM

PR #3021, opened 2019-09-27
Massive unused/unnecessary variable sweep-up, and more

ferdnyc:
As the branch name indicates, this _started_ as just a set of fixes for one specific issue: the throwaway half of the tuple returned by `QFileDialog.getOpenFileName()`. But, it ballooned into a lot more. All of the changes are centered around the theme of reducing unused-and-unnecessary variables (both to quiet warnings from static analysis, and to streamline the code).

#### Fix `QFileDialog.getOpenFileName()` calls
  - Ignore `file_type` return value, when not used (which is always)
    Apparently the typical pattern for this is:
    ```python
    filename, _ = QFileDialog.getOpenFileName(...)
    ```
    But we can't use that, because we've bound `_` to the translation wrapper. So, I just went with:
    ```python
    filename = QFileDialog.getOpenFileName(...)[0]
    ```
  - simplify some logic around empty return values (canceled dialogs)
  - Create fewer variables we don't need/use

#### Similar cleanup with the use of `os.path` standard library functions

* I went through **EVERY** `.py` file in the repo, and looked at all of the uses of the `os.path` library functions, looking to reduce the number of one-sided `os.path.split()` / `os.path.splitext()` calls and the unused variables they engendered. Meaning...

  ```python
  # If x or y is unused in:
  (x, y) = os.path.split(f)
  # then replace it with whichever is appropriate of:
  x = os.path.dirname(f)
  y = os.path.basename(f)
  
  # If m or n is unused in:
  (m, n) = os.path.splitext(f)
  # then replace it with whichever is appropriate of:
  m = os.path.splitext(f)[0]
  n = os.path.splitext(f)[1]
  ```

* Also, cut down on indentation by reworking some loops like:
    Old...
    ```python
    for x in y:
        if A:
            ...
            if B:
                do_real_work()
    ```

    New...
    ```python
    for x in y:
        if not A:
            continue
        ...
        if not B:
            continue
        do_real_work()
    ```

- Rewrap some really wide / heavily indented lines

#### Clean up `app` vs `get_app()` usage in `main_window.py`
* In functions that call `get_app()` frequently, make sure we've set `app = get_app()` at the top, and replace every remaining instance of `get_app()` with `app`.
* In functions where we _don't_ access `app`/`get_app()` often, eliminate `app` and just use `get_app()` directly.


jonoomph:
Looking good! conflicts though

ferdnyc:
Resolved, but the `final_cut_pro.py` exporter conflict was _huge_, like70% of the entire file. I just blindly took the PR branch version, I'll check now whether that accidentally lost any changes from `develop`.

ferdnyc:
Looks fine, and Travis is happy. Crossing my fingers and merging.

PR #3022, opened 2019-09-27
JavaScript: Fixes to comparisons, iteration

ferdnyc:
As Codacy was kind enough to point out, we should:

### Use === and !== with bool, numbers
We should really be using `===` and `!==`, instead of `==` and `!=` (respectively) when comparing values to Number or Boolean, to ensure that we're not accidentally type-converting values unexpectedly.

### Iterate over arrays with `for...of` loops
`variable[key]` syntax [is evil](https://web.archive.org/web/20150430062816/https://blog.liftsecurity.io/2015/01/15/the-dangers-of-square-bracket-notation), which makes loops of the form `for (var index = 0; index < array.length; index++) { }` inherently problematic, because they're _going_ to access `array[index]` inside the loop.

Fortunately, there's a better way: [`for...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/statements/for...of). Not only is is better/safer, but it often makes the code _so_ much cleaner. So, I've replaced all of the indexed `for()` loops that looped over arrays from `[0 ... length]` with `for (const element of array) { }` loops. (Which is another advantage to `for...of` loops: The iterator variable can be `const`, if the array contents don't need to be modified.)

So `for...of` both cuts down on temporary variables and helps keep everything block-scoped. (`var` has [scoping issues](https://dev.to/johnwolfe820/should-you-never-truly-use-var-bdi). In fact, by changing to `const` iterator variables I already caught one function where a variable was being redeclared inside a large-ish loop. In that particular case it didn't matter, but the fact that `var` allows that to happen silently is a problem.)

As I said, I replaced all of the `for (index = 0; index < array.length; index++)` loops. I left a few indexed `for ()` loops alone, though, like the ones that iterated backwards over the list elements from `length - 1` to `0`. Basically, if it couldn't be trivially converted to `for...of`, I let it be.

### DIFF
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -397,7 +397,7 @@ App.controller('TimelineCtrl',function($scope) {
 
 	  // Determine actual x coordinate (over timeline)
       var center_x = Math.max(cursor_x - track_labels_width, 0);
-      if (cursor_x == 0) {
+      if (cursor_x === 0) {
 		center_x = 0;
       }
 
@@ -977,7 +977,7 @@ $scope.SetTrackLabel = function (label) {
 
 		// Compare position of track to Y param (for unlocked tracks)
 		if (!layer.lock) {
-			if ((top < layer.y && top > bounding_box.track_position) || bounding_box.track_position==0) {
+			if ((top < layer.y && top > bounding_box.track_position) || bounding_box.track_position === 0) {
 				// return first matching layer
 				bounding_box.track_position = layer.y;
 			}
@@ -1205,7 +1205,7 @@ $scope.SetTrackLabel = function (label) {
 	}
 	
 	// no nearby found?
-	if (smallest_diff == 900.0) {
+	if (smallest_diff === 900.0) {
 		smallest_diff = 0.0;
 	}
 	
@@ -1440,7 +1440,7 @@ $scope.SetTrackLabel = function (label) {
 // ############ DEBUG STUFFS ################## //
 
  $scope.ToggleDebug = function() {
-	 if ($scope.debug == true) {
+	 if ($scope.debug === true) {
 		 $scope.debug = false;
 	 }
 	 else {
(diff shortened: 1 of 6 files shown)


src/timeline/js/controllers.js
```diff
@@ -425,12 +421,12 @@ App.controller('TimelineCtrl',function($scope) {
...
-		if ($scope.project.clips[clip_index].id == clip_id) {
+	for (const clip of $scope.project.clips) {
+		if (clip.id == clip_id) {
 			// Set audio data
 			$scope.$apply(function(){
-				$scope.project.clips[clip_index].audio_data = audio_data;
-				$scope.project.clips[clip_index].show_audio = true;
+				clip.audio_data = audio_data;
+				clip.show_audio = true;
 			});
```
ferdnyc: I mentioned that `for...of` often makes the code cleaner: This change makes for a good representative example. The updated version is _so_ much more readable.

jonoomph: :+1:

jonoomph:
LG

jonoomph:
@ferdnyc Well, on actual testing of this PR, it breaks all HTML rendering / JS syntax issues. Maybe a Qt version issue? Not sure, but I've kept some of the PR (the === comparisons), but reverted the loop code.

ferdnyc:
@jonoomph 

Hrmf, that's too bad. I suspect it's a WebKit version issue. Everything worked fine in my local testing, but that's running with a QtWebKit build that's less than two months old, based on presumably a very recent WebKit. (Though I can't seem to find anyplace that actually indicates specifically _what_ version, which is annoying.)

`for...of` [was introduced](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/statements/for...of#Specifications) in ECMAScript 2015 6/ed, and is supported in Chrome starting with version 38. That's fairly recent, and I guess I shouldn't be surprised that not all of our WebKit-based engines are new enough for it to be usable. Oh well.

PR #3025, opened 2019-10-02
New marker icon (purple, SVG); new add-marker toolbar icon; other icon fixes

ferdnyc:
This PR serves two purposes:

### New marker icon
I replaced the venerable old `blue.png` marker icon with a new one, created in Inkscape. It looks like this:

image

There's also a corresponding new "Add Marker" toolbar icon, to replace the nonsense green one:

image

The new marker, when combined with supporting changes in this PR, has several advantages over the old one:
* It's SVG, so it will scale cleanly, even in HiDPI situations
* The width on the timeline is raised from `12px` to `13px`, so the playhead can center over it properly
* The purple color means that the red playhead line won't appear off-center due to subpixel misalignment<sup>**1**</sup>
* It's raised up two pixels higher on the ruler, allowing the cache map to show underneath
* The icon on the ruler has some minor 3D-effect shading, so that multiple markers overlapping in close proximity can be differentiated visually. (The toolbar icon is completely flat, since the shading is unnecessary there.)
* It is _very_ simple, and therefore quick to render. 
   <strike>Here's the _entire_ contents of the `marker.svg` file, which is exported from Inkscape as Optimized SVG:
   ```xml
   <svg width="64mm" height="64mm" version="1.1" viewBox="0 0 64 64"
   xmlns="http://www.w3.org/2000/svg">
    <g transform="translate(0,-233)">
     <path d="m1.8e-6 265 32-32 32 32v32h-64z"
   style="fill-rule:evenodd;fill:#bb48eb;paint-order:normal"/>
    </g>
   </svg>
   ```

   </strike>

   The code is *slightly* more complex since I added the 3D shading, but it's still very simple. The icon is implemented as three overlapping polygons, something even the lowest-end GPUs are capable of rendering at a rate of many hundreds of thousand per second.

Here's how it all looks in practice:

image

#### Notes
1. The subpixel misalignment issue is due to the fact that, on LCD Screens, primary color pixels are never _exactly_ aligned with each other — because the subpixels are placed side-by-side, they are literally out of line. It's the same reason the playhead line always looks slightly misaligned with the playhead itself. (If you haven't noticed it before, now you always will. I'm so sorry.) The playhead itself is blue, while the playhead line is red, meaning that they're always a third of a pixel out of alignment with each other.
   
   Previously, that same issue affected the alignment of the markers (also blue) and the playhead line. But this purple marker activates both red and blue subpixels, so the appearance of misalignment is greatly reduced.

### Clean up custom icons in `images/`
For a while now, we've been placing our custom action icons in `src/images/Humanity/actions/custom/`. I feel this is wrong. Those files are _not_ part of the Humanity theme set, and they shouldn't be mixed in with it. (For one thing, if we ever wanted to update the bundled icons to a newer release of Humanity, or swap it out for a different icon set, it's much easier if we can just overwrite/remove the entire `src/images/Humanity/` directory and all of its contents.)

Well, now we could, because I **believe** I've moved _all_ of our custom additions out of the `Humanity` directory. Mostly they ended up in `src/images/actions/`. The paths in `openshot.qrc` and the `main_window.ui` file's references are updated to match.

I also eliminated the PNG export of the custom razor-tool pointer (`razor_line_with_razor`), tightened up its SVG some (mostly in terms of the coordinates used), and updated the timeline JS code to use the SVG version for its cursor when razor mode is activated.

(NOTE: Images and details updated 2019-11-02 to incorporate latest commits.)


ferdnyc:
Whoop. Well, dropping this back to WIP status, as it seems I forgot about something: The marker image filename is actually _stored_in_ the project file data.

Which means:
1. Old project files won't use the new marker image, or (worse) will have a _mix_ of old and new marker images if they're subsequently edited with a post-merge version of OpenShot
1. Project data saved from a post-merge version of OpenShot will have with broken marker images, if loaded into an older version of OpenShot

The solution, I think, is to just _ignore_ the image filename stored in the JSON `markers` list, but still **set it** to `blue.png` when creating _new_ markers, for backwards-compatibility. So, I need to add a commit to that effect.

ferdnyc:
> The solution, I think, is to just _ignore_ the image filename stored in the JSON `markers` list, but still **set it** to `blue.png` when creating _new_ markers, for backwards-compatibility. So, I need to add a commit to that effect.

That's now done, and I added some shading to the ruler marker icon to make overlapping markers visible. Bringing this back out of WIP status, I believe it's ready to merge..

SuslikV:
The purple color is awful here. Make white outline for the image if you what centered precision for your display device.

SuslikV:
I don't think that the color of the markers require changes.

Next is few words about the playhead image.
>... It's the same reason the playhead line always looks slightly misaligned with the playhead itself. (If you haven't noticed it before, now you always will. I'm so sorry.)

Where you are looking at? Upload your images to GIMP at least and notice that the playhead is drawn this way (optical illusion).

Compare this two:
_original_
play_head

_mirrored_
play_head_mirrored

Did you see the difference? The red dot at the bottom is part of the playhead image, thus misalignment not possible - they both (dot and line) drawn same color.

ferdnyc:
@SuslikV You're right, it's somewhat more complicated than that. Starting with the fact that, although the playhead was drawn with a width of `25px` to make it centerable, the JS code doesn't take that into account. It simply computes the playhead offset as being half the width (should've been half `(width - 1)`), causing it to shift the playhead left by `12.5px`:

image

Which is presumably why the playhead is drawn with the red dot slightly off center.

(But in a way that subpixels still affect. At least on my screen, the upper (original) image looks slightly off center to the right, but the _mirrored_ image DOESN'T really appear to be off center to the left, the way you'd expect. It appears to be pretty much centered.)

But you're correct: the playhead line is aligned with the dot, because the dot is part of the playhead image. I never disputed that.

It's still impossible to have perfectly-aligned red and blue on an LCD screen, though. Like, this will never be a  perfectly straight line:

RedBlueLine

On my screen, which uses R-G-B subpixels, it appears to be a black line with blue to its right, over a red line with black to its right.

ferdnyc:
All that being said, I'm very open to changing the color. It's certainly easy enough, given that it's an SVG file. Just a matter of editing the code and replacing the RRGGBB values:

https://github.com/OpenShot/openshot-qt/blob/094e3aa3825a18a306c61c1a27716df5278a204f/src/timeline/media/images/markers/marker.svg#L3-L5

jonoomph:
I would also vote for a color change. :wink: Perhaps something that doesn't jump out as much. Also, the icon on the toolbar looks very much like a "home" icon, as I've been conditioned to see. Maybe we could make it slightly more narrow? Just some thoughts... but I do like these changes!

jonoomph:
Actually, I don't hate the color at all. But I do think this would look better with a more narrow marker. What do you think? :smile:

SuslikV:
Why it points to up? It wasn't placed above the ruler.
It looks like design of the tag. Why not just triangle, upside down?

TharenaMelishka:
>     1. The subpixel misalignment issue is due to the fact that, on LCD Screens, primary color pixels are never _exactly_ aligned with each other — because the subpixels are placed side-by-side, they are literally out of line. It's the same reason the playhead line always looks slightly misaligned with the playhead itself. (If you haven't noticed it before, now you always will. I'm so sorry.) The playhead itself is blue, while the playhead line is red, meaning that they're always a third of a pixel out of alignment with each other.

Thank you for explaining this, @ferdnyc .  I've always wondered why they never line up properly so I can make clean slices based on other clips on other tracks (also blue vs red).  <insert GI Joe Moment Here>  I know it's a separate issue, but is there a better way to optimize playhead vs clip alignment?  It might not be industry standard, but what about a white or light grey playhead line?  Would that solve both alignment issues at once?

>        Previously, that same issue affected the alignment of the markers (also blue) and the playhead line. But this purple marker activates both red and blue subpixels, so the appearance of misalignment is greatly reduced.

I like the purple.  As a user, I agree with the concept of it being a little narrower, it doesn't need to be such a chunky monkey to be effective.  And I agree with @SuslikV -- Upside down to point at the ruler makes more sense, since you'll be aligning the point with your moment in question.

jonoomph:
@ferdnyc We are going to do a pretty big refresh of the OpenShot UI soon, and I think this PR should probably also be closed, unless it gives us other benefits than just a new icon. If you disagree, please re-open the PR and we'll discuss. Thanks!

PR #3026, opened 2019-10-02
Make onLogTheEnd() a Python atexit function

ferdnyc:
In #2936 @SuslikV added some much-needed bookending to our session logging, which is great to have.

The exit logging made use of a Qt signal `aboutToQuit`, which _seemed_ great at the time — the final log messages would be output when the program was, well, about to quit. Unfortunately, I subsequently discovered that In PyQt, things are a bit more complicated.

When running OpenShot in the interactive Python console (not a typical use case, but a helpful debugging technique I often do make use of), it seems that the `aboutToQuit` signal fires _every_ time the Qt event loop is suspended, including anytime it yields to the console! Literally every keypress in the interactive session would be followed by another repetition of the final end-of-session log messages, which made the console unusable.

Fortunately, Python to the rescue. This PR takes `onLogTheEnd()` and converts it from a Qt slot, into a Python function registered as an [`atexit`](https://docs.python.org/3.4/library/atexit.html) exit handler (via the `@atexit.register` decorator). This means that _Python_ (not Qt) will handle running it at termination of the program. That way, it will only be called _once_ per session, no matter how many times the Qt event loop yields. It'll also be called if Python aborts with a traceback, unlike the Qt slot. (@SuslikV: It even gets called after the unit tests complete.)

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -31,6 +31,7 @@ import os
 import sys
 import platform
 import traceback
+import atexit
 from uuid import uuid4
 from PyQt5.QtWidgets import QApplication, QStyleFactory, QMessageBox
 from PyQt5.QtGui import QPalette, QColor, QFontDatabase, QFont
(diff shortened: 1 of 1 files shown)


ferdnyc:
Merging this, since it only affects logging and should be pretty safe. (And I need my debug console back.)

PR #3031, opened 2019-10-05
freeze.py: Don't hardcode Python extension paths

ferdnyc:
This is a _start_ at making `freeze.py` reusable and more resilient, by removing the hardcoded paths to Python extension modules (PyQt5 components). Instead, each needed module is imported in turn, and `sys.modules.get(module_name).__spec__.origin` is used as the path to the `.so` file.

Also, some indentation is reduced by replacing positive conditionals in the loop with negative conditionals and `continue`, then outdenting the previous contents.

### DIFF
--- a/freeze.py
+++ b/freeze.py
@@ -191,17 +191,20 @@ elif sys.platform == "linux":
     src_files.append((os.path.join(PATH, "installer", "launch-linux.sh"), "launch-linux.sh"))
 
     # Get a list of all openshot.so dependencies (scan these libraries for their dependencies)
+    pyqt5_mod_files = []
+    from importlib import import_module
+    for submod in ['QtWebKit', 'QtSvg', 'QtWebKitWidgets', 'QtWidgets', 'QtCore', 'QtGui', 'QtDBus']:
+        mod_name  = "PyQt5.{}".format(submod)
+        import_module(mod_name)
+        pyqt5_mod_files.append(sys.modules.get(mod_name).__spec__.origin)
+
+    lib_list = [os.path.join(libopenshot_path, "libopenshot.so"),
+                "/usr/local/lib/libresvg.so",
+                "/usr/lib/x86_64-linux-gnu/qt5/plugins/platforms/libqxcb.so"
+                ] + pyqt5_mod_files
+
     import subprocess
-    for library in [os.path.join(libopenshot_path, "libopenshot.so"),
-                    "/usr/lib/python3/dist-packages/PyQt5/QtWebKit.cpython-34m-x86_64-linux-gnu.so",
-                    "/usr/lib/python3/dist-packages/PyQt5/QtSvg.cpython-34m-x86_64-linux-gnu.so",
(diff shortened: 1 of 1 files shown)


ferdnyc:
(BTW, I'm pretty sure two of the paths were typoed, which may explain some of the troubles with freeze.py finding dependencies. Note how the `PyQt5.QtGui` and `PyQt5.QtDBus` extensions have "<code>34<b>d</b>m</code>" in their filenames, instead of "<code>34m</code>":

https://github.com/OpenShot/openshot-qt/blob/7c8925b5c3002cd8cebef77868d43b47bf93242b/freeze.py#L203-L212

It's _possible_ that's correct, but it seems unlikely. Regardless, this new code will protect against any issues there, and allow upgrading Python itself and/or PyQt5 without breaking the freeze script, by retrieving the correct filenames direct from Python.

ferdnyc:
> BTW, I'm pretty sure two of the paths were typoed

(Ooh, or maybe the `34dm` libraries are debug versions? That sounds plausible — though they still shouldn't be the ones included in the packaging, I'd think.)

jonoomph:
As long as this still works, I'm good with it! Hoping this has been tested, but gonna merge and see what happens.

ferdnyc:
> As long as this still works, I'm good with it! Hoping this has been tested, but gonna merge and see what happens.

I'll be honest: changes to `freeze.py` and (_especially_) `installer/build-server.py` often **_aren't_** tested when I submit them, for the simple reason that they currently contain a _ton_ of hardcoded stuff that only applies on the GitLab builder hosts. (I mean, `build-server.py` uses paths in `/home/ubuntu/`, which unsurprisingly enough are not valid on my Fedora system.)

But, at lot of these PRs are attempts at _changing_ that, and making the code more generic so that it _can_ be run in other places. This one is a perfect example, since previously the Python module freezing would fail on any system that didn't use the exact same filenames. (i.e. not even most other _Ubuntu_ installs, where the version numbers would be different.) All of which is to say, the _concept_ behind these changes was tested locally, in developing the new code. But the changes themselves were not tested end-to-end, because I'm still not _able_ to run the full AppImage creation process locally — at all! (A situation I'm slowly working towards improving, I hope.)

I guess what I'm saying is: No, very often these PRs aren't tested, but that's because they modify code that is itself inherently untestable. But a liberal merge-and-revert-if-necessary approach to them is probably the only way that situation's ever going to improve, so I'd encourage taking that approach to them. I'm always happy to fix what I do break, when I break things. ("Move fast and break stuff", wasn't that Facebook's internal motto? Not that I'd want to pattern myself after them, but it's hard to deny they must've been on to _something_...)

Some of my PRs can't ever carry a promise that they won't break anything. (Not an **honest** one, anyway. I _could_ lie and say, "Suuuure, I guarantee the code will work, complete confidence!" I'm just not that guy.) Sometimes, in fact, the whole _point_ of them is to break things. (In the way you sometimes have to re-break a fractured bone, in order to set it properly.) But ultimately, worst-comes-to-worst... that's exactly why version control exists, right? :angel:

PR #3033, opened 2019-10-05
Fix playhead being cut off at left edge of ruler

### LINKED ISSUE #453: Pointer handle tries to hide
The handle to the pointer in the timeline is half hidden at time 00:00:00:01
This is inelegant!
Also the time at the start should be 00:00:00:00    "Shurely shum mishtake?"

ferdnyc:
This makes the ruler container slightly wider than the timeline container, extending `13px` farther left than before (the `ruler_time` display is now only `127px` wide instead of `140px`) — this allows the playhead top to extend into that space. A corresponding `margin-left: 13px` on the playhead-top and the ruler shift everything back into alignment.

image

Fixes #453


ferdnyc:
(Oh, right, and any references to `#progress_container` were removed, because no such element exists in the HTML. The progress display is a `canvas` named `progress`, which is inside the `#scrolling_ruler` container.)

SuslikV:
Now it has other issue that the track start is hardly visible.

ferdnyc:
> Now it has other issue that the track start is hardly visible.

I'm not sure what you mean, can you explain the issue better?

This is a timeline with the middle track positioned at `0.00`, scrolled all the way to the left — everything looks fine to me, but maybe I'm missing something.

image

(Here it is after using <kbd>Ctrl</kbd>+<kbd>Home</kbd>)

image

ferdnyc:
@SuslikV  ?

SuslikV:
Visual perception. At the left when handle half hidden, I'm sure that I it already at the corner (start of the track) and no sense to try to move it to the left any more.

ferdnyc:
@SuslikV Ah, now I understand!

Hmm, I'll think about that. It's a fair point. There's probably something we can do to restore that visual feedback. Maybe even change the playhead handle color, when it's "homed" to the first frame. (Either that,or some sort of oversized "boundary" ruler tick that it visually lines up with.)

ferdnyc:
In light of #652, the way we draw the ruler needs to be rethought anyway, as it turns out. Plenty of opportunity to mess with things while we're in there.

ferdnyc:
I haven't quite tackled @SuslikV 's point about the playhead "docking" (though I do have some idead percolating), but I pushed additional commits to do the following

1. **Prevent the playhead from being dragged left past the timeline edge**

2. **Correct the marker positioning (which was way off)**
    In the process I:
    * Added a container DIV to hold all of the markers (for convenience when correcting the positioning)
    * Removed the hardcoded position offset in favor of appropriate CSS
    * Shifted the markers up `3px` so they no longer obscure the cache progress display

3. **Add styling for the cache progress readout**
    I had to shift it left slightly, so I added CSS for the canvas and applied a negative margin.

    Then I noticed that, with the markers raised up, it was suddenly obvious how overloaded that space was with shades of blue. So I made the color of the progress line styleable, instead of being hardcoded. (The `color` styling of the `#progress` rule will be used as the fill color, when drawing.) After experimenting with a bunch of different colors, I settled on a slightly warm green.

    It's not _so_ green that it looks like a christmas display gone bad when it's next to the red outline of a selected track, it's more visible than the blue when it's next to an _unselected_ track, and I figure green is a more encouraging color for the progress readout anyway.

    image

ferdnyc:
Now I have to decide how much it bothers me that, when the timeline is scrolled right of the origin point, the ruler elements are visible sticking over the left edge of the timeline area by `13px`:

image

ferdnyc:
In other news, @DylanC , @jonoomph — in the process of working on this and the keyframe div stuff in #3039 I worked out where all the craziness around clip sizing, dragging, etc. has been coming from.

To keep their borders from sticking out past the defined dimensions, the `.clip` and `.transition` containers are styled `box-sizing: border-box`. But that means that you can't trust the element's `width` attribute — formulas like this are no longer correct:
```js
var item_bottom = item.position().top + item.height() + vert_scroll_offset;
var item_top = item.position().top + vert_scroll_offset;
var item_left = item.position().left + horz_scroll_offset;
var item_right = item.position().left + horz_scroll_offset + item.width();
```
mostly because that's no longer interchangeable for:
```js
var clip_left_position = clip.position * $scope.pixelsPerSecond;
var clip_right_position = (clip.position + (clip.end - clip.start)) * $scope.pixelsPerSecond;
```

That's why all of the bounding box math was off, why things would resize themselves when they were dragged or snapped, and why this line exists in the code (complete with telltale comment):
https://github.com/OpenShot/openshot-qt/blob/4431bd5c0af4ca705b4e958c193513be749e6bf3/src/timeline/js/functions.js#L291

I think I have a _relatively_ painless way we can remove the `box-sizing: border-box` styling from the clips/transitions. If we insert an _inner_ div, slightly smaller, just inside the main `.clip` container, and set the borders on **that** div instead, then the same effect will be achieved (borders that don't cause the element to grow) without mucking with `box-sizing` and breaking `width` and `height`. 

Then the bounding-box math becomes way simpler, since it can just work with the dimensions of the outer `.clip` div directly, instead of having to "fuzz" the numbers.

ferdnyc:
@SuslikV Well, I came up with a working idea to restore visual indication that the Playhead is positioned at the start of the timeline:

image

That red line over the playhead  handle will only appear at exactly `playhead_position === 0`. As soon as the playhead moves off that spot it disappears.

The positioning, stacking, opacity, etc. can all be changed, but the functionality is the main thrust.

SuslikV:
I think - that item not allowed to move. Try again. First impression is important, believe me.

ferdnyc:
> I think - that item not allowed to move.

Mmm, fair enough. Now that you say it I can definitely see where you're coming from.

I could put the line _behind_ the handle, but my concern is that it won't really be visible enough to serve its purpose.

Maybe my initial instinct to change the color of the playhead **handle** was the correct one after all. It would also be the closest analogue to how things work currently, where the playhead handle's "appearance" (in a manner of speaking) signals when it's at the timeline's starting point.

The handle is a PNG image, so right now we only have one color (and size). For more reasons than just this feature (sharpness on HiDPI screens being another), it would actually be a good thing to replace that static bitmap with a scalable SVG implementation. Then, recoloring it as needed would be child's play.

But I don't feel like diving into that _quite_ yet, so for now I'll just whip up an alternate PNG and swap them using `ng-switch`. Maybe...?

play_head_docked

ferdnyc:
Ohh-kay. Still not 100% sold on the red, but it matches the rest of the line so I'm sticking with it for now. (Though, one idea I did have was to change _both_ the playhead handle AND the playhead line itself to green, when they're at the left edge. That'd make the "docking" even more visible.)

image

image

As you can see, I'm still using the enlarged ruler tick-mark as well, but now it appears behind the handle (as mentioned) _and_ it's white instead of red.

I haven't pushed this yet because I still have one last bug to track down: When a project file _loads_ into an existing OpenShot window, the color doesn't reset to red even if `playhead_position` is 0. Looks like we're loading it from the JSON, but not calling `MovePlayhead()` in the JavaScript (which is what chooses the color of the handle).

I don't really want to have to resort to setting a watch on the playhead position (expensive), so I'll probably just brute-force in a call to `MovePlayhead(project.playhead_position)` after a project file load. That appears to be the only time the position is set _without_ calling `MovePlayhead()`, so far.

ferdnyc:
> I haven't pushed this yet because I still have one last bug to track down: When a project file _loads_ into an existing OpenShot window, the color doesn't reset to red even if `playhead_position` is 0. Looks like we're loading it from the JSON, but not calling `MovePlayhead()` in the JavaScript (which is what chooses the color of the handle).
> 
> I don't really want to have to resort to setting a watch on the playhead position (expensive), so I'll probably just brute-force in a call to `MovePlayhead(project.playhead_position)` after a project file load. That appears to be the only time the position is set _without_ calling `MovePlayhead()`, so far.

Hah! Looks like I need to do that anyway. We have an existing bug I somehow failed to notice until now, where loading a project file in place of a previously-loaded one resets the playhead position to what's stored in the file (which appears to always be `0`, we don't actually store it when saving) — buuut, the time index next to the ruler doesn't get refreshed, so it still displays the previous position from before the load.

SuslikV:
Bumping into red... Maybe it's not so bad.

jonoomph:
I'm a little hesitant on this PR, just because it appears to change some styling and behavior, and more importantly, we will likely be doing a big timeline UI refresh in the next version of OpenShot. So might make sense to table these ideas, and see if they still apply to our new UI ideas. Sorry, but let's pick this back up soon. :pray:

ferdnyc:
Fair 'nuff.

PR #3037, opened 2019-10-10
Remove default-project progress key

ferdnyc:
Project files contain an unused, empty list key `"progress"`, and the only reason it's there is because it's in the default file template. This PR removes it.

### DIFF
--- a/src/settings/_default.project
+++ b/src/settings/_default.project
@@ -66,7 +66,6 @@
     }
   ],
   "markers": [  ],
-  "progress": [  ],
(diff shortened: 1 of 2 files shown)


ferdnyc:
I'm rethinking this. `"progress"` _is_ used internally on the JavaScript side to hold the preview cache status, and while it's a little silly to store it (always empty) in the project files, I'm not sure it's a good idea to just remove it without changing the code to compensate. I'll revisit after some testing.

PR #3038, opened 2019-10-10
Update timeline zoom directly, on project load

ferdnyc:
For reasons that are not _entirely_ clear to me, when a project file is loaded on the command line (or by double-clicking it in the OS file manager), the saved zoom level isn't applied to the timeline. The `update_zoom` slot _should_ be triggered in response to the `setValue()` call on `slider_zoom`, it's just... not. Perhaps we're jumping the gun on it being connected or something?

Regardless, in my testing just calling the JS `setScale()` function directly, immediately after setting the slider value, causes zoom to be set properly even when a project file is loaded at launch.


jonoomph:
LGTM

Issue #689, opened 2017-05-29
Ability to move keyframes forward/back in time within a clip


justarandomgeek:
I'm building animations using lots of images positioned and moved around the screen. However, when I get the timing on a keyframe wrong I have to track to it, write it all down, delete it, track to the new time and recreate it by hand. It would be nice if there was a way to move the keyframes around within the clip to fine-tune animations!

DylanC:
@justarandomgeek - I think this feature exists but I will double check.

sjjh:
@DylanC I was just looking for the same feature and couldn't find it either (openshot 2.4.2). If it does exist, could you add some explanation to the documentation? thx!

peanutbutterandcrackers:
I'm not sure if it does exist. However, this really would be a great addition for sure!

Hopefully we'll get enough volunteers/contributors to have this feature implemented soon enough.

hrimhari:
This is really important. I do fade-ins and outs over a music track based on what's on the screen and if I move my images around it's not easy to manage the volume keyframes.

frederikheld:
I was just looking for this feature as well and I'm surprised that it isn't possible to move keyframes.

The whole keyframe handling is a bit cumbersome to be honest. Selecting a keyframe needs a lot of patience unless there's another way than fiddling around with the red line until it hits the keyframe marker that I can barely see without a microscope :-P Would be nice to have keyframes explicitly displayed on the timeline below each track so that I can select one by clicking it, move it around by dragging it, etc. Similar to [how they did it in Adobe Flash](https://www.youtube.com/watch?v=UPluh8Mre-c).

henning:
I would need that too! 

And also some way of navigating to keyframes to check/change their values, too.

Grigor77:
This would be brilliant! I thought I was missing something having to take screenshots of my properties, deleting the keyframes, then recreating them elsewhere.

naaraanjaa:
For sure it would be a great improvement to keyframe management. I'm moving images in sync with the music, and when I realise that the keyframe is not precisely at the right time, it would feel very natural to just move it and adjust the timing.

BruceFast:
I really believe that keyframes deserves a properties page, or a dock.  You should be able to see a list of all keyframes in the project, you should be able to see/edit all animations that begin at a keyframe, and all animations that end at the keyframe.  You should be able to precisely adjust the keyframe in time.  You should be able to have an animation end on any future keyframe, not just "the next one".

[...]

PR #3039, opened 2019-10-10
Draw keyframe points as styled divs

ferdnyc:
As a first step towards possibly making them more interactive (see #689, others), this does away with the keyframe-point image file entirely, instead displaying keyframes as styled divs. A background gradient supplies the shading. 

- The `keyframePointOffset` variable is also done away with, replaced instead with a negative left margin in the CSS. Some style values are moved from the HTML to the CSS for maintainability and cleanliness.
- Each point is actually a small "display div" inside a larger div container, again planning ahead for possible future interactivity.

A composite image of clip and transition keyframes, using the old image-based code:

OpenShot-image-keyframes

The same timeline, drawn with the code from this PR:

OpenShot-div-keyframes

- As an added bonus, if you uncomment the `point_region:hover` styling in the CSS file, you can have keyframe points that get huge when you hover over them:
   
   Screenshot from 2019-10-10 14-58-32
   
   (Note: it's not really that pronounced; the mouse pointer would be obscuring a big chunk of it.)


jonoomph:
LGTM

PR #3042, opened 2019-10-13
Unify the getDocks searching algorithm

SuslikV:
Just found this in attempt to add new dock (simple curve editor or viewer for the animation keys).

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -1885,7 +1885,8 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
     def removeDocks(self):
         """ Remove all dockable widgets on main screen """
         for dock in self.getDocks():
-            self.removeDockWidget(dock)
+            if self.dockWidgetArea(dock) != Qt.NoDockWidgetArea:
+                self.removeDockWidget(dock)
 
     def addDocks(self, docks, area):
         """ Add all dockable widgets to the same dock area on the main screen """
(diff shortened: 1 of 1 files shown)


ferdnyc:
@SuslikV 

This won't work. Problem is, the Tutorial popup is a dock widget, which shouldn't be included in the `getDocks()` list when other parts of the `main_window.py` code request it. (The couple of functions that iterate over `self.findChildren(QDockWidget)` **should** include the tutorial popup, though, if it also matches the other criteria.)

ferdnyc:
If you're adding a new dock (which sounds awesome, BTW) best thing is to just add it to the `getDocks` list.

SuslikV:
Where?

OpenShot_all_docks_panes

ferdnyc:
Here:

https://github.com/OpenShot/openshot-qt/blob/4431bd5c0af4ca705b4e958c193513be749e6bf3/src/windows/ui/main-window.ui#L387-L473

It's disabled+hidden except when in use, to keep it off that list.

(Though I assume that, when the tutorial _IS_ visible, it would appear.)

SuslikV:
I only found that the tutorial window hides when switching from Simple UI to Advanced with this change. I will revert my changes.

Anyway, I wish to modify comment that describes the _getDocks()_
from:
`""" Get a list of all dockable widgets """`
to:
`""" Get a list of some dockable widgets """`
because the Tutorial dock lies in Qt::NoDockWidgetArea (not main window) and thus isn't listed on the screenshot.

SuslikV:
It should work as intended. So updated code again to exclude Tutorial from the manipulations.

I want to add dock that lies in area of main window but not available until user's explicitly request.

SuslikV:
The mentioned curve editor tool: https://github.com/OpenShot/openshot-qt/pull/3084

SuslikV:
And it's not "work in progress" but "complete".

jonoomph:
LGTM!

Issue #2114, opened 2018-09-14
Cannot Import MP4s? (RESOLVED)


DecalsIAm:
Doesn't matter the quality of the video, if it is an mp4 video, it will not import. Did I miss something in the program description that it doesn't work for mp4s?

Windows 10
OpenShot Version 2.4.2

DylanC:
@DecalsIAm - Try v2.4.3 and let us know if you still have an issue.

DylanC:
@jonoomph @peanutbutterandcrackers @ferdnyc - This seems to be the first instance of import issues regarding mp3/mp4. I have linked all of the dups to this original one and updated the priority list I have here:
https://gist.github.com/DylanC/73ecfa2f6f99fd7828604e4f8b65bb3f

DecalsIAm:
I don’t use the portable app.

Pedals2Paddles:
First time using Open Shot.  Imported some mov files from an action camera, which worked fine.  Then imported two MP4 files from my phone, which has rendered the application totally unusable. it literally broke the app and cannot be recovered. The application crashed upon trying to play the files back.  And you can't even reopen the app because it gets stuck in some backup state.  Presumably, I have to delete application and start over.  So this is a non-starter for me and will have to use a different application.

corelgott:
I'm also seeings openshot crash when trying to import a mp4 on OSX (High Sierra)

oisinmulvihill:
First time user on OSX High Sierra (10.13.6) attemting to use OpenShot v2.4.4. Any attempt to import (drag-n-drop or import from menu) gopro generated mp4 files results in immediate crash. I can import images no problem. I tried converting the file with Quicktime to MOV format however this resulted in a crash as well :(

Its a shame as OpenShot looks like it has everything I need to move away from iMovie and much more.

SuslikV:
@oisinmulvihill can you add your openshot log-file? Can you import images from the same location, with the same filename? What if you try to import some .mp3 files?

oisinmulvihill:
I've tried openshot v2.4.4 on a new macbook air running OSX Mojave (10.14.3). It works out of the box with the same gopro created mp4s.

On my OSX 10.13 machine, I can import images from the same directory. I haven't tried importing if they have the same filename. I try this evening along with some mp3s and see what happens.

I had a look at the osx logging console to see if I could catch errors/logs but nothing came up. Where do I find the openshot log?

oisinmulvihill:
[...]

PR #3045, opened 2019-10-18
CONTRIBUTING: Correct filename of openshot logfile

ferdnyc:
@SuslikV mentioned way back in https://github.com/OpenShot/openshot-qt/issues/2114#issuecomment-487891990 that the filename was wrong (`openshot_qt.log` instead of `openshot-qt.log`), but nobody ever corrected it.


ferdnyc:
Heh, yep. (Thanks for commenting, BTW, I completely forgot to merge this!)

ferdnyc:
@SuslikV Actually, I may have misspoken — `CONTRIBUTING.md` _isn't_ part of the site (if you meant openshot.org), the manual has its own copy of the same content — which is also stored in the repo, in the `docs/` dir. And since that copy contained the same error, #3046 fixes it there. (Though it won't go live on the website until the next time @jonoomph regenerates the manual HTML.)

PR #3052, opened 2019-10-28
Stop messing with users' filename formatting

ferdnyc:
In a couple of places:
* The code to set the main window title
* The Export dialog generating its initial base export filename

we were processing the user's assigned project file name, replacing `_` and `-` characters with spaces and altering the capitalization. That's simply rude, and there's no reason to mess with the user's file naming.

If they want spaces in the name, they'll insert them. If they want their project file to be named `MY_OPENSHOT_PROJECT.osp`, then we should default to `MY_OPENSHOT_PROJECT.mp4` as the export filename.


jonoomph:
Agreed! LG!

PR #3070, opened 2019-11-09
Prevent accidental properties edit by mouse move

SuslikV:
Set threshold for the first mouse move before change the values of properties and lock the row selection.

It pisses me off when I clicking properties and every single parameter tries to change its value. Also, it is not a funny when you drag mouse over the vertical list of properties and everything is changing.

I have plans to re-use the property selection to display some stuff related to the property value in time. So it had to work smooth and robust.

### DIFF
--- a/src/windows/views/properties_tableview.py
+++ b/src/windows/views/properties_tableview.py
@@ -227,10 +227,21 @@ class PropertiesTableView(QTableView):
 
                     # Huge range - increment / decrement slowly
                     if self.previous_x == -1:
-                        # init previous_x for the first time
+                        # Init previous_x for the first time
                         self.previous_x = event.x()
-                    # calculate # of pixels dragged
+                        self.diff_length = 10
+                    # Calculate # of pixels dragged
                     drag_diff = self.previous_x - event.x()
+
+                    if abs(drag_diff) < self.diff_length:
+                        # If threshold exceed then allow any small incriment for the next time
+                        # In few steps, because the first click may has any far position
+                        self.diff_length -= 1
+                        if self.diff_length < 0:
+                            self.diff_length = 0
+                        self.previous_x = event.x()
+                        return
+
                     if drag_diff > 0:
                         # Move to the left by a small amount
                         self.new_value -= 0.50
@@ -240,8 +251,21 @@ class PropertiesTableView(QTableView):
                     # update previous x
                     self.previous_x = event.x()
                 else:
+                    if self.previous_x == -1:
+                        # Init previous_x for the first time
+                        self.previous_x = event.x()
+                        self.diff_length = 10
+                    if abs(self.previous_x - event.x()) < self.diff_length:
+                        # If threshold exceeded then allow any small incriment for the next time
+                        # In few steps, because the first click may has any far position
+                        self.diff_length -= 1
+                        if self.diff_length < 0:
+                            self.diff_length = 0
+                        self.previous_x = event.x()
+                        return
                     # Small range - use cursor % to calculate new value
                     self.new_value = property_min + (min_max_range * cursor_value_percent)
+                    self.previous_x = event.x()
 
                 # Clamp value between min and max (just incase user drags too big)
                 self.new_value = max(property_min, self.new_value)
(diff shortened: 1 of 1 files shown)


src/windows/views/properties_tableview.py
```diff
@@ -152,8 +152,18 @@ class PropertiesTableView(QTableView):
...
+
+        # Do not change selected row during mouse move
+        if self.lock_selection:
+            row = self.prev_row
+        else:
+            row = self.indexAt(event.pos()).row()
+            self.prev_row = row
+            self.lock_selection = True
+
+        if row is None:
```
jonoomph: Looks like `row` could be undefined at this point. Probably still need a declaration outside these `if/else` statements

jonoomph:
Looking good! I like the concept!

jonoomph:
I'm gonna go ahead and merge this and test it. If we need to fix anything, we'll do it in a new PR. Thx @SuslikV!

src/windows/views/properties_tableview.py
SuslikV: I googled a bit and it seems that the code like this is common in Python.

The `None` check (the later one) was left from the debugging and not a big deal right now. Feel free to remove it. I placed it just as reminder for myself when code wasn't complete yet.

By the way, the `self.currentIndex().row()` instead of the `indexAt` - works OK too with this QTableView.

ferdnyc:
I'd meant to comment on this one, it's a great fix and that mouse-dragging 'drift' thing was **_SO_** incredibly frustrating. Huge improvement having that fixed. Thanks @SuslikV !

src/windows/views/properties_tableview.py
```diff
@@ -230,8 +251,21 @@ def mouseMoveEvent(self, event):
...
+                        self.previous_x = event.x()
+                        self.diff_length = 10
+                    if abs(self.previous_x - event.x()) < self.diff_length:
+                        # If threshold exceeded then allow any small incriment for the next time
+                        # In few steps, because the first click may has any far position
+                        self.diff_length -= 1
+                        if self.diff_length < 0:
+                            self.diff_length = 0
+                        self.previous_x = event.x()
+                        return
```
ferdnyc: **_All_** of these lines are _identical_ (effectively) to the code in the `if min_max_range > 1000.0:` side of the condition, and none of it depends on `min_max_range` in any way. That means it should really be outside of the condition entirely. By moving the code up before the condition, instead of copy-pasting it into both paths _inside_ the condition, it can be processed up front, before we ever have to worry about the range of the parameter. Meaning:

```python
            # For numeric values, apply percentage within parameter's allowable range
            if property_type in ["float", "int"] and property_name != "Track":

                # Start tracking previous position
                if self.previous_x == -1:
                    self.diff_length = 10
                    # Init previous_x for the first time
                    self.previous_x = event.x()

                # Calculate # of pixels dragged
                drag_diff = self.previous_x - event.x()

                # update previous x
                self.previous_x = event.x()

                # Ignore small initial movements
                if abs(drag_diff) < self.diff_length:
                    # Lower threshold to 0 incrementally, to guarantee it'll eventually be exceeded
                    self.diff_length = max(0, self.diff_length - 1)
                    return

                # Compute size of property's possible values range
                min_max_range = float(property_max) - float(property_min)

                if min_max_range < 1000.0:
                    # Small range - use cursor to calculate new value as percentage of total range
                    self.new_value = property_min + (min_max_range * cursor_value_percent)
                else:
                    # range is unreasonably long (such as position, start, end, etc.... which can be huge #'s)

                    # Get the current value and apply fixed adjustments in response to motion
                    self.new_value = QLocale().system().toDouble(self.selected_item.text())[0]

                    if drag_diff > 0:
                        # Move to the left by a small amount
                        self.new_value -= 0.50
                    elif drag_diff < 0:
                        # Move to the right by a small amount
                        self.new_value += 0.50

                # Clamp value between min and max (just incase user drags too big)
                self.new_value = max(property_min, self.new_value)
                self.new_value = min(property_max, self.new_value)
```

That code should work **exactly** the same as the merged version — I'm not saying there's any functional difference at all, in fact it's the whole _point_ that there isn't. 

Anyway, since this is already merged I'll submit the change as a new PR. I still think it makes sense to do even though it won't affect anything from a user perspective. It's purely a question of code cleanliness and maintainability. (If nothing else, it'll mean that the next time I have Codacy scan the repo, I won't have to face its scowl of disapproval because the "Code duplication" stat crept up by however-many lines of code.)

PR #3074, opened 2019-11-11
Timeline: Refresh position on JSON load

ferdnyc:
This corrects a bug I discovered when working on #3033, where loading a new project file into an existing session doesn't refresh the time index display next to the ruler.


ferdnyc:
I'm going to merge this, as it's a one-line change that fixes a long-standing bug. (Even if we only just noticed it.)

PR #3084, opened 2019-11-20
Curve Editor for keyframe animations

SuslikV:
- [x] View Bezier interpolation curves
- [x] View Linear interpolation curves
- [x] View Constant interpolation curves
- [x] View keyframes with `property_type == "color"`
- [x] Edit Bezier interpolation curves
- [x] Edit Linear interpolation curves
- [x] Edit Constant interpolation curves
- [x] Value axis and timeline scrolling and other small improvements and features
- [x] Code clean up (not much of course)

It takes a lot of time, but I'll show to you what I have right now.
No time deadlines specified for this work. Of course, it is possible to re-brand `Editor` to `Viewer`, temporary...

System theme screen:

<details>

OpenShot Curve Viewer system theme04

</details>

Dark theme screen:
<details>

OpenShot Curve Viewer03

</details>


jonoomph:
@SuslikV Looking good!

src/windows/curve_editor.py
```diff
@@ -0,0 +1,657 @@
...
+                        else:
+                            obj_title = "title"
+
+                        # Get translation function
+                        _ = get_app()._tr
+                        self.crvEdt.setInfoLabels("{} ({})".format(c.data[obj_title], item_type), "{} ({})".format(property_name, _("No animation")))
+                        return
+
+                    # Number of animated points (keys of animation)
+                    self.newCurvPointsNum = len(keyframe["Points"])
```
musteresel: Hm ... `keyframe["Points"]` will access (in C++) `Keyframe::Points`, no? I'm about to remove this (public) member!

musteresel:
I think this is a very good addition, being able to see the curves will make working with them MUCH easier.  I don't think that the current approach is that usefull, though!

The Python code here recreates what the C++ code (in libopenshot) already does: Calculating the curves.  Instead of duplicating this calculation this (UI) code should rather query libopenshot to get the calculated values, e.g. via `Keyframe::GetValue(frame)`.  Depending on the zoom level it doesn't need to get values from all frames of course (few values and use linear or smooth interpolation between those).

SuslikV:
@musteresel 
> The Python code here recreates what the C++ code (in libopenshot) already does: Calculating the curves...

Not quite true. The only calculation available here is the clipping of the bezier curve while preserving its shape - the math action that isn't available in libopenshot.

The calculations inside the `addGraphCurveItem` - is only here because of the way the libopenshot's animation keys stored (there, for bezier curves, the control points are in %% of the coordinates - not actual values, that you can use directly from the reading of the parameters).

SuslikV:
@musteresel 
The first intention was to edit curves (change its shape not only keyframe values) the way its done in any CAD application - by moving control leverages. So, some implementations was based on this, nothing more. I even didn't knew that the control points in libopenshot has limits when started to write this code (I looked through the libopenshot's code, found Bezier and 2 Handle words and I decided that this is cubic spline that is available in Qt too). Of course, there maybe better solutions to show how animation lasts in time.

Its questionable assumption what is faster to manage and draw one object or to manage and draw few smaller objects. Custom painting is possible, but this:

> ...(few values and use linear or smooth interpolation between those)...

is close to what is done in Qt for bezier rendering itself.

Any way, the worst scenario is here - if part of the curve that will be modified by user didn't fits into the current scale of the grid - I should either:

- clip the full modified curve and show that clipped part (that has new shape)

or 

- take few values recalculated in libopenshot within this grid and show them with the linear or smooth interpolation

The later one only possible after keyframe modification complete, while the first one doesn't  changes anything in keyframes until you explicitly save new data to keyframes. Saving keyframes should trigger the preview update as I understand. Not a big deal as for me but what do you think about it?

--------------

> * no de-sync possible ... if the code in libopenshot changes but this here doesn't, then the user sees different curves as those which are used to render

I can limit interpolation type to 0,1,2, so if there will be 3, 4, or 5 it will be ignored and not shown - that will warn the user.

------------------

src/windows/curve_editor.py
SuslikV: You need to look at the `properties_model.py` code too then...

musteresel: Thanks for the hint, turned out to be no issue, `keyframe["Points"]` does *not* access (in C++) `Keyframe::Points` (almost everything seems to be passed back and forth using Json)

musteresel:
@SuslikV yes, sorry, my misunderstanding was due to the fact that openshot mostly uses its own representation (Json values) for keyframes, so the best approach is probably to draw the bezier curves using Qt as you do.

Regarding the "de-sync": I meant that if the curves shown in the viewer/editor are generated by different code than the curves in the actual rendering, then it can happen that the curves differ.  This is mostly for the Bezier curves, where the approximation done by libopenshot (currently) isn't that great when it comes to precision. And only applies in a case where a user *absolutely needs* the keyframe to behave as smoothly as displayed.

I'm about to change the curve code for libopenshot though, so if that gets accepted then the curves will hopefully be as smoothly as the ones of Qt.

Regarding the editing part: Do you already know you you want to handle e.g. when a user slowly (e.g. dragging with the mouse) changes one handle? Do you want to update the keyframe in the model (i believe that's what it's called in openshot-qt, I hope you understand what I mean) smoothly as the user changes the value or just once the change is complete (e.g. mouse button release)?

The first will provide a better experience IMO (e.g. update in the preview window) but will send updates of the Keyframe at much higher rate.  That means I need to make sure `Keyframe::SetJsonValue()` (in libopenshot) is up to the task of handling these updates fast.

SuslikV:
@musteresel probably, it will be another layer of keyframes (third in total) that will manage the changes. I hoping for fast enough response in curve display but not so fast updating of the data that affects video preview. If not - you have a nice auto-scaling _Viewer_... ^_^

SuslikV:
This is longest work-in-progress pull request I had. Here some info just to keep you on track.

I have plans to reduce number of commits later, so they will include only groups of key features (add window, viewing abilities, editing abilities etc.) in hope that it will be easier to remove/disable parts of the code if needed. This is history modifications and will take place only when all work will be done.

The curve edit will be split into 4 steps (remember curve coefficients, add point, restore curve coefficients, modify the new segment) due to way the libopenshot and OpenShot's existing logic works.

Color property edit will be tricky. More time needed for me to write the code. I set less priority on this task.

Some calculations and logic will be modified (movement restrictions etc.). Some features waits to be added (value axis and timeline scrolling). It's a lot of work. But every work should come to the end.

I think, the edit of  Bezier interpolation will be done this month. Cross the fingers for me!

SuslikV:
I have updated the tasks list to reflect current changes. And, of course, some thoughts below.

There should be right-click menus over the keys to remove keyframes and over the segments to change interpolation type. Restriction on axis move etc. Many thing was planned. Not sure that all will see the green light.

Right now the code uses _Properties_ table and depends on it a lot. The edit itself is defers from what you may seen in other editors (when key moved by time axis it actually creates new point), but this is what I was able to do using available OpenShot logic and my knowledge of the program.

SuslikV:
Number of fixes implemented, number of new bugs found.

Known issues:
* Step interpolation curve drawn with the 1 pixel step instead of 1 frame step at the edge. Noticeable at high zoom out.
  - Will be fixed soon.
* At some levels of zoom some parts of the curve not drawn at all.
  - Can be fixed, but not so fast... I had wrong assumption that if the both ends (points) of the segment isn't lies on the grid - it can be skipped from the draw. Instead, intersection of the segment curve and grid area should be checked to skip the draw of the segment.

------------

Limitations:
* The levers are limited to max/min point position.
   - While OpenShot can handle presets that can exceed this limit (`Ease In/Out (Back)`), it is hard to explain to the user why curvature of the constant lever position, for example, depends on the distance between the points. Also, it's just hard to use, as for me.

* The move of the Bezier curve is shown as it has constant levers positions but finally the initial value applied.
   - Just simplification to speed up rendering and everything.

* Curves that are clipped at grid borders are edited slightly different from original counterparts.
   - The clipped point in not shown and substituted with the point at the grid (it simplifies method I used in the program), so in general, don't attempt to modify the keyframe that lies on the edge. Zoom-out a bit and try again.

* When colors property edited with the adding the point or updating interpolation - all three RGB components changes.
   - This edit (as others) completely transferred to the _Properties_ pane. There it works this way.

------------

Anyway, the Curve Editor is only UI layer over the available math, thus it is possible to fallback to the general _Properties_ pane and its options. I hardly believe that at least 1% of the OpenShot users will use this Curve Editor for some effects. It can help to find (or even resolve) some bugs in the scene but... this is not very complex tool. Thus **there will be not much effort in developing this further**.

SuslikV:
78 commits... If I'm even able to combine all changed parts into 3-4 commits. O_o

jonoomph:
Thanks @SuslikV! This is looking awesome. Let me check it out locally, and get more familiar with it prior to merging.

jonoomph:
@SuslikV Looking good so far! Tested it out, and here are some thoughts:

 - I **love** the clip outline you draw on the curve editor
 - I'm getting double scrollbars on my screen. Perhaps we can get the curve editor to expand without using it's own scrollbars.
 - It appears that the curve editor uses it's own zoom slider. It seems a little confusing switching back and forth, it's easy to loose the context I think. I think I would prefer share the zoom slider (timeline and curve editor).
 - Is it possible to always enable the editable curve handles? I think it's more obvious what the user can interact with when they are visible.

Screenshot from 2020-02-27 16-17-45

SuslikV:
@jonoomph I'm glad that you like it from the fist sight. Later you will hate it, sure. It is very poor editor with a number of bugs and limitations. It is not designed for frequent usage - just to resolve some errors in scene animation.

About  double scrollbars:
* Maybe regular modeless Window here will be better than dock pane I used. I will try to change this.

Own zoom slider:
* Nope. I tried the shared slider first from all and it confuses me even more. So I decided to made independent one. The change itself is only in the signal connection - maybe later it can be added to the Curve Editor settings (grid settings)? Don't know if this needed. It can work in one way (Timeline -> Curve Editor) or in both directions (Timeline <-> Curve Editor). The solution with the only Timeline slider - is not handy and was tested many times.

Always enable the editable curve handles:
* Of course it is possible but I don't think this is good idea. At high zoom out it causes total mess of overlapping objects on the screen.

---------

Other bug I found, that when I'm used (1.0, 1.0) values as the defaults for the levers, the handle (circle item) that belongs to the paired point (p1 - p2 of the Bezier curve) can lie under the point itself (behind the rectangle item) and thus unreachable to the mouse move. This happens because I don't modifying Z order of child items on mouse click, but only make rising whole segment by modifying parent item. Not sure, if I will able to change this soon.

N3WWN:
I have to admit, this is my first time glancing at the curve editor, but I think the X and Y icons are a little confusing (at least to me).

Generally, I see X as left and right and Y as up and down.  The X with the horizontal line makes me think that X is the vertical component and the Y with the vertical line makes me think that Y is the horizontal component, even if that's not the intent.  

The reason is that I see the lines as moving perpendicular to their lengths, not along them.

Perhaps adding arrowheads to the lines (or perpendicular arrows in a different color) would help clear that up.

SuslikV:
> The reason is that I see the lines as moving perpendicular to their lengths, not along them.

Navy? ^_^

These lines can be removed. What other people think about icons?

SuslikV:
Small addition.
* New auto-scale of the grid (default now) in attempt get rid of double scrolls.
* New option to enable Timeline's Scale slider sharing.

OpenShot Curve Viewer system theme settings02

I tried the regular window against the dock pane - it no way better. Only in case you have two display devices new window may be more comfortable than dock pane.

The pros of window over the dock, is that the users will keep it closed more often. ^_^

So, I didn't made changes to the Curve Editor's window type.

SuslikV:
Lets hope I didn't messed up with the recent updates...

Rebased on top of current _develop_ branch. New icons and auto-grid settings (that was here many moons ago) are merged into corresponding commits of Curve Editor viewing and editing abilities.

SuslikV:
Current _develop_ branch was updated. I will update the code later.

SuslikV:
Updated on top of the current _develop_ branch.

SuslikV:
Wow! O_o new checks!
Codacy check is a bit paranoid as I see. Also, some of the code was just copied from different parts of the OpenShot itself ^_^ ^_^ ^_^

jonoomph:
Closing PR due to conflicts and lack of updates. Also, keyframe and curve editing will be coming soon, and are not based on this PR.

PR #3085, opened 2019-11-20
json_data: Eliminate utf_path

ferdnyc:
The `os.fsencode()` conversion and the use of `utf_path` is causing exceptions when mixing and matching regular expressions applied to `utf_path` and `path`:

```
   json_data:ERROR Error while converting relative paths to absolute paths:
 a bytes-like object is required, not 'str'
```

`path` will be a `str` which is already Unicode, there's no reason to apply conversions.

Also removed the `re.UNICODE` flag from the regular expression, as it's [deprecated and ignored](https://docs.python.org/3/library/re.html#re.ASCII):
> Note that for backward compatibility, the re.U flag still exists (as well as its synonym re.UNICODE and its embedded counterpart `(?u)`), but these are redundant in Python 3 since matches are Unicode by default for strings (and Unicode matching isn’t allowed for bytes).

### DIFF
--- a/src/classes/json_data.py
+++ b/src/classes/json_data.py
@@ -38,7 +38,7 @@ from classes.logger import log
 from classes import info
 
 # Compiled path regex
-path_regex = re.compile(r'\"(image|path)\":.*?\"(.*?)\"', re.UNICODE)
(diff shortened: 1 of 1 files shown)


ferdnyc:
Turns out this fixes #3094 (who knew?!?), merging.

PR #3087, opened 2019-11-20
Thumbnail server: Some fixes

ferdnyc:
* Use `send_response_only()` to avoid logging (at ERROR level) successful thumbnail loads
* Name regular expression groups, access matches by name
* Use `match()` instead of `findall()`, to avoid unnecessary nesting of result
* Look up File object from supplied ID earlier, return 404 on lookup failures
  (This covers both the previous "undefined" condition, and IDs that aren't recognized)
* Fix output path for thumbnail generation (don't override with legacy/compatibility path)
* Add overrides for logging functions:
  -  log to Python `logging` instance
  - don't log server address (always `127.0.0.1`), timestamp

### DIFF
--- a/src/classes/thumbnail.py
+++ b/src/classes/thumbnail.py
@@ -122,8 +122,14 @@ class httpThumbnailHandler(BaseHTTPRequestHandler):
         file_frame = int(url_output.group('file_frame'))
         only_path = url_output.group('only_path')
 
-        # Catch undefined calls
-        if file_id == "undefined":
+        try:
+            # Look up file data
+            file = File.get(id=file_id)
+
+            # Ensure file location is an absolute path
+            file_path = file.absolute_path()
+        except AttributeError:
+            # Couldn't match file ID
             self.send_error(404)
             return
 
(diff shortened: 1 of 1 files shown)


src/classes/thumbnail.py
```diff
@@ -136,19 +151,19 @@ def do_GET(self):
...
+        if not os.path.exists(thumb_path):
+            if file_frame == 1:
+                # Try ID with no frame # (for backwards compatibility)
+                alt_path = os.path.join(info.THUMBNAIL_PATH, "%s.png" % file_id)
+            else:
+                # Try with ID and frame # in filename (for backwards compatibility)
+                alt_path = os.path.join(info.THUMBNAIL_PATH, "%s-%s.png" % (file_id, file_frame))
+
+            if os.path.exists(alt_path):
+                thumb_path = alt_path
```
ferdnyc: @jonoomph 

This relates to this item: "Fix output path for thumbnail generation (don't override with legacy/compatibility path)"

I noticed that, because `thumb_path` was being overridden by the legacy locations **before** it was passed to `GenerateThumbnail()`, the thumbnails were always being generated at the "backwards compatibility" path. That's now fixed, and thumbnails should be created as `ProjectName_assets/thumbnail/FILE_ID/frame#.png` as presumably intended.

ferdnyc:
This fixes the issue, at least for me, of an otherwise non-functional thumbnail server on Win7 (resulting in an OpenShot that can't import any media files). So, merging...

PR #3088, opened 2019-11-20
Export: Don't double file extensions

### LINKED ISSUE #2510: Name of the exported video without double extension
**Is your feature request related to a problem? Please describe.**
immagine

As you can see I named my exported video t1.mp4, but the it's name, in the folder, is "t1.mp4.mp4", can we do something to detect redundant things like this? 

**Describe the solution you'd like**
If the name ends with .mp4, it should be trimmered

ferdnyc:
If the user includes the format string (extension) at the end of their filename, don't add it again.

Fixes #2510

### DIFF
--- a/src/windows/export.py
+++ b/src/windows/export.py
@@ -711,10 +711,15 @@ class Export(QDialog):
         # Determine final exported file path (and replace blank paths with default ones)
         default_filename = "Untitled Project"
         default_folder = os.path.join(info.HOME_PATH)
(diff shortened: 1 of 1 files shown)


SuslikV:
Should be merged this year, I think. This is simple fix. With this the behavior is much better than without (at least for me).

ferdnyc:
I will count that as a review, thanks! Merging. :grin:

PR #3095, opened 2019-11-23
Add socket, socketserver, http modules to freeze.py list

ferdnyc:
The latest Daily Builds don't seem to be able to import media on Windows 7, due to issues with the thumbnail server. I notice that the http and socketserver modules don't seem to be included in the distribution, attempting to add them explicitly.


ferdnyc:
Mmm, I think I'm wrong, that isn't the issue.

Merged my PR #3087 fixes into this branch, as it looks like some of what's going on may be addressed there.

ferdnyc:
#3087 looks to have been the proper fix.

PR #3103, opened 2019-11-28
Created security page

speedytechdev:
Used to tell users what to do if they find a security risk.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+All versions of Openshot are security supported. This means we keep all versions of Openshot update to date with the latest security features around.
+
+
+
+## Reporting a Vulnerability
+
+Use this section to tell people how to report a vulnerability.
+If you find a security vulnerability we would like to here from you asap.
+Keeping Openshot and our users safe is very important to us.
+
+If you do find a security vulnerability you can report it to jonathan@openshot.org.
+It should take us a day to the latest a week to get back to you. 
+If we do comfirm it to be a real ssecurity vulnerability we will get a patch out hopefully asap. 
+
+Thank you for taking the time to care about Openshot and the saftey of it.
+
+


SECURITY.md
```diff
@@ -0,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+All versions of Openshot are security supported. This means we keep all versions of Openshot update to date with the latest security features around.
```
ferdnyc: Just out of curiosity — devil's advocate, here — what is that claim based on? What facts provide the basis for it? As we say on Wikipedia, `[citation needed]`.

SECURITY.md
```diff
@@ -0,0 +1,21 @@
...
+
+
+
+## Reporting a Vulnerability
+
+Use this section to tell people how to report a vulnerability.
```
ferdnyc: Pretty sure this line was supposed to be edited out.

SECURITY.md
```diff
@@ -0,0 +1,21 @@
...
+If you find a security vulnerability we would like to here from you asap.
```
ferdnyc:
```suggestion
If you find a security vulnerability we would like to hear from you as soon as possible.
```

(It's a bad idea to assume that colloquialisms like "asap" are universally understood.)

SECURITY.md
```diff
@@ -0,0 +1,21 @@
...
+Keeping Openshot and our users safe is very important to us.
```
ferdnyc:
```suggestion
Keeping OpenShot and our users safe is very important to us.
```

SECURITY.md
```diff
@@ -0,0 +1,21 @@
...
+
+If you do find a security vulnerability you can report it to jonathan@openshot.org.
+It should take us a day to the latest a week to get back to you. 
+If we do comfirm it to be a real ssecurity vulnerability we will get a patch out hopefully asap.
```
ferdnyc:
```suggestion
If we confirm the reported vulnerability exists in OpenShot, we will get a patch out in a reasonable timeframe.
```

(In an ideal world this would be a concrete timeframe, "within 30 days" or "in the next release, no later than 6 weeks" or whatever. But this is a largely-volunteer effort, and commitments like that are meaningless if there isn't someone who can **promise** they'll be held to. ...Failing that, non-committal commitments are understandable. But "hopefully asap" is not a response commitment.)

SECURITY.md
```diff
@@ -0,0 +1,21 @@
...
+
+Use this section to tell people how to report a vulnerability.
+If you find a security vulnerability we would like to here from you asap.
+Keeping Openshot and our users safe is very important to us.
+
+If you do find a security vulnerability you can report it to jonathan@openshot.org.
+It should take us a day to the latest a week to get back to you. 
+If we do comfirm it to be a real ssecurity vulnerability we will get a patch out hopefully asap. 
+
+Thank you for taking the time to care about Openshot and the saftey of it.
```
ferdnyc:
```suggestion
Thank you for taking the time to care about OpenShot and its users.
```
("Safety" is a loaded word, it's impossible to define in these contexts. Say your email account is hacked. Does that threaten your _safety_, necessarily? Well, it **might**, or it might not, in a lot of different ways. Depends on who hacked it, what they saw, and what they intend to do with it. Etc...)

SECURITY.md
ferdnyc:
```suggestion
All versions of OpenShot are security supported. This means we keep all versions of OpenShot update to date with the latest security features around.
```

(Also, "all versions"? Not only is that not true, it _shouldn't_ be true. If someone finds a security vulnerability in OpenShot 2.1.0, the correct response is, "That version is no longer supported, as it's out of date. Upgrade to a newer version." There is absolutely no way anyone's going to fix an OpenShot 2.1.0 security issue. So, given that _all_ versions are **not** supported, the question becomes: Which ones _are_. Which versions, concretely, are developers willing to **promise** they'll update in response to vulnerability notifications?

SuslikV:
Public license already contains disclaimer part. OpenShot depends on third party software and some of the developers only can claim that they have had no intention to make some backdoors or in other way put security risks to the software. And all this info can be shortened to the address where to report any vulnerability if found.

ferdnyc:
@SuslikV 
> Public license already contains disclaimer part.

That's an excellent point. The reality of OpenShot's _promised_ commitment is laid out in the GPL:

>   15. Disclaimer of Warranty.
> 
>   THERE IS NO WARRANTY FOR THE PROGRAM, TO THE EXTENT PERMITTED BY
> APPLICABLE LAW.  EXCEPT WHEN OTHERWISE STATED IN WRITING THE COPYRIGHT
> HOLDERS AND/OR OTHER PARTIES PROVIDE THE PROGRAM "AS IS" WITHOUT WARRANTY
> OF ANY KIND, EITHER EXPRESSED OR IMPLIED, INCLUDING, BUT NOT LIMITED TO,
> THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
> PURPOSE.  THE ENTIRE RISK AS TO THE QUALITY AND PERFORMANCE OF THE PROGRAM
> IS WITH YOU.  SHOULD THE PROGRAM PROVE DEFECTIVE, YOU ASSUME THE COST OF
> ALL NECESSARY SERVICING, REPAIR OR CORRECTION.

While there is a "when otherwise stated..." exception there, for exactly that reason it should be carefully considered when making claims which could contradict what's _dis_-claimed by the GPL (which is there for the protection of **both** the developers and the users).

The fact is, we — this open-source development project — _don't_ support **any** version of OpenShot in any way, shape, or form. Nor do we promise that it's been subjected to any sort of security standards or evaluation. And since the code is constantly changing, it would be foolish to claim otherwise. (I mean, a non-secure HTTP server was just added as part of the new thumbnailing code. There could be dozens of vulnerabilities there.)

If supported versions are made available by @jonoomph or anyone else, either through https://openshot.org/ or other means, then those responsible can make whatever promises they deem fit regarding security or support. But those commitments shouldn't really be attached to the open-source project, where they don't apply.

jonoomph:
@ferdnyc Agree completely. Attaching claims that contradict the GPL is probably not a good idea.

jonoomph:
Closing this one, since we don't offer support guarantees. :+1:

PR #3106, opened 2019-11-30
freeze.py: Win CLI launcher, filepath fixes

ferdnyc:
#### Windows debugging CLI

So, I discovered experimentally that the only thing required to run OpenShot in a console session on Windows is to configure the build that way.

By adding a second `Executable` to the cx_Freeze configuration which _unsets_ the `base = Win32GUI` option that selects the graphical launcher, running `python3 freeze.py build` places two `.exe` files in the destination directory. With our current Inno Setup configuration, the extra executable will automatically be packaged and installed. So you end up with:
* `openshot-qt.exe`, still the graphical launcher, the target of the `OpenShot Video Editor` Start menu entry / optional Desktop icon / `.osp` file associations
* `openshot-qt-cli.exe`, which is an effectively-"hidden" alternate launcher. (In the sense that it's present in the application dir for all to see, but nothing calls attention to it so you have to either know about it or discover it on your own).

If `openshot-qt-cli.exe` is launched graphically (by double-clicking) or by typing its name into the search field of the Start menu, it will open a terminal window before launching, which automatically closes when it exits. But it can also be run from a Windows COMMAND.COM terminal, or even an MSYS shell, so that the output is preserved after exit. (It can even be run from an MSYS shell, however it becomes confused and uses the MSYS homedir instead of the Windows one as the path to `.openshot_qt`, meaning no settings and etc.)

While running, the experience is exactly the same as running from a Linux terminal: All of the standard console output is displayed in the Windows terminal. In addition, **the FFmpeg STDERR output is _also_ displayed**.

image

When errors occur... well, now that we've fixed most of our Windows launcher reporting issues (like the QMessageBoxes that displayed in place of useful error info), the experience is roughly the same as the graphical launcher. Yes, you'll see tracebacks on the console, but they'll also be displayed in popup dialogs with the graphical launcher.

Still, for debugging and testing purposes this can be an invaluable tool. I think we should just silently include it in our WIndows installers (which merging this PR will accomplish), so it's available both for our own use, and for any users who have difficult-to-diagnose problems running OpenShot on Windows.

#### Other changes

 I also made a couple of changes to the freeze script that should make it easier to run from locations _other_ than the GitLab builders, without affecting the normal build-server runs.

* I fixed my CPython-extension-file lookup code to use `inspect.getfile()`, instead of manually pulling the filename out of Python class metadata.
* I added a variable ARCHLIB to represent the architecture-specific library path on the build system. It defaults to `/usr/lib/x86_64-linux-gnu/`, the correct value for the Linux builders, _but_ it can be overridden by setting an `ARCHLIB` environment variable. All of the paths that were formerly hardcoded as `/usr/lib/x86_64-linux-gnu/somefile.so` and etc. are replaced with `ARCHLIB + "somefile.so" which allows them to be overridden with a different library path.
* I made the bundling of a number of hardcoded library files conditional on the _existence_ of the file, so that if the machine doesn't have a `libssl.so.1.0.0` file or whatever it'll freeze the program without that file, instead of throwing an exception and aborting the build.
* I reformatted a long list of Python modules to be one-per-line, making them easier to sort through and examine individually.

This PR was created in a branch of the project repo, so the Gitlab builders have already produced an installer package that includes the CLI launcher. I can make that available online if anyone would like to try it out, just let me know.

### DIFF
--- a/freeze.py
+++ b/freeze.py
@@ -76,7 +76,7 @@ python_packages = ["os",
                    "openshot",
                    "time",
                    "uuid",
-                   "shutil", 
+                   "shutil",
                    "threading",
                    "subprocess",
                    "re",
@@ -86,7 +86,7 @@ python_packages = ["os",
                    "urllib",
                    "requests",
                    "zmq",
-                   "webbrowser", 
+                   "webbrowser",
                    "json"
                    ]
 
@@ -301,7 +301,9 @@ elif sys.platform == "linux":
                       ]:
         if os.path.exists(added_lib):
             external_so_files.append((added_lib, os.path.basename(added_lib)))
-                      
+        else:
+            log.warning("{}: not found, skipping".format(added_lib))
+
 elif sys.platform == "darwin":
     # Copy Mac specific files that cx_Freeze misses
     # JPEG library
--- a/freeze.py
+++ b/freeze.py
@@ -64,6 +64,11 @@ import shutil
 
 print (str(cx_Freeze))
 
+# Set '${ARCHLIB}' envvar to override system library path
+ARCHLIB = os.getenv('ARCHLIB', "/usr/lib/x86_64-linux-gnu/")
+if not ARCHLIB.endswith('/'):
+    ARCHLIB += '/'
+
 # Packages to include
 python_packages = ["os",
                    "sys",
@@ -221,7 +226,7 @@ elif sys.platform == "linux":
 
     lib_list = [os.path.join(libopenshot_path, "libopenshot.so"),
                 "/usr/local/lib/libresvg.so",
-                "/usr/lib/x86_64-linux-gnu/qt5/plugins/platforms/libqxcb.so"
+                ARCHLIB + "qt5/plugins/platforms/libqxcb.so"
                 ] + pyqt5_mod_files
 
     import subprocess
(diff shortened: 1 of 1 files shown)


ferdnyc:
> * I made the bundling of a number of hardcoded library files conditional on the _existence_ of the file, so that if the machine doesn't have a `libssl.so.1.0.0` file or whatever it'll freeze the program without that file, instead of throwing an exception and aborting the build.

And in fact I just pushed a commit to add a `WARNING` log message whenever a file is skipped, so it won't happen silently. Now, for any nonexistent files left out of the bundle, a message will be displayed:
```console
$ ARCHLIB=/usr/lib64 python3 ./freeze.py build
<module 'cx_Freeze' from '/home/ferd/.local/lib/python3.7/site-packages/cx_Freeze/__init__.py'>
Copying modules to openshot_qt directory: /home/ferd/rpmbuild/REPOS/openshot-qt/openshot_qt
Loaded modules from openshot_qt directory: /home/ferd/rpmbuild/REPOS/openshot-qt/openshot_qt
      freeze:INFO Execution path: /home/ferd/rpmbuild/REPOS/openshot-qt/freeze.py
ldd: /usr/local/lib/libopenshot.so: No such file or directory
ldd: /usr/local/lib/libresvg.so: No such file or directory
      freeze:WARNING /usr/lib64/libssl.so.1.0.0: not found, skipping
      freeze:WARNING /usr/lib64/libcrypto.so.1.0.0: not found, skipping
      freeze:WARNING /usr/lib64/libthread-2.0.so: not found, skipping
running build
running build_exe
```

nanohtpc:
I wanna test this...

If this works, i can make a *.bat to ´QUEUE´  all my *.OSP, instead of waiting each one to finish.. one by one!!!!

ferdnyc:
@nanohtpc I think that may be jumping a gun or three... this doesn't magically make OpenShot _scriptable_, it's still the same GUI application, just printing its console log messages into a window. 

You could _load_ an `.osp` file into OpenShot from a `.BAT` script, sure, but that's as far as you'd get, and I'm pretty sure you can batch _that_ step even today with the GUI version. (Or, if not, it's still hard to see what real benefit would come from automating **only** the "Open Project..." step.) I still _manually_ clicked the Export button, set my encoding parameters, and clicked "Export Video", in order to reach the point shown in that screenshot.

Still, I'm not going to get in the way of anyone testing the code, so here's a Google Drive download link for the latest 64-bit installer with `openshot-qt-cli.exe` included: https://drive.google.com/open?id=1KxvbtKFUJNAK_ej1om-NDhr1GwFiIkoD

nanohtpc:
or the optimal solution could be a REAL QUEUE . like Shotcut...

jonoomph:
LGTM!

PR #3112, opened 2019-12-05
Undo/Redo history: don't track import/export path changes, don't clear redo for untracked changes

### LINKED ISSUE #3111: Importing a file into Project Files requires TWO Undos to reverse
Discussing Project Files list management in https://github.com/OpenShot/openshot-qt/issues/444#issuecomment-561882075 I happened to notice that for every file you import into Project Files, **two** undo operations are required to remove the item from the list.

Looking at the stored undo data, the second item pushed onto the stack for each file imported (which would correspond to the first, seemingly-ineffectual Undo operation) is an update to the `import_path`, which is performed each time even if it's being updated to the same path.

```json
            {
                "type": "update",
                "key": [
                    "import_path"
                ],
                "value": "/home/ferd/Pictures",
                "partial": false,
                "old_values": "/home/ferd/Pictures"
            }
```

So, at the very least we should avoid updating `import_path` unless it actually changes, or even better just exclude changes to `import_path` from update tracking.

ferdnyc:
This PR makes some small changes to the undo/redo tracking:
1. Add a convenience method `update_untracked`, which takes the same args as `update()` and is equivalent to:
    ```python
    app.updates.ignore_history = True
    app.updates.update(...)
    app.updates.ignore_history = False
    ```
    Except that it preserves and restores the previous value of `ignore_history`, so it's safe to use even when `ignore_history` is already `True` without losing that state.
3. Only clear the `redoHistory` if `ignore_history` is `False`, since there's no reason to clear it if no new `actionHistory` (undo) items are being created. This means that e.g. changes to the timeline zoom will no longer erase the redo history.

In addition, all updates to `import_path` and `export_path` are changed to use `update_untracked()`, so they won't appear on the undo list.

Fixes #3111

### DIFF
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -939,7 +939,7 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
                 if path and is_modified and not is_skipped:
                     # Found file, update path
                     file["path"] = path
-                    get_app().updates.update(["import_path"], os.path.dirname(path))
+                    get_app().updates.update_untracked(["import_path"], os.path.dirname(path))
                     log.info("Auto-updated missing file: %s" % path)
                 elif is_skipped:
                     # Remove missing file
--- a/src/classes/updates.py
+++ b/src/classes/updates.py
@@ -306,8 +306,8 @@ class UpdateManager:
         """ Insert a new UpdateAction into the UpdateManager (this action will then be distributed to all listeners) """
 
         self.last_action = UpdateAction('insert', key, values)
-        self.redoHistory.clear()
         if not self.ignore_history:
+            self.redoHistory.clear()
             self.actionHistory.append(self.last_action)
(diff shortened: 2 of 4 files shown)


ferdnyc:
Hmm. The **only** sticking point with this is, now exporting the project doesn't set it as "unsaved" — which we really want it to do, sice after exporting you really want to save again to store the `export_path` in the project file.

Let me see if I can find a way to mark the project data as "dirty" _without_ having to add an entry to the undo stack.

ferdnyc:
Well, that was easy.

ferdnyc:
That fix makes me feel better about this PR, it was the only thing bugging me. So, with that sorted out, merging...

ferdnyc:
Even though this is merged, it's related to the need for history transactions (which would've prevented it, and might make it worth restoring the older functionality as part of a transaction).

PR #3114, opened 2019-12-06
Revert "thumbnail: Fix path for thumb output"

ferdnyc:
This reverts commit 5a1419d52be760e50da65e27ea1b4a71ac72ec10.

Turns out that, even if the multi-directory organization was the
INTENT, it can't be used. The asset-consolidation logic in
`project_data:move_temp_paths_to_project_folder expects` only
files, not directories, in the thumbnail dir.

(Note that this revert means that the program logic goes back to being incorrect, as the thumbnail path will never _actually_ be `os.path.join(info.THUMBNAIL_PATH, file_id, "%s.png" % file_frame)` — but that's OK, because we don't actually want it to be. Still, it would be better to fix that part of the code.)


ferdnyc:
Merging this, since it fixes my mistake in applying the original change.

PR #3120, opened 2019-12-09
Replace the Center-on-Playhead icon with the quiet colors

SuslikV:
new:
center_icon_new_dark
center_icon_new_light

old:
center_icon_old_dark
center_icon_old_light

Old one was an eyesore for both themes light and dark. On the light browser background the difference is less noticeable (here on the site), but for the Dark theme in OpenShot it feels better and more organic.


ferdnyc:
Nice! Works for me, I was never happy with the old icon. It was very much slapped together in as little time as possible.

I'm not worried about the light theme, it's really not well supported period — especially considering the timeline doesn't match it, so it just looks like a disaster no matter what.

Thanks!

SuslikV:
Thank you for rebuilding resource file. I don't have where to.

PR #3122, opened 2019-12-09
Restore saving of thumbnail paths

ferdnyc:
To preserve backwards compatibility with OpenShot 2.4.4, it's best if project files continue to store the thumbnail path for now.

This is especially true because, in setting the fields to `""`, they were subsequently being normalized into the current working directory of the OpenShot process, a nonsensical value.


ferdnyc:
It's been three weeks, so I'm going to merge this, because backwards-compatibility is worth preserving.

Issue #3118, opened 2019-12-08
Cutting needs to implement a SetPlayheadFollow() interface (or, no-op)


ferdnyc:
@SuslikV noticed this in a user's logs:

```
preview_thread:INFO QThread Start Method Invoked
preview_thread:INFO initPlayer
preview_thread:INFO onModeChanged
  exceptions:ERROR Unhandled Exception
Traceback (most recent call last):
  File "C:\Program Files\OpenShot Video Editor\windows\preview_thread.py", line 61, in onModeChanged
    self.parent.SetPlayheadFollow(True)
AttributeError: 'Cutting' object has no attribute 'SetPlayheadFollow'
```

Seems there's a bug in the playhead-follow code, in that we didn't think to implement it (or, to implement a dummy function that ignores it) for the Cutting interface, since Cutting can also be a parent of the preview_thread's `onModeChanged`. 

The issue appears to be harmless (the Cutting window still works fine on my system), but it's easy enough to fix and will eliminate those exceptions.

PR #3125, opened 2019-12-15
Check if playhead follow code is available

SuslikV:
Fixes https://github.com/OpenShot/openshot-qt/issues/3118


ferdnyc:
Makes sense to me, and it passes the "an exception is expected to be a relatively rare condition" test for using `try` blocks, so I can't see any reason not to go with it. Thanks!

PR #3131, opened 2019-12-19
Always make attempts to set option only for existing stream

SuslikV:
When clip exported to the Audio Only file the error:

>The stream was not found. Be sure to call PrepareStreams() first.

may appear because the export video has no video stream where program attempts to apply the _SetOption()_. The regression was introduced in https://github.com/OpenShot/openshot-qt/pull/2860 and some masked version ("crf" text field) in https://github.com/OpenShot/openshot-qt/pull/2543

This change allows Audio Only export again.

### DIFF
--- a/src/windows/export.py
+++ b/src/windows/export.py
@@ -835,12 +835,15 @@ class Export(QDialog):
             # These extra options should be set in an extra method
             # No feedback is given to the user
             # TODO: Tell user if option is not available
-            # Set the quality in case crf was selected
-            if "crf" in self.txtVideoBitRate.text():
-                w.SetOption(openshot.VIDEO_STREAM, "crf", str(int(video_settings.get("video_bitrate"))) )
-
-            # Muxing options for mp4/mov
-            w.SetOption(openshot.VIDEO_STREAM, "muxing_preset", "mp4_faststart")
+            if export_type in [_("Audio Only")]:
+                # Muxing options for mp4/mov
+                w.SetOption(openshot.AUDIO_STREAM, "muxing_preset", "mp4_faststart")
+            else:
+                # Muxing options for mp4/mov
+                w.SetOption(openshot.VIDEO_STREAM, "muxing_preset", "mp4_faststart")
+                # Set the quality in case crf was selected
+                if "crf" in self.txtVideoBitRate.text():
+                    w.SetOption(openshot.VIDEO_STREAM, "crf", str(int(video_settings.get("video_bitrate"))) )
 
             # Open the writer
             w.Open()


ferdnyc:
Huh! I never even realized you could use faststart for M4A files as well, but a quick google confirms that it works. So, makes sense to me.

src/windows/export.py
```diff
@@ -835,12 +835,15 @@ def accept(self):
...
-            w.SetOption(openshot.VIDEO_STREAM, "muxing_preset", "mp4_faststart")
+            if export_type in [_("Audio Only")]:
+                # Muxing options for mp4/mov
+                w.SetOption(openshot.AUDIO_STREAM, "muxing_preset", "mp4_faststart")
+            else:
+                # Muxing options for mp4/mov
+                w.SetOption(openshot.VIDEO_STREAM, "muxing_preset", "mp4_faststart")
+                # Set the quality in case crf was selected
+                if "crf" in self.txtVideoBitRate.text():
+                    w.SetOption(openshot.VIDEO_STREAM, "crf", str(int(video_settings.get("video_bitrate"))) )
```
ferdnyc: Since these are export-type dependent, as it turns out, is there any reason these can't be moved up to the previous section where video and audio options are set? I'm thinking something like,
```python
          # Set video options
            if export_type in [_("Video & Audio"), _("Video Only"), _("Image Sequence")]:
                w.SetVideoOptions(True,
                                  video_settings.get("vcodec"),
                                  openshot.Fraction(video_settings.get("fps").get("num"),
                                                    video_settings.get("fps").get("den")),
                                  video_settings.get("width"),
                                  video_settings.get("height"),
                                  openshot.Fraction(video_settings.get("pixel_ratio").get("num"),
                                                    video_settings.get("pixel_ratio").get("den")),
                                  False,
                                  False,
                                  video_settings.get("video_bitrate"))
                # Muxing options for mp4/mov
                w.SetOption(openshot.VIDEO_STREAM, "muxing_preset", "mp4_faststart")
                # Set the quality in case crf was selected
                if "crf" in self.txtVideoBitRate.text():
                    w.SetOption(openshot.VIDEO_STREAM, "crf", str(int(video_settings.get("video_bitrate"))) )

            # Set audio options
            if export_type in [_("Video & Audio"), _("Audio Only")]:
                w.SetAudioOptions(True,
                                  audio_settings.get("acodec"),
                                  audio_settings.get("sample_rate"),
                                  audio_settings.get("channels"),
                                  audio_settings.get("channel_layout"),
                                  audio_settings.get("audio_bitrate"))
                # Muxing options for mp4 audio
                w.SetOption(openshot.AUDIO_STREAM, "muxing_preset", "mp4_faststart")
```
Or do they need to be after `PrepareStreams()`? Is there any issue setting `mp4_faststart` on both / setting it twice?

ferdnyc: Neeeeeevermind, I see that indeed `SetOption()` 
> must be called after the `PrepareStreams()` method, otherwise the streams have not been initialized yet. ([source](https://openshot.org/files/libopenshot/classopenshot_1_1FFmpegWriter.html#a1973624cd6d2d45a71787586f8bf78e9))

ferdnyc:
In my testing this fixes audio-only exports, without affecting video or image sequence exports. Merging, thanks!

PR #3141, opened 2019-12-25
Remove ModuleNotFoundError exception check

ferdnyc:
Turns out `ModuleNotFoundError` was only added in Python 3.6, and since our AppImages are built with 3.4 this causes problems.

But `ModuleNotFoundError` is a subclass of `ImportError` anyway, so just checking for `ImportError` covers all cases.


ferdnyc:
Passes Travis, merging.

PR #3142, opened 2019-12-25
README: Mention PYTHONPATH, reformat commands

ferdnyc:
- Add a note about running OpenShot with an uninstalled libopenshot by setting PYTHONPATH.
- Replace space-indented code blocks with fenced blocks, and mark as `sh` code to enable syntax highlighting.
- Remove `$` prompts from shell code to make copy-pastable.


ferdnyc:
Merging, it's just the README.

PR #3144, opened 2019-12-25
A bunch of small User Guide tweaks: Formatting, link updates, etc.

ferdnyc:
This _started_ when I decided I wanted to change the User Guide's 'Developers' page, replacing a link to the outdated "install instructions" PDF (from 2014!) with links to the Mac and Windows build instructions pages in the `libopenshot` wiki.

I did that. And then I kept spotting other things.

- Sphinx has a `make linkcheck` tool that will check whether the hyperlinks in your documentation are valid. Some of ours weren't. I fixed them all.
- Part of that was adding the `sphinx.ext.extlinks` extension, which lets you define templates for repeated links in the `conf.py` file and use them in the document source. I did, where appropriate.
- I also conditionally enabled the `sphinx_copybutton` extension. If found when building the source, it'll add a button to each `code-block` section that copies its contents to the clipboard for easy copy-pasting. (If not found, it'll just be skipped.)
- A lot of inline literals were wrapped with single \` characters (Markdown-style), which in ReStructuredText italicizes the contents. Switched those to double \` pairs, which formats monospaced.
- There are notes on the Developers  page about developing on Ubuntu as the preferred environment, I turned that into a Warning admonition and moved it up some.
- Wrapping paragraphs at periods or commas is an AsciiDoc convention (they call it "sentence-per-line") that makes a lot of sense in ReST and Markdown as well, instead of arbitrarily fitting text to a certain character width. Tried to do more of that.
- Also try to use shorter sentences where possible. (See previous point, plus it's just easier to read, especially for non-native English speakers.)

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -19,8 +19,9 @@
 #
 import os
 import sys
-sys.path.insert(0, os.path.dirname(os.path.abspath('.')))
-from src.classes import info
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath('.')), "src"))
+
+from classes import info
 
 # -- General configuration ------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


ferdnyc:
> * Wrapping paragraphs at periods or commas is an AsciiDoc convention (they call it "sentence-per-line") that makes a lot of sense in ReST and Markdown as well

...But not necessarily GitHub Flavored Markdown, since there are some places where line breaks will be preserved in the formatted output, and others where they won't. (GFM is extremely frustrating that way.)

ferdnyc:
Oh, right, and I also updated the build commands for `libopenshot-audio` and `libopenshot`, to avoid requiring that the user install their builds system-wide in order to test changes. It's perfectly workable to run OpenShot entirely off non-installed libraries (though I'm hoping to make it even easier), we just need to explain how to do it.

jonoomph:
@ferdnyc This looks great!! Thx

PR #3156, opened 2020-01-03
Removing usage of `-webkit-fill-available` in keyframe CSS

jonoomph:
Support older versions of webkit, since nothing renders for keyframes using the `-webkit-fill-available` value. Not really sure what this does, but all I know is that it doesn't work for me. :wink:. Fixing the size to very narrow (1 px) and the same green gradient as before. Feels pretty good to me.

### DIFF
--- a/src/timeline/media/css/main.css
+++ b/src/timeline/media/css/main.css
@@ -119,18 +119,11 @@ img {
     margin-left: -7px;
 }
 .point_icon {
-    width: -webkit-fill-available;
-    height: -webkit-fill-available;
+    width: 1px;
+    height: 6px;
     margin: 0 4px;
     background: -webkit-gradient(linear, left top, left bottom, color-stop(40%,rgba(0,255,0,1)), color-stop(100%,rgba(0,255,0,0)));
 }
-
-/*
-.point_region:hover {
-    height: 20px;
-    width: 14px;
-}
-*/
 
 /* Transitions */
 .transition_top { background: -webkit-gradient(linear, left top, left bottom, color-stop(0%,rgba(255,255,255,0.75)), color-stop(100%,rgba(255,255,255,0))); border-radius: 8px; overflow: hidden;}


ferdnyc:
@jonoomph 

Ah, yeah, sorry about that... `-webkit-fill-available` came up as one of the possible values in the Chrome CSS inspector. It was the only predefined choice that achieved what I wanted, so I went with it. When what I _should've_ done is checked the Caniuse entry for [intrinsic & extrinsic sizing](https://caniuse.com/#feat=intrinsic-width) which documents that `-webkit-fill-available` is _supposed_ to eventually become `stretch` (unprefixed), but isn't yet ready.

The **idea**, originally, was a poorly-executed — and obviously failed — experiment in trying to make use of implicit, rather than explicit, dimensions for elements.

Mostly because explicit pixel dimensions are the devil in UI design. Because of both mobile UIs and HiDPI displays, typical UIs today won't include a single explicit pixel size: everything is either sized in abstract units, or (more often lately) not sized at all, and just trusted to the layout system to render. (Though the latter obviously can't work for a timeline that's using dimensions to represent concrete values like time positions.)

I realize OpenShot is currently riddled with pixel dimensions, of course, but it would be better if it was... less so. And since keyframe marks are pretty much unique among the Timeline elements in that their **size** really _doesn't matter_ (only their position), I wanted to see if I could avoid giving them one, at least explicitly.

This was also around the time I was having issues with the Preferences window not being tall enough to show the entire Performance tab, after the hardware settings were added to that, and right after I'd discovered in https://github.com/OpenShot/openshot-qt/pull/3033#issuecomment-541369649 that the `box-sizing: border-box` styling on Clip `div`s is the source of so many Timeline layout difficulties and bugs.

So the idea went:
1. I created `div.point_region` to be the "interactive surface" for each keyframe point, as a bit of future-proofing for the day that they hopefully become clickable. The clickable area obviously has to be larger than the mark itself, because they're so tiny.
1. Since `div.point_region` had a size, and `.point_icon` had internal margins, the width and height of whatever space remained within the container was defined implicitly, so I wanted to specify that `.point_icon` should simply fill that space, without specifying explicit pixel sizes.
1. (The `width` of `.point_region` is `10px` and the left/right `margin`s on `.point_icon` are `4px`, so that's admittedly sort of nonsensical at the moment. But my goal was to then replace _those_ with abstract dimensions, maybe in `em` or `vw` units — anything but explicit pixels, really.)
1. So, point is, `width` on `.point_icon` was meant to be `10px` - `4px` - `4px` or `2px`, but specified in a way that would size itself automatically even if the scaling/resolution changed, or if I changed the `width` of `.point_region` or the `margin`s on `.point_icon`.

A deeper, goal-oriented search of Caniuse might've led me to `width: calc(100% - 8px)` as a semi-solution to the problem — it duplicates the explicit sizing of the margins, which isn't ideal, but could also be used with abstract units at least. And `calc()` has apparently been well-supported [for a surprisingly long time](https://caniuse.com/#feat=calc) on non-IE, non-Opera web platforms. Though it may still not work in WebKit versions old enough to not support `-webkit-fill-available`, I suppose.

I am  curious _how old_ these "older WebKits" are, though, and where they're used. Were there bug reports about missing keyframe marks? I didn't see any reports of issues, other than one email where you mentioned it in passing.

Caniuse doesn't track WebKit itself per se, but according to them `-webkit-fill-available` is supported since Chrome 22 which came out in July 2012. I'm used to our _most_ outdated components being frozen in early 2014, roughly, so this is unexpected even by those standards!)

jonoomph:
@ferdnyc No worries, I'm running the following version of QWebKit:
```
>>> PyQt5.QtWebKit.qWebKitVersion()
'538.1'
```
I tend to run older libraries on my development system, partially to ensure what I build runs on the most systems. But the dark side of this is, sometimes things get too old to be useful. I probably need to update my dev system. :wink:

ferdnyc:
I'm always confused how WebKit version numbers seem to have no relationship to reality, I guess Because Apple™. I'm running version `602.1`, apparently. `¯\_(ツ)_/¯`

 Teh Internetz™ seem to indicate that `538.1` dates back to around mid-2013, and was probably bundled with Qt 5.2, so that's consistent with the version still being used in the AppImage builds. I guess I'm still surprised it doesn't support `-webkit-fit-available`, but not very surprised.

But, as the AppImage team themselves pointed out in a discussion I had with them, although they recommend building on the oldest _systems_ available for the sake of compatibility, they _also_ recommend updating **components** of the build to the latest available versions for that system, basically "Because **#sekurrity**".

Since Qt/PyQt5/WebKit are never used as a system component in any of our release builds (they're bundled with the Windows/Mac installers, and into the AppImage), it's up to us what version we use, since we could bundle any version we like. Even Travis's Ubuntu Xenial image at least uses PyQt5 / Qt 5.5.1, from October 2015.

...Though, after pulling down the archived sources, it appears even Qt 5.5.1 (the last version before QtWebKit was removed from the official Qt distribution) was _still_ being bundled with the same WebKit `538.1`! The `Source/` trees in `qtwebkit-opensource-src-5.2.1` and `qtwebkit-opensource-src-5.5.1` differ by less than 20 lines. And unless someone bothered to update it, even more recent QtWebKit builds are version "`5.212`", which corresponds to WebKit 2.1.2 which is that _same_ `538` source from early-mid 2013!!

Fedora/RPM Fusion's packaged `602.1` appears to correspond to Safari 10 (OS X 10.11), which is at least from September 2016.

Aaanyway, long story short outdated WebKit versions are apparently still the _norm_ in Qt builds, so we're probably stuck with pretty poor support for modern web features until we manage to get off QtWebKit entirely.

PR #3158, opened 2020-01-03
Updating image cache logic and icons

jonoomph:
Reducing the file size of icon/image cache (85% quality vs the old 100% quality), and fixing a major issue which generated 20MB of blender/animated title icons. Finally, updating the actual icon cache to be consistent with our updated Blender animations and Titles.


jonoomph:
Not sure why Travis is having trouble running the unit tests... hmmm. @ferdnyc Any ideas on why the unittests are hanging for 10 min? I can't imagine it's because of the updated icons. :wink:

jonoomph:
I'm adding additional debugging, and getting closer to the problem. Looks like an issue with a missing file, or something along those lines.

jonoomph:
The additional debugging helped a bunch:
`effects_model:INFO Invalid effect image file: /home/travis/build/OpenShot/openshot-qt/src/effects/icons/color shift.png`

Apparently, the Travis tests use the PPA, which must be pulling in the old 'Color Shift.png' file name, and then breaking when trying to find or generate a thumbnail image. Because a QMessageBox was being shown with no corresponding log statement, it was hanging.

jonoomph:
Just when you think it's safe to commit only images... and would bet a million dollars that nothing would break. :money_with_wings: :astonished: !

SuslikV:
Hardly believable, that when you get that million, you will be committing images here. Be more realistic - one hundred or so it's nice sum to bet on.

ferdnyc:
> Hardly believable, that when you get that million, you will be committing images here.

Not wrong! Unless maybe it's an image of the view from your deluxe tropical vacation suite (just to be cruel).

> Because a QMessageBox was being shown with no corresponding log statement, it was hanging.

Yeah, I've been trying to root out and obliterate those with extreme prejudice — it doesn't _help_ to show a message box about something that isn't logged, even in the interactive case. It only leads to bug reports from users that go like this:

"Yeah, I was editing my video and I clicked on the thing I wanted, but instead of it doing what I wanted it just popped up this message box about invalidated something or other."

If the message is _also_ in the log, then at least we can say "JUST POST YOUR LOGS." in our quiet indoor voice that totally doesn't make it sound like we're talking to four-year-olds, and get the actual message from there.

...Though, in this particular case, where it's an effect icon that's not loading... **Why** exactly are we showing a message box to the user _**at all**_? There's nothing they can be expected to do about it! As presented it's completely unactionable information, so it's of no interest to them whatsoever.

If we actually consider it that dire a situation, the error should really be a more imperative message about needing to reinstall OpenShot due to a possible application corruption, because user shouldn't be expected to _infer_ the appropriate response. (Just as long as we don't tell them to delete their configuration! This is  yet another problem we cannot hope to solve by obliterating the user's settings.)

... _Or_ — hear me out here — we say "Eh, it's only an icon", fire off a `log.warning()` message about the issue, and then proceed merrily on our way as if nothing happened. Because if the _only_ issue is a missing icon file, the actual impact on OpenShot functionality is exactly none. So, what sense is there in showing an error dialog? An effect doesn't show an icon, big whoop. That's not a critical error, that's a quirk. The effect will still operate normally, and nothing about their actual project will be impacted in any way.

We need to strike a better balance between too cavalier (blindly advising users to blow away their settings as the solution to every problem, collecting tracking data without their explicit consent*) and too alarmist (treating a missing icon file as an error that warrants interrupting the user with a dialog box).

* – (I did see the commits to queue metrics until after the user's seen the welcome box and had a chance to opt out. Which I was very relieved by, it's a huge step in the right direction. _Personally_, I'd still prefer **opt-in** instead of opt-out, but hey...)

PR #3160, opened 2020-01-05
Add ordered keyframe points only

SuslikV:
Currently OpenShot always attempts to insert not ordered keyframe point when adding new one, resulting in not optimized list of keys (X --> Frame, Y --> Value, frames out of order):
```
'has_video': 
{'Points': [
{'co': {'X': 1.0, 'Y': -1.0}, 'interpolation': 2},
{'co': {'X': 9, 'Y': 0}, 'interpolation': 1},
{'co': {'X': 4, 'Y': 1}, 'interpolation': 1}
]}
```
**But as soon as algorithm looping through all existing points**, it is possible to insert the new keyframe at right "X" (frame) position.

Steps to create similar list of keyframes:
1) Add new video clip to the _Timeline_
2) Set timeline cursor to frame 9
3) Set _Enable Video_ property of the clip to ON
4) Set Timeline cursor to frame 4
5) Set _Enable Video_ property of the clip to ON
6) Look at the `has_video` points (in the log, for example).

This attempt to resolve the issue on the creation side. It is not like strict requirement by _libopenshot_ but it is much easier to work with ordered lists when it was just created. Especially, if you attempting to create Curve Editor for animation (https://github.com/OpenShot/openshot-qt/pull/3084).

### DIFF
--- a/src/windows/models/properties_model.py
+++ b/src/windows/models/properties_model.py
@@ -309,8 +309,12 @@ class PropertiesModel(updates.UpdateInterface):
                         # Keyframe
                         # Loop through points, find a matching points on this frame
                         found_point = False
+                        last_point_idx = 0
                         for point in c.data[property_key][color]["Points"]:
                             log.info("looping points: co.X = %s" % point["co"]["X"])
+                            if point["co"]["X"] < self.frame_number:
+                                # Remember index(+1) of the point previous to the current frame
+                                last_point_idx += 1
                             if interpolation == -1 and point["co"]["X"] == self.frame_number:
                                 # Found point, Update value
                                 found_point = True
@@ -350,7 +354,7 @@ class PropertiesModel(updates.UpdateInterface):
                         if not found_point:
                             clip_updated = True
                             log.info("Created new point at X=%s" % self.frame_number)
-                            c.data[property_key][color]["Points"].append({'co': {'X': self.frame_number, 'Y': new_value}, 'interpolation': 1})
+                            c.data[property_key][color]["Points"].insert(last_point_idx, {'co': {'X': self.frame_number, 'Y': new_value}, 'interpolation': 1})
 
                 # Reduce # of clip properties we are saving (performance boost)
                 c.data = {property_key: c.data[property_key]}
@@ -436,8 +440,12 @@ class PropertiesModel(updates.UpdateInterface):
                     # Loop through points, find a matching points on this frame
                     found_point = False
                     point_to_delete = None
+                    last_point_idx = 0
                     for point in c.data[property_key]["Points"]:
                         log.info("looping points: co.X = %s" % point["co"]["X"])
+                        if point["co"]["X"] < self.frame_number:
+                            # Remember index(+1) of the point previous to the current frame
+                            last_point_idx += 1
                         if interpolation == -1 and point["co"]["X"] == self.frame_number:
                             # Found point, Update value
                             found_point = True
@@ -486,7 +494,7 @@ class PropertiesModel(updates.UpdateInterface):
                     elif not found_point and new_value != None:
                         clip_updated = True
                         log.info("Created new point at X=%s" % self.frame_number)
-                        c.data[property_key]["Points"].append({'co': {'X': self.frame_number, 'Y': new_value}, 'interpolation': 1})
+                        c.data[property_key]["Points"].insert(last_point_idx, {'co': {'X': self.frame_number, 'Y': new_value}, 'interpolation': 1})
 
             if not clip_updated:
                 # If no keyframe was found, set a basic property


ferdnyc:
I'm not sure I see the point of this unless you can rely on the list _always_ being sorted — which you can't, with just this change, because all sorts of other operations are still going to be `.append()`ing points to the keyframe property lists indiscriminately. (Like, for instance, _all_ of the Clip context menu property selections, like **Volume**, **Fade**, etc. If you select a **Fade > Start of Clip > Fade In (Fast)** operation, those keyframes are going to end up on the _end_ of the list.

Because of that, if you need the list sorted at any point, it's probably best to just sort it _then_, and not worry about trying to keep it sorted **all** the time.

If your curve editor wants the list of keyframe points for a property to be sorted, you can just sort it in-place:
```python
>>> it = {}
>>> it['has_video'] = {'Points': [
... {'co': {'X': 1.0, 'Y': -1.0}, 'interpolation': 2},
... {'co': {'X': 9, 'Y': 0}, 'interpolation': 1},
... {'co': {'X': 4, 'Y': 1}, 'interpolation': 1}
... ]}
>>> from pprint import pprint
>>> pprint(it)
{'has_video': {'Points': [{'co': {'X': 1.0, 'Y': -1.0}, 'interpolation': 2},
                          {'co': {'X': 9, 'Y': 0}, 'interpolation': 1},
                          {'co': {'X': 4, 'Y': 1}, 'interpolation': 1}]}}
>>> it['has_video']['Points'].sort(key=lambda k: k['co']['X'])
>>> pprint(it)
{'has_video': {'Points': [{'co': {'X': 1.0, 'Y': -1.0}, 'interpolation': 2},
                          {'co': {'X': 4, 'Y': 1}, 'interpolation': 1},
                          {'co': {'X': 9, 'Y': 0}, 'interpolation': 1}]}}
```

Unless you can _rely_ on a sorted list, you're better off sorting it yourself when needed. (You can also make a sorted copy, instead of doing the in-place sorting:
```python
listcopy = sorted(alist, key=...)
# or
for point in sorted(alist, key=...):
    ...
```

ferdnyc:
(Sorting a list like that is _very_ fast, BTW:
```python3
>>> pprint(it)
{'has_video': {'Points': [{'co': {'X': 1.0, 'Y': -1.0}, 'interpolation': 2},
                          {'co': {'X': 4, 'Y': 1}, 'interpolation': 1},
                          {'co': {'X': 9, 'Y': 0}, 'interpolation': 1},
                          {'co': {'X': 2, 'Y': 3}, 'interpolation': 1},
                          {'co': {'X': 47, 'Y': 13}, 'interpolation': 1},
                          {'co': {'X': 3, 'Y': 99}, 'interpolation': 1},
                          {'co': {'X': 3000, 'Y': 1}, 'interpolation': 1},
                          {'co': {'X': 17, 'Y': 45}, 'interpolation': 1},
                          {'co': {'X': 555, 'Y': 55}, 'interpolation': 1},
                          {'co': {'X': 554, 'Y': 54}, 'interpolation': 1},
                          {'co': {'X': 553, 'Y': 53}, 'interpolation': 1},
                          {'co': {'X': 552, 'Y': 52}, 'interpolation': 1},
                          {'co': {'X': 551, 'Y': 51}, 'interpolation': 1},
                          {'co': {'X': 550, 'Y': 50}, 'interpolation': 1},
                          {'co': {'X': 556, 'Y': 56}, 'interpolation': 1},
                          {'co': {'X': 558, 'Y': 58}, 'interpolation': 1},
                          {'co': {'X': 557, 'Y': 57}, 'interpolation': 1},
                          {'co': {'X': 549, 'Y': 49}, 'interpolation': 1},
                          {'co': {'X': 6, 'Y': 0}, 'interpolation': 1},
                          {'co': {'X': 2999, 'Y': 12}, 'interpolation': 1}]}}
>>> import timeit
>>> res = timeit.repeat("alist = sorted(it['has_video']['Points'], key=lambda k: k['co']['X'])",
    number=10000, globals=globals()); pprint(res); print("Min: {}".format(sorted(res)[0]))
[0.044432253984268755,
 0.040496447996702045,
 0.0398482580203563,
 0.039978381944820285,
 0.04034478100948036]
Min: 0.0398482580203563
```
Python can make a sorted copy of that list 10,000 times barely 4/100 of a second, on my system. (Even though it's a relatively poor case, in terms of sorting.)

It's not easy to test the _in_-place sort using `timeit.timeit()`, since it modifies the original data, but according to the Python docs in-place sorting with `.sort()` **should** be even faster.

SuslikV:
> ..._all_ of the Clip context menu property selections, like **Volume**, **Fade**, etc. If you select a **Fade > Start of Clip > Fade In (Fast)** operation, those keyframes are going to end up on the _end_ of the list...

Unfortunately, the right-click menus has other (more terrible) bugs that this PR doesn't cover. And it will be solved in a different way. Just for reference, the scale animation  Zoom-in 50% doing double frame keys that impossible to modify in OpenShot (only remove all in once).

SuslikV:
@ferdnyc can you check if sorted list is sorted faster then unsorted?

ferdnyc:
> @ferdnyc can you check if sorted list is sorted faster then unsorted?

It is, yes. Marginally, for such a short list, but it's definitely faster. And an in-place `.sort()` on an already-sorted list is even _faster_ still.

```python
>>> from pprint import pprint
>>> pprint(it)
{'has_video': {'Points': [{'co': {'X': 1.0, 'Y': -1.0}, 'interpolation': 2},
                          {'co': {'X': 4, 'Y': 1}, 'interpolation': 1},
                          {'co': {'X': 9, 'Y': 0}, 'interpolation': 1},
                          {'co': {'X': 2, 'Y': 3}, 'interpolation': 1},
                          {'co': {'X': 47, 'Y': 13}, 'interpolation': 1},
                          {'co': {'X': 3, 'Y': 99}, 'interpolation': 1},
                          {'co': {'X': 3000, 'Y': 1}, 'interpolation': 1},
                          {'co': {'X': 17, 'Y': 45}, 'interpolation': 1},
                          {'co': {'X': 555, 'Y': 55}, 'interpolation': 1},
                          {'co': {'X': 554, 'Y': 54}, 'interpolation': 1},
                          {'co': {'X': 553, 'Y': 53}, 'interpolation': 1},
                          {'co': {'X': 552, 'Y': 52}, 'interpolation': 1},
                          {'co': {'X': 551, 'Y': 51}, 'interpolation': 1},
                          {'co': {'X': 550, 'Y': 50}, 'interpolation': 1},
                          {'co': {'X': 556, 'Y': 56}, 'interpolation': 1},
                          {'co': {'X': 558, 'Y': 58}, 'interpolation': 1},
                          {'co': {'X': 557, 'Y': 57}, 'interpolation': 1},
                          {'co': {'X': 549, 'Y': 49}, 'interpolation': 1},
                          {'co': {'X': 6, 'Y': 0}, 'interpolation': 1},
                          {'co': {'X': 2999, 'Y': 12}, 'interpolation': 1}]}}
>>> it['has_video']['Points'].sort(key=lambda k: k['co']['X'])
>>> pprint(it)
{'has_video': {'Points': [{'co': {'X': 1.0, 'Y': -1.0}, 'interpolation': 2},
                          {'co': {'X': 2, 'Y': 3}, 'interpolation': 1},
                          {'co': {'X': 3, 'Y': 99}, 'interpolation': 1},
                          {'co': {'X': 4, 'Y': 1}, 'interpolation': 1},
                          {'co': {'X': 6, 'Y': 0}, 'interpolation': 1},
                          {'co': {'X': 9, 'Y': 0}, 'interpolation': 1},
                          {'co': {'X': 17, 'Y': 45}, 'interpolation': 1},
                          {'co': {'X': 47, 'Y': 13}, 'interpolation': 1},
                          {'co': {'X': 549, 'Y': 49}, 'interpolation': 1},
                          {'co': {'X': 550, 'Y': 50}, 'interpolation': 1},
                          {'co': {'X': 551, 'Y': 51}, 'interpolation': 1},
                          {'co': {'X': 552, 'Y': 52}, 'interpolation': 1},
                          {'co': {'X': 553, 'Y': 53}, 'interpolation': 1},
                          {'co': {'X': 554, 'Y': 54}, 'interpolation': 1},
                          {'co': {'X': 555, 'Y': 55}, 'interpolation': 1},
                          {'co': {'X': 556, 'Y': 56}, 'interpolation': 1},
                          {'co': {'X': 557, 'Y': 57}, 'interpolation': 1},
                          {'co': {'X': 558, 'Y': 58}, 'interpolation': 1},
                          {'co': {'X': 2999, 'Y': 12}, 'interpolation': 1},
                          {'co': {'X': 3000, 'Y': 1}, 'interpolation': 1}]}}
>>> import timeit
>>> res = timeit.repeat("alist = sorted(it['has_video']['Points'], key=lambda k: k['co']['X'])",
 number=10000, globals=globals()); pprint(res); print("Min: {}".format(sorted(res)[0]))
[0.03604586300025403,
 0.03283630500027357,
 0.03252098400025716,
 0.0327066099998774,
 0.03253189499992004]
Min: 0.03252098400025716
>>> res = timeit.repeat("it['has_video']['Points'].sort(key=lambda k: k['co']['X'])",
 number=10000, globals=globals()); pprint(res); print("Min: {}".format(sorted(res)[0]))
[0.03523642700019991,
 0.030781944999944244,
 0.030187516000296455,
 0.030085716999565193,
 0.03025336199971207]
Min: 0.030085716999565193
```

SuslikV:
Now, we need to compare `.insert()` single item in the middle, and `.append()` of the same item. The item's size twice (or more) times lower than the list.

ferdnyc:
@SuslikV 

We _could_, but I don't really feel like designing those tests. But if you're curious, here: https://www.python.org/shell/

My point wasn't _about_ which method is **fastest**. And it's kind of impossible to apply speed numbers practically anyway, since it's about how you use it. `.append()` is super fast, if you don't have to worry about sorting. Inserting into a sorted list will always be slower, but still pretty fast. Appending and sorting will be slowest, but if the list doesn't have to be sorted after every addition (it doesn't), then the most optimal method is going to be dependent on how _often_ you sort.

But optimizing for best performance is silly, when you're talking about a difference of single-digit microseconds between two methods that both execute in under 40 microseconds.

So my point was that, since sorting is **quick** under all circumstances, rather than worrying about keeping the list sorted (and unless every line of code that works with the list is carefully coded to ensure that it stays that way, you still can't **rely** on it being sorted), it's simpler and more reliable if code that cares about sorting just sorts the list explicitly, only when needed.

src/windows/models/properties_model.py
```diff
@@ -436,8 +440,12 @@ def value_updated(self, item, interpolation=-1, value=None, interpolation_detail
                     # Loop through points, find a matching points on this frame
                     found_point = False
                     point_to_delete = None
+                    last_point_idx = 0
                     for point in c.data[property_key]["Points"]:
                         log.info("looping points: co.X = %s" % point["co"]["X"])
+                        if point["co"]["X"] < self.frame_number:
+                            # Remember index(+1) of the point previous to the current frame
+                            last_point_idx += 1
```
ferdnyc: And, in fact, your code as written here won't work correctly, because it assumes the list _is_ already sorted. If the list is **not** sorted, like for instance my example `it` from our conversation in the comments, then you get this:

```python3
>>> pprint(it)
{'has_video': {'Points': [{'co': {'X': 1.0, 'Y': -1.0}, 'interpolation': 2},
                          {'co': {'X': 4, 'Y': 1}, 'interpolation': 1},
                          {'co': {'X': 9, 'Y': 0}, 'interpolation': 1},
                          {'co': {'X': 2, 'Y': 3}, 'interpolation': 1},
                          {'co': {'X': 47, 'Y': 13}, 'interpolation': 1},
                          {'co': {'X': 3, 'Y': 99}, 'interpolation': 1},
                          {'co': {'X': 3000, 'Y': 1}, 'interpolation': 1},
                          {'co': {'X': 17, 'Y': 45}, 'interpolation': 1},
                          {'co': {'X': 555, 'Y': 55}, 'interpolation': 1},
                          {'co': {'X': 554, 'Y': 54}, 'interpolation': 1},
                          {'co': {'X': 553, 'Y': 53}, 'interpolation': 1},
                          {'co': {'X': 552, 'Y': 52}, 'interpolation': 1},
                          {'co': {'X': 551, 'Y': 51}, 'interpolation': 1},
                          {'co': {'X': 550, 'Y': 50}, 'interpolation': 1},
                          {'co': {'X': 556, 'Y': 56}, 'interpolation': 1},
                          {'co': {'X': 558, 'Y': 58}, 'interpolation': 1},
                          {'co': {'X': 557, 'Y': 57}, 'interpolation': 1},
                          {'co': {'X': 549, 'Y': 49}, 'interpolation': 1},
                          {'co': {'X': 6, 'Y': 0}, 'interpolation': 1},
                          {'co': {'X': 2999, 'Y': 12}, 'interpolation': 1}]}}
>>> idx = 0
>>> for point in it['has_video']['Points']:
...     if point['co']['X'] < 27:
...         idx += 1
... 
>>> print(idx)
7
```

The correct way to keep track of your current list index when looping over a list in Python is with `enumerate()`, e.g.:
```python3
>>> for (i,  point) in enumerate(it['has_video']['Points']):
...     if point['co']['X'] == 553: print("At index {}".format(i))
... 
At index 10
```

...But that would still require the list to be already-sorted, if you wanted to insert into the correct location. If it's not sorted, there _is_ no "correct location".

ferdnyc: In fact, the existing code really shouldn't be looping over the list _anyway_, just to search for a row. That's inefficient. A more Pythonic method would be to use `filter()` to find an existing element:

```python3
>>> for row in list(filter(lambda i: i['co']['X'] == 557, arr)):
...     print(row)
... 
{'co': {'X': 557, 'Y': 57}, 'interpolation': 1}
>>> for row in list(filter(lambda i: i['co']['X'] == 510, arr)):
...     print(row)
... 
>>>
```

SuslikV:
You misunderstood intention of the request. I don't trying to sort the list. I'm trying to reduce chaos during point creation. It makes actual sorting working faster when ever you call for it. That was the main purpose.

As was said, currently:

> algorithm looping through all existing points

why not to use small benefits of it? I'm not insist, of course.

ferdnyc:
> You misunderstood intention of the request. I don't trying to sort the list. I'm trying to reduce chaos during point creation. It makes actual sorting working faster when ever you call for it. That was the main purpose.

Ah, OK. Now I better understand the _intention_, but I'm afraid it won't have the intended **effect**. Python uses [Timsort](https://en.wikipedia.org/wiki/Timsort), and [that's not how it works](https://hackernoon.com/timsort-the-fastest-sorting-algorithm-youve-never-heard-of-36b28417f399).

Because Timsort is all about "_runs_" of elements that are already in sorted order, placing something in its _final position_ (the one it will have when the list is sorted) will have no benefit. The sort operation will end up MOVING that element out of that position, in fact, only to later move it back again.

The only "optimal" (and I use that word loosely) insertion point(s) for an element in a non-sorted list would be to stick it between two elements it "belongs" between, meaning a position where the element preceding it has a smaller sort key, and the element that follows it sorts higher. That will help in the construction of a "run" of sorted values.

IOW, if you wanted to insert a `27` into the example list above, the places that would be "slightly more optimal" would be:
1. Between `2` and `47`
1. Between `3` and `3000`
1. Between `17` and `555`
1. Between `6` and `2999`

But even those positions would provide minimal optimization, really, and in the general case could do more harm than good because it's hard to predict how Timsort will interact with the value in each of those positions. So you're better off just sticking it at the end of the list, because `.append()` is way quicker than _any_ `.insert()` (even when the list is being walked anyway).

SuslikV:
If you never start doing - nothing will change.

Even properties of handles in OpenShot can be written in reverse order of dictionary keys, for example: `'handle_left': {'Y': 0.0, 'X': 1.0} vs 'handle_left': {'X': 0.5, 'Y': 1.0}` (Y preceded X) you think it is easy to read in the log or in the `.osp` file?

ferdnyc:
@SuslikV:

> If you never start doing - nothing will change.

That has nothing to do with anything. If you want to sort the keyframe point list, then submit a PR to sort the keyframe point list. But do it using `.sort()`, when it needs to be sorted.

_Even if_ we were to do that, inserting items into the middle of a list is counterproductive. Python favors `.append()` and `.sort()`, both of which are optimised. If a list is already sorted, or nearly-sorted, Timsort makes a sort operation just as fast as the `insert()` would be.

I have no objection to sorting lists. I object to doing it the wrong way.

As far as the Json output, that's a separate issue entirely, since we just pass a dict to `json.dumps()`, and Python `dict`s didn't provide **any** control over storage order for elements until Python 3.7. (Even now, the only guarantee is that keys will be returned in insertion order, there's no reordering of an existing dict's keys.)

So we'd either have to use an `OrderedDict`, to control the stored order, or manually write out the individual elements (choosing the order based on the parent key, or similar) instead of letting `json.dumps()` just sort everything alphabetically. I definitely _don't_ think that's even a tiny bit worth it.

SuslikV:
This is `dict` control?

https://github.com/OpenShot/openshot-qt/blob/a8a641c2798918ede1c7d9c2d446c5ff4f0474da/src/windows/models/properties_model.py#L329

It written this way by OpenShot creators.

jonoomph:
@SuslikV I'm struggling to merge your 19 open PRs. Many of the PRs comments contain disagreements about whether the PR actually works, or will possibly not function correctly, and then the conversation eventually dies, and I'm left in an uncomfortable position... not confident in the completeness or accuracy or reliability of the final PR.

**Open** PRs:
Could you do me a favor, and please add a comment to each of your open PRs, something like `This PR is ready for merging in my opinion`. And please retest the PR branches, and actually make sure that they still work as designed, prior to me merging.

**Close** PRs:
And finally, if there are PRs that are no longer needed, please close them for me, so we don't waist any time reviewing them again. For example, any PRs relating to `models` and `views` are probably going to conflict with the 'Emoji' branch, where they have been heavily refactored.

**WIP** PRs:
If any of your PRs are not ready for merging, and you are still **actively** working on them, please edit the title to start with "WIP: ".  Thanks!

SuslikV:
@jonoomph Look, my PRs that isn't has "Fix" word in it are just suggestions. They are all optional changes. It is not like you give me the task and I made report if it was complete. I'm not in your team. The "Fix" marked PRs are intended to fix some issues that surely exist in the code. All my PRs can contain errors (I cannot predict everything) - this is unavoidable.

Because of long time my PRs were in opened state - main code moved forward number of times. All this time I made updates and added number of fixes/improvements to the already opened PRs. But still they all were ready to merge in every single step of it (only some builds of libopenshot based on old FFmpeg got commits that was not work in each step and I made special fix commit to ensure that all will work OK - during this period of time the PR had "WIP" word to inform that it has pending changes). I'm tying to make changes in the code and group them so, that each commit works OK with the previous. This is my own rule and I'm trying to follow it.

The PRs that depends on others - all has "WIP" word in the header. All PR are based on the _develop_ branch (no nesting changes), so when some are accepted - others may need changes before merge. This is no wonder that you'll be unable to merge everything in one hour(day). Otherwise there was no job named programmer...

_Conclusion:_
PRs with "Fix" word - requires attention (important)
PRs without - are some features that can improve experience with OpenShot or boost performance.

"Fixes" or not - it is your decision to accept them or not.

I appreciate, that you and other developers shared their code so I can use it for my own needs.

jonoomph:
Closing PR due to lack of updates and no confidence this is actually needed.

Issue #2697, opened 2019-03-21
Openshot 2.4.4 Some title text colours display wrong after changing via the Openshot title editor


davlaw0902:
Title text colour is wrong in some titles after changing it via the Openshot title editor

1. Start Openshot.  
2. Click on 'Title ->Title'
3. Titles template chooser opens. Choose 'Film rating 1' title. (Applies to all 'Film rating' and 'Creative Commons' titles)
4. Click on 'Text Colour'
5. Choose a text colour e.g. #0000ff (blue)
6. Click OK

The title viewer should display the text as blue.
On the default green background, it displays as orange.
If you click 'Save', the title is saved with orange text and the text shows as orange in the Project Files area and when the title is added to a track.

Edit the title to change the text to orange and it displays as blue.
Edit the title to change the text to yellow and it displays as light blue.
Black, white and some other text colour changes display as expected but several are wrong.

If you create then edit a title, choose a text colour and then click on the 'Use Advanced Editor' button, this opens Inkscape with the file temp.svg and the text colour is always correct within Inkscape.

However, within the Openshot title editor, the text colour is wrong and it saves with the wrong text colour.

Background colours appear to work as expected however, depending on the chosen background colour, different text colours are displayed.
For instance, select a white background colour.
Change the text colour to blue (#0000ff).
Text will display as red.
The text previously displayed as orange against a green background. 

There is no issue with Openshot 2.4.3 titles.
They work as expected.

**System Details**
- OpenShot Version: 2.4.4
- libopenshot Version:  0.2.3
- Platform: Linux-4.13.16-041316-generic-x86_64-with-LinuxMint-19-tara
- Processor: x86_64
- Machine: x86_64
- Python version: 3.4.3
- Qt5 version: 5.2.1
- PyQt5 version: 5.2.1
- Qt Detected Languages: ['en-GB', 'en']
- LANG Environment Variable: en_GB.UTF-8
- LOCALE Environment Variable: 
- Daily Build Installed: http://github.com/OpenShot/openshot-qt/releases/download/OpenShot-v2.4.4-1553153174-d04a71ee-fbfac789-x86_64.AppImage

**Log Files**
- http://logs.openshot.org/4zyqauvr7535po4gt2ss35eftj8bcr48/openshot-qt.zip
- http://logs.openshot.org/4zyqauvr7535po4gt2ss35eftj8bcr48/libopenshot.zip

ferdnyc:
@jonoomph 
[...]

PR #3161, opened 2020-01-05
Release of openshot-qt 2.5.0 (min libopenshot version: 0.2.4)

jonoomph:
This PR represents the release branch for **openshot-qt**, version 2.5.0 (min libopenshot version: 0.2.4), and will be used during the release testing. Feel free to comment, or send PR's for this release branch.

Related PR's:
 * libopenshot-audio release: https://github.com/OpenShot/libopenshot-audio/pull/80
 * libopenshot release: https://github.com/OpenShot/libopenshot/pull/407

**Testing Plan**: https://github.com/OpenShot/openshot-qt/wiki/Release-Testing-Plan

 - **Failing on All OSes** (checked when fixed):
   - [x] Save Project -> New Project -> Recent Projects -> Choose original one -> Frozen _(not able to always reproduce). This is related to rapidly opening projects (and closing the old one), one after another. A crash in FFmpegReader::CheckMissingFrame(). Looking into this now._
   - [x] SVG alpha gradients not working with **resvg**
   - [x] Help -> About -> Changelog shows all japanese/chinese characters if you scroll down to the bottom. (Mac and Linux). Perhaps a UTF-8 or character set issue)

 - **Failing on Appimage** (checked when fixed):
   - [x] 6.8 - Export a 30 FPS project to 60 FPS (verify Fade in & Fade out is exported correctly). _Our keyframes are being scaled on the parent project, and not scaled back after export._
   - [x] 6.9 - Export a 30 FPS project to 24 FPS (verify Fade in & Fade out is exported correctly)
   - [x] 7.2 - Create title, it should appear in Project Files, and preview correctly on timeline. _Some of our titles are not rendering correctly with gradients (Bar 1, Flames, Oval 4, Standard 2)_

 - **Failing on Mac** (checked when fixed):
   - [x] Most titles are rendered without text with **resvg** (incompatible fonts I think)
   - [x] Most fonts do not render with **resvg** (a few do render)
   - [x] ~Frozen export at 99.8% (could not reproduce)~
   - [x] 8.4 - Verify credits load with names and data, and filters correctly. _Help -> About -> Credits | 
      License: Both fail to find a file (contributors.json | license.txt) It's looking for /Contents/Resources/resources/, but that folder is currently at /Contents/MacOS/resources/._
   - [x] 7.1 - Set SVG editor path and Blender path in preferences. _Preferences: Choose Blender exe not really possible, because it needs the following appended: ...Blender.app/Contents/MacOS/Blender_

 - **Failing on Windows** (checked when fixed):
   - [x] File -> New Project: Title of project defaults to "Copy text  Untitled Project"???
      Deleting ~/.openshot_qt/openshot.settings fixed this, but still weird


ferdnyc:
Awesome! I'll try to circle back to this later tonight.

>
> * **Failing on Appimage**:
>   
>   * [ ]  6.8 - Export a 30 FPS project to 60 FPS (verify Fade in & Fade out is exported correctly). _Our keyframes are being scaled on the parent project, and not scaled back after export._
>   * [ ]  6.9 - Export a 30 FPS project to 24 FPS (verify Fade in & Fade out is exported correctly)
>   * [ ]  7.2 - Create title, it should appear in Project Files, and preview correctly on timeline. _Some of our titles are not rendering correctly with gradients (Bar 1, Flames, Oval 4, Standard 2)_

...Does this imply that all the **rest** of the tests were run, and passed? Maybe we should copy the entire checklist from that wiki page into a comment here, and check off tests as they _are_ confirmed passed, so we know we've covered all of them? (Might even make sense to do it three times, one for each OS. Though I don't know if we have any testers on MacOS...)

jonoomph:
@ferdnyc Yes, all tests passed on the *.AppImage except those. I'm going to run through the tests on Windows and Mac shortly, and figured it would be less verbose to only post the failures.

jonoomph:
Fixed keyframe scaling issues after an export (it was rescaling them all twice... oops). Updated comment at top of PR.

jonoomph:
Refactored some Export logic, so we no longer modify the active project data when using different FPS (i.e. keyframe rescaling). This seems much safer now. We just make a copy of the rescaled data, and use that copied data during the export.

jonoomph:
Updated **resvg** on our Ubuntu Gitlab runner (used for our Linux AppImage builds). Hopefully this will fix any SVG gradient rendering issues. I'll test once the new build is ready.

jonoomph:
The latest version of resvg does not appear to solve the SVG gradient rendering issue. I've sent an email to @RazrFalcon seeking guidance. :wink: On my local system, resvg renders everything correctly. However, when using the AppImage, gradients appears to be inverted.

Screenshot from 2020-01-07 15-17-09

jonoomph:
Finished my initial testing run on all 3 OSes. Updated the PR with the results. A few failures, but mostly everything works good. :+1:

ferdnyc:
> * SVG alpha gradients not working with **resvg**

I see what you mean about the resvg rendering. I've reproduced the issue with the `rendersvg` utility in the latest code from the resvg repo as well. The weird thing is, resvg [is documented](https://razrfalcon.github.io/resvg-test-suite/svg-support-table.html) as fully supporting all of those features, so maybe it's some quirk of the way Inkscape is constructing the SVG source.

(It is a bit odd, with every gradient `fill` and `stroke` referencing a `<linearGradient>` that's built in two stages. There's one that sets the color stops, then a second one — the one the object references — that links to the _first_ gradient and applies geometry to it. Maybe resvg has some issue correctly interpreting that many levels of indirection.)

Regardless, I'll take `Bar_1.svg` over to https://github.com/RazrFalcon/resvg and file an issue with samples of the Inkscape and `rendersvg` rendering, see if they can figure out what's going on and, more importantly, if there's any way we can work around it.

RazrFalcon:
>I've reproduced the issue with the rendersvg utility in the latest code from the resvg repo as well.

What OS do you use? I cannot reproduce it on linux. Also, what Qt version do you have?

ferdnyc:
@RazrFalcon I just filed https://github.com/RazrFalcon/resvg/issues/196 with details and steps to reproduce.

ferdnyc:
@RazrFalcon has confirmed the issue (in RazrFalcon/resvg#196), and is on the case! :smile:

ferdnyc:
@jonoomph The gradient _inversion_ thing is actually an issue that's been happening from the very beginning, with the AppImages. (And it's filed as #2697.) And, I found the same: it's _only_ in the AppImage builds — works fine when rendering anywhere else, either from our code or directly with resvg's `rendersvg`. I've never succeeded in reproducing it outside of OpenShot-as-AppImage.

I suppose it _could_ be down to the very, very old Qt version that's bundled into the AppImages builds, in that case.

RazrFalcon:
@ferdnyc Can you try to render a simple gradient manually to test if it actually resvg or Qt bug?

ferdnyc:
@RazrFalcon The color-inversion? Oh, that's not even confined to gradients, it happens on _solid_ colors too.

But I've tried to reproduce it every way I know how, and just can't find a way. It doesn't happen with OpenShot+resvg running on any system I've used (outside of the AppImage environment), it doesn't happen with OpenShot+resvg on Windows, and it doesn't  happen with any build of `rendersvg` that I've tried.

I can't imagine it's actually a resvg bug, it's just way too isolated. Plus, our AppImages are built with Qt 5.2.1, soooo... :wink:

RazrFalcon:
@ferdnyc I thought that you can try rendering a gradient manually, without resvg. An if error persists than this is definitely not a resvg bug.

ferdnyc:
@RazrFalcon Oh, it doesn't happen if we fall back to our QtSvg implementation, but that's such a different animal that I feel like it doesn't really mean all that much. QtSvg just `load()`s the file straight into a `QImage`. With resvg we pass it a reference to `QPainter`, for an image in `QImage::Format_RGBA8888` — could the format be an issue? Maybe older Qt versions don't handle conversion automatically/correctly? 

libopenshot [src/QtImageReader.cpp](https://github.com/OpenShot/libopenshot/blob/4ebda593320d89c85ab4bdb0137d2625be1c5a6e/src/QtImageReader.cpp#L82):
```c++
ResvgRenderer renderer(path);
if (!renderer.isValid()) {
	// Attempt to open file (old method using Qt5 limited SVG parsing)
	success = image->load(path);
	if (success) {
		image = std::shared_ptr<QImage>(new QImage(
			image->convertToFormat(QImage::Format_RGBA8888)));
	}
} else {
	image = std::shared_ptr<QImage>(new QImage(renderer.defaultSize(),
		QImage::Format_RGBA8888));
	image->fill(Qt::transparent);

	QPainter p(image.get());
	renderer.render(&p);
	p.end();
```

I'd love to be able to test it more thoroughly, but like I said it only happens in that one specific environment. AppImages are read-only packages, and I don't have the tools to build them myself, so testing is sorely limited.

RazrFalcon:
>With resvg we pass it a reference to QPainter, for an image in QImage::Format_RGBA8888 — could the format be an issue?

Of course it is :smile: `resvg` works only with `QImage::Format_ARGB32_Premultiplied`. You should prefer rendering SVG on image, and then rendering the image, because in many cases it will be faster and safer.

ferdnyc:
A-HA! Glad I thought to ask. I'll make that change and see what that does for us, thanks.

ferdnyc:
Haven't quite finished the changes on our end yet, but here's `Bar_1.svg` rendered with `rendersvg` from the resvg git HEAD:

Bar_1

So, yup. That's fixed.

(And now that I've built it, I can test-build a libopenshot linked with it.)

jonoomph:
@ferdnyc Can you push the fix to our release branch? :1st_place_medal:

ferdnyc:
@jonoomph 

Absolutely, as soon as I'm done completely rewriting `FindRESVG.cmake` to use targets and be modern and stuff. :wink: (It was vexing me.)

jonoomph:
I've updated resvg on our AppImage linux builder, but I'm having trouble compiling the latest version on our 64-bit and 32-bit Windows builders. I've opened a ticket: https://github.com/RazrFalcon/resvg/issues/197.

ferdnyc:
And I have an AppImage built with the color-inversion fix (not sure if it got the newer resvg or not, but meh), time to test out http://gitlab.openshot.org/OpenShot/openshot-qt/-/jobs/14993/artifacts/raw/build/OpenShot-v2.4.4-dev2-1578524577-309c49b6-f38389fc-x86_64.AppImage

(Link only accessible to project members, sorry rest-of-world.) :wink:

ferdnyc:
OK, I've re-confirmed that the test file in #2697 imports as orange-on-green in `OpenShot-2.4.4-x86_64.appimage`...

Ah, poo. But in the AppImage I just built, I end up with this crash on launch:
```
OpenShotApp::Import Error: /usr/local/lib/libresvg.so: cannot open shared object file:
No such file or directory
```

Mmm, looks like a link path issue. `libopenshot.so.18` is linked with the full path to `libresvg.so`. Inside the AppImage mount:
```sh
$ cd /tmp/.mount_qiwDqQ/
$ ls -l usr/bin/libresvg.so
-rwxr-xr-x. 1 root 1001 5.5M Jan  8  2020 usr/bin/libresvg.so
$ ldd usr/bin/libopenshot.so.18 | head -n 20
	linux-vdso.so.1 (0x00007ffe9b1fd000)
	libavcodec.so.57 => not found
	libavformat.so.57 => not found
	libavutil.so.55 => not found
	libswscale.so.4 => not found
	libswresample.so.2 => not found
	libzmq.so.3 => not found
	/usr/local/lib/libresvg.so => not found
	libopenshot-audio.so.7 => not found
	libMagick++.so.5 => not found
	libMagickCore.so.5 => not found
	libQt5Widgets.so.5 => /lib64/libQt5Widgets.so.5 (0x00007f9ae217e000)
	libQt5Gui.so.5 => /lib64/libQt5Gui.so.5 (0x00007f9ae1c42000)
	libQt5Core.so.5 => /lib64/libQt5Core.so.5 (0x00007f9ae1722000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007f9ae1528000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f9ae13e2000)
	libgomp.so.1 => /lib64/libgomp.so.1 (0x00007f9ae13a8000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f9ae138e000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f9ae136c000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f9ae11a1000)
```

I'll try to figure out what's causing that.

In the meantime, an `LD_PRELOAD` of the library might work to launch it for testing. Stay tuned...

ferdnyc:
> I'll try to figure out what's causing that.

A-HAH!! Got it. A quirk of CMake with `IMPORTED TARGET`s. Because `libresvg.so` has no SONAME versioning, I have to set [`IMPORTED_NO_SONAME`](https://cmake.org/cmake/help/v3.3/prop_tgt/IMPORTED_NO_SONAME.html) on the `RESVG::resvg` target for it to link without the full path. I'll fix that now.

ferdnyc:
@jonoomph  Um... did I break the whole world over @ the GitLab builders, somehow, or are you working on them? Suddenly everything's failing... I don't _think_ it's my fault? (Hopefully not!)

jonoomph:
@ferdnyc You broke the whole world

jonoomph:
Nope, it's my fault. :wink:. Mac builder is currently updating rustc, and Windows are in a bad state due to a broken rustc install. Fun times!

ferdnyc:
S'good timing anyway, I have to stop to drill a hole in a door (don't ask), and I could use some dinner.  So, no rush. And whenever things are up and building again, the `resvg-image-format` branch should spit out an AppImage with the color-inversion fix. Working, I think! (Hopefully. :crossed_fingers: )

jonoomph:
Mac is working again (for the moment), but your resvg-image-format branch is still breaking on Windows. I am able to re-run older versions of that branch without issue, but the latest version of that branch crashes the Windows build. So, this looks like something specific to Cmake if I had to guess?

freeze.py
```diff
@@ -292,8 +292,8 @@ def find_files(directory, patterns):
     # for certain distros (like Fedora, openSUSE, Debian, etc...)
     # Also add Glib related files (required for some distros)
 
-    for added_lib in [ARCHLIB + "libssl.so.1.0.0",
```
jonoomph: Apparently these files no longer have versions appended to them on my Linux build server. Hmmm. Also, I can't find `ARCHLIB + "libthread-2.0.so"`. Is that supposed to be `libpthread.so`?

ferdnyc: @jonoomph 

I think it's supposed to be `libpthread-2.30.so`, actually, AKA `libpthread.s.0`. That exists on my Fedora machine. Or there's `libthread_db.so.1` `libthread_db-1.0.so`, also part of glibc. But it's most likely meant to be the former.

The unversioned `libname.so` files are usually installed by `-dev` packages, they're for use when linking builds against the libraries. If you don't have the versioned `.so.M.N...` files, though, it's likely you either have devel packages installed without the corresponding libs, or you really need to run `ldconfig`. :laughing:

ferdnyc:
FTR, the `resvg-image-format` build issues were cleared up with the merge of my latest `find-resvg` branch changes, AppImages are again building, and... the code changes to QtImageReader did nothing to fix the color inversion. (See Openshot/libopenshot#407 for more.)

@jonoomph and I are now operating on the assumption that Qt 5.2.1 is just too damn old, and working on updating the Qt version on the AppImage builder. I've got Qt 5.12.6 compiling from source in another window, hopefully that'll sort us out.

jonoomph:
@ferdnyc @RazrFalcon Looks like Qt 5.2 still works fine, as long as we are correctly using `Format_BRGA32_Premultiplied`. Just tested a build, and it finally works correctly! Thank you so much to both of you for the support! You both rock!

Screenshot from 2020-01-11 23-37-48

RazrFalcon:
Nice!

ferdnyc:
@jonoomph Great news! Qt 5.2 is still moldy and ancient (as evidenced by the fact that any later versions of Qt were saving us from ourselves, and just automatically converting the image), but nice to know that the issue is cleared up.

It's also worth pointing out that the use of `Format_ARGB_Premultiplied` is a documented performance optimization according to Qt. From [the QImage docs](https://doc.qt.io/qt-5/qimage.html#Format-enum):
> **Note:** Avoid most rendering directly to most of these formats using `QPainter`. Rendering is best optimized to the `Format_RGB32` and `Format_ARGB32_Premultiplied` formats, and secondarily for rendering to the `Format_RGB16`, `Format_RGBX8888`, `Format_RGBA8888_Premultiplied`, `Format_RGBX64` and `Format_RGBA64_Premultiplied` formats.

You'll notice that our internal QImage format and the one we convert all pixel data to, `Format_RGBA8888`<sup>1</sup>, is nowhere on that list. Considering how much we rely on both QImage and QPainter in composing frame images, sticking with `RGBA8888` may have a measurable impact on performance. If it'd be possible to at _least_ switch to the `RGBA8888_Premultiplied`<sup>2</sup> version, a similar format but one that at least makes their second-tier list, it's worth looking into.

#### Notes
1. There's also a caveat in the table of formats about `RGBA8888` specifically:
    > The image is stored using a 32-bit byte-ordered `RGBA` format (`8-8-8-8`). Unlike `ARGB32` this is a byte-ordered format, which means the 32bit encoding differs between big endian and little endian architectures, being respectively (`0xRRGGBBAA`) and (`0xAABBGGRR`). The order of the colors is the same on any architecture if read as bytes `0xRR,0xGG,0xBB,0xAA`. (added in Qt 5.2)
    
    So, not only is the format inefficient, but its underlying storage _differs_ between little-endian and big-endian systems. (Which may be the source of the inefficiency, since it sounds like Qt has to do conversion on access to ensure that bytes are always read in the correct order.)

2. (_Especially_ since I'm pretty sure I remember seeing code in at least one place where libopenshot actually goes through and _manually **multiplies**_ the pixel color values by the alpha channel — in other words, converts the data to Premultiplied "by hand", when it would not only be simpler but more efficient to keep it in that format in the first place.)
    The table's notes on `ARGB32_Premultiplied` give more details:
    > The image is stored using a premultiplied 32-bit `ARGB` format (`0xAARRGGBB`), i.e. the red, green, and blue channels are multiplied by the alpha component divided by 255. (If `RR`, `GG`, or `BB` has a higher value than the alpha channel, the results are undefined.) Certain operations (such as image composition using alpha blending) are faster using premultiplied `ARGB32` than with plain `ARGB32`.

ferdnyc:
> 2. (_Especially_ since I'm pretty sure I remember seeing code in at least one place where libopenshot actually goes through and _manually **multiplies**_ the pixel color values by the alpha channel — in other words, converts the data to Premultiplied "by hand", when it would not only be simpler but more efficient to keep it in that format in the first place.)

I remembered incorrectly. What we're _actually_ doing is going through and manually applying the configured alpha level for the frame image, by multiplying the image's existing alpha channel by our alpha setting. From `Timeline::add_layer()` in [`Timeline.cpp@387:403`](https://github.com/OpenShot/libopenshot/blob/0bc87c00bcce2e1703eeca889fa685e743413785/src/Timeline.cpp#L387-L403)...
```c++
/* ALPHA & OPACITY */
if (source_clip->alpha.GetValue(clip_frame_number) != 1.0)
{
	float alpha = source_clip->alpha.GetValue(clip_frame_number);

	// Get source image's pixels
	unsigned char *pixels = (unsigned char *) source_image->bits();

	// Loop through pixels
	for (int pixel = 0, byte_index=0; pixel < source_image->width() * source_image->height(); pixel++, byte_index+=4)
	{
		// Get the alpha values from the pixel
		int A = pixels[byte_index + 3];

		// Apply alpha to pixel
		pixels[byte_index + 3] *= alpha;
	}
```

Which isn't **that** bad, I suppose. But it is still a _bit_ silly, because at the end of `add_layer()` we pick up the `QPainter` for the output frame, then use `drawImage()` to paint on top of it. `QPainter` _**has**_ a [`setOpacity()` method](https://doc.qt.io/qt-5/qpainter.html#setOpacity), and you'll never guess what it does.

So, I'll probably whip up a quick PR to make that use `QPainter::setOpacity()` instead. (And since Qt claim the performance is so heavily dependent on format, maybe experiment with using `ARGB32_Premultiplied` to store our frame `QImage`s?)

jonoomph:
About to merge for final release

ferdnyc:
Just updated the release date to 2020-02-07 in the AppData file, though if it pushes into tomorrow it should really be updated again. (And we should really come up with some way of scripting that part of the process, among others.)

PR #3166, opened 2020-01-08
launch.py: Add a --path arg to set PYTHONPATH

ferdnyc:
This adds an argument `--path` to the command-line processing, providing an alternate way to prepend paths to `sys.path`, rather than setting them in the environment via `PYTHONPATH`.

It can be used multiple times, e.g.:
`src/launch.py --path=/my/local/lib/python3.7/site-packages --path=../libopenshot/build/src/python`


ferdnyc:
It's something I have to test first, but I'm _hopeful_ the new Windows CLI launcher calls `launch.py` and will accept its command-line arguments, in which case this would allow the path to be adjusted without having to futz with a PYTHONPATH environment variable. (Which are a bigger pain in Windows than in Linux/MacOS.)

ferdnyc:
This is relatively harmless, and everyone's had over two months to raise any concerns. At worst, it just isn't actually useful on Windows. (But it's still handy for Linux, if nothing else.) Merging.

PR #3168, opened 2020-01-08
Effects/Transitions: Remove redundant column

ferdnyc:
- The thumbnail column already contains the "Name" string (as the icon caption), so rather than having a separate column, rename the "Thumb" column to "Name" and only display that.
- The Effects `TreeView` requires special handling, since the width of the column has to be set explicitly to balance with the "Details" column. A `resizeEvent()` implementation sets it to 40% of the widget width, constrained within the range <code>[150:<strike>260</strike>280]</code>.

### DIFF
--- a/src/windows/models/effects_model.py
+++ b/src/windows/models/effects_model.py
@@ -50,7 +50,7 @@ class EffectsStandardItemModel(QStandardItemModel):
         files = []
         for item in indexes:
             selected_row = self.itemFromIndex(item).row()
-            files.append(self.item(selected_row, 4).text())
+            files.append(self.item(selected_row, 3).text())
         data.setText(json.dumps(files))
         data.setHtml("effect")
 
@@ -73,7 +73,7 @@ class EffectsModel():
             self.model.clear()
 
         # Add Headers
-        self.model.setHorizontalHeaderLabels([_("Thumb"), _("Name"), _("Description")])
+        self.model.setHorizontalHeaderLabels([_("Name"), _("Description"), _("Type")])
 
         # Get the folder path of effects
         effects_dir = os.path.join(info.PATH, "effects")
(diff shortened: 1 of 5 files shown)


ferdnyc:
The motivation for this is here:

image

See the "&" next to the third row? That's part of the effect name; if we widen the "Thumb" column, we can see why it shows up there:

image

Applying this PR, it becomes this:

image

ferdnyc:
I raised the max width slightly to `280` — it'll be dependent on language and font size, of course, but on my system that's the point where the longest effect name, "Alpha Mask / Wipe Transition", fits on one line. At that point, going wider accomplishes nothing.

I'd love to have Qt auto-size the column, but it doesn't seem to want to do that sensibly. (It didn't really do it sensibly _before_, either, hence the "&" sign that started all this.)

That's possibly a consequence of how we're using the model. Specifically, that we're destroying and regenerating it constantly, including when we switch views / filter.

Explicitly using `resizeColumnToContents()` doesn't work, but that forces the columns to fit the _entire_ **unwrapped** width of their longest string. (IOW, it makes them waaaaay too wide.)

ferdnyc:
> I'd love to have Qt auto-size the column, but it doesn't seem to want to do that sensibly. (It didn't really do it sensibly _before_, either, hence the "&" sign that started all this.)

To anticipate the question, "So why did it look _mostly_ OK before?", it's because both columns were just always given the default column width of `100` pixels. Which luckily was a workable size for those  two columns.

jonoomph:
@ferdnyc I'm wondering what to do with this PR. We've made some pretty big changes to the view/models in the Emoji branch, and I'm worried about conflicts if we merge this. What do ya think is the easiest way forward? Close this, and re-implement in the Emoji branch? Or merge it, and fix all the conflicts?

ferdnyc:
@jonoomph Oh, if we do this, we should definitely do it in the Emoji branch. That thing will never be mergeable otherwise. (Plus, I'd rather consolidate _all_ the models' columns at once. And probably will, some time this week.)

Issue #3164, opened 2020-01-07
Video clip timeframe and fps not imported or exported correctly


john-davies:
_This is possibly related to #3138 but I also have some additional problems not shown in that issue._

**Describe the bug:**
Video clip timeframe and fps not imported or exported correctly

**Steps to reproduce the behavior:**
1. I created a short time lapse movie containing 150 frames at 25 fps using ffmpeg:

`ffmpeg -framerate 25 -start_number 0000 -i %04d.jpg -c:v libx264 -profile:v high -crf 20 -pix_fmt yuv420p e.mp4`

2. If I open the file with VLC or HandBrake then it reports the expected 25 fps and duration of 6s

3. I then import this file into a blank OpenShot project with a profile of 720p 25fps 

![issue1](https://user-images.githubusercontent.com/16657559/71919299-3d28a100-317c-11ea-9c50-1a035923438e.jpg)

4. Opening the "Preview" window gives the correct frame count but the overall time is reported as 5.25s rather than 6s.

![issue2](https://user-images.githubusercontent.com/16657559/71919352-61847d80-317c-11ea-9da4-fd8d74f1d36c.jpg)

5. If I add this clip to a blank track and export it then ffmpeg reports a frame rate of 25.17 fps and a duration of 5.98s for the resulting output file.

6. Additionally there are jumps in the timeline when I'm previewing the clip in OpenShot. For example starting at frame 125:

![issue3](https://user-images.githubusercontent.com/16657559/71919378-76f9a780-317c-11ea-9a55-cfe55ba18dc4.jpg)

then 1 click of the right arrow key gets me correctly to frame 126 but the timeline jumps from 4.25s to 5.01s:

![issue4](https://user-images.githubusercontent.com/16657559/71919398-824cd300-317c-11ea-8f06-5e0df852554a.jpg)

**Expected behavior:**
I would expect the clip and fps setting to remain constant throughout these operations.

**System Details:**
- app:INFO openshot-qt version: 2.4.4 _( Latest AppImage version downloaded from the website )_
- app:INFO libopenshot version: 0.2.3
- app:INFO platform: Linux-4.4.0-170-generic-x86_64-with-Ubuntu-16.04-xenial

**Log Files:**

Initial input file: https://drive.google.com/open?id=1EXuKU6o4mmXW2bV-UQ1NcNZiJ0KfsBYY

[...]

PR #3170, opened 2020-01-09
Change separator for the frame number display

SuslikV:
**Edit:**
`:` to `[ ]`  (for frames only)

before
Timecode old

after
Timecode new 02

While OpenShot uses own 1-31 markup for frames notation (instead of standard 0-29) for the 30fps videos, it is OK to make it more noticeable for the end users.

Mentioned in: https://github.com/OpenShot/openshot-qt/issues/3164#issuecomment-571758511

### DIFF
--- a/src/timeline/index.html
+++ b/src/timeline/index.html
@@ -39,7 +39,7 @@
 		
 		<!-- RULER NAME (left of screen) -->
 		<div tl-rulertime id="ruler_label">
-			<div id="ruler_time">{{playheadTime.hour}}:{{playheadTime.min}}:{{playheadTime.sec}}:{{playheadTime.frame}}</div>
+			<div id="ruler_time">{{playheadTime.hour}}:{{playheadTime.min}}:{{playheadTime.sec}} [{{playheadTime.frame}}]</div>
 		</div>
 		<!-- RULER (right of screen) -->
 		<div id="scrolling_ruler">
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -93,7 +93,7 @@ App.controller("TimelineCtrl", function ($scope) {
     // Use JQuery to move playhead (for performance reasons) - scope.apply is too expensive here
     $(".playhead-top").css("left", (($scope.project.playhead_position * $scope.pixelsPerSecond) + $scope.playheadOffset) + "px");
     $(".playhead-line").css("left", (($scope.project.playhead_position * $scope.pixelsPerSecond) + $scope.playheadOffset) + "px");
-    $("#ruler_time").text($scope.playheadTime.hour + ":" + $scope.playheadTime.min + ":" + $scope.playheadTime.sec + ":" + $scope.playheadTime.frame);
+    $("#ruler_time").text($scope.playheadTime.hour + ":" + $scope.playheadTime.min + ":" + $scope.playheadTime.sec + " [" + $scope.playheadTime.frame + "]");
   };
 
   // Move the playhead to a specific frame
--- a/src/windows/add_to_timeline.py
+++ b/src/windows/add_to_timeline.py
@@ -404,7 +404,7 @@ class AddToTimeline(QDialog):
 
         # Update label
         total_parts = time_parts.secondsToTime(total, fps["num"], fps["den"])
-        timestamp = "%s:%s:%s:%s" % (total_parts["hour"], total_parts["min"], total_parts["sec"], total_parts["frame"])
+        timestamp = "%s:%s:%s [%s]" % (total_parts["hour"], total_parts["min"], total_parts["sec"], total_parts["frame"])
         self.lblTotalLengthValue.setText(timestamp)
 
     def reject(self):
--- a/src/windows/cutting.py
+++ b/src/windows/cutting.py
@@ -191,7 +191,7 @@ class Cutting(QDialog):
 
         # Convert seconds to time stamp
         time_text = time_parts.secondsToTime(seconds, self.fps_num, self.fps_den)
-        timestamp = "%s:%s:%s:%s" % (time_text["hour"], time_text["min"], time_text["sec"], time_text["frame"])
+        timestamp = "%s:%s:%s [%s]" % (time_text["hour"], time_text["min"], time_text["sec"], time_text["frame"])
 
         # Update label
         self.lblVideoTime.setText(timestamp)


ferdnyc:
First, a point of order:
> While OpenShot uses own 1-31 markup for frames notation (instead of standard 0-29) for the 30fps videos, it is OK to make it more noticeable for the end users.

You mean 1-30, instead of 0-29. We don't add an extra frame to each second! :laughing: 

I have thought about this on occasion, yeah. I don't personally think the `##:##:##:##` notation is a _problem_, really, but at the same time there's no harm in changing it if it prevents the occasional confused user. (Especially since it's one of those topics where said confusion tends to get blown out of proportion, because clearly we MALICIOUSLY intended to LIE to them by formatting the playhead position in a way that they managed to misinterpret. But I digress...)

So, yeah, in terms of changing the display, I'm on board. But I have to admit, I **hate** that slash character, the way it's used there. (For starters, it can be read as "16 ÷ 30", and the current position isn't sixteen thirtieths of _anything_. So, we haven't really reduced confusion at all.)


Occasionally you'll see video tools that display the frame number as a fraction, but that's as a fraction of the _total frames_, and is a way of noting not only the frame _position_, but also the frame **_rate_** in one fell swoop. Which, actually, is kind of handy when editing on a timeline.

So, for the above position, what you'd see is:
#### <code>00:01:16 <sub>30/30</sub></code>

The time code for that second would range from
##### <code>00:01:16 <sub>1/30</sub></code>
to
##### <code>00:01:16 <sub>30/30</sub></code>
Then the next second would start back over again,
##### <code>00:01:17 <sub>1/30</sub></code>

One thing I like about that format is, it's one of the only situations where our 1-based frame numbering is actually a benefit, rather than a source of confusion. Showing the intra-second frames as `1/30` to `30/30`, at 30fps, is arguably _more_ natural than showing them as `0/29` - `29/29`, as opposed to when they're mixed into the time code where the opposite is true. (I'd still rather we used 0-based numbering, but as long as we're stuck with 1-based, might as well play to our strengths...)

Other ideas I've had, with varying degrees of relevance to this PR, regarding the playhead position display formatting:
* Use a period, instead: <code><b>00:01:16.30</b></code>
    * Disadvantages: Looks a bit too much like standard millisecond-based timecode, which is formatted <code><b>HH:MM:SS.UUU</b></code> where `UUU` is a 3-digit value in thousandths of a second.
* Use an underscore: <code><b>00:01:16_30</b></code>
    * Disadvantages: ...Just ugly?
* Make the frame number _smaller_ (as in my first example of frame-rate fractions)
    <code><b>00:01:16 <sup>30</sup></b></code>
* Use textual units labels as the separators, i.e.: <code><b>00</b><sub>h</sub> <b>01</b><sub>m</sub> <b>16</b><sub>s</sub> <b>30</b><sub>f</sub></b></code>
    * Disadvantages: Requires translations for each locale. Possibly doesn't work in all languages?
* Make the frame number a different <i>color</i> / <i>font</i> / <i>size</i> (the latter as in some of my examples above) — or some combination. Can be combined with any of the above, except probably for the period and/or underscore separators.
* Put the frame number in brackets (and possibly smaller): <code><b>00:01:16 <sup>[30]</sup></b></code>
    * Could also be combined with the <code>30/30</code> fractional-style display

And finally, separate from and/or in addition to any of the above:
* Add the raw _linear_ frame number to the display, in some way, to supplement the per-second frame counter. So that, when you're at that <b>00:01:16:30</b> position in a 30FPS video, you can _also_ see at a glance that you're at frame #<b>2280</b>.

That last one would actually be really valuable, since it's currently needed in the Export dialog if you want to export a section of the Timeline. If we're making users _input_ raw frame number values, we could at LEAST provide them with a convenient way to **discover** the correct values for those inputs. (We'd still have the problem that if they export at a different frame rate, the values would be _wrong_, but let's burn one bridge at a time...)

ferdnyc:
> Use a period, instead: `00:01:16.30`
> * Disadvantages: Looks a bit too much like standard millisecond-based timecode, which is formatted `HH:MM:SS.UUU` where `UUU` is a 3-digit value in thousandths of a second.

Ooh, I thought of another disadvantage to that one: Requires localization as well, probably, since the conventions for when you can get away with using period separators vs. comma separators vs. etc also vary from region to region.

ferdnyc:
> So, for the above position, what you'd see is:
> 
> #### `00:01:16 30/30`
> The time code for that second would range from
> 
> ##### `00:01:16 1/30`
> to
> 
> ##### `00:01:16 30/30`
> Then the next second would start back over again,
> 
> ##### `00:01:17 1/30`

BTW, it _seems_ like this would become messy when you're dealing with complex frame rates like `1001:30000` (aka `29.97`fps), but in reality it doesn't. It's just that occasionally, you have a second with only 29 frames instead of 30. So, leap-year rules, effectively. You'd imagine that it would be a lot more confusing than it actually turns out to be in practice.

ferdnyc:
> you have a second with only 29 frames instead of 30.

Yes, I should have said "a second with 31 frames instead of 30". *\*sigh\** Like I said: More difficult _conceptually_ than practically. :grin:

SuslikV:
What about?
`00:00:00 (00)`

There is timecodes in the industry of the television. Where `00:00:00:00` is normal.


> You mean 1-30, instead of 0-29. We don't add an extra frame to each second!

Of course.
But you know, sometimes I see 31 at the end in OpenShot's _Timeline_. And next frame is 02 then, instead of 01. Did you even seen this before?

ferdnyc:
> What about?
> `00:00:00 (01)`

I like it! Another one to add to the list.
 
> But you know, sometimes I see 31 at the end in OpenShot's _Timeline_. And next frame is 02 then, instead of 01. Did you ever seen this before?

Yep, in fact it turns out there was a bug in the `secondsToTime()` code ­— I just submitted #3174 to fix it. Thanks for bringing that up!

SuslikV:
Updated to top of the _release-20200105_ branch. New view (square brackets).

jonoomph:
@SuslikV Can we please rebase this PR on `develop`. I would like to delete the release branch. :smile:

ferdnyc:
@jonoomph Hm. Looks like we can do that, actually. (If you click Edit on the PR, the target branch becomes a dropdown list.)

ferdnyc:
Retargeted to develop.

SuslikV:
> @jonoomph Hm. Looks like we can do that, actually. (If you click Edit on the PR, the target branch becomes a dropdown list.)

I didn't see it before too. It is nice feature!

SuslikV:
Current _develop_ branch was changed. I will update the code later.

SuslikV:
Updated on top of the current _develop_.

jonoomph:
Closing PR. While there are lots of great suggestions, I feel like we'll revisit this soon with our new UI. THx.

PR #3174, opened 2020-01-13
WIP: Use Math.floor() to compute intra-second frame#

ferdnyc:
Turns out, because it's always adding 1 to the value computed, the JS `secondsToTime()` function should be using `Math.floor()`, _**not**_ `Math.round()`, to compute the displayed frame number. (Same as the other time components.)

The rounding was pushing the frame# one higher than its max should be, so some seconds would run from frames 2 - 31, instead of 1-30, at 30fps or 29.97fps. (Worse at 30fps, actually.)

Pointed out by @SuslikV in https://github.com/OpenShot/openshot-qt/pull/3170#issuecomment-573645461

### DIFF
--- a/src/timeline/js/functions.js
+++ b/src/timeline/js/functions.js
@@ -1,6 +1,6 @@
 /**
  * @file
- * @brief Misc Functions used by the OpenShot Timeline 
+ * @brief Misc Functions used by the OpenShot Timeline
  * @author Jonathan Thomas <jonathan@openshot.org>
  * @author Cody Parker <cody@yourcodepro.com>
  *
@@ -38,12 +38,12 @@ function findElement(arr, propName, propValue) {
 
 // Get the height of the track container (minus bottom margin of last track)
 function getTrackContainerHeight() {
-	
+
 	var track_margin = 0;
 	if ($(".track").length)
 		if ($(".track").css("margin-bottom"))
 			track_margin = parseInt($(".track").css("margin-bottom").replace("px",""));
-	
+
 	return $("#track-container").height() - track_margin;
 }
 
@@ -73,10 +73,10 @@ function drawAudio(scope, clip_id){
 
         // Clear canvas
         ctx.clearRect(0, 0, ctx.canvas.width, ctx.canvas.height);
-        
+
         color = "#2a82da"; // rgb(42,130,218)
         color_transp = "rgba(42,130,218,0.5)";
-        
+
         ctx.strokeStyle = color;
 
         // Find the midpoint
@@ -89,15 +89,15 @@ function drawAudio(scope, clip_id){
         ctx.moveTo(0, mid_point);
         ctx.lineTo(audio_canvas.width(), mid_point);
         ctx.stroke();
-        
-        
+
+
         var sample = 0.0,
             // Variables for accumulation
             avg = 0.0,
             avg_cnt = 0,
             max = 0.0,
             last_x = 0;
-        
+
         // Go through all of the (reduced) samples
         // And whenever enough are "collected", draw a block
         for (var i = start_sample; i < end_sample; i++) {
@@ -109,19 +109,19 @@ function drawAudio(scope, clip_id){
             avg += sample;
             avg_cnt++;
             max = Math.max(max, sample);
-            
+
             if(x >= last_x + block_width || i == end_sample-1){
                 // Block wide enough or last block -> draw it
 
                 // Draw the slightly transparent max-bar
                 ctx.fillStyle = color_transp;
                 ctx.fillRect(last_x, mid_point, x-last_x, -(max * scale));
-                
+
                 // Draw the fully visible average-bar
                 ctx.fillStyle = color;
                 ctx.fillRect(last_x, mid_point, x-last_x, -(avg/avg_cnt * scale));
-                
-                // Reset all the variables for accumulation 
+
+                // Reset all the variables for accumulation
                 last_x = x;
                 avg = 0;
                 avg_cnt = 0;
@@ -152,7 +152,7 @@ function secondsToTime(secs, fps_num, fps_den)
 	var week = Math.floor(day/7);
 	var day = day % 7;
 
-	var frame = Math.round((milli / 1000.0) * (fps_num / fps_den)) + 1;
+	var frame = Math.floor((milli / 1000.0) * (fps_num / fps_den)) + 1;
     return { "week":padNumber(week,2), "day":padNumber(day,2), "hour":padNumber(hour,2), "min":padNumber(min,2), "sec":padNumber(sec,2), "milli":padNumber(milli,2), "frame":padNumber(frame,2) };
 }
 
@@ -202,7 +202,7 @@ var bounding_box = Object();
 function setBoundingBox(scope, item){
 	var vert_scroll_offset = $("#scrolling_tracks").scrollTop();
 	var horz_scroll_offset = $("#scrolling_tracks").scrollLeft();
-	
+
     var item_bottom = item.position().top + item.height() + vert_scroll_offset;
     var item_top = item.position().top + vert_scroll_offset;
     var item_left = item.position().left + horz_scroll_offset;
@@ -221,7 +221,7 @@ function setBoundingBox(scope, item){
         if (item_left < bounding_box.left) bounding_box.left = item_left;
         if (item_bottom > bounding_box.bottom) bounding_box.bottom = item_bottom;
         if (item_right > bounding_box.right) bounding_box.right = item_right;
-        
+
         // compare height and width of bounding box (take the largest number)
         var height = bounding_box.bottom - bounding_box.top;
         var width = bounding_box.right - bounding_box.left;
@@ -305,7 +305,7 @@ function moveBoundingBox(scope, previous_x, previous_y, x_offset, y_offset, left
             bounding_box.right -= nearby_offset;
             snapping_result.left -= nearby_offset;
         }
-		
+
     } else {
 	    // Hide snapline
 		scope.HideSnapline();


ferdnyc:
I've set this to WIP, because it appears that it's not quite that simple. With this code in place, now I'm seeing jumps of 2 frame-numbers between some frames, and other times two frames in a row will have the same frame number.

I tested the JS `secondsToTime()` function with a script, though, and everything came out exactly right, with this change in place. Which makes me think that the issue is with the "fuzzing" code for the positioning of the timeline. (This stuff):

https://github.com/OpenShot/openshot-qt/blob/5e90b39e2e781b1b159d357e5f63aefd4c685fec/src/timeline/js/controllers.js#L302-L309

_With that code in place_, rounding the frame number rather than `floor()`ing it may be "more" "correct", meaning it gives the expected result more of the time. Because the position values used by OpenShot aren't exact multiples of the frame rate.

The problem is that it's still not really _correct_-correct, which means it will lead to odd values in the position display sometimes. I'm just not sure what to do about that.

ferdnyc:
Actually, the more I think about it the more I think those should all be `Math.floor()` as well. If we're going to adjust the position to conform to frame boundaries, it should be an adjustment to the start of the frame we're _currently_ in, not the "nearest".

ferdnyc:
Looking through the history, that code (just an example, there are other similar bits of code throughout the JavaScript) evolved from this:

```js
 $scope.PreviewClipFrame = function(clip_id, position_seconds) {
	  // Determine frame
	  var frames_per_second = $scope.project.fps.num / $scope.project.fps.den;
	  var frame = (position_seconds * frames_per_second) + 1;
```

To this:
```js
  $scope.PreviewClipFrame = function(clip_id, position_seconds) {
	  // Determine frame
	  var frames_per_second = $scope.project.fps.num / $scope.project.fps.den;
	  var frame = Math.round(position_seconds * frames_per_second) + 1;
```

To what we have today.

But the problem is, I think the `Math.round()` was _**always**_ wrong — it should've been `Math.floor()` right from the start. (Well, right from the middle, where `Math.round()` was added in.) That's what caused the perceived issues with frame accuracy, that led to the current convoluted version.

SuslikV:
`frame = Math.round(position_seconds_rounded * frames_per_second) + 1;`

in python I wrote similar code as:

`return int(round(secondsF * fps_float, 6)) + 1`

ferdnyc:
> `frame = Math.round(position_seconds_rounded * frames_per_second) + 1;`
> 
> in python I wrote similar code as:
> 
> `return int(round(secondsF * fps_float, 6)) + 1`

Not actually similar at all, though. Because with a precision set, `round()` will still produce a floating-point number, and `int()` is effectively the same as `floor()` — so that's basically identical to the `Math.floor()` version, really. The only time it'll round up is when the result's decimal portion is > `.995` or so. (For 3-digit frame numbers.)

SuslikV:
So, what caused actual "jumps" in display number?

And it not similar, of course - it is exact code that returns number from time.

ferdnyc:
> So, what caused actual "jumps" in display number?

Still looking into that, but basically AFAICT a lot of it comes from imprecision in floating-point math — every time a position value is converted from a frame number to a time in seconds and back again, there's a chance the same value won't be arrived at.

Which is actually why, I now realize, it _is_ necessary to round when converting seconds into frame numbers — the precision issues mean that you can end up just **slightly** below your target frame#, and `floor()` (or truncation) will drop you an entire frame early. So the choices are:
1. Use `round()` for the _final_ conversion to integer
1. Add a small value (half a frame or less), and then use `floor()`, relying on the addition to push you over your target in all instances.

The second option (which we don't take) may actually be preferable, because it means that you can also _overshoot_ your target by a lot more — if the value you add is only `.2`, say, then everything from `(N-1).80` through `N.79` will become `N`. As opposed to `round()` which will transform `(N-1).5` through only `N.5` into `N`. — but if you end up on any `X.5` or `X.6` you should _probably_ snap to `X` instead, when doing frame-snapping.

Observe, in Python:
```python
>>> fps = [30000,1001]
>>> for f in range(505,515):
...     print( ( (f - 1) / (fps[0]/fps[1]) * (fps[0]/fps[1]) + 1 ))
... 
505.0
506.0
506.99999999999994
508.00000000000006
509.0
510.0
511.0
511.99999999999994
513.0
514.0
```
Your Python actually suffers from the same issue, depending on the precision of the stored time-in-seconds value:
```python
>>> for f in range(505,515):
...     s = round((f - 1) / (fps[0]/fps[1]), 3)
...     fs = int(round(s * (fps[0]/fps[1]), 6)) + 1
...     print(f" {f} => {s} => {fs}")
... 
 505 => 16.817 => 505
 506 => 16.85 => 505
 507 => 16.884 => 507
 508 => 16.917 => 508
 509 => 16.95 => 508
 510 => 16.984 => 510
 511 => 17.017 => 511
 512 => 17.05 => 511
 513 => 17.084 => 513
 514 => 17.117 => 513
>>> for f in range(505,515):
...     s = round((f - 1) / (fps[0]/fps[1]), 6)
...     fs = int(round(s * (fps[0]/fps[1]), 6)) + 1
...     print(f" {f} => {s} => {fs}")
... 
 505 => 16.8168 => 505
 506 => 16.850167 => 506
 507 => 16.883533 => 506
 508 => 16.9169 => 508
 509 => 16.950267 => 509
 510 => 16.983633 => 509
 511 => 17.017 => 511
 512 => 17.050367 => 512
 513 => 17.083733 => 512
 514 => 17.1171 => 514
>>> for f in range(505,515):
...     s = round((f - 1) / (fps[0]/fps[1]), 8)
...     fs = int(round(s * (fps[0]/fps[1]), 6)) + 1
...     print(f" {f} => {s} => {fs}")
... 
 505 => 16.8168 => 505
 506 => 16.85016667 => 506
 507 => 16.88353333 => 507
 508 => 16.9169 => 508
 509 => 16.95026667 => 509
 510 => 16.98363333 => 510
 511 => 17.017 => 511
 512 => 17.05036667 => 512
 513 => 17.08373333 => 513
 514 => 17.1171 => 514
```
Unless the precision of the stored value is at least _eight_ decimal places, you'll get wrong frame numbers converting back if you use `int()`. `round()` solves it for any precision:
```python3
>>> for f in range(505,515):
...     s = round((f - 1) / (fps[0]/fps[1]), 2)
...     fs = round(round(s * (fps[0]/fps[1]), 6)) + 1
...     print(f" {f} => {s} => {fs}")
... 
 505 => 16.82 => 505
 506 => 16.85 => 506
 507 => 16.88 => 507
 508 => 16.92 => 508
 509 => 16.95 => 509
 510 => 16.98 => 510
 511 => 17.02 => 511
 512 => 17.05 => 512
 513 => 17.08 => 513
 514 => 17.12 => 514
```
...With the caveat being, when your time indexes _aren't_ on frame boundaries in exact inverse-FPS intervals, `round()` will bias how the time indexes map to frames at the current rate.

ferdnyc:
However, `secondsToTime()` isn't converting _whole_ time values into frame numbers, it's only converting remainders — fractional seconds — into frame numbers _within the current second_. `Math.floor()` **SHOULD** be correct, for _that_ application, because the imprecisions that can shift our time values slightly in front of the frame boundary are already corrected for. (IOW, even if `16.983633` seconds is just _slightly_ before the position of frame `510`, at `29.97fps`, `.983633` seconds is still _well_ above the frame boundary: in fact, it's almost 29.5 frames into the second:
```python
>>> 16.98363333 * (fps[0]/fps[1])
508.9999999000999
>>> 0.98363333 * (fps[0]/fps[1])
29.47952037962038
```
So, truncation/flooring _**should**_ be correct, when computing intra-second frame numbers, rather than absolute frame numbers —  just like it is for all of the other time components.

The fact that `Math.floor()` doesn't produce correct indexes seems to indicate that our time signature _inputs_ to `secondsToTime()` aren't correctly mapped to frame boundaries, still.

ferdnyc:
The reason I say `Math.floor()` should be correct for intra-second frames: There are no slippages for values under 1 second, even operating at exact precision...
```python3
>>> for f in range(0,30):
...     print( ( (f - 1) / (fps[0]/fps[1]) * (fps[0]/fps[1]) + 1 ))
... 
0.0
1.0
2.0
3.0
4.0
5.0
6.0
7.0
8.0
9.0
10.0
11.0
12.0
13.0
14.0
15.0
16.0
17.0
18.0
19.0
20.0
21.0
22.0
23.0
24.0
25.0
26.0
27.0
28.0
29.0
```
The values start slipping only once you get up into higher values for frame number and/or seconds, where the whole-value component begins to rob the decimal value of its full precision.

ferdnyc:
ARGH, ignore that last comment, it's an invalid test. I forgot that Python seems to reduce math where it can. So an equation like `( (N / fps[1]/fps[0]) * (fps[0]/fps[1]) )`, it will actually reduce to `N`. The actual precision of the conversion will depend on how you perform the math.

(Which I _also_ find interesting. Check out the difference between these:)
```python3
>>> for f in range(0,30):
...     print( ( (f-1) * (fps[1]/fps[0]) * (fps[0]/fps[1]) + 1))
... 
0.0
1.0
2.0
3.0
4.0
5.0
6.0
7.0
8.0
9.0
10.0
11.0
12.000000000000002
13.0
14.0
15.000000000000002
16.0
17.0
18.0
19.0
20.0
21.0
22.000000000000004
23.000000000000004
24.000000000000004
25.0
26.0
27.0
28.0
29.000000000000004
>>> for f in range(0,30):
...     print( ( (f-1) / (fps[0]/fps[1]) / (fps[1]/fps[0]) + 1))
... 
0.0
1.0
2.0
3.0
4.0
5.0
6.0
7.0
8.0
9.0
10.0
11.0
11.999999999999998
13.0
14.0
15.0
16.0
17.0
18.0
19.0
20.0
21.0
21.999999999999996
22.999999999999996
23.999999999999996
25.0
26.0
27.0
28.0
29.0
```

_Mathematically_, they're identical. Practically, not so much.

ferdnyc:
Based on those last tests, I'd say one issue is that we should _always_ be multiplying by the inverse of the FPS fraction, _never_ dividing by the fraction _or_ by the floating-point expression of the FPS. Doing that introduces floating-point imprecision.

The whole reason FPS values are expressed in fractions like `30000/1001` is that timing circuits back in the day couldn't even _do_ floating-point math, they had to rely exclusively on integer arithmetic. Which is still the best approach, when dealing with frame numbers.

ferdnyc:
cc: @jonoomph , on my previous comment in particular.

SuslikV:
> ```python
> s = round((f - 1) / (fps[0]/fps[1]), 6)
> ```

Where did you get this math?
the OpenShot uses this:
https://github.com/OpenShot/openshot-qt/blob/d0397096b280d52152e07b7620d77d614bebca4a/src/windows/models/properties_model.py#L170-L179

Floating point nature is where you should take care of the precision of the result (or be aware of it).

SuslikV:
In my point of view, If you don't know how floating point operations is done in hardware, next general rules should apply:
1) If result is float - never do rounding
2) If result is integer - look at the float result first, maybe it is not perfect for rounding

SuslikV:
I did some tests.

The 31-th frame (when I see it on the screen at some cursor position) with 30 fps profile, at about 1 second time in browser of OpenShot's timeline, it actually has:
`milli = 990`
and raw float (without rounding)
`frame = 30.7`

the 01-th frame at about 1 second time:
`milli = 0`
and raw float (without rounding)
`frame = 1`

the 30-th frame at about <1 second time (the highest value I able to aim with mouse at):
`milli = 980`
and raw float (without rounding)
`frame = 30.4`

Up to 10 signs after the point is preserved in float `milli` at length of the timeline 5 minutes at 1 second scale.
Up to 9 signs after the point is preserved in float `milli` at length of the timeline 60 minutes at 1 second scale.

And in range 0-60 minutes, only `milli = 990` produces 31-th frame display. For some values, the `milliseconds % 1000` returns float, like: `milli = 966.6666666666279`. (about 5 minutes position).

SuslikV:
The notation of frames in OpenShot should be mentioned somewhere, from the working code I can only assume that it was intended to be like this
`hh:mm:ss (frames) -> [(frame_number - 1) * 1/fps; frame_number * 1/fps)`, see :
```
00:00:00 (01) -> [0.0;  1 * 1/fps)
00:00:00 (02) -> [1 * 1/fps;  2 * 1/fps)
00:00:00 (03) -> [2 * 1/fps;  3 * 1/fps)
...
00:00:00 (28) -> [27 * 1/fps;  28 * 1/fps)
00:00:00 (29) -> [28 * 1/fps;  29 * 1/fps)
00:00:00 (30) -> [29 * 1/fps;  30 * 1/fps)
...
00:00:01 (01) -> [30 * 1/fps;  31 * 1/fps)
00:00:01 (02) -> [31 * 1/fps;  32 * 1/fps)
00:00:01 (03) -> [32 * 1/fps;  33 * 1/fps)
```
[ means includes the number, or ">="
) means not includes the number, or "<"
and the values represents the time in seconds.

To add my 2 cents for 30000/1001 fps calculations, analyzing last 3 digits after the point for time in seconds is not enough even for floating calculations, look:
```
 999.99900 sec -> is       100.00% of 29970-th frame already played.
 999.99901 sec -> is about   0.03% of 29971-th frame already played.
1000.03236 sec -> is about  99.98% of 29971-th frame already played.
1000.03237 sec -> is about   0.01% of 29972-th frame already played.
...
1000.96663 sec -> is about  99.99% of 29999-th frame already played.
1000.96664 sec -> is about   0.02% of 30000-th frame already played.
1000.99999 sec -> is about  99.97% of 30000-th frame already played.
1001.00000 sec -> is       100.00% of 30000-th frame already played.
```
**Conclusion.**
Neither floor() nor round() nor trunc() cannot apply as current fix. But different combination of the functions can get right result that will has nice precision. `sec - trunc(sec)` is good remainder to show only milliseconds, but `secs`, as most precise variable, should be used for frame number calculations, without simplifications.

jonoomph:
This is an interesting discussion for sure! However, as mentioned a few times above, none of the solutions work perfectly, and OpenShot converts time to frame #'s in many places, and vice verse. I'm leaning towards a libopenshot solution, perhaps a utility class/function, which can convert between time and frame numbers, consistently, and then we can implement that everywhere we needed it.

That doesn't really solve the problem, but it will provide a consistent place for this logic to happen, and a single place for us to tweak and improve in the future, as we find solutions that are appropriate. 

In general though, I like the idea of math.floor vs round, since it makes sense we are "inside" a specific frame, vs just finding the closest frame #.

jonoomph:
@ferdnyc However, I'm not sure what to do about this PR though. This might need to be closed as well, if we don't ever plan on merging it.

ferdnyc:
Yeah, I've got about 0% confidence that this is the correct approach, anymore. (Maybe it _should_ be, but the use of frame numbers _mixed_ with time values makes things... complex, because not every `HH:MM:SS:01` corresponds to `HH:MM:SS.000` milliseconds. And how far _after_ that time it actually lands can affect what frame you want to say you're in, at the start of that second time-wise.

ferdnyc:
(Really, the issue is that there are times when, if we're at `NN` fps, so much of `HH:MM:SS:NN` spills over into `HH:MM:SS+1` that _that_ frame arguably should be called `HH:MM:SS+1:01` — but we don't have any way of supporting that right now, since the `floor()` can only go to the start of the current second of time. For `floor()` to do what we wanted, it'd have to be able to go _below_ the `0ms` point of the current second, when a frame straddles the seconds divider and its "floor" is so low it occupies the previous second.)

PR #3178, opened 2020-01-15
WIP: New add-track implementation

ferdnyc:
**NOTE** @jonoomph : I eventually get around to having a question for you, at the end of this ramble.

(Yeah, yeah, I know, I'm supposed to be working on the libopenshot QImage format conversion. And, I am! But, this is in a state where I need input on it, but it's otherwise ready, so I wanted to get it out there for review in the meantime.)

This PR contains a revised implementation of `actionAddTrack_trigger`, `actionAddTrackAbove_trigger`, and `actionAddTrackBelow_trigger`, which tweaks how they handle the unusual condition where layer numbering has become compressed enough that there's no room between two layers to insert a new one (in numerical order).

Previously, when this happened, the `add..._trigger()` callbacks would do a complex shifting operation, to slide layer numbers around and make room for an insertion. The problem is, they could potentially do this for one insertion, only to have to do it _again_ for the next one.

In this new code, a new utility method `renumber_all_layers()` is called the moment this extraordinary condition is detected. As the name implies, `renumber_all_layers()` reindexes the **entire** layer stack, spreading each layer back out to the default initial distance of `1000000` per layer (leaving a gap for the newly-created layer, as well). This way, when that condition _does_ occur, for the cost of slightly more complexity in the response, we ensure that it won't be encountered **again** anytime soon.

The new algorithm works well, in my testing. The only remaining issue is connecting it in to the undo/redo history.

The old code didn't make any accommodations for the history stack, meaning it would create separate history entries for each track, clip, and transition it renumbered, which is a problem to begin with. But, since now the code is touching **every** Timeline item, it's completely unworkable. 

So, right now I switch off history tracking (`self.updates.ignore_history = True`) for the entire renumbering process, only switching it back on right before the _new_ track for insertion is created and `Track.save()`d. But that's really not correct either, from the POV of the users' undo history.

### (Question here:)
So I wanted to ask @jonoomph, is it possible to set up a single Undo item that has basically the **entire** project data as its `old_values`, and what would be the best way to do that? Obviously it's not really what you'd want to be doing regularly, but this _is_ an exceptional condition that should rarely occur. And, due to the complete renumbering of all Timeline elements, it's basically the _reality_ of the situation, plain and simple — in these (exceedingly rare) cases.


ferdnyc:
P.S> A big part of the impetus for doing this is a bug report / feature request, which of course now I can't find _anywhere_, regarding the rearranging of layers. Getting the insertion logic ironed out, and having an algorithm for renumbering that's less ad-hoc than the cut-and-paste duplicates that previously existed in the `addTrackAbove/addTrackBelow` callbacks, is a necessary prerequisite to implementing any sort of layer-rearranging (aka Track reordering) algorithm.

jonoomph:
We don't currently have a method for wrapping up **all changes** into a single undo/redo action. We do have a way to stop and start the undo/redo system, which is similar but not exactly what we need. We use that method when a user is sliding properties around (waiting until they are done) before creating the undo/redo action. But I imagine we could add in this functionality, so it's possible for an undo/redo action to contain the entire project JSON, for really **extreme** changes like reordering tracks, adding multiple items to the timeline, or splitting clips, all of those add multiple undo/redo actions currently.

jonoomph:
But let's target this PR for our next release (i.e. develop branch), including any changes to the undo/redo system to accommodate a full project undo/redo. How does that sound?

ferdnyc:
@jonoomph Absolutely, makes sense.

I took a quick look at how Qt does it in their [QUndoStack](https://doc.qt.io/qt-5/qundostack.html). There are basically two parts to it, an implicit implementation and an explicit one:
> To support command compression, [`QUndoCommand`](https://doc.qt.io/qt-5/qundocommand.html) has an [`id()`](https://doc.qt.io/qt-5/qundocommand.html#id) and the virtual function [`mergeWith()`](https://doc.qt.io/qt-5/qundocommand.html#mergeWith). These functions are used by [`QUndoStack::push()`](https://doc.qt.io/qt-5/qundostack.html#push).
> 
> To support command macros, a QUndoCommand object can have any number of child commands. Undoing or redoing the parent command will cause the child commands to be undone or redone. A command can be assigned to a parent explicitly in the constructor. In this case, the command will be owned by the parent.
> 
> The parent in this case is usually an empty command, in that it doesn't provide its own implementation of [`undo()`](https://doc.qt.io/qt-5/qundocommand.html#undo) and [`redo()`](https://doc.qt.io/qt-5/qundocommand.html#redo). Instead, it uses the base implementations of these functions, which simply call `undo()` or `redo()` on all its children. The parent should, however, have a meaningful [`text()`](https://doc.qt.io/qt-5/qundocommand.html#text).
>
> Another way to create macros is to use the convenience functions [`QUndoStack::beginMacro()`](https://doc.qt.io/qt-5/qundostack.html#beginMacro) and [`QUndoStack::endMacro()`](https://doc.qt.io/qt-5/qundostack.html#endMacro).

Given that our undo/redo is managed centrally (the history manager is responsible for both tracking and implementing each history action, and none of the rest of the code has any understanding of the stack), a slightly modified implementation of implicit transaction functionality, triggered by an explicit request similar to Qt's `beginMacro()`, would probably fit well.

> #### `void QUndoStack::beginMacro(const QString &text)`
> Begins composition of a macro command with the given text description.
> 
> An empty command described by the specified text is pushed on the stack. Any subsequent commands pushed on the stack will be appended to the empty command's children until `endMacro()` is called.
> 
> Calls to `beginMacro()` and `endMacro()` may be nested, but every call to `beginMacro()` must have a matching call to `endMacro()`.
> 
> While a macro is being composed, the stack is disabled. This means that:
> 
> * `indexChanged()` and `cleanChanged()` are not emitted,
> * `canUndo()` and `canRedo()` return false,
> * calling `undo()` or `redo()` has no effect,
> * the undo/redo actions are disabled.
> 
> The stack becomes enabled and appropriate signals are emitted when `endMacro()` is called for the outermost macro.
> ```c++
> stack.beginMacro("insert red text");
> stack.push(new InsertText(document, idx, text));
> stack.push(new SetColor(document, idx, text.length(), Qt::red));
> stack.endMacro(); // indexChanged() is emitted
> ```
> This code is equivalent to:
> ```c++
> QUndoCommand *insertRed = new QUndoCommand(); // an empty command
> insertRed->setText("insert red text");
> 
> new InsertText(document, idx, text, insertRed); // becomes child of insertRed
> new SetColor(document, idx, text.length(), Qt::red, insertRed);
> 
> stack.push(insertRed);
> ```

So, instead of setting `get_app().updates.ignore_history = True` at the start of `renumber_all_layers`, like I'm doing now, it could call `app.updates.begin_transaction()`. That would tell the update manager to:
* implicitly suspend update tracking
* allocate a transaction ID
* create a separate transaction stack
* redirect all subsequent updates to that stack until the transaction is completed

Every `{Track,Clip,Transition}.save()` in `renumber_all_layers()` then pushes a change onto the transaction stack.

A call to `get_app().updates.commit_transaction()` replaces the `get_app().updates.ignore_history = False` at the end of `relabel_all_tracks()`. (It should also come _after_ the creation and insertion of the new track, since that's now part of the transaction.<sup>1</sup>)

When `commit_transaction()` is called, the update manager:
* collects all of the changes on the transaction stack
* packages them up together into a single undo action
* pushes it onto the main stack
* deletes the transaction stack
* re-enables normal history tracking and undo/redo.

1 — (And, wow, now I'm really glad I made the track insertion part of `renumber_all_layers()` itself. Originally it just left a gap in the numbering, and the new track would be inserted by the caller after it returned, same as the non-renumbering case. But when I realized that meant the layer number for the new track had to be passed back to the caller, I decided to skip that and just do the insertion right there, so that `renumber_all_layers()` would be more... well... transactional.)

ferdnyc:
(Note to any and all interested parties following this: The moment we _do_ have `begin_` and `end_transaction()`, we should also wire up the **Slice > Keep Both Sides** functionality as a transaction. That way each of those won't create _three_ undo entries per sliced track, like they do now.)

ferdnyc:
I think there may be some other Clip context-menu actions that also create multiple undo entries. Someone should really make a list. **#NotIt**

jonoomph:
@ferdnyc I love the idea of creating a undo/redo "transaction", which could contain many, many smaller changes. Love it!

jonoomph:
@ferdnyc Also, I've tried to step through this code in my head many times and keep getting distracted, but I'm just going to merge it, and test it locally. :+1: My brain has certain limitations...

PR #3180, opened 2020-01-17
Fix main menu actions state (UI)

SuslikV:
After the main menu _Edit>Clear History_ clicked the main menu _Edit>Undo_ action stays enabled.

This change is attempt to fix the UI inconsistency to the available features.

**Edit:** designed to work with the Curve Editor: https://github.com/OpenShot/openshot-qt/pull/3084

### DIFF
--- a/src/classes/updates.py
+++ b/src/classes/updates.py
@@ -191,6 +191,7 @@ class UpdateManager:
         """ Reset the UpdateManager, and clear all UpdateActions and History. This does not clear listeners and watchers. """
         self.actionHistory.clear()
         self.redoHistory.clear()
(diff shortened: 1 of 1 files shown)


ferdnyc:
Oh. Heh, I hadn't even seen this, and included the same fix in #3181.

SuslikV:
It was included into the https://github.com/OpenShot/openshot-qt/pull/3187 not the 3181

I have made few runs of it and it works fine for me. And because your PR also includes this fix as separate commit, I'm going to close this in favor of #3187. But, if for some reason, it wouldn't be merged... I will make new one again.

Issue #3171, opened 2020-01-10
'NoneType' object is not subscriptable


technotrance911:
**Describe the bug:**
Hello!
I can not open the saved project. Error: 'NoneType' object is not subscriptable

**Steps to reproduce the behavior:**
I did my project, and at this time the program began to behave strangely. Objects stopped selecting. Objects did not respond to a mouse click. I close program. Then start program again and I began to get an error when opening a project. 'NoneType' object is not subscriptable

**System Details:**
- OpenShot Version 2.4.4-2.fc28
- Operating System Fedora 28 x86_64
- Qt 5.11.3
- Python 3.6.8

**Log Files:**
[openshot_qt.zip](https://github.com/OpenShot/openshot-qt/files/4045525/openshot_qt.zip)

technotrance911:
Is there any way to restore the project? Even without the latest changes. I did it for several days.

SuslikV:
It's depending on how much data were damaged. Where is you project file? Can you share it?

a3nm:
I had the same problem with Openshot 2.4.3 from Debian. I'm not sure what caused the problem, as I could re-do the same steps in a new file without triggering the problem.

I attach the project file which was obtained by saving and raises the error when opening. [video.osp.zip](https://github.com/OpenShot/openshot-qt/files/4393635/video.osp.zip)

Thanks!

SuslikV:
@a3nm your repaired project file:

About 3KB to down load, CRC of the zip archive MD5:CE373548057B39A67BDA9A1D754C7EC7
[video - repaired.zip](https://github.com/OpenShot/openshot-qt/files/4394912/video.-.repaired.zip)

v2.5.1 compatible:
About 3KB to down load, CRC of the zip archive MD5:FF5123CD992CC21D99E6625BDDA92D9D
[video - repaired-251compatible.zip](https://github.com/OpenShot/openshot-qt/files/4394922/video.-.repaired-251compatible.zip)

When you complete your project, consider to upgrade to the latest version available (v2.5.1)

**Info for the developers:**
project file has`null` object after 16750 position in the file (end of the `"clips"` section)
Also, may suffer form the https://github.com/OpenShot/openshot-qt/issues/3227 that was solved in v2.5.1

a3nm:
[...]

PR #3181, opened 2020-01-17
Protect filter() against damaged projects

ferdnyc:
In #3171 a user uploaded logs that contained  a _bunch_ of these errors:
```
 main_window:INFO main::addSelection: item_id: , item_type: effect, clear_existing: True
  File "/usr/lib/python3.6/site-packages/openshot_qt/windows/views/properties_tableview.py", line 699, in select_item
    clip = Clip.get(id=self.item_id)
  File "/usr/lib/python3.6/site-packages/openshot_qt/classes/query.py", line 162, in get
    return QueryObject.get(Clip, **kwargs)
  File "/usr/lib/python3.6/site-packages/openshot_qt/classes/query.py", line 135, in get
    matching_objects = QueryObject.filter(OBJECT_TYPE, **kwargs)
  File "/usr/lib/python3.6/site-packages/openshot_qt/classes/query.py", line 110, in filter
    if key in child and not child[key] == value:
TypeError: argument of type 'NoneType' is not iterable
```

It doesn't appear they're fatal, though they may have _led_ to the subsequent project-loading issues experienced. Regardless, it's a good idea to protect against them. This should.

The unit tests still pass, so that should hopefully be a good sign that it at least doesn't make things _worse_.

### DIFF
--- a/src/classes/query.py
+++ b/src/classes/query.py
@@ -1,26 +1,26 @@
-""" 
+"""
  @file
  @brief This file can easily query Clips, Files, and other project data
  @author Jonathan Thomas <jonathan@openshot.org>
- 
+
  @section LICENSE
- 
+
  Copyright (c) 2008-2018 OpenShot Studios, LLC
  (http://www.openshotstudios.com). This file is part of
  OpenShot Video Editor (http://www.openshot.org), an open-source project
  dedicated to delivering high quality video editing and animation solutions
  to the world.
- 
+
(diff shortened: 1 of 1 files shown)


SuslikV:
The log import part
https://github.com/OpenShot/openshot-qt/blob/f4c3aa6f82cd44c9705e3645c55b1fdf00ce6bdc/src/classes/query.py#L33
not required, I think.

ferdnyc:
Whoops, did I leave that in? Quite right, thanks. Fixed.

Issue #3185, opened 2020-01-18
Slice issue with media files opened from command line


Raph-2018:
I have just tried the release candidate v2.5.0-1578960720 on linux xubuntu 14.04, and I noticed the following problem:

If I open a media file with Openshot from the file manager or via command line, and then I press Ctrl+K to slice it on the timeline, the video file is sliced a few seconds before the position of the red marker (see image attached). Of course, this is very annoying and prevents the user from being able to cut the video at an accurate time (moreover, the lapse of time seems to vary depending on the position of the red marker on the video).

Strangely enough, there is no such behavior when the media file is opened from inside Openshot.

I guess the issue could be related to the bug #2162 or #2355. I remember noticing a similar behavior with version 2.4.4 (the bug was so annoying I had to downgrade to the previous version).

![Screenshot](https://user-images.githubusercontent.com/42306896/72668188-6e2e8e80-3a24-11ea-8c8f-a062d0209e7d.jpg)

SuslikV:
The screenshot says that cursor at `00:02:30` and frame `15`. What log-file says?

Raph-2018:
@SuslikV: Here is what the log says when I press Ctrl+K:

[Log.txt](https://github.com/OpenShot/openshot-qt/files/4081520/Log.txt)

Raph-2018:
I just noticed another strange thing: if I click on the menu "slice all - keep both" instead of pressing Ctrl+K, the video is properly sliced!

SuslikV:
I think it is not affected by the keypress at all. But for sure, there is difference in how the profile applied. 
The default one called from the recover_backup() function...

`get_app().project.load("")`

What a mess. This string is also needed in `app.py` right before the file import:
```
                # Auto import media file
                self.window.filesTreeView.add_file(path)
```

PR #3186, opened 2020-01-19
Apply the profile settings when file opened from command line

SuslikV:
Fixes: https://github.com/OpenShot/openshot-qt/issues/3185

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -231,7 +231,8 @@ class OpenShotApp(QApplication):
                 # Auto load project passed as argument
                 self.window.OpenProjectSignal.emit(path)
             else:
-                # Auto import media file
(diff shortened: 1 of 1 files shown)


ferdnyc:
Mmm, I'm not sure how I feel about command-line media imports _preventing_ backup recovery from happening.

But, currently they **do** prevent recovery, so that's unrelated really. And without `recover_backup()` being run, nothing will load the default project data, so this fix makes sense. The issue is especially obvious if you have a user default project file — I do, to start my projects out with three tracks instead of five, but when I import media on the command line I end up with 5 tracks.

This PR fixes that. Good catch, thanks!

ferdnyc:
I'm going to merge this, as it's a small  change that fixes a clear bug. I'll also push it to the release branch.

PR #3187, opened 2020-01-19
Add pending_action tracking to update manager

ferdnyc:
This adds a new `self.pending_action` variable to the update manager, as described in https://github.com/OpenShot/openshot-qt/pull/3179#issuecomment-575932045 and followups, to correct an issue with the update state as identified by @SuslikV in that PR. <strike>It also corrects an issue with the update manager not correctly disseminating the new state of the history stack after making certain changes to it. (`reset()`, `apply_last_action_to_history()`)</strike> Handled in #3194.

In my testing, this corrects the bugs targeted by PR #3179 _without_ needing to change the property manager, IOW it represents an alternative to that PR that (IMHO) solves the issue where it actually occurs (and potentially for other possible trigger points elsewhere in the code, as well). 

However, there's also no harm in merging both this PR and #3179, for a little extra protection.

### DIFF
--- a/src/classes/updates.py
+++ b/src/classes/updates.py
@@ -28,16 +28,16 @@
  """
 
 from classes.logger import log
-from classes import info
 import copy
-import os
 import json
+
 
 class UpdateWatcher:
     """ Interface for classes that listen for 'undo' and 'redo' events. """
 
     def updateStatusChanged(self, undo_status, redo_status):
-        """ Easily be notified each time there are 'undo' or 'redo' actions available in the UpdateManager. """
+        """ Easily be notified each time there are 'undo' or 'redo' actions
+        available in the UpdateManager. """
         raise NotImplementedError("updateStatus() not implemented in UpdateWatcher implementer.")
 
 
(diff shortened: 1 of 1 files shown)


ferdnyc:
(This change should also leave us in a better place for implementing combined / multi-change update transactions, as described in https://github.com/OpenShot/openshot-qt/pull/3178#issuecomment-575375343, since `self.pending_action` can easily become a transaction _stack_ instead of a variable holding only a single action, and `apply_last_action_to_history()` can easily become `commit_transaction()`.)

ferdnyc:
Hey, while we're on this topic ­— anyone know what the heck `"partial_update"` is? @jonoomph ?

I can see where it's passed  around the update manager a whole bunch, but _not_ where it's ever used for ANYTHING... and I'm not sure there's ever **any** situation where it would be `True`, even. It's also not really documented in any meaningful way.

...Does it even have any purpose? (Anymore? Yet?)

SuslikV:
> ...what the heck `"partial_update"` is?

It was used to not overwrite whole dictionary but to update only values of existing keys. Not used anymore. Rudiment of the `project_data.py` storage `_set()`.

ferdnyc:
@SuslikV Aha! Thanks.

jonoomph:
LGTM

Issue #2552, opened 2019-01-22
Changing keyframe type is also affecting previous keyframes


schauveau:
**Please fill out this template** 

**Describe the bug**
When changing the type of the numerical keyframe (e.g. linear, bezier or constant) I noticed that the type of the previous keyframe is also changed. 

**System Details (please complete the following information):**
 - Operating System / Distro: Debian/buster
 - OpenShot Version 2.4.3

**To Reproduce**
1. In a video element, create 5 keyframes for the 'Scale X' property with the default type (Linear)
2. Move the cursor to the 3rd keyframe. The 'Scale X' name should turn green. 
3. Right click on the value and change its type to 'Constant'
4. Move the cursor to the 2nd keyframe (or anywere between the 1st and the 2nd keyframes) 
5. The type of 'Scale X' for the 2nd keyframe was also set to 'Constant' which is unexpected.

**Expected behavior**
Changing a keyframe type should NOT affect the previous keyframe.

**Workaround**
With that bug, changing the type of single keyframe X can be done as follow:
1. Insert a dummy keyframe before keyframe X. 
2. Edit the type of  keyframe X (Remark: the bug will affect the type of the dummy keyframe)
3. Remove the dummy keyframe

peanutbutterandcrackers:
I can confirm. This is present.

However, this wasn't present before. So this must have slid in somehow. Certainly very strange.

It seems that only the keyframe immediately preceding the keyframe being modified is being affected, not the other ones.

schauveau:
I had a quick look at the code. There is a function value_updated in file properties_model.py
There is a 'for point' loop within which the value of point["interpolation"] is updated at 2 different locations for previous_point_x and closest_point_x.
https://github.com/OpenShot/openshot-qt/blob/develop/src/windows/models/properties_model.py#L443

I commented out the first instance (so for previous_point_x) and that seems to fix the bug at least for switching between 'linear' and 'constant' interpolation. The new interpolation is now applied to the current or to the next keyframe point.

Remark: There is one exception. When the interpolation is changed after the last keyframe point, the modification is applied to that last keyframe point so BEFORE the requested location. That probably makes sense but this is a bit confusing. In fact, there no transition is taking place after the last keyframe so should not provide the Bezier, Linear and Constant entries. This is already the case when there are no keyframes and so no transitions in the whole segment. 

[...]

PR #3192, opened 2020-01-21
Fix the updating of the interpolation type of previous keyframe point

SuslikV:
Fixes: https://github.com/OpenShot/openshot-qt/issues/2552

This fixes issue when current and previous segment both modifies its
interpolation type, when only current was modified.

If you try to modify the segment of the _Constant_ type keyframed animation that lies between the two  same _Constant_ type of animation keys (red vertical line) with the Bezier _Ease In (Expo)_ preset :
OpenShot step to bezier 01
you will end with 2 Bezier interpolations:
**Before the fix:**
OpenShot step to bezier 03

(left, where the was Const is now Bezier, straight line but Bezier interpolation, no Properties Table shown on the screen but it is Bezier, sure. The backward, 3 _Linear_ segments to _Constant_ in the middle will produce 2 _Constant_ + 1 _Linear_, instead of 1 _Linear_ + 1 _Constant_ + 1 _Linear_).

**After the fix:**
And how it should be
OpenShot step to bezier 02

**Edit:** the change will be useful for Curve Editor too (https://github.com/OpenShot/openshot-qt/pull/3084).

### DIFF
--- a/src/windows/models/properties_model.py
+++ b/src/windows/models/properties_model.py
@@ -324,13 +324,12 @@ class PropertiesModel(updates.UpdateInterface):
                                 # Only update interpolation type (and the LEFT side of the curve)
                                 found_point = True
                                 clip_updated = True
-                                point["interpolation"] = interpolation
                                 if interpolation == 0:
                                     point["handle_right"] = point.get("handle_right") or {"Y": 0.0, "X": 0.0}
                                     point["handle_right"]["X"] = interpolation_details[0]
                                     point["handle_right"]["Y"] = interpolation_details[1]
 
-                                log.info("updating interpolation mode point: co.X = %s to %s" % (point["co"]["X"], interpolation))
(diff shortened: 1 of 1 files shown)


SuslikV:
I reinvented this fix. It was mentioned in: https://github.com/OpenShot/openshot-qt/issues/2552 a long time ago, before I started too work on Curve Editor and noticed this bug too.

The mentioned last keyframe modification by the @schauveau is same to the clip boundaries limit (keyframe cannot be modified by existing logic out of clip boundaries thus every modification is lies inside the clip or at _closest point_) and this fix doesn't change the behavior - the last keyframe is still closest to anything that lies behind it (in terms of _libopenshot_).

SuslikV:
Just added co-author, because the solution was the same and I didn't found any other related PRs.

jonoomph:
Closing PR. I cannot reproduce the problem on develop. Perhaps this was already fixed in another PR? If someone can verify this, please reopen.

Issue #3190, opened 2020-01-21
Project auto save exception


millpaub:
**Describe the bug**
On export, the process shuts down about 40% through the process. No errors are displayed.
Loaded the lasts daily build with no noticeable changes.

**Steps to reproduce the behavior:**

**Expected behavior**

**System Details**
- OpenShot Version: 2.5.0
- libopenshot Version:  0.2.4
- Platform: Windows-10-10.0.18362
- Processor: Intel64 Family 6 Model 94 Stepping 3, GenuineIntel
- Machine: AMD64
- Python version: 3.7.4
- Qt5 version: 5.12.4
- PyQt5 version: 5.12.3
- Qt Detected Languages: ['en-US']
- LANG Environment Variable: 
- LOCALE Environment Variable: 
- Daily Build Installed: http://github.com/OpenShot/openshot-qt/releases/download/OpenShot-v2.5.0-release-candidate-1579485252-x86_64.exe

**Log Files**
- http://logs.openshot.org/u1865iylv5rpuxeaeugln87qurlf78u9/openshot-qt.zip
- http://logs.openshot.org/u1865iylv5rpuxeaeugln87qurlf78u9/libopenshot.zip

**Exception / Stacktrace**
```No stacktrace found in log files```

**Screenshots** (Optional)
If applicable, add screenshots to help explain your problem. You can include screenshots by
copy/pasting them on GitHub or dragging-and-dropping into the GitHub page. All images are public,
so please don't post screenshots containing personal information.

**Comment by ferdnyc** [comment]
```
Traceback (most recent call last):
  File "C:\Program Files\OpenShot Video Editor\windows\main_window.py", line 643, in auto_save_project
    recover_path_with_timestamp = os.path.join(info.RECOVERY_PATH, "%d-%s.osp" % (int(time()), file_name))
TypeError: 'module' object is not callable
     metrics:INFO Track exception: [200] http://www.openshot.org/exception/json/ | *failed-recording-exception*: Control/space characters not allowed (key=b":1:Windows-2.5.0-TypeError: 'module' object is no")
 main_window:INFO auto_save_project
  exceptions:ERROR Unhandled Exception
Traceback (most recent call last):
  File "C:\Program Files\OpenShot Video Editor\windows\main_window.py", line 643, in auto_save_project
    recover_path_with_timestamp = os.path.join(info.RECOVERY_PATH, "%d-%s.osp" % (int(time()), file_name))
```

[...]

PR #3193, opened 2020-01-22
Add confirmation dialog when canceling export

SuslikV:
Just fully localized temporary solution that targets this small request: https://github.com/OpenShot/openshot-qt/issues/3190#issuecomment-576935708

Export paused until decision taken.

Is it works for all systems? I can only test it in Windows.

### DIFF
--- a/src/windows/export.py
+++ b/src/windows/export.py
@@ -979,6 +979,21 @@ class Export(QDialog):
             super(Export, self).accept()
 
     def reject(self):
+        # Handle cancel
+        if self.exporting and not self.close_button.isVisible():
+            _ = get_app()._tr
+            ret = QMessageBox(self)
+            ret.setIcon(QMessageBox.Question)
+            ret.setWindowTitle(_("Video Export"))
+            ret.setText(_("Are you sure you want to abort the video export?"))
+            ret.addButton(_("Yes"), QMessageBox.AcceptRole)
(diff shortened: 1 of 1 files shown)


ferdnyc:
Works well, thanks! I pushed a commit with a couple of tweaks:
- Change title to "Export Video", since we already have a translation for that string
- Use `QMessageBox.question()` and the `QMessageBox.No | QMessageBox.Yes` convenience objects, which automatically decorates the "No" and "Yes" buttons with icons, uses Qt built-in translations, and adds accelerator keys (localized) for each response.

Merging.

SuslikV:
In this short notation the default button is missing ("No"):
```
QMessageBox.No | QMessageBox.Yes, QMessageBox.No)
```

ferdnyc:
> In this short notation the default button is missing ("No"):

Not missing, just unspecified. (The default value, `QMessageBox::NoButton`, is used.) [Which means](https://doc.qt.io/qt-5/qmessagebox.html#question):
> If _defaultButton_ is [`QMessageBox::NoButton`](https://doc.qt.io/qt-5/qmessagebox.html#StandardButton-enum), QMessageBox chooses a suitable default automatically.

(It picks "No".)

SuslikV:
> (It picks "No".)

^_^ It's for you only. When I hit <kbd>Enter</kbd> twice - export terminates.

PR #3202, opened 2020-02-03
Cherry-pick 5dbb993a from release branch

ferdnyc:
Cherry-pick 5dbb993ae59dca6bc5bfb03144ae57ac7ec8b226:

Removing verbose debug statement when getting properties for property dialog (which happens as the user scrubs the timeline)

### DIFF
--- a/src/windows/models/properties_model.py
+++ b/src/windows/models/properties_model.py
@@ -1,26 +1,26 @@
-""" 
+"""
  @file
  @brief This file contains the clip properties model, used by the properties view
  @author Jonathan Thomas <jonathan@openshot.org>
- 
+
  @section LICENSE
- 
+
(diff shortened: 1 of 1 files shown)


SuslikV:
Can this be transformed to the debug log level?

ferdnyc:
Merging, because this is hella annoying.

ferdnyc:
> Can this be transformed to the debug log level?

If someone implements off-by-default debug logging, sure, I suppose it could be. Though I somewhat question the usefulness of spewing out the _entire_ property definition of *every* frame the user previews, even for debugging.

But, I won't disagree that we could really use another level of logging, one that defaults to disabled. Along with a command-line switch to enable it for just the current session, so you don't have to edit your settings (and then remember to turn it off again.)

Issue #3198, opened 2020-01-24
Switching between transition and image - property editor keeps old properties


TharenaMelishka:
Could not find reported

**Describe the bug:**
Clicking to change editing from a transition to an image, the properties changes the selection to the image but the properties remain the transition's and not the image properties.  

**Steps to reproduce the behavior:**
I have a file with multiple clips (30+) split between images and short video, with transitions for each clip.  While this could be an artifact of heavy load on OS, here is reproduction.

1.  Drag png image (screen shot from OS imported back into OS) to track.
2.  Drag transition to start of image and edit length to 2 sec (any transition)
3.  Click directly on image attached to transition.  -- this should show issue in properties (see screenshot)

Note:  this doesn't seem to happen every time, and I'm not sure how reproduce-able it is.

Workaround - Click from the transition to empty track space to clear properties, then click on the image.

**Expected behavior:**
I expected to see the properties of my image, not the transition.

**System Details:**
- OpenShot Version 2.5.0 Dev 1
- Operating System Windows 10

**Log Files:**
No log

**Exception / Stacktrace:**
N/A

**Screenshots:** (Optional)
See attached where image is selected, but properties show transition properties instead of image properties.

![2020-01-24 (2)](https://user-images.githubusercontent.com/9803487/73098335-280b7c00-3eb7-11ea-9180-a871ed8e1acd.png)

SuslikV:
As soon as it happens, please post the log-file from the OpenShot session. Maybe it will help to resolve the issue.

TharenaMelishka:
[openshot-qt.log](https://github.com/OpenShot/openshot-qt/files/4146120/openshot-qt.log)
[openshot-qt.log.1.log](https://github.com/OpenShot/openshot-qt/files/4146127/openshot-qt.log.1.log)
[openshot-qt.log.2.log](https://github.com/OpenShot/openshot-qt/files/4146128/openshot-qt.log.2.log)
[openshot-qt.log.3.log](https://github.com/OpenShot/openshot-qt/files/4146131/openshot-qt.log.3.log)

..... Libopenshot.log is currently 130,680,295kb... even Word told me to go to hell.  :) So... I'm assuming you want one of the openshot-qt logs.  Cause I'm not giving Git the option of giving me the bird on that one. :)  I saw the .1 .2 and .3 and wasn't sure if you needed those, so I renamed extension to log to upload, then reverted them so I could keep editing without blowing up my session.  

[...]

PR #3203, opened 2020-02-03
Fix the interruption of the Properties update

SuslikV:
Fixes: https://github.com/OpenShot/openshot-qt/issues/3198

Is there any reason to stop the timer exactly inside the `update_model` call?


jonoomph:
Closing PR due to unneeded change. I'm unclear if this is needed or not, and I'm not convinced this will do anything.

PR #3206, opened 2020-02-08
Fix auto-creation of animation data for scaling

SuslikV:
Fixes the issue that you cannot modify value at start frame of the clip if you at least once used the Zoom options from the _Animate_ context menu over the clip in the Timeline.

Steps to reproduce:
1) Add video clip at least 2 sec long to the Timeline
2) Right-click over the added clip and click in context menu _Animate > Start of Clip > Zoom > Zoom In (50% to 100%)_
3) Place Timeline cursor at first keyframe of the clip (start of the Zoom)
4) Modify the `Scale X` value from 0.5 to 0.3
5) Notice that nothing changes.


jonoomph:
Addressing in a more complete way in a new PR: https://github.com/OpenShot/openshot-qt/pull/3714

PR #3209, opened 2020-02-09
Small correction in Doxyfile.in

albert-github:
It doesn't make sense to exclude a directory starting with a '/' when it is not in the INPUT, the intended directory is the one without the starting '/'.

### DIFF
--- a/Doxyfile.in
+++ b/Doxyfile.in
@@ -687,7 +687,7 @@ RECURSIVE              = YES
 # run.
 
 EXCLUDE                = "_darcs" \
-						 "/src/blender/scripts/"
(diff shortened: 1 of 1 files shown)


ferdnyc:
@albert-github Thanks for the submission, but we don't use Doxygen to document `openshot-qt` (preferring to use Python's built in documentation, viewable using the `pydoc` command). That file was just an early experiment that hasn't been maintained, and it should really just be deleted.

ferdnyc:
I've removed the `Doxyfile.in` from the repo, so I'm going to close this. Thanks again!

PR #3214, opened 2020-02-11
Tutorial: Mouse cursor, painting fixes; clean up widget objects

### LINKED ISSUE #3212: Wrong cursor style on Tutorial popup
**Describe the bug:**
The cursor style is set to "not allowed" when hovering over Tutorial windows.

1. Start OpenShot for the first time or bring up tutorials
2. Hover over Tutorial window
3. ???

**Expected behavior:**
Reg cursor style

**System Details:**
- OpenShot Version [e.g. 2.4.3]: {{ 2.5.0 }} fa699d77
- Operating System / Distro: Win7

**Screenshots:** (Optional)
openshot-qt

ferdnyc:
This fixes the issue reported in #3212 where the mouse cursor displayed while the tutorial is open would either be the Forbidden symbol (inadvertently set for `dockTutorial` in the UI file) or some seemingly-random resizing cursor. We now explicitly `setCursor(Qt.ArrowCursor)` in the tutorial `QWidget`.

It also cleans up some other aspects of the tutorial code:
1. The `QWidget` objects for closed tutorial popups are now cleaned up, previously they were building up and hanging around until the application terminated
1. Painting of the dialog is streamlined by using a pen and brush, instead of stacking two filled rectangles
1. A bunch of single-use variables (data copied into them, then referenced exactly one time) are eliminated by referencing the original source of the data directly
1. The tutorial data structure is reformatted for readability

Fixes #3212

### DIFF
--- a/src/windows/ui/main-window.ui
+++ b/src/windows/ui/main-window.ui
@@ -400,9 +400,6 @@
      <height>38</height>
     </size>
    </property>
-   <property name="cursor">
-    <cursorShape>ForbiddenCursor</cursorShape>
-   </property>
    <property name="focusPolicy">
     <enum>Qt::ClickFocus</enum>
    </property>
--- a/src/windows/views/tutorial.py
+++ b/src/windows/views/tutorial.py
@@ -154,7 +154,9 @@ class TutorialDialog(QWidget):
         # Make transparent
         self.setAttribute(Qt.WA_NoSystemBackground, True)
         self.setAttribute(Qt.WA_TranslucentBackground, True)
-        #self.setWindowFlags(Qt.FramelessWindowHint)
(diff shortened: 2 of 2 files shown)


ferdnyc:
One more: Turns out `drawPath()` and `fillPath()` are redundant — `fillPath()` draws the filled path (without its outline), whereas `drawPath()` will draw the outline of the path (using the current QPen), and also fill with the Painter's current brush.

ferdnyc:
I'm going to merge this, because:

1. It fixes the reported bug
1. I'm a _tiny_ bit worried that some of the drawing changes may not work correctly in really ancient Qt, so I want to test the resulting AppImage. If it breaks, I'll revert.

ferdnyc:
> 2\. I want to test the resulting AppImage.

I did, and all is well.

PR #3218, opened 2020-02-12
Modernize .travis.yml, add Python 3.6/Bionic

ferdnyc:
Upgrade the Travis YAML control file to:
- Use the 'language: python' test environment
- Install deps with the Travis apt plugin, instead of sudo
- Install cx_Freeze with pip3, rather than from a URL
- Add a Python 3.6 test build, running on Bionic

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,46 @@
-language: cpp
-dist: xenial
-sudo: required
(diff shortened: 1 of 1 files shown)


ferdnyc:
Merging, as it only affects the CI system.

PR #3223, opened 2020-02-13
WIP: Package standard .desktop file in AppImage

### LINKED ISSUE #3222: Appimage .desktop file needs Categories=
Appimage 2.5.0

This link creator is a great feature >> kudos.

$HOME/.local/share is fine but this categories line is needed or the desktop link ends up classified as 'Other' and it doesn't get placed in 'sound and video' section:

Categories=AudioVideo;AudioVideoEditing;  

Also, a request:  I downloaded the image, not the .deb, deliberately so I didn't install.  Try: "Would you like to add a desktop menu link for this appimage y/n/later " on the popup.    Not   "Do you want to install bla bla bla "  I found myself auto-clicking 'no' the first time round ....

_Originally posted by @TuxVinyards in https://github.com/OpenShot/openshot-qt/issues/2440#issuecomment-585607302_

ferdnyc:
@TuxVinyards pointed out that the `.desktop` file in our AppImage doesn't contain the `Categories=` line required for it to properly show up in application menus. It was also missing a bunch of other fields, because instead of using the standard file from `xdg/`, the `build-server.py` script was generating its own abbreviated version.

But reading over the code for `openshot-qt.wrapper` in the AppImage, it _seems_ like we could just use the standard file — it rewrites the `Exec=` line anyway, when copying it into the user's applications directory.

So, this PR is an experiment with just copying `xdg/org.openshot.OpenShot.desktop` into the AppImage, instead. (It's a branch in the project repo, so an AppImage will get generated by the builders.)

(Hopefully) fixes #3222

### DIFF
--- a/installer/build-server.py
+++ b/installer/build-server.py
@@ -322,26 +322,32 @@ try:
         os.mkdir(os.path.join(app_dir_path, "usr", "share", "pixmaps"))
         os.mkdir(os.path.join(app_dir_path, "usr", "share", "mime"))
         os.mkdir(os.path.join(app_dir_path, "usr", "share", "mime", "packages"))
-        os.mkdir(os.path.join(app_dir_path, "usr", "lib"))
-        os.mkdir(os.path.join(app_dir_path, "usr", "lib", "mime"))
-        os.mkdir(os.path.join(app_dir_path, "usr", "lib", "mime", "packages"))
 
         # Create AppRun file
         app_run_path = os.path.join(app_dir_path, "AppRun")
         shutil.copyfile("/home/ubuntu/apps/AppImageKit/AppRun", app_run_path)
 
-        # Create .desktop file
-        with open(os.path.join(app_dir_path, "org.openshot.OpenShot.desktop"), "w") as f:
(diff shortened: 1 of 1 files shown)


ferdnyc:
Whoops, dropped a paren. Let's try that again...

ferdnyc:
This PR also removes `$(APPIMAGE_MOUNT)/usr/lib/mime/packages/openshot-qt`, because:
1. It contains a hardcoded `/usr/bin/` path
1. _Nothing_ uses it, that I can find — on my (Fedora) system, it is the **ONLY** file in `/usr/lib/mime/packages/`
1. Everything works just fine without it, as far as I can tell.

I'd really like to get rid of it _entirely_ (including removing it from the `setup.py`-generated installs), but I'm using the AppImage as a proving ground since I'm almost certain that it's broken _there_ even if it DOES work elsewhere.

ferdnyc:
First attempt failed miserably, not really sure why TBH. Trying again, but now rewriting the `Exec=` line on the way in.

ferdnyc:
Going to merge this, so our AppImage .desktop file is generated with Categories and other goodness.

PR #3228, opened 2020-02-14
Docs: Outdent table code a bit

ferdnyc:
Might fix some build issues with the manual. (Might not! But it's probably more correct.)

### DIFF
--- a/doc/animation.rst
+++ b/doc/animation.rst
@@ -32,17 +32,17 @@ Overview
 .. image:: images/animation-overview.jpg
 
 .. table::
-     :widths: 5 32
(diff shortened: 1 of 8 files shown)


ferdnyc:
Added another fix, I noticed on my system that our tables are being generated with this huge 24px margin under the headings (inside the header row) — looked weird. I don't remember that always being there. Regardless, hopefully gone now.

ferdnyc:
Guess I'll merge this to finish up the remainder of the docs, since #3236 fixed the `titles.rst` file, but not the rest. :wink:

PR #3231, opened 2020-02-16
Fix readonly text fields visibility for light system themes

SuslikV:
Fixes an issue when read only fields of properties are hard to read on bright backgrounds.

**Before** (the _Duration_ field):

OpenShot properties view system01

**After** (the _Duration_ field):

OpenShot properties view system02

The Dark theme changes is not significant, bright text replaced with the "disabled" style, see below:

<details>

**Before** (the _Duration_ field):

OpenShot properties view dark01

**After** (the _Duration_ field):

OpenShot properties view dark02

</details>

Actually, I was thinking that these fields is just empty... (I'm using system theme for OpenShot).

### DIFF
--- a/src/windows/views/properties_tableview.py
+++ b/src/windows/views/properties_tableview.py
@@ -105,7 +105,10 @@ class PropertyDelegate(QItemDelegate):
             else:
                 painter.setBrush(QBrush(QColor("#3e3e3e")))
 
-        if not readonly:
+        if readonly:
+            # Set text color for read only fields
+            painter.setPen(QPen(get_app().window.palette().color(QPalette.Disabled, QPalette.Text)))
+        else:
             path = QPainterPath()
             path.addRoundedRect(QRectF(option.rect), 15, 15)
             painter.fillPath(path, QColor("#3e3e3e"))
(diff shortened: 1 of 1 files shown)


ferdnyc:
Makes total sense. And the built-in "Humanity" theme suffers from the same issue, I just found, so this change should fix that as well.

One small request: Can you swap the conditional around? So...
```python
if readonly:
    # Set text color for read only fields
    painter.setPen(QPen(get_app().window.palette().color(QPalette.Disabled, QPalette.Text)))
else:
    # ...everything else...
```

It's a lot more readable/maintainable that way, than getting to an 'else' after 25 lines of code and having to remember what was even tested in the first place.

SuslikV:
I can. But it would be harder to understand (follow) the changes I have made because github more likely will show that whole block of the code was removed and new added, despite only few lines were changed.

ferdnyc:
It should just show an insertion at the top and removal at the bottom, but regardless.
Maintainability of the CODE is more important than readability of the PRs, ultimately.

SuslikV:
OK. I will do it. (I'm typing slowly).

ferdnyc:
Thanks, and no rush. (I'll squash the commits when I merge the PR, so in the git history it will just show the end result of both edits as a single change.)

SuslikV:
I'm rebase it rather. Because when you learn something bad (as rebase) you can't stop using it...

ferdnyc:
That didn't work, it seems. (Maybe you forgot to `git add` the edited file before making the updated commit? I do that _all_ the time.)

ferdnyc:
Never mind, now it's there.

SuslikV:
If I even had a mind.

ferdnyc:
Hrm. The text's a lot lighter with my system theme than with yours. (And same with Humanity (non-dark). I wonder why?

Still, it's definitely a lot better than invisible. Merging!

image

ferdnyc:
Yeesh, I never noticed how crappy this orange looks, either, in the light theme. Maybe not the best color choice, considering. (Or we should theme the icons as well.)
image

SuslikV:
This is system settings for the disabled text. Only dark theme sets its own colors in `app.py` of OpenShot. That means that everything else is Qt defaults for the platform.

PR #3232, opened 2020-02-16
Include new codecs

eisneinechse:
Initial changes made to include newer codecs

### DIFF
--- a/src/presets/format_mp4_librav1e.xml
+++ b/src/presets/format_mp4_librav1e.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!DOCTYPE openshot-export-option>
+<export-option>
+	<type translatable="True">All Formats</type>
+	<title translatable="True">MP4 (rav1e)</title>
+	<videoformat>mp4</videoformat>
+	<videocodec>librav1e</videocodec>
+	<audiocodec>libvorbis</audiocodec>
+	<audiochannels>2</audiochannels>
+	<audiochannellayout>3</audiochannellayout>
+	<videobitrate
+	 low="200 qp"
+	 med="100 qp"
+	 high="50 qp"></videobitrate>
+	<audiobitrate
+	 low="96 kb/s"
+	 med="128 kb/s"
+	 high="192 kb/s"></audiobitrate>
+	<samplerate>48000</samplerate>
+</export-option>
--- a/src/presets/format_mp4_librav1e.xml
+++ b/src/presets/format_mp4_librav1e.xml
@@ -2,7 +2,7 @@
 <!DOCTYPE openshot-export-option>
 <export-option>
 	<type translatable="True">All Formats</type>
-	<title translatable="True">MP4 (rav1e)</title>
+	<title translatable="True">MP4 (AV1 rav1e)</title>
 	<videoformat>mp4</videoformat>
 	<videocodec>librav1e</videocodec>
 	<audiocodec>libvorbis</audiocodec>
--- a/src/presets/format_mp4_libsvtav1.xml
+++ b/src/presets/format_mp4_libsvtav1.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!DOCTYPE openshot-export-option>
+<export-option>
+	<type translatable="True">All Formats</type>
+	<title translatable="True">MP4 (SVT-AV1)</title>
+	<videoformat>mp4</videoformat>
+	<videocodec>libsvt_av1</videocodec>
+	<audiocodec>libvorbis</audiocodec>
+	<audiochannels>2</audiochannels>
+	<audiochannellayout>3</audiochannellayout>
+	<videobitrate
+	 low="60 qp"
+	 med="50 qp"
+	 high="30 qp"></videobitrate>
+	<audiobitrate
+	 low="96 kb/s"
+	 med="128 kb/s"
+	 high="192 kb/s"></audiobitrate>
+	<samplerate>48000</samplerate>
+</export-option>
--- a/src/presets/format_mp4_libsvtav1.xml
+++ b/src/presets/format_mp4_libsvtav1.xml
@@ -2,7 +2,7 @@
 <!DOCTYPE openshot-export-option>
 <export-option>
 	<type translatable="True">All Formats</type>
-	<title translatable="True">MP4 (SVT-AV1)</title>
+	<title translatable="True">MP4 (AV1 svt)</title>
 	<videoformat>mp4</videoformat>
 	<videocodec>libsvt_av1</videocodec>
 	<audiocodec>libvorbis</audiocodec>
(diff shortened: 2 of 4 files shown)


ferdnyc:
Very cool!

(Although I have to say, as an end user, being presented with the choice between "MP4 (rav1e)" and "MP4 (SVT-AV1)", my response would probably be, "... ... ...Gesundheit?" We may be asking a lot of the users, to make those kind of choices, especially in the so-called "simple" interface.)

TBH this list feels like it's starting to get unwieldy in general:
image

Not claiming I have any brilliant solutions to offer, just throwing it out there. The _code_ looks good as always, of course. No comments there.

(Though maybe it's time for the increasingly-misnomered `convert_to_bytes` to be given a name that represents its function better?)

eisneinechse:
And again I have to agree.
It's even worse webm(AV1) is AV1 too, just libaom. This can easily be fixed it's only in the preset file. But the root cause would be overlooked. This is really just a initial proposal of what might be used. The code is not as clean as I liked. Again it is just something I threw in to check it out and a lot didn't even work until not too long ago.
Please feel free to change as much as you like. And as for the names, I completely agree. The names become misleading. Another reason why I presented this. It's time to make changes. The newer codec have option that work differently.
In libopenshot you can see that we have to make further changes.
As for the three AV1 encoders: Well they all have their pros and cons. libaom is still the standard to go after, and svt_av1 is very fast. Right now as a test I am encoding a 50 min clip in 720p 30 fps with medium (good quality at least as good as the input file) settings and on a R5 3600 it is at about 15 fps encoding speed. 5 years ago I had that with h.264. A threadripper 3990X would blast through this, SVT scales very well with threads!
Please feel completely free to make changes. This is only a starting point.

eisneinechse:
@ferdnyc I forgot to mention that I might be a good idea when renaming procedures etc. to keep in mind that we at some point have to go for more than 8 bit color depth. No need to include 10 bit, 12 bit etc but if something is changed we should avoid renaming it to something that we know will have to be renamed again later.

SuslikV:
It will be only in the way you'll do it. The usage window in 1 year - is good result too. Everything can change later. Forget about 10-12 bits for at least for this year.

ferdnyc:
> The names become misleading. Another reason why I presented this. It's time to make changes. The newer codec have option that work differently.
> In libopenshot you can see that we have to make further changes.

Yeah, I don't know if you saw it, but I actually have an open PR (OpenShot/libopenshot#421) to add `SetVideoOptions` and `SetAudioOptions` overloads that don't require **all** of the arguments, and just pick sensible defaults instead.

(Though the overloads in the current PR still have the downside that, since there are no setters for the parameters `Set___Options` defines, there's no way to assign values _other_ than those "sensible" defaults either before or after calling the methods. Not unless you're into directly manipulating the `WriterInfo` values directly, anyway. )

I didn't remove `bit_rate` from the `SetVideoOptions()` overload, but I kind of wanted to since you're _not_ always encoding at a fixed bitrate, and setting it up front like that feels "off". Especially with all that special-casing based on the `bit_rate` value, because of the way it's overloaded to _also_ carry the `crf` and/or `qp` rate setting — which then, depending on the codec selected, has different magical effects the **actual** bitrate parameter ultimately passed to ffmpeg.

Seems to me that FFmpegWriter could use extending, so it can store either or both of:
1. Field(s) for `qp` and `crf` value(s) separate from the `video_bit_rate`. (Which can then just be directly set to `0` when it needs to be, for those codecs that work that way. That way all of the magic interpretation of `video_bit_rate` can go away.)
    1. And maybe some more, for additional parameters that some modes accept (like `qmin` and `qmax`).
2. One or more enums(?) to select between / activate `CBR`, `QP`, `CRF`, and any other methods of defining encoding quality/bandwidth parameters. (Hmm. Would a single, xor-style enum be sufficient, or are there ways to combine some of those modes?)

Whether it would be better to add them to the base `WriterInfo` struct, or we should add them to the FFmpegWriter class's data storage so they're kept separate from other writer classes where they really make no sense, I don't feel like I know the code well enough to say.

eisneinechse:
All I can say is I agree.
The thing is that might be a good start for 3.0. Or at least 2.6.

eisneinechse:
I just now confirmed that openshot can import 10 bit color depth files if ffmpeg has the support. I could also export to 10 bit, but that's pointless for now. The important part is that openshot can import files from sources that have 10 bit. The rest can wait for now.
Also confirmed that svt does its own parallel execution pretty well, any parallel executions in ffmpeg and/or OMP does not have any effect.
The fps problem is now my main concern.
It would also be nice to have people join in the development of a good FFmpegReader::IsHardwareDecodeSupported in FFmpegReader.cpp or better move that method into an extra file and improve it there. That would solve the h.265 decode in hardware problem.
The new codecs can use more options and I hope that soon SVT-VP9 will be usable. I haven't checked SVT-HEVC which probably some people would like to have (fast h.265 export!).

eisneinechse:
FFmpegReader::IsHardwareDecodeSupported in FFmpegReader.cpp could be changed to something that looks up a list, at least until we come up with something better.
For linux there is "lspci -vnn | grep VGA" that gives in brackets VendorID:ProductID (like [1002:1309]). This IDs should also be accessible from C++ and they give the exact GPU. Now we could whiteflag codecs according to the GPU. Not perfect at all but better than trying to connect graphic cards to codecs. This could even be done in an extra file that is read on program start.

eisneinechse:
@ferdnyc in the performance tab of the preferences window we already set the max width and height that the GPU can handle to decode. I highly doubt that most of the users know what to set as the values. To be honest I have to google it myself for my graphics card. Just an idea, but we could move this in an expert section and in this expert section we could have the codecs that the card can use in a kind of list where each one can be selected.
Something you could use to include h265 and you could set the maximum dimension that the GPU could decode there too. Maybe a rest button that sets everything to sane minimum values.
Having the product ID and vendor ID might not be enough. Things like the maximum dimension might be different for different cards.
I also thought about having the ability to set the decoding option for each imported file individually in the window where the files are listed. It would be forced hw decode, forced sw decode, or decode according to the settings in preferences (the default).
Decoding looks easier but its way harder to implement and it can be different for every imported file as each file can have different dimensions and codecs.

groslaine:
Hi, I download the last daily build and I dont have the decode encode available. In the preference there is "no GPU acceleration".
I'm on windows 10, with a gtx 980M and the latest nvidia driver.

jonoomph:
Merging this, and will deal with the larger codec list in our UI refresh at some point. Some of these might end up as only advanced options.

SuslikV:
Wait a moment, the https://github.com/OpenShot/libopenshot/pull/445 is in WIP state and isn't even merged yet. Why this comes first?

SuslikV:
@jonoomph Please, revert it back until _libopenshot_ will get required changes. "qp" - settings isn't supported in _libopenshot_.

ferdnyc:
@SuslikV @eisneinechse 

I haven't been following this closely, but any reason we can't just merge the libopenshot PR to make this work? It doesn't seem to be WIP in the sense that it doesn't work, any further changes can be done in subsequent PRs.

PR #3234, opened 2020-02-17
Retire Windows envvar-deletion code (again)

ferdnyc:
(I actually did this once before, over the summer, but it subsequently got lost during conflict resolution on another branch.)

It's been three years since OpenShot stopped inserting those envvars.
If anyone still has them set today, it's likely they were created by
some other software and we shouldn't be removing them.


SuslikV:
Then merge it and forget.

Issue #3227, opened 2020-02-14
Cannot open properly just saved project


hitrome:
![Снимок экрана от 2020-02-14 18-00-18](https://user-images.githubusercontent.com/48879999/74542109-ee78de80-4f53-11ea-9109-a403cd9d19bb.png)
[test.osp.zip](https://github.com/OpenShot/openshot-qt/files/4204983/test.osp.zip)

**Describe the bug**
Cannot open just saved project. Got a message: "file_name.MTS not found".

**Steps to reproduce the behavior:**
1. Start the OpenShot
2. Go to "Project files"
3. Right-click "Import files"
4. Select a file
5. Click "Open" button
6. Go to "File -> Save project"
7. Enter the project name ( "test.osp for example)
8. Click "Save"
9. Close OpensShot
10. Start OpenShot
11. Go to "Open Project"
12. Choose the project ( "test.osp for example)
13. Click "Open" button
14. See error: "file_name.MTS not found".

**Expected behavior**
The project should open without errors.

**System Details**
- OpenShot Version: 2.5.0-dev1
- libopenshot Version:  0.2.4-dev1
- Platform: Linux-4.15.0-43-generic-x86_64-with-Ubuntu-16.04-xenial
- Processor: x86_64
- Machine: x86_64
- Python version: 3.4.3
- Qt5 version: 5.2.1
- PyQt5 version: 5.2.1
- Qt Detected Languages: ['ru-RU']
- LANG Environment Variable: ru_RU.UTF-8
- LOCALE Environment Variable: 
- Daily Build Installed: http://github.com/OpenShot/openshot-qt/releases/download/OpenShot-v2.5.0-dev1-1581583925-99fda013-836e49fa-x86_64.AppImage

**Log Files**
- http://logs.openshot.org/6ge3aman8zt9wg6991k0dck4max9wtfd/openshot-qt.zip
- http://logs.openshot.org/6ge3aman8zt9wg6991k0dck4max9wtfd/libopenshot.zip

**Exception / Stacktrace**
```No stacktrace found in log files```

**Screenshots** (Optional)
If applicable, add screenshots to help explain your problem. You can include screenshots by
copy/pasting them on GitHub or dragging-and-dropping into the GitHub page. All images are public,
so please don't post screenshots containing personal information.

hitrome:
This issue is connected with localized file names. If a file structure consist only English characters - the project is opened without errors.

SuslikV:
[...]

PR #3247, opened 2020-02-25
Improve UTF-8 file path support for OpenShot 2.5

jonoomph:
This fixes a bug reported in https://github.com/OpenShot/openshot-qt/issues/3227. File paths were being incorrectly returned from the regex re.sub function, and also incorrectly returned from the new thumbnail server.

### DIFF
--- a/src/classes/thumbnail.py
+++ b/src/classes/thumbnail.py
@@ -153,7 +153,7 @@ class httpThumbnailHandler(BaseHTTPRequestHandler):
         if not only_path:
             self.send_header('Content-type', 'image/png')
         else:
-            self.send_header('Content-type', 'text/html')
+            self.send_header('Content-type', 'utf-8')
         self.end_headers()
 
         # Locate thumbnail
--- a/src/classes/json_data.py
+++ b/src/classes/json_data.py
@@ -164,19 +164,16 @@ class JsonDataStore:
         # Find absolute path of file (if needed)
         if "@transitions" in path:
             new_path = path.replace("@transitions", os.path.join(info.PATH, "transitions"))
-            new_path = json.dumps(new_path)  # Escape backslashes
-            return '"%s": %s' % (key, new_path)
+            return '"%s": "%s"' % (key, new_path)
 
         elif "@assets" in path:
             new_path = path.replace("@assets", path_context["new_project_assets"])
-            new_path = json.dumps(new_path)  # Escape backslashes
-            return '"%s": %s' % (key, new_path)
+            return '"%s": "%s"' % (key, new_path)
 
         else:
             # Convert path to the correct relative path
             new_path = os.path.abspath(os.path.join(path_context.get("new_project_folder", ""), path))
-            new_path = json.dumps(new_path)  # Escape backslashes
-            return '"%s": %s' % (key, new_path)
+            return '"%s": "%s"' % (key, new_path)
 
     def convert_paths_to_absolute(self, file_path, data):
         """ Convert all paths to absolute using regex """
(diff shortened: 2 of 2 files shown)


SuslikV:
In attempt to save the project file with this fix I got the error:

openshot_unicodefix_new_code01

related part of the log:
```
project_data:INFO Saving project file: O:/Temp/OpenShotCrashTest/рус текст/хіба ревуть воли як ясла повні Їїї中文/проект17_ 中文new code01.osp
      assets:INFO Asset dir created as O:/Temp/OpenShotCrashTest/рус текст/хіба ревуть воли як ясла повні Їїї中文\проект17_ 中文new code01_assets
      assets:INFO New thumbnails folder: O:/Temp/OpenShotCrashTest/рус текст/хіба ревуть воли як ясла повні Їїї中文\проект17_ 中文new code01_assets\thumbnail
      assets:INFO New titles folder: O:/Temp/OpenShotCrashTest/рус текст/хіба ревуть воли як ясла повні Їїї中文\проект17_ 中文new code01_assets\title
      assets:INFO New blender folder: O:/Temp/OpenShotCrashTest/рус текст/хіба ревуть воли як ясла повні Їїї中文\проект17_ 中文new code01_assets\blender
project_data:INFO Checking clip IR9E2H23OK path for file SLGP7MT2GO
   json_data:ERROR Couldn't save project data file:
O:/Temp/OpenShotCrashTest/рус текст/хіба ревуть воли як ясла повні Їїї中文/проект17_ 中文new code01.osp
'charmap' codec can't encode characters in position 4271-4272: character maps to <undefined>
 main_window:ERROR Couldn't save project O:/Temp/OpenShotCrashTest/рус текст/хіба ревуть воли як ясла повні Їїї中文/проект17_ 中文new code01.osp. Couldn't save project data file:
O:/Temp/OpenShotCrashTest/рус текст/хіба ревуть воли як ясла повні Їїї中文/проект17_ 中文new code01.osp
'charmap' codec can't encode characters in position 4271-4272: character maps to <undefined>
```

I need to see why it happens. The names including Chinese, Ukrainian and Russian letters, of course, it is some nonsense but it fails for me...

src/classes/thumbnail.py
```diff
@@ -153,7 +153,7 @@ def do_GET(self):
         if not only_path:
             self.send_header('Content-type', 'image/png')
         else:
-            self.send_header('Content-type', 'text/html')
+            self.send_header('Content-type', 'utf-8')
```
ferdnyc: The actually-correct header:
```suggestion
            self.send_header('Content-type', 'text/html; charset=UTF-8')
```
Ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type

ferdnyc:
The root problem here is that the JSON data is being modified by regular expression, AFTER encoding/decoding. That's always going to cause problems, because `json.dumps()` and `json.loads()` will escape the JSON string representation for safe encoding to file. The two functions are complementary, so `json.loads()` will restore an encoded string representation saved with `json.dunps()` -- but NOT if the serialization has been altered in between.

The right way to do the path fix-up is to modify the Python `dict` representation, _before_ passing it to `json.dumps()` or _after_ it's processed by `json.loads()`. After, or **during**.

My branch [`json-coders`](https://github.com/ferdnyc/openshot-qt/tree/json-coders) contains a WIP implementation of custom `JSONEncoder` and `JSONDecoder` subclasses, which walk the dict object hierarchy created/consumed by `json.{loads,dumps}`, fixing up any pathname-containing dict keys it finds. It's not quite finished, but the idea is there.

My take is, that approach is **far** safer than fucking around with encodings. JSON is already encoded safely by default, we just have to stop corrupting the data.

SuslikV:
The longer fix will be developed, then more users will complain.

Do you have **quick** fix @ferdnyc ? I have the one but @jonoomph said that it doesn't works for him and he has other solution. I don't have much info, so I'm unable to improve the one I have. And I seeing problems with his code (actually, it refuses to open any file .osp file as soon as the first slash `/` happens in the project file on Windows. Windows File Explorer application, for example, accepts both slash and backslash symbols in the path string and converts them into backslash'es).

jonoomph:
Closing this PR in favor of https://github.com/OpenShot/openshot-qt/pull/3229.

PR #3262, opened 2020-02-29
Fix formatting for the json_data Autorepair output

SuslikV:
Fixes an issue where repaired file saved as single line of symbols on Windows (no CR LF endings).

Quick fix to: https://github.com/OpenShot/openshot-qt/pull/3259


ferdnyc:
Makes sense, thanks for catching that. I missed that the original save had an `indent` argument.

jonoomph:
Nice catch!

Issue #3182, opened 2020-01-17
Name Sorting Project Files (Quality of Life Enhancement)


TharenaMelishka:
When I select multiple files to import (Sequentially numbered), they don't import sequentially, so if I import say 12 files, I have to hunt them for the correct order.  

Can we get it so when we click on "name" at the top of project files it will sort alphabetically, and invert sort if we click it again?  

Also support the idea in #2436 where project files get folders, so if I have a multi segment video I could sort the files (audio / video / image) into the folders for ease of editing.  

Thanks!

SuslikV:
This is feature request. Sorting option not implemented yet

Issue #444, opened 2016-12-26
Importing many pictures slow


brmmm3:
Openshot 2.2.0:
Importing just 74 pictures (each 1024x768) is very slow.
I've observed that the project files view is updated (completely redrawn) for each new imported file.
Better: Do a delayed redraw and update only the region which has changed. If sorting matters then sort the files when the import has finished.
I'm not sure if you use multithreading/multiprocessing for file import. If not then please use this.

Best regards,

Martin

jebeld17:
#550

DylanC:
@brmmm3 - I agree. It is very slow and also does redraw the files area after every file import.

mikelacsa:
This issue has been tested with the latest version. I confirm that it does not exist anymore.

ferdnyc:
> This issue has been tested with the latest version. I confirm that it does not exist anymore.

Please test again. It most definitely does still exist.

ferdnyc:
(Also, just to be a pedant, you can't "confirm" something that nobody has claimed before.)

ferdnyc:
Oh, wow. And actually, if you want to see _really_ slow, select all of those `$many_dozen` imported images and choose "Remove from Project...". That trundles along at about **half** the speed of the (slow) import process. This is all due to the fact that OpenShot unnecessarily rebuilds (not just redraws, but actually _deletes_ and completely recreates) the Files model for each and every individual add/remove operation..

mikelacsa:
> (Also, just to be a pedant, you can't "confirm" something that nobody has claimed before.)

copy that @ferdnyc thanks!

Twenkid:
> Oh, wow. And actually, if you want to see _really_ slow, select all of those `$many_dozen` imported images and choose "Remove from Project...". That trundles along at about **half** the speed of the (slow) import process. This is all due to the fact that OpenShot unnecessarily rebuilds (not just redraws, but actually _deletes_ and completely recreates) the Files model for each and every individual add/remove operation..

:/ Do you know the reason for that design decision? It doesn't make sense to me, isn't it fixable with a few lines? Or it is an issue with the GUI control/widget itself?

ferdnyc:
@Twenkid

> It doesn't make sense to me, isn't it fixable with a few lines?

It desperately _needs_ to be fixed, but it won't be easy, as unfortunately the current implementation sort of backed itself into a corner. 

> Or it is an issue with the GUI control/widget itself?

[...]

Issue #3240, opened 2020-02-23
"Add to timeline" dialog contains no files or only one


PontiacCZ:
**Describe the bug:**
The "Add to timeline" (CTRL + W) dialog contains no files or only one depending on how you select files in the "Project Files" window.

**Steps to reproduce the behavior:**
1. download these two free videos in FullHD:
    - [Golden Gate Bridge](https://www.pexels.com/video/golden-gate-bridge-at-san-francisco-1994828/)
    - [Burning Charcoal Fire](https://www.pexels.com/video/burning-charcoal-fire-854692/)
2. import then into OpenShot 2.5.0 by dragging
3. activate the OpenShot app but don't click on any of the two files. Press CTRL + A to select them and then press CTRL + W - "Add to timeline" dialog appears
4. there are no files in the dialog. It looks like this:

![image](https://user-images.githubusercontent.com/9153595/75114557-3637ee00-5657-11ea-902a-7957d58a9854.png)

5. return back to the Project Files window. Now click the Golden Gate video and then, with SHIFT, the Fire video. They're both selected so they should both go into the "Add to timeline" dialog, right?
6. press CTRL + W. There is only the first selected video (Golden Gate) there. The same happens if you select the Fire video as first - then only the Fire video is present there:

![image](https://user-images.githubusercontent.com/9153595/75114595-a21a5680-5657-11ea-819c-6830bf0779df.png)

**Expected behavior:**
All selected videos should be displayed in the "Add to timeline" window

**System Details:**
- OpenShot Version [e.g. 2.4.3]: 2.5.0
- Operating System / Distro: Windows 10 Pro x64

ferdnyc:
Hmm. I couldn't immediately reproduce the bug as reported, but there _is_ something weird going on with selection in the Project Files list and Add to Timeline. You can see here that, after selecting all videos in the list, I _deselected_ the first one — but it was still on the list when I opened the dialog. (That's an easier problem to work around, since you can always use the :heavy_minus_sign: icon to remove the extra video from the list, but it still points to a bug in how the selection list is handled.)

![Screenshot from 2020-02-23 10-58-50](https://user-images.githubusercontent.com/538020/75115367-c5c7a780-562b-11ea-89ee-e1f9e93ec288.png)

nanohtpc:
same problem here.... it only happens when you used the keyboard shorcuts....

SuslikV:
Small update. Info for the developers.

[...]

PR #3264, opened 2020-03-01
Add Sorting for project files Views

SuslikV:
**Edit 2:** Add sorting for Project Files (by Name, Path etc.). Nothing else is here. Other will be in other PRs.

Some ideas were adopted in current _develop_, so no sense to read below.

---------------

This removes excessive listener (double of the FilesModel) from the updateListeners list.

This is first step to resolve the https://github.com/OpenShot/openshot-qt/issues/444 **Edit:** ~~(it wouldn't be solved in this PR in part of add files but at remove).~~ ready for multiprocessing

---------------

The FilesModel adds itself to the listeners during initialization:

https://github.com/OpenShot/openshot-qt/blob/777cf1f1547f48bcddf55da2a766c05e9d28376f/src/windows/models/files_model.py#L216-L218

Thus, the first copy of FilesModel class is added in:

https://github.com/OpenShot/openshot-qt/blob/777cf1f1547f48bcddf55da2a766c05e9d28376f/src/windows/views/files_listview.py#L320-L321

or

https://github.com/OpenShot/openshot-qt/blob/777cf1f1547f48bcddf55da2a766c05e9d28376f/src/windows/views/files_treeview.py#L370-L371

via main window creation on application startup:

https://github.com/OpenShot/openshot-qt/blob/777cf1f1547f48bcddf55da2a766c05e9d28376f/src/windows/main_window.py#L2530-L2534

Everything else (from PropertiesTableView initialization) - just result of action updates (it seems that it works good and keeps both copies up to date whole time).

It can reduce twice the number of messages `updating files model` in the log.

### DIFF
--- a/src/settings/_default.settings
+++ b/src/settings/_default.settings
@@ -106,6 +106,20 @@
     "setting": "file_view"
   },
   {
+    "value": -1,
+    "title": "",
+    "type": "hidden",
+    "category": "UI",
+    "setting": "files_view_sorting"
+  },
+  {
+    "value": 0,
+    "title": "",
+    "type": "hidden",
+    "category": "UI",
+    "setting": "files_view_sorting_order"
+  },
+  {
     "value": "thumbnail",
     "title": "",
     "type": "hidden",
--- a/src/windows/views/files_listview.py
+++ b/src/windows/views/files_listview.py
@@ -30,7 +30,7 @@ import os
 
 from PyQt5.QtCore import QSize, Qt, QPoint, QRegExp
 from PyQt5.QtGui import QDrag, QCursor
-from PyQt5.QtWidgets import QListView, QAbstractItemView, QMenu
+from PyQt5.QtWidgets import QListView, QAbstractItemView, QMenu, QActionGroup
 
 from classes.app import get_app
 from classes.logger import log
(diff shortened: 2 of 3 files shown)


SuslikV:
It turns out that by switching the View mode the number of listeners will only grow.

ferdnyc:
Yeah... that's true regardless to some extent. But it can actually grow faster with this change applied, as more model references build up that can't be garbage-collected. I had OpenShot logging 7 "updating" messages per deleted item, at one point.

SuslikV:
> ...But it can actually grow faster with this change applied

How? It doesn't creates any new data.

ferdnyc:
> > ...But it can actually grow faster with this change applied
> 
> How? It doesn't creates any new data.

No, but when switching Project Files views, the abandoned FilesModel from the old view will usually be garbage collected, if it hasn't been "in contact" with the rest of the code. (No drags-and-drops, no dialogs or list appearances, etc.

But if the Properties table view is still holding a reference to that old model from the previous view, even after the new one is created, then the old model can't be destroyed until **after** the reference in Properties also gets removed... at the _earliest_.

(PyQt5 tends to be skittish about cleaning up objects — understandably so, since destroying an object that's still in use can easily crash the application — so if something doesn't get reaped right away, I've noticed that it can take quite a while before all of the references to it are accounted for and it can finally be garbage-collected. Having signals still connected — which, as those log messages show, the abandoned models do — even seems like it might be enough to make things persist indefinitely.)

SuslikV:
> ...things persist indefinitely

It can't be destroyed because it updates constantly! How you don't understand that it will never work this way.

SuslikV:
Added sorting, because it was simple.

OpenShot sorting01

Other views will be changed in few days, I think.

SuslikV:
Do you have any negative thoughts?

Code clean up is ahead. Maybe rebase too. Or maybe not.

The general _FilesModel_ changes are complete. Other models and views can be updated in different PRs. Some wouldn't be updated because they are rarely in use. This totally depends on benefits of such changes.

SuslikV:
https://github.com/OpenShot/openshot-qt/pull/3266 was integrated here

SuslikV:
Added fix for: https://github.com/OpenShot/openshot-qt/issues/3240

SuslikV:
The PR also fixes: https://github.com/OpenShot/openshot-qt/issues/3182

SuslikV:
Minor improvement to the filtering of the Details view (user defined tags were missing).

SuslikV:
Asynchronous thumbnails generation (and thus less Qt calls) speeds up file loading (import to the project) about x2 times. Thus this addresses the  #444

SuslikV:

SuslikV:
This PR. It did a good job by the way... Proxy models, removing doubles of objects, selections and listeners. All this was like a punch (or maybe kick)... And finally, the code got its new form. Hooray! Maybe this PR wouldn't be closed but updated, just to add few small features (sorting, asynchronous thumbnails creation).

jonoomph:
Closing PR for now. Sorting adds mostly unneeded options to the project context menu, making things more complex looking. Filtering accomplishes the same basic task faster... in my opinion. In sorting becomes a highly requested feature (which it is currently not), I would be happy to reconsider adding in some sorting... but more likely it would be clicking column headers, etc... and simply methods like that.

PR #3266, opened 2020-03-01
Ignore file model update during files remove

SuslikV:
Fixes the performance issue during group of files delete.

Steps to reproduce the issue:
1) Add 40 image files to the Project files list
2) Select any 6 of them
3) In right-click menu over the files click _Remove from Project_
4) Notice that list updates one by one in very slow manner (even slower than during adding them).

If you don't see changes after this fix, probably you forgot to apply the: https://github.com/OpenShot/openshot-qt/pull/3264

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -1578,6 +1578,9 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
     def actionRemove_from_Project_trigger(self, event):
         log.info("actionRemove_from_Project_trigger")
 
+        # Tell file model to ignore updates
+        self.filesTreeView.files_model.ignore_update_signal = True
+
         # Loop through selected files
         for file_id in self.selected_files:
             # Find matching file
@@ -1594,6 +1597,10 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
 
         # Clear selected files
         self.selected_files = []
+
+        # Re-enable updates and refresh view
+        self.filesTreeView.files_model.ignore_update_signal = False
+        self.filesTreeView.refresh_view()
 
         # Refresh preview
         get_app().window.refreshFrameSignal.emit()


ferdnyc:
Unfortunately this change (with or without #3264) causes problems due to model disconnection, because even with this change / these changes, our models are being rebuilt instead of reused — and when there are multiple of them, that has to be done in tandem or bad things happen.

Try this:
#### A. Setup
1. Import a bunch of files
1. Add a file <var>file1</var> to the Timeline
1. Add a Mask effect to that clip
1. Set the effect's Source property to a file <var>file2</var>
#### B. So far so good, now...
1. Right-click on Project Files and change the view mode from Thumbnails to Details, or vice versa.
1. Find <var>file2</var> on the list
1. Right-click it and select "Remove from project"

What _doesn't_ happen is, the effect's Source parameter isn't altered in any way. It still lists the removed <var>file2</var> as its Source, and the list you get by right-clicking it still includes <var>file2</var> — it no longer matches the Project Files list.

What's happening is, anytime you switch the view mode of Project Files, the model for the previous view is _dropped_, and a completely new one is built. The same thing happens whenever you filter the list using the text field, or change its category.

If #3264 is applied, Properties is still holding a reference to the **old** model, from the _previous_ view mode — not the one currently being displayed. Because it's still referenced, it can't be deleted. But actions on the Project Files list will no longer modify it, so it falls out of sync. But even if Properties has a separate FilesModel of its own (IOW, no #3264 applied), without the update signals things get out of whack. (In the `devel` branch code, deleting <var>file2</var> actually removes the effect _and_ the clip hosting it from the Timeline, which I suppose is drastic, but also safest...)

We _definitely_ need to preserve our models instead of destroying them, but that goes way beyond the Properties view. There needs to be only _one_ model, that's **never** destroyed, and that means changes to both of the files View types, to not construct their own private models every time they're updated.

ferdnyc:
> There needs to be only _one_ model, that's **never** destroyed, and that means changes to both of the files View types, to not construct their own private models every time they're updated.

(As well as changes to filtering and even selection, so they use ProxyModel classes which _reference_ the main Model. Instead of destroying and rebuilding it in-place based on their parameters.)

SuslikV:
For what reason OpenShot creates many file models? They aren't different at all.

ferdnyc:
No good reason. It's just brute-force code, the model is treated as disposable data since it can always be regenerated from the _real_ data, which canonically lives in the Python `project._data` object.

Unfortunately, while that code works fine for OpenShot's data management, it breaks a lot of things on the QT side. (It's also why we don't have list sorting, why selections aren't retained when switching views...)

SuslikV:
Models never dropped they all works and each executes `app.processEvents()` from the `files_model.py` that slows down the application (by redrawing UI multiply times etc.).

ferdnyc:
(Yup. Which was probably a workaround to force-sync the various copies, ironically.)

SuslikV:
This will be changed. I see nothing in the files model that are unique and cannot be shared.

SuslikV:
Right now this PR is harmless. but as soon as other parts of code isn't ready yet I'm going to close this and reopen when ready.

ferdnyc:
There isn't anything.

Ideally, the FilesModel would live in the `MainWindow` class, and it would construct _itself_ during `__init__()`. Then, when either view is activated, it gets a `setModel(app.windows.main_window.models['files'].model` (or whatever). The model is never brute-force emptied and rebuilt, only updated, and filtering uses a `QSortFilterProxyModel` to create a constrained partial _view_ of the data. All of the model-management code would come out of both `files_*view` modules, because they're **VIEWS** and they're not _supposed_ to be altering the model. Ever.

...But, that'll take some doing. It's certainly possible, but the trick is getting it wired up right. Which means fixing how OpenShot uses signals to communicate between pieces of itself, since right now it's using them... just kind of _wrong_.

ferdnyc:
> Ideally, the FilesModel would live in the `MainWindow` class, and it would construct _itself_ during `__init__()`. Then, when either view is activated, it gets a `setModel(app.windows.main_window.models['files'].model)` (or whatever).

(The Properties view would also pick it up the same way, _not_ from the View.)

ferdnyc:
> Ideally, the FilesModel would live in the `MainWindow` class, and it would construct _itself_ during `__init__()`.

Hmm, that was confusing. What I meant was, a persistent _reference_ to the files model would be held in the MainWindow class, instantiated during the MainWindow `__init__()`. The model itself would still be its own class, of course.

SuslikV:
Opening for the test. The #3264 already has all required changes (it may be edited in the future, but in general everything the files model needs is already there).

SuslikV:
Now I see no sense to make separate from the #3264 (anyway it requires update).

PR #3273, opened 2020-03-02
Blender: Capture stderr, log output

### LINKED ISSUE #3268: OpenShot Animated Title Rendering
**Describe the bug**
Animated Titles rendering

1. Go to "Title - Animated Title"
2. Click on "Dissolving Text"
3. Input "Gilligan" for the "File Name"
4. Input "Gilligan's Island" in "Title"
5. Input "1.50" for the "Text Size"
6. Try changing  Fonts in the "Font Name" (Nothing works there except the Bfont) - Why is this?
7. Slide the slider back to frame 1
8. Click on Render
9. The rendering only goes to frame 13 and stops (I left it on all night and it was still on frame 13 the next day)
10. I have gone back to version 2.4 and it rendered the animated title.

**Expected behavior**
I expected it to render the title for me so that I could put it into the timeline and export it to use in another video editing program that I use all the time.
It did not render in version 2.5 and in version 2.5.1 - it only worked in version 2.4.
Also I could not change to any other font (which, by the way, is very limited) - Can you not make it where it will accept any font that I have installed on my computer?
So there is the font issue as well as the rendering issue.
BTW...the font issue is on all versions 2.4 and forward.

**System Details**
- OpenShot Version: 2.5.1
- libopenshot Version:  0.2.5
- Platform: Windows-7-6.1.7601-SP1
- Processor: Intel64 Family 6 Model 158 Stepping 9, GenuineIntel
- Machine: AMD64
- Python version: 3.7.4
- Qt5 version: 5.12.4
- PyQt5 version: 5.12.3
- LANG Environment Variable: 
- Daily Build Installed: http://github.com/OpenShot/openshot-qt/releases/download/OpenShot-v2.5.1-release-candidate-1583026317-x86_64.exe

**Log Files**
- http://logs.openshot.org/730u3oip38wca5kvefjbez8pt74d319t/openshot-qt.zip
- http://logs.openshot.org/730u3oip38wca5kvefjbez8pt74d319t/libopenshot.zip

**Exception / Stacktrace**
```No stacktrace found in log files```

**Screenshots** (Optional)
If applicable, add screenshots to help explain your problem. You can include screenshots by
copy/pasting them on GitHub or dragging-and-dropping into the GitHub page. All images are public,
so please don't post screenshots containing personal information.

ferdnyc:
This is an attempt to better track the progress of our Blender subprocess, by:

1. Logging the output of the command as it runs (this will be somewhat verbose, unfortunately)
1. Capturing `stderr` as well as `stdout`.

The latter is intended to avoid an [ugly deadlock situation](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.stderr) that can occur if either output buffer is left to accumulate messages without being read.

I also made use of the `universal_newlines` argument to `subprocess.Popen()`, to capture the streams as text instead of as bytes. Fingers crossed that Windows and the AppImage are happy with that.

Fixes #3268 (I hope)

### DIFF
--- a/src/windows/views/blender_listview.py
+++ b/src/windows/views/blender_listview.py
@@ -722,7 +722,7 @@ class Worker(QObject):
 
             # Check the version of Blender
             import shlex
-            log.info("Checking Blender version, command: {}".format(" ".join([shlex.quote(x) for x in command_version])))
+            log.info("Checking Blender version, command: {}".format(" ".join([shlex.quote(x) for x in command_get_version])))
 
             self.process = subprocess.Popen(command_get_version, stdout=subprocess.PIPE, stderr=subprocess.PIPE, startupinfo=startupinfo, universal_newlines=True)
 
--- a/src/windows/views/blender_listview.py
+++ b/src/windows/views/blender_listview.py
@@ -720,14 +720,14 @@ class Worker(QObject):
             command_get_version = [self.blender_exec_path, '-v']
             command_render = [self.blender_exec_path, '-b', self.blend_file_path, '-P', self.target_script]
 
-            # debug info
-            # NOTE: If the length of the command_render list changes, update to match!
-            log.info("Blender command: {} {} '{}' {} '{}'".format(*command_render))
+            # Check the version of Blender
+            import shlex
+            log.info("Checking Blender version, command: {}".format(" ".join([shlex.quote(x) for x in command_version])))
 
-            self.process = subprocess.Popen(command_get_version, stdout=subprocess.PIPE, stderr=subprocess.PIPE, startupinfo=startupinfo)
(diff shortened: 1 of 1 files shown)


jonoomph:
@ferdnyc I was just looking into this issue also. I'll give this PR a quick test on my Windows system... since that is the only place I can reproduce the freeze on frame 13.

jonoomph:
FYI: In my tests, on Windows, it's freezing on the stdout.readline() command.

ferdnyc:
> @ferdnyc I was just looking into this issue also. I'll give this PR a quick test on my Windows system... since that is the only place I can reproduce the freeze on frame 13.

I'm guessing the issue is just that Linux and MacOS have _much_ larger buffers for both stdout and stderr, since the Blender run is definitely producing stderr output on all platforms. (Probably worth figuring out how to squash those messages anyway, but their cause wasn't immediately obvious to me when I first discovered them, and fixing the process to be resilient against stderr output leading to deadlocks seemed like the higher priority overall.

Issue #3280, opened 2020-03-06
Openshot stopped running unexpectedly


Zloyparen:
[openshot-qt.log](https://github.com/OpenShot/openshot-qt/files/4298125/openshot-qt.log)
![shot-2020-03-06_15-35-54](https://user-images.githubusercontent.com/61851365/76080868-50f85400-5fc1-11ea-8b89-60593ca4c92d.jpg)

SuslikV:
`OPENSHOT_VERSION_FULL` AttributeError

Look for the solution (libopenshot lib update) here: https://github.com/OpenShot/openshot-qt/issues/3235#issuecomment-590540210

ferdnyc:
Ugh. Ok, that's it, we're definitely fixing the PPA. `openshot-qt` has a VERSIONED dependency on `python3-openshot` (which in turn depends on a specific libopenshot version), and if that's not codified in the package metadata then things break.

@jonoomph, are you able to give me commit access to the daily PPA repo, or should I shoot patches your way?

In terms of your install, @Peacepill, just run
```sh
sudo apt upgrade python3-openshot
```

That should solve it.

Peacepill:
> Ugh. Ok, that's it, we're definitely fixing the PPA. `openshot-qt` has a VERSIONED dependency on `python3-openshot` (which in turn depends on a specific libopenshot version), and if that's not codified in the package metadata then things break.
> 
> @jonoomph, are you able to give me commit access to the daily PPA repo, or should I shoot patches your way?
> 
> In terms of your install, @Peacepill, just run
> 
> ```shell
> sudo apt upgrade python3-openshot
> ```
> 
> That should solve it.

thanks it solved my problem

pleabargain:
installing from PPA 6JULY.
for
OpenShot version 2.5.1-dev2

errors:

<pre>openshot-qt 
Loaded modules from installed directory: /usr/lib/python3/dist-packages/openshot_qt
         app:INFO ------------------------------------------------
         app:INFO             Mon Jul  6 21:49:25 2020            
         app:INFO               Starting new session              
         app:INFO ------------------------------------------------
         app:INFO          OpenShot (version 2.5.1-dev2)          
         app:INFO ------------------------------------------------
         app:INFO openshot-qt version: 2.5.1-dev2
     metrics:ERROR Error determining OS version in metrics.py
      logger:ERROR Traceback (most recent call last):
[...]

PR #3281, opened 2020-03-07
Docs: Generate User Manual HTML without Google Analytics (which is only needed for the official website version)

ferdnyc:
Working on fixes to the OpenShot PPA definition (see #3280), one of the things I'd like to do is fix `openshot-qt-doc` so that it actually contains a local copy of the User Guide HTML.

However, `lintian` complains about the generated docs, because they include Google Analytics tracking.

With these changes, instead of unconditionally including GA in the generated output, we revert to using the theme's built-in support for conditional analytics (controlled by the
`html_theme_options` config dictionary).

Instead of hardcoding the analytics embed, `html_theme_options.analytics_id` is set on the documentation-rebuild command line for the official OpenShot site updates, in `.gitlab-ci.yml`. This ensures that GA will only be included in _those_ documentation builds, not local/user copies of the manual.


ferdnyc:
Merging this, for the benefit of PPA builds (and all other local doc builds).

PR #3285, opened 2020-03-09
Allow *.osp project files to be imported as a Clip

jonoomph:
- Allow *.osp project files to be imported as a Clip. 
- Better handle unknown "media_type"'s.  
- Trigger "has unsaved changes" when clearing history from a Project, so the user can immediately save the *.osp file without history.

A few misc things, but all related to my work on allowing *.osp files to be imported as a Clip, without deadlocking the system, or crashing anything. :wink: 

Related PR for `libopenshot`: https://github.com/OpenShot/libopenshot/pull/459

### DIFF
--- a/src/windows/views/files_listview.py
+++ b/src/windows/views/files_listview.py
@@ -1,27 +1,27 @@
-""" 
+"""
  @file
  @brief This file contains the project file listview, used by the main window
  @author Noah Figg <eggmunkee@hotmail.com>
  @author Jonathan Thomas <jonathan@openshot.org>
- 
+
  @section LICENSE
- 
+
  Copyright (c) 2008-2018 OpenShot Studios, LLC
  (http://www.openshotstudios.com). This file is part of
  OpenShot Video Editor (http://www.openshot.org), an open-source project
  dedicated to delivering high quality video editing and animation solutions
  to the world.
- 
+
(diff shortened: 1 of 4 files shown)


ferdnyc:
> * Trigger "has unsaved changes" when clearing history from a Project, so the user can immediately save the *.osp file without history.

:+1: I added the same thing to the Export dialog at some point, too, so users would be reminded to commit their updated `export_path` to the project file.

(Now I have to go back and fix things so it only sets the flag if the export path is _different_ from the previous value. Because it's a bit annoying having to save every time you export, if you use the same output directory each time.)

src/windows/views/files_listview.py
```diff
@@ -151,6 +151,9 @@ def add_file(self, filepath):
                 file_data["media_type"] = "image"
             elif file_data["has_audio"] and not file_data["has_video"]:
                 file_data["media_type"] = "audio"
+            else:
+                # If neither set, just assume video
+                file_data["media_type"] = "video"
```
ferdnyc: This exact same code is copypasted into `files_treeview.py` as well, it probably makes sense to keep them in sync. (Better would be to eliminate the redundancy, but short of that...)

ferdnyc: Done, pushed a commit.

PR #3291, opened 2020-03-11
A lot of code cleanup, in timeline_webview

ferdnyc:
- Rewrap lots of long lines, reindent, fix spacing
- Replace `print()` with `log.info()`
- Eliminate unused imports, variables
- No bare Excepts
- Eliminate `prop_name` variables, which are unnecessarily obtuse
- Get rid of `translations` list, incorporate translated strings into regular flow of code
  (sometimes using tuples, to pair them with untranslated partners)
- Reformat several long lists as item-per-line
- Use `.format()` to assemble JS calls

### DIFF
--- a/src/windows/views/timeline_webview.py
+++ b/src/windows/views/timeline_webview.py
@@ -28,12 +28,10 @@
  """
 
 import os
-import sys
 import time
 from copy import deepcopy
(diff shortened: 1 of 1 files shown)


jonoomph:
LGTM!

PR #3306, opened 2020-03-21
classes.info: Use current year in COPYRIGHT

ferdnyc:
This way, at least _that_ string will stay up to date. (Especially since it gets encoded into the User Guide when it's generated for the website, and displayed in the "About" dialog.)

### DIFF
--- a/src/classes/info.py
+++ b/src/classes/info.py
@@ -26,6 +26,7 @@
  """
 
 import os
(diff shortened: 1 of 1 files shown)


ferdnyc:
Merging, to fix the displayed copyright year in both the application and the official online copy of the User Guide.

PR #3312, opened 2020-03-23
Fix Edit Title, Duplicate Title shortcut overlap with Add Track

### LINKED ISSUE #3240: "Add to timeline" dialog contains no files or only one

### LINKED ISSUE #3310: "duplicate titles" hotkey does not work
**Describe the bug:**

"duplicate titles" hotkey does not work.
 I wanted to reassign it, but there is not even such an option in the hotkey settings.
The function itself works if you click on the context menu.

**System Details:**
- Operating System:  Windows 10

ferdnyc:
A user reported that the Duplicate Title shortcut both (a) didn't work, and (b) wasn't editable. Turned out, it was set to <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>T</kbd>, which is a problem because so was Edit Title. Worse, _neither_ of those actually responded to <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>T</kbd>, because <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>T</kbd> is Add Track! They also appeared nowhere in the settings file.

I managed to fix that, made them both configurable, with Duplicate Track set to <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>C</kbd> — it was free — and Edit Title not set by default.

As part of making the shortcuts properly react to selection in the file views, I rewrote a bunch of the selection-handling logic so it's less easily confused, and accidentally fixed #3240 in the process.

Fixes #3310 
Fixes #3240

### DIFF
--- a/src/settings/_default.settings
+++ b/src/settings/_default.settings
@@ -687,10 +687,6 @@
     "value": "-",
     "type": "text"
   },
-
-
-
-
   {
     "category": "Keyboard",
     "title": "Previous Frame",
@@ -905,6 +901,22 @@
     "restart": true,
     "setting": "nudgeRight",
     "value": "Shift+Right",
+    "type": "text"
+  },
+  {
+    "category": "Keyboard",
+    "title": "Edit Title",
+    "restart": true,
+    "setting": "actionEditTitle",
+    "value": "",
+    "type": "text"
+  },
+  {
+    "category": "Keyboard",
+    "title": "Duplicate Title",
+    "restart": true,
+    "setting": "actionDuplicateTitle",
+    "value": "Ctrl+Shift+C",
     "type": "text"
   }
 ]
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -395,20 +395,13 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         file = File.get(id=selected_file_id)
         file_path = file.data.get("path")
 
-        # Delete thumbnail for this file (it will be recreated soon)
-        thumb_path = os.path.join(info.THUMBNAIL_PATH, "{}.png".format(file.id))
-
-        # Check if thumb exists (and delete it)
-        if os.path.exists(thumb_path):
-            os.remove(thumb_path)
-
         # show dialog for editing title
         from windows.title_editor import TitleEditor
         win = TitleEditor(file_path)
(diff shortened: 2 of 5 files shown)


SuslikV:
Conflict of interests with the: https://github.com/OpenShot/openshot-qt/pull/3264

ferdnyc:
> Conflict of interests with the: #3264

Is it? Darn. I haven't looked at that one yet.

They may be reconcilable — the stuff I touched mostly uses the <code>Q<b>Selection</b>Model</code> (which will still be there regardless, in some form), and anything that's retained gets immediately translated into OpenShot File IDs. (As is required with regular `QModelIndex`es, since they're not persistent.)

We may need to shuffle some things around a bit between us, but hopefully solvable.

SuslikV:
I'm using slightly different solution for the selection - instead of Qt signal, I'm running selection update before the function that will use the results of the selection (it's only few places). Thus, some Qt events may come later if PC is busy and this is OK.

jonoomph:
@ferdnyc I'm worried this might also conflict with the Emoji branch, but I'm gonna just merge it, and then reconcile the changes on the Emoji branch. Merging now. Thanks!

PR #3318, opened 2020-03-24
Enhance security by preferring defusedxml to the standard xml, if it's available

ferdnyc:
The Python stdlib xml module has several known vulnerabilities in its XML parsing code, which defusedxml corrects. We attempt to load defused, and if it fails to import, just continue with stdlib.

Python docs for the `xml` module, noting its vulnerabilities to maliciously-crafted XML: https://docs.python.org/3/library/xml.html

### DIFF
--- a/src/classes/ui_util.py
+++ b/src/classes/ui_util.py
@@ -1,35 +1,40 @@
-""" 
+"""
  @file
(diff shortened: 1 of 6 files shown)


jonoomph:
LGTM!

PR #3320, opened 2020-03-26
Move openshot-arrow logo (for User Guide) to xdg dir

ferdnyc:
The logo design in `src/images/openshot-arrow.png` file is only ever used when building the User Guide sidebar (it's never referenced in the OpenShot code at all), so this change moves it to the `/xdg` dir and includes it from there.

I thought about moving it into the `/doc` dir instead, but it might theoretically have other uses at some point. (We should really have a `/data` or `/images` directory for things like this that aren't part of the application source, but _are_ part of the project. We **should** — but we don't, yet.)


ferdnyc:
Merging this as it's simple housekeeping.

PR #3321, opened 2020-03-26
Mangle hardware library builtin paths, in libs bundled into AppImage builds

### LINKED ISSUE #3210: Crash when launching Preferences
**Describe the bug:**
Openshot crashed when I tried to launch the Preferences.

1. Launch AppImage via terminal window.
2. Click on Preferences
3. See error in the terminal

**Expected behavior:**
No crash when launching Preferences.

**System Details:**
- OpenShot Version v.2.5.0 Pre-Release
- Operating System / Distro: PopOS/Ubuntu 19.10

**Exception / Stacktrace:**
```
libva info: Trying to open /usr/lib/x86_64-linux-gnu/dri/i965_drv_video.so
libva error: /usr/lib/x86_64-linux-gnu/dri/i965_drv_video.so has no function __vaDriverInit_0_32
libva info: va_openDriver() returns -1
[AVHWDeviceContext @ 0x21b35c0] Failed to initialise VAAPI connection: -1 (unknown libva error).
```

**Screenshots:** (Optional)
Screenshot from 2020-02-09 17-07-32

ferdnyc:
This PR adds a script (`installer/mangle-hw-libs.sh`) which uses `sed` to modify the contents of `libvdpau.so.1` and `libva.so.1`, replacing the hardcoded driver paths they contain with modified ones that point into the AppImage, instead of at the system drivers.

Since we don't bundle hardware driver libs, this will have the effect of disabling those libraries' access to the hardware drivers, preventing crashes due to mismatched driver versions on many systems.

See the extensive discussion in #3210 for rationale and details.

Fixes #3210, possibly others (though not any Windows crash-on-opening-Preferences issues)


SuslikV:
Bit-hack is bad decision. No better workarounds?

ferdnyc:
> Bit-hack is bad decision. No better workarounds?

There really aren't. The paths are compiled into the libraries, which we _need_ because they're linked with other libraries in use, but we also need them to **not** try to access those paths on the host system. Short of a fully-virtualized container that restricts access to the filesystem beyond the sandbox, monkeypatching the hardcoded paths is the best solution anyone's come up with. It's actually a time-honored technique that goes back decades.

Someday it may stop working (like, when library objects are signed and authenticated), but that day isn't anywhere in sight for enduser, non-hardened Linux systems generally.

The only other option would be to build our own custom versions of the libs from patched sources, and there are plenty of downsides to that approach as well.

ferdnyc:
I wrote the expressions as safely as possible, out of an abundance of caution, and `sed` will simply do nothing at all if it doesn't find the exact paths it's expecting in the binaries it scans.

Like, for this replacement:
https://github.com/OpenShot/openshot-qt/blob/7613425802525747322daf3fb2f83ee7d133a398/installer/mangle-hw-libs.sh#L32-L35

That _**could**_ be written as `sed -i -e 's,/usr/lib,./../lib,g' ${hwlib}`, if you feel like breaking your libraries at some point. (And for sure don't accidentally run it on the WRONG library, or you're screwed.) So, despite the temptation, I decided to do exactly not that.

ferdnyc:
(Although, come to think of it all that backslash-escaping on the periods in the replacement strings really _is_ totally unnecessary, and sure makes it harder to read.)

SuslikV:
Madness. I think, that AppImage usage ups hype and decreasing reputation.

ferdnyc:
> Madness. I think, that AppImage usage ups hype and decreasing reputation.

Yeah, I'm not particularly a fan either. My issue is that, no matter how well-designed the AppImage format/tools are, they're based on a faulty premise: That it's _possible_ to create distribution-agnostic Linux binary packages. I think that's been pretty handily disproven.

Their punt has been, "Build them for the oldest system available" on the theory that they should then be forwards-compatible on newer systems. That's fine for simple software applications, but the moment you mix in any sort of hardware drivers or APIs it totally breaks down.

Not to mention, it forces the use of outdated and feature-limited APIs even when they _are_ compatible. I'm constantly feeling held back by the need to ensure that **new** code we write is still compatible with 3-year-old Ubuntu releases. ...Why? If people insist on running 3-year-old Ubuntu releases, they should stick to running 3-year-old OpenShot releases. Otherwise, keep up!

Aaanyway. Merging this, since disabling hardware acceleration across the board in AppImage-land seems to be the safest course of action.

PR #3335, opened 2020-03-31
Make screenshots at export-like quality

SuslikV:
Small improvement for _Save Current Frame_ function of _Preview_ to get better quality.

Not much. It less noticeable for 4:2:0 inputs. That's all.

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -1021,6 +1021,9 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
             self.statusBar.showMessage(_("Save Frame cancelled..."), 5000)
             return
 
+        # Set scale mode to higher quality (to skip chroma optimizations of preview and get Export-like quality)
+        openshot.Settings.Instance().HIGH_QUALITY_SCALING = True
+
         # Append .png if needed
         if not framePath.endswith(".png"):
             framePath = "%s.png" % framePath
@@ -1031,8 +1034,8 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         # Pause playback (to prevent crash since we are fixing to change the timeline's max size)
         app.window.actionPlay_trigger(None, force="pause")
 
-        # Save current cache object and create a new CacheMemory object (ignore quality and scale prefs)
-        old_cache_object = self.cache_object
+        # Create a new CacheMemory object (ignore quality and scale prefs)
+        self.timeline_sync.timeline.ClearAllCache()
         new_cache_object = openshot.CacheMemory(settings.get_settings().get("cache-limit-mb") * 1024 * 1024)
         self.timeline_sync.timeline.SetCache(new_cache_object)
 
@@ -1055,13 +1058,13 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         else:
             self.statusBar.showMessage(_("Failed to save image to %s" % framePath), 5000)
 
+        # Return scale mode to lower quality scaling (for faster previews)
+        openshot.Settings.Instance().HIGH_QUALITY_SCALING = False
+
         # Reset the MaxSize to match the preview and reset the preview cache
         viewport_rect = self.videoPreview.centeredViewport(self.videoPreview.width(), self.videoPreview.height())
         self.timeline_sync.timeline.SetMaxSize(viewport_rect.width(), viewport_rect.height())
-        self.cache_object.Clear()
-        self.timeline_sync.timeline.SetCache(old_cache_object)
-        self.cache_object = old_cache_object
-        old_cache_object = None
+        self.InitCacheSettings()
         new_cache_object = None
 
     def actionAddTrack_trigger(self, event):


src/windows/main_window.py
```diff
@@ -1021,6 +1021,12 @@ def actionSaveFrame_trigger(self, event):
             self.statusBar.showMessage(_("Save Frame cancelled..."), 5000)
             return
 
+        # Set scale mode to higher quality (to skip chroma optimizations of preview)
+        openshot.Settings.Instance().HIGH_QUALITY_SCALING = True
+
+        # Clear current frame from cache to get export-like quality image
+        self.timeline_sync.timeline.Remove(self.preview_thread.current_frame)
```
ferdnyc:
```
  File "/home/ferd/rpmbuild/REPOS/openshot-qt/worktree-testing/src/windows/main_window.py", line 1028, in actionSaveFrame_trigger
    self.timeline_sync.timeline.Remove(self.preview_thread.current_frame)
AttributeError: 'Timeline' object has no attribute 'Remove'
```

...I don't think it should be necessary to flush the cache anyway. The entire cache gets _replaced_ with a temporary one a few lines later, **before** the call to save the image to disk. So, whatever's in the cache when this line is executed isn't what gets saved anyway.

ferdnyc: https://github.com/OpenShot/openshot-qt/blob/accfd6a137bcc6bbfc8986a383f0705f8ae6ff75/src/windows/main_window.py#L1034-L1041

src/windows/main_window.py
```diff
@@ -1021,6 +1021,12 @@ def actionSaveFrame_trigger(self, event):
             self.statusBar.showMessage(_("Save Frame cancelled..."), 5000)
             return
 
+        # Set scale mode to higher quality (to skip chroma optimizations of preview)
+        openshot.Settings.Instance().HIGH_QUALITY_SCALING = True
```
ferdnyc: Other than fixing the traceback, the only thing I'd suggest is storing the old value here, and restoring it after — so that, if it's _not_ `False` already, you don't accidentally **force** it false.
```suggestion
        prev_scaling = openshot.Settings.Instance().HIGH_QUALITY_SCALING
        openshot.Settings.Instance().HIGH_QUALITY_SCALING = True
```

src/windows/main_window.py
```diff
@@ -1055,6 +1061,9 @@ def actionSaveFrame_trigger(self, event):
         else:
             self.statusBar.showMessage(_("Failed to save image to %s" % framePath), 5000)
 
+        # Return scale mode to lower quality scaling (for faster previews)
+        openshot.Settings.Instance().HIGH_QUALITY_SCALING = False
```
ferdnyc:
```suggestion
        openshot.Settings.Instance().HIGH_QUALITY_SCALING = prev_scaling
```

ferdnyc:
Oh, wow. When a still image is placed on the Timeline at full resolution (e.g. a `1920x1080` PNG file on a timeline set up for `1080p`), the difference in snapshot-tool output quality when the clip is set to `Scale: Best Fit` vs. `Scale: None` is **IMMENSE**. Like, disturbingly so. That's gotta be a bug.

src/windows/main_window.py
SuslikV: I copied wrong code here. But...

> The entire cache gets _replaced_ with a temporary one a few lines later, **before** the call to save the image to disk...

this doesn't work as it written (or intended to work).

src/windows/main_window.py
SuslikV: Changes not required. It is set during init of the main window and switched back and force only in Export.

src/windows/main_window.py
SuslikV: More over, the https://github.com/OpenShot/libopenshot/commit/968e472c73bccac9179a139fa7584a05acf8a95b
set some limits on: `Timeline::SetCache` method, so mentioned "replace" is questionable thing. VERY questionable THING.

I think that I need simple:
```
        # Clear timeline preview cache (to set new quality)
        self.timeline_sync.timeline.ClearAllCache()
```
 (as I actually did in my own code) This thing is surely works.

SuslikV:
Small update. Cache handling was erroneous after the OpenShot/libopenshot@968e472 , so this PR fixing this too. Cache is discarded.

Screenshots is useful when you making preview thumbnail for the YouTube. I didn't found better usage for it.

SuslikV:
Hmm, something wrong with the scaling. This still not like export quality... I'm going to see if image export (similar to sequence export can be done here).

SuslikV:
It turns out that _libopenshot_ is missing SWS_ACCURATE_RND and SWS_FULL_CHR_H_INT flags (equivalent of `-vf scale=flags=accurate_rnd+full_chroma_int`) for high quality output.

I have plans to add required flags together with the color range/space fixes to _libopenshot_.

So, re-opening it again ^_^

src/windows/main_window.py
ferdnyc: For now! But can you guarantee that code won't be added in the future which sets it to `True` somewhere, for some reason? (Or maybe it's just _always_ set to `True`?)

Why should whoever makes that change then have to go and find **_this_** code, so they can fix it to stop forcibly resetting the value to `False`? It can easily be made safe for any existing value of the setting, and then we _do_ have a guarantee that there will be no such problem in the future.

My favorite kind of future problems are the ones that get solved before they even happened.

SuslikV: If I would secure any possibilities - the only thing this application will do is checking, checking, and checking. Without any useful result in the end. OpenShot is by design has low quality until saving. I'm just using this rule.

jonoomph:
Closing PR. The "screenshot" feature is already complex and flawed, no need to make it more complex. I am planning on either removing the screenshot button, or drastically simplifying it's implementation with some new Clip refactoring that I'm working on now.

PR #3340, opened 2020-04-02
Editable time label

ElDonad:
I've slightly modified the time label in the timeline to make it editable by the user. 

I'm not a developper by profession, but I needed this functionnality, and though I could share it with the community... I havn't found any issue so far, but I've not extensively tested it.

### DIFF
--- a/src/windows/title_editor.py
+++ b/src/windows/title_editor.py
@@ -354,7 +354,6 @@ class TitleEditor(QDialog):
         if col.isValid():
             self.set_font_color_elements(col.name(), col.alphaF())
             self.update_font_color_button()
-            self.font_color_code = col
 
             # Something changed, so update temp SVG
             self.writeToFile(self.xmldoc)
@@ -374,7 +373,6 @@ class TitleEditor(QDialog):
         if col.isValid():
             self.set_bg_style(col.name(), col.alphaF())
             self.update_background_color_button()
-            self.bg_color_code = col
 
             # Something changed, so update temp SVG
             self.writeToFile(self.xmldoc)
@@ -461,13 +459,13 @@ class TitleEditor(QDialog):
             else:
                 text_color = QtGui.QColor(Qt.black)
 
-            # Convert the opacity into the alpha value
-            alpha = int(opacity * 65535.0)
-
-            # Set the colors of the button
+            # Set the color of the button, ignoring alpha
             self.btnFontColor.setStyleSheet(
                 "background-color: %s; opacity: %s; color: %s;"
-                % (color.name(), alpha, text_color.name()))
+                % (color.name(), 1, text_color.name()))
+
+            # Store the opacity as the color's alpha level
+            color.setAlphaF(opacity)
             self.font_color_code = color
 
     def update_background_color_button(self):
@@ -519,13 +517,13 @@ class TitleEditor(QDialog):
             else:
                 text_color = QtGui.QColor(Qt.black)
 
-            # Convert the opacity into the alpha value
-            alpha = int(opacity * 65535.0)
-
-            # Set the colors of the button
+            # Set the colors of the button, ignoring opacity
             self.btnBackgroundColor.setStyleSheet(
                 "background-color: %s; opacity: %s; color: %s;"
-                % (color.name(), alpha, text_color.name()))
+                % (color.name(), 1, text_color.name()))
+
+            # Store the opacity as the color's alpha level
+            color.setAlphaF(opacity)
             self.bg_color_code = color
 
     def set_font_style(self):


src/windows/views/timeline_webview.py
```diff
@@ -2595,15 +2595,8 @@ def update_zoom(self, newValue):
...
-        if (key_value == Qt.Key_Shift or key_value == Qt.Key_Control):
-
-            # Only pass a few keystrokes to the webview (CTRL and SHIFT)
-            return QWebView.keyPressEvent(self, event)
-
-        else:
-            # Ignore most keypresses
-            event.ignore()
+        #keyboards events are necessary for the time input field to work. No apparent conflict with the rest of the interface.
+        return QWebView.keyPressEvent(self, event)
```
ElDonad, line 2599: Not sure if a more specific filter is needed. Doesn't seems to conflict with any other shortcut in the editor.

ferdnyc: No, _conflicts_ per se don't seem to be a problem, agreed. One issue I have hit on in testing, though, is that the editable field makes it much easier to inadvertently end up focusing it — which will then interfere with several keyboard shortcuts, including <kbd>Space</kbd> to toggle Play/Pause.

To be fair, the Properties table has the same issue. After it's been focused for keyboard navigation, it can have a tendency to "re-grab" focus on subsequent inputs that you wouldn't expect to go to the table. But it's especially disconcerting here, because any click on the ruler will also focus the time field. So, you can't perform common actions like positioning the playhead with a ruler click, followed by <kbd>Space</kbd>to play from that point.

The timeline arrow-key navigation bindings, like <kbd>←</kbd>, <kbd>→</kbd>, <kbd>Ctrl</kbd>+<kbd>←</kbd>, <kbd>Ctrl</kbd>+<kbd>→</kbd>, etc. are also blocked by the focus grab, unless you first tab out of the field. (The last two <kbd>Ctrl</kbd> combos I mentioned will jump left or right to the next `:` in the time field when it's focused, interestingly.)

I'll comment more generally below, but I wanted to raise that issue here since it pertains to keypress events in particular. I don't think these issues are _unsolvable_, by any means, but... well, like I said, I'll leave an overview response in a separate PR comment.

ferdnyc:
> I've slightly modified the time label in the timeline to make it editable by the user.
> 
> I'm not a developper by profession, but I needed this functionnality, and though I could share it with the community... I havn't found any issue so far, but I've not extensively tested it.

@ElDonad Thanks! Always exciting to see users contributing back to the project, you are after all the ones who know best what you need from OpenShot. I ran OpenShot with your PR branch code and tried it out. It's definitely an interesting approach, and I think it's got a lot of potential to be a great feature addition.

##### Instead of Always-on, sometimes-off?
I ran into a few bugs that I'll detail specifically (I already reported one, above), but before going too far down that road, I wanted to bring up a more general issue. I... _harbor some doubts_, let's just say, that having an always-editable time field is the safest approach to implementing the necessary functionality. As things currently stand it's freely editable under all circumstances, even when OpenShot is in playback mode, which is probably my single biggest concern. I think if we're going to do it this way, if _nothing_ else it needs to be disabled when playback starts, and re-enabled when it ends. Which is, again, not impossible, it just requires some additional code.

##### Updating Qt
The input-processing code also needs to update the Python/Qt side of things, which currently it isn't doing. The ruler event handler's `scope.PreviewFrame()` call was the part that notified Qt of the position change, so that's a critical function on the JS side. Without it, position changes are confined to the JavaScript code, which becomes out of sync with the Qt side — meaning that if you (for example) edit the time field to change position, then add a keyframed Property value, that keyframe won't be placed at the (visibly-current) playhead position. When you start playback after moving the playhead, it also doesn't begin from the repositioned location. Restoring `scope.PreviewFrame(playhead_seconds)` to the `tlRulertime` directive in `ruler.js` should hopefully solve that, so not a major concern.

#### "Edit the time field", or "Go to this time index"?
I wanted to take a step back and ask a more general question, though: You say the motivation for this was needing an editable time field, but I'm wondering if that's the real need here. Would you say that what you _actually_ need from a functionality standpoint is a "go to time" feature, with an editable field being just one way to implement that function? Or does it really need to take the form of editing the field itself. (I'm honestly asking — I'll accept either answer, just felt the question should be asked and considered.)

I'm just thinking that it might be less disruptive to the general editing process, and make "go to time" actions easier to process, if it were driven by the Qt side of the world, and the JS left as mostly just a presentation layer. For instance, "Go to time" could take input in a popup edit field or even possibly a dialog window, something that's triggered by Qt in response to a click event or keyboard shortcut, then processed and forwarded to the JavaScript for visualization.

But I'm _honestly_ not trying to dismiss or meddle with your approach, and if an editable field is really determined to be the best/only way to accommodate the need then like I said, I can accept that and I feel the issues are all solvable. I just think it will probably take a bit of effort to cover all these bases. (To use a very played-out baseball expression.) That's not me saying there's anything wrong with or lacking in your code, it's just the nature of any interactions across the Python/JavaScript barrier, which are inherently complex and tend to be best avoided where possible.

But I'm impressed by what's here so far, and especially appreciate you taking the time and initiative to give back to the project and, by extension, the rest of the users. I'm hoping we can work together on polishing this up and incorporating it into OpenShot so that it can continue to benefit all of us. Thanks!

SuslikV:
The "frame" field has no much sense for edit, because it's only per frame number not global value.

The best approach here is small pop-up dialog with the three fields for time to edit with the header `Navigate to:` (and maybe later for global frame number). It's my own thoughts about this.

ElDonad:
Thank you kindly for your reviews ! 

Yes @ferdnyc, a "go to time" function is exactly what I need. I considered both approaches and went for this because it seemed the most intuitive for me, but it's only my subjective point of view and a dialog box would be totally acceptable too. And about disabling the field during playback, I will see what I can do. Since I'm totally inexperienced with this codebase, I can't garantee anything until I've tried it ^^.

Furthermore, as @SuslikV pointed, the frame label is editable too, which is not ideal. I added it since I needed fine positionning of the cursor, but I would have preferred hundredth of seconds. In facts that's what I initially implemented, but it didn't make sense since it was translated back to frame number when exiting the input field. So maybe a dialog box which would ask specifically for a timestamp is a better approach. 

For the focus "re-grabbing", I think the problem could be solved by allowing the user to easily focus out the input field, whith 'escape' for example. When I use the input field, hitting escape to get out is a natural reflex for me, so maybe this could be implemented. 

So I will do a few bugfixes following your reviews, but the popup window might end up the best solution. However I like the simplicity of this approach. When I use OpenShot, being able to click the timestamp in order to edit it seems natural to me. Maybe the popup box could appear on clicking the timestamp instead of it being directly editable ? What do you think ?

ElDonad:
Ok, here we are. I have also a strange bug where the frame is shifted by 1 whenever the input is updated, not sure where it comes from...

Anyway, I will get into implementing a dialogBox soon, my Qt is a bit too rusty for that right now ^^.

ferdnyc:
> EDIT : What happened to Travis ? Not me I swear !

It just does that sometimes.

Normally I'd have the ability to just restart the job, and it would go through... but I'm not seeing that option! Pushing any random commit (or even just force-pushing the branch in the same state, probably) should kick off a new build, which will probably be fine.

ferdnyc:
> Yes @ferdnyc, a "go to time" function is exactly what I need. I considered both approaches and went for this because it seemed the most intuitive for me, but it's only my subjective point of view and a dialog box would be totally acceptable too.
The biggest advantage to a dialog is... well, two major ones, that I can think of:
1. It's driven from the Qt side, which is really where you want the positioning to be controlled from
1. It can be triggered with a keyboard shortcut.

Binding a Qt shortcut to "enable and focus this input field in the Timeline WebView" is probably doable, but it's likely going to be a hassle. And having a keyboard combo that triggers "go-to-time" seems like a usability win (in addition to handily sidestepping the keyboard-focus issue).

Using a [QLineEdit](https://doc.qt.io/qt-5/qlineedit.html) would also allow the use of an [input mask](https://www.ics.com/blog/qt-support-input-masks-and-validators) on the field, so we could enforce the `99:99:99:99` format as the only valid one.

(It is possible to take input of frame numbers, as well — the values just have to be converted correctly. There's an existing function on the Python side to properly convert to linear seconds, if the go-to-time feature stays on the JavaScript side then `timecodeToSeconds()` should be reimplemented in JS and used to do the conversion.)

https://github.com/OpenShot/openshot-qt/blob/33fd76bd9429155357dd051d21e47d25a059abac/src/classes/time_parts.py#L56-L68

ferdnyc:
> Using a [QLineEdit](https://doc.qt.io/qt-5/qlineedit.html) would also allow the use of an [input mask](https://www.ics.com/blog/qt-support-input-masks-and-validators) on the field, so we could enforce the `99:99:99:99` format as the only valid one.

There _is_ also [a jQuery plugin for that](http://igorescobar.github.io/jQuery-Mask-Plugin/).

ElDonad:
Yeah, it seems the most reasonable approach. I will now reset this branch and start working on this solution (as soon as git let me do that)

ferdnyc:
@ElDonad I'm going to assume that this was closed accidentally, and reopen it. (If it was closed intentionally because you're planning to submit a new PR from a different branch, instead, feel free to re-close.)

My branch [`ferdnyc:gototime`](https://github.com/ferdnyc/openshot-qt/tree/gototime) contains a WIP commit 14c815832ac272db310702fd9dbf7d6cc7d4eea8 which (in addition to adding a 'goToTime' keyboard shortcut, by default bound to "G", to the `_default.settings` file) includes a demonstration dialog UI in `src/windows/ui/go-to-time.ui`. I've used `pyuic5` to generate runnable python code, which is also included as `src/windows/ui/go-to-time.py`.

You can run it with just `python3 go-to-time.py` from the `src/windows/ui` directory, as long as your Python installation includes PyQt5.

Regarding the dialog, or more specifically the time editing, in my mind there are three ways it could be done, each of which is demonstrated in the UI:

image

1. **We could use a QLineEdit with an input mask of `99:99:99:99`.**
    This is nice because, when combined with VisualMoveStyle cursor movement and the setCursorPosition feature, we can automatically position the cursor in the section with the largest non-zero value, and as the user types it'll automatically overwrite digits (insert-mode). We can't easily prevent them entering bad values (like `00:70:00:00`), though we _can_ add logic to auto-correct on focus out — whether in that example "corrected" would mean `00:59:00:00`, `00:59:59:30`, or `01:10:00:00`, it's our choice how we wanted to handle that.

1. **We could use four separate QSpinBoxes, one for each segment of the time**
    Again, the cursor can be positioned in the first section with a non-zero value, by default. There are a few big advantages here:
    * Each segment can have the appropriate min/max set — `0`-`999` for hours, `0`-`59` for minutes and seconds, and `1`-FPS for frames.
    * Values can be adjusted up/down using both the arrow keys and the mouse wheel, in addition to the spinbox buttons and by typing
    There are also some disadvantages, though:
    * You have to <kbd>Tab</kbd> between sections
    * Because of that, you can't easily just type in a time value, nor can you cut-and-paste one
    Advanced correction/streamlining of user input might be possible. (For instance, if they raise the Seconds value to 60, we automatically increment the Minutes value by 1, and reset Seconds to 0.) However, that would require significant additional code.

1. **We could use QDateTimeEdit, or a customized subclass**
    QDateTimeEdit is a specialized Qt widget that implements basically a combination of the previous two options. It automatically segments the time string into H:M:S components and allows values to be typed like in QLineEdit, or raised/lowered with the arrow keys or mouse just like a QSpinBox. It automatically handles the proper min/max values for each segment.
    A few concerns or potential issues with this approach, some or all of which _might_ be surmountable by subclassing and reimplementing a few functions:
    * Supporting frame numbers is obviously the biggest hurdle.
       QDateTimeEdit _does_ have support for millisecond precision, which is what I've used in the example dialog, but there's no 2-digit format — QDateTime's format strings only offer two options: `z` which is an unpadded raw milliseconds value (so, `3` for 3ms); or `zzz` for a padded **three**-digit representation (`003` for 3ms). The latter is intended to be used in the standard `hh:mm:ss.zzz` time representation.
       In addition to the representation, we'd also have to override some code to be able to specify a minimum value of 1 and the appropriate max for the frame rate.
    * Technically every QDateTimeEdit value is a full, valid calendar time. We can arbitrarily choose a date (1/1/2000 by Qt default) and just use that as our "starting time", but we have to be careful to convert the input value based on the _difference_ from that starting point, which is a little weird.
    * It **is** possible to set a maximum time value, so we could prevent the user entering values that exceed the Timeline's max length, but again we'd have to do that relative to 1/1/2000 12:00:00 AM.

The QAbstractSpinBox class actually has an API to specify arbitrary multi-section values with grouping and separator characters, so that it's possible to implement "QDateTimeEdit-like" functionality with arbitrary, non-datetime-based formats — dotted-decimal IP addresses are one example they've used — on top of the abstract class. But, some of the features we'd need weren't added until Qt 5.3 which unfortunately makes them unavailable to us since the AppImage builds are still on Qt 5.2.

Without access to QAbstractSpinBox, and despite the hurdles, I'm kind of partial to the QDateTime-based option myself, as it seems like it presents the most flexible UI for the user. But I'm curious to hear your impressions, especially if you have a chance to try out the demo dialog.

ferdnyc:
> @ElDonad I'm going to assume that this was closed accidentally, and reopen it. (If it was closed intentionally because you're planning to submit a new PR from a different branch, instead, feel free to re-close.)

Apparently I can't do that, either. Weird.

ferdnyc:
Hmm, I found another downside to QDateTimeEdit.

When you type a value in to one of the segments, by default the cursor is positioned on the **next** segment automatically. If you then _type_ the `:` segment separator, the cursor gets repositioned to the **following** segment. (So, by typing `:` characters you accidentally skip over segments.)

With QLineEdit and a format string, `:` (or any unrecognized character) is just silently ignored, which is what I was hoping QDataTimeEdit would do as well. Perhaps there's some way to override the default behavior, though...

PR #3353, opened 2020-04-06
dissolve.py: Keep track of particles for duration

### LINKED ISSUE #3350: Animated Title: "Dissolving Text" is not working properly
**Describe the bug:**
Using the animated title "Dissolving Text" result in an incomplete animation or probably.
I tried to make a rendering using all defeault values, even for the title and the name
But the result is not correct.
I do not know how this effect should be, but for sure the should not be half title reapparing at the end.
1. Basicalli it start with the whole title
2. Then the title start dissolving
3. When it is almost dissolved the title start to reappear one char per time 
4. The title do not show completely again, it stop after some char

Probably the title should show char by char and then start dissolving, but for sure it should not half show at the end of the animation

1. Go to Anomated Title
2. Select "dissolving Text" effect 
3. Click on render

**Expected behavior:**
A proper animated text.

**System Details:**
- OpenShot-v2.5.1-dev2-1585970041-166f6938-232f3f79-x86_64
- Blender: 2.82

**Screenshots:** 
Here is a screenshot of the last frame of the animation.
openshot-qt_spauUMrvsP
I attach also the video export 
[export.zip](https://github.com/OpenShot/openshot-qt/files/4434211/export.zip)

ferdnyc:
Because the particle system starts affecting particles in frame 10, a lifetime of 80 means that some of the earliest particles will begin to die off at frame 90, well before the 128-frame animation is complete. It appears that, when particles die, they "reset" to their starting point.

This change raises the lifetime to 120, to ensure that the particle system keeps control of all particles for the entire duration of the animation.

Fixes #3350

### DIFF
--- a/src/blender/scripts/dissolve.py
+++ b/src/blender/scripts/dissolve.py
@@ -113,7 +113,7 @@ def createDissolveText(title, extrude, bevel_depth, spacemode, textsize, width, 
     ActiveObjectText.particle_systems[0].settings.count = NbQuads
     ActiveObjectText.particle_systems[0].settings.frame_start = 10
     ActiveObjectText.particle_systems[0].settings.frame_end = 60
-    ActiveObjectText.particle_systems[0].settings.lifetime = 80
+    ActiveObjectText.particle_systems[0].settings.lifetime = 120
     ActiveObjectText.particle_systems[0].point_cache.frame_step = 1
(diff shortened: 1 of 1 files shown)


ferdnyc:
Merging to hopefully fix the Dissolve title template

PR #3354, opened 2020-04-06
classes.app: Set icon, desktop file

ferdnyc:
Do a couple of things to establish our application identity with various window managers, desktop session managers, etc

Fixes: 1112

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -36,7 +36,7 @@ import traceback
 from PyQt5.QtCore import PYQT_VERSION_STR
 from PyQt5.QtCore import QT_VERSION_STR
 from PyQt5.QtCore import Qt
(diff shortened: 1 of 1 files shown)


SuslikV:
```from images import openshot_rc```  - is excessive check.

It seems that it works for Windows without changes. Thus, all OK.

ferdnyc:
@SuslikV 

Cool, thanks, I was pretty confident it was cross-platform.

> `from images import openshot_rc` - is excessive check.

Not sure what you mean -- that import establishes the resource file for the application. When I later use `QIcon(":/OpenShot.svg")` to set the window icon, that import is what makes the `:/...` path resolve to the resource file.

The resource file does end up registering as an unused import, annoyingly, but that's just how PyQt set things up.

PR #3357, opened 2020-04-07
Blender length, import fixes

### LINKED ISSUE #3356: Animated Title: "Dissolving Text" is not working properly when Animation Length != default
**Describe the bug:**
When creating a Dissolving Text animated title and setting animation length to a value different from default it results in an incorrect animation.
The result seems that have the animation frames mixed, I attach the resulting video (all other parameter are left as default)
[export.zip](https://github.com/OpenShot/openshot-qt/files/4444094/export.zip)

**Side bug:**
Changing the animation length do not update the total frames
UmcYRLnM7Y

1. Go to Animated Title
2. Select "dissolving Text" effect
3. Select a value different from default for Animation Length (the attached video is with 2x)
4. Click on render

**Expected behavior:**
A proper animated title

**System Details:**
- OpenShot-v2.5.1-dev2-1586151841-166f6938-232f3f79-x86_64
- Blender: 2.82

ferdnyc:
This PR makes three major changes to the Blender code:

1. Computing animation length (fixes #3356)
    * The `animation_speed` parameter is renamed to `length_multiplier` throughout, as the former name seems almost intentionally confusing.
    * `length_multiplier` is used to compute the progress slider end value and labeling (frame #/#) everywhere, to properly account for adjusted lengths

1. Render selection is now done from the command line
    The blender `.py` files no longer trigger the actual rendering process. Instead, they merely set up the scene, and then stop. In addition, they don't limit the frame range for previewing. Blender wil be executed with a `-a` argument (or `-f frame#` in preview mode) to render all or one of the configured frames.
    This is primarily a debugging convenience, as it means that Blender can now be launched with one of the Python scripts applied. This allows us to examine the scene in its **final** configuration immediately before rendering.
    
    For example, this command loads `dissolve.blend` into the Blender GUI, **with** all of the `dissolve.py` parameters applied:
    ```sh
    blender dissolve.blend -y -P dissolve.py
    ```
    These commands render from a configured scene (since `start_frame` and `end_frame` are no longer set to the same frame, this even works from a preview script):
    ```sh
    # Render just one frame
    blender -b dissolve.blend -y -P dissolve.py -f 150
    # Render the entire animation
    blender -b dissolve.blend -y -P dissolve.py -a
    ```
1. Instead of having its own `add_file()` implementation, `animated_title.py` now passes the information off to the current files view for importing. This required constructing an `image_seq_details` dict and populating it, then passing it to the `add_file()` method of the view as a new parameter with a default value of `None`. If `add_file()` is called _without_ that parameter, it will still attempt to divine the image sequence details from the selected filename.

The last change, in addition to just being a good idea, was an attempt to get Blender animations to import at the project frame rate. It doesn't _work_, in part because I was mistaken about sequence importing in general, but even still I'm really not sure why there's a difference.

Even with the new code:
* A newly-rendered animation always loads at 25 FPS.
* If I load the same set of images as an image sequence _directly_, they import at 30fps, just like any other image sequence.

(I was under the mistaken impression that image sequences loaded at the project rate, whereas it seems they always load at 30fps. Except when the sequence is a Blender animation.)

As a late-in-the-process change while I was working on this, I also happened to fix the inclusion of project FPS data in the parameters passed to Blender. (It wasn't working, as it was trying to pass an unsupported dict object (`project.get("fps")`) in with `params`. I changed it to set Blender's `fps` and `fps_base` to `fps["num"]` and `fps["den"]`, respectively.

Perhaps setting the FPS will change how OpenShot reads in the output, since time signatures _are_ embedded in the PNG files' metadata. `¯\_(ツ)_/¯`


jonoomph:
@ferdnyc Quite a few conflicts on this one. I started fixing them, but maybe you should give this a try (since you are more familiar with your new changes). Thanks!

ferdnyc:
@jonoomph I'll take a look, thanks — though not until this evening, probably. I spent too long on #3462 last night / this morning, and now I have an offline commitment for the day.

ferdnyc:
Yeah, there's no clean way to reconcile the `blender_listview.py` changes with the current HEAD. I'll redo this and resubmit.

Issue #2420, opened 2018-12-02
Change property without setting keyframe


stefan-reich:
**Describe the bug**
I want to change a clip property such as ScaleX without generating an animation

**System Details (please complete the following information):**
 - Operating System / Distro: [e.g. Windows 10, Linux Mint 17.1]
Peppermint Linux 8
 - OpenShot Version [e.g. 2.4.1]
2.4.3

**To Reproduce**
Add a video clip to timeline.
Slice off the beginning.
Try to change property ScaleX.
Now it always makes a keyframe, and when I say "remove keyframe", the value of ScaleX is reverted.

**Expected behavior**
Properly remove keyframe, but keep new value

ferdnyc:
That's not possible, the value adjustment will always be applied as a keyframe. It's inherent in the animatable properties that their value adjustments are stored in keyframes.

The secret is in where you set it. If you make the adjustment at the _very first_ frame of your clip, the parameter won't be animated — it'll be set by the first keyframe to whatever value you select, and then because there are no other keyframes to further adjust the parameter, it will hold on the same value throughout the rest of the clip.

stefan-reich:
OK. The thing is, it didn't seem to work.

If the clip starts at 2:00 and I put the timeline pointer (?) at 0:00 for setting the value - is that supposed to work? 'cause it didn't...

It neither worked when the clip was at 0:00 and the pointer at 0:00 too.

It only worked when I added the clip again from the videos list.

So it seems slicing a clip was what caused the problem setting the value.

I can make a video demo if you like.

ferdnyc:
"Remove Keyframe" is somewhat broken in that if you use it on a clip that only has one setting for a value, that value is returned to 0 instead of to its default/initial value. 

Like, for instance, if you do this...
1. Place a Clip on the Timeline
2. Select the Clip
3. Right-click on "Scale X"
4. Choose "Remove Keyframe"

..."Scale X" will become `0` throughout the duration of the clip, which is wrong. Same thing with "Enable Audio/Video" (which also become `0`, instead of resetting to `-1`), etc. So, to avoid that issue, don't use "Remove Keyframe" to remove the last keyframe of a clip. 

[...]

Issue #1040, opened 2017-11-15
Rotate then transform


moneill4068:
Whenever I rotate an image within a video then try to transform it to get it to the right position the effect and the transformation icons disappear.  Do I have to transform it then rotate it?  If so, any advice on ensuring the rotated image was moved to the correct location?   Thank you.

DylanC:
@moneill4068 - Do you want to rotate while keeping the center in the same position?

moneill4068:
Ok, here is the situation.  I have a downloaded special effect that is a fireball going from left to right.  I need to have that same effect also go from right to left.  So I load the image and rotate 180degrees so it is now heading in the right direction.  I then attempt to hit transform to drag the fireball so it leaves from the correct position.  I click on transform and I get the icons with the boxes in the corner and the circle in the center.  I then try to drag it to the correct location and the fireball and the transform circle/boxes disappear.

moneill4068:
The original effect (going from left to right) can be transformed and moved easily.

moneill4068:
https://www.youtube.com/watch?v=cmGcOgdpLYk&app=desktop

DylanC:
@moneill4068 - Thanks.

peanutbutterandcrackers:
@moneill4068 - Hey! This is a very easy issue to resolve! You should use the 'Location X' and 'Location Y' parameters in the special effect's clip-properties. The transform tool is apparently a new addition and it isn't that mature yet. And also, it's just a GUI wrapped around the actual parameter-setting, which is what makes openshot such a beloved video editor.

So, here's what you should do: After you rotate that thing, go to it's properties and drag the handle of the 'Location X' parameter. It'll move the clip along x-axis. Leave the 'Location Parameter' to 0. [I kinda' nicked it (screenshot) by a 0.01. 😆 ]
![screenshot from 2018-02-12 16 22 14](https://user-images.githubusercontent.com/26764547/36093021-69effeec-1011-11e8-9fc2-dd23647b07c1.png)


I have actually been able to do some pretty neat stuff using the parameters themselves. And, frankly, I think, once a person has actually mastered the use of keyframes and parameters, transform tool is for sissies. You'll understand that too, someday. No offense though. :D 

I hope that helps.
P. S: I'm super glad that you actually made a video. That is quality issue-reporting. Please do continue to submit such quality issues. You're one of the best reporters here. :)

P. P. S: Here are some links to master keyframes and parameters:
[...]

Issue #728, opened 2017-06-14
feature request: warning during removal of used items in Project Files


openpianist:
**System Details:**
* Operating System / Distro: Ubuntu 17.04
* OpenShot Version: 2.3.4
* _Please attach log files if crash_

**Issue Description and steps to reproduce:**

If items are removed in the window Project Files, but are used in the project, it would be nice to give a warning. Currently it s a silent death, as the files are removed from the Project Files and the Timeline without any info.

- O.

ferdnyc:
Reopening this, as it's actually a really good idea.

ferdnyc:
Aack, I let Stale sneak in and close this again.

Assigning to myself, which will block Stale activity. Not that I'm committing to any timeframe for actually getting this done.

Colorjet3:
I think this is a good enhancement.  I have added this Trello:  https://trello.com/c/VOFkTbbt/125-a-warning-to-the-user-when-a-file-is-removed-from-the-project-files-window-and-it-has-been-dragged-onto-a-track

PR #3366, opened 2020-04-10
Emojis and Transform Improvements

jonoomph:
NOTE: This is a massive PR, with lots of new files. This might not work great in GitHub. :cry: 

**Emojis!**
Integrated ~3000~ 900+ emoijs from OpenMoji project. This started out a bit experimental, and I'm still very open to suggestions and feedback on it. This PR adds a new "Emoji" dock, and allows for users to drag and drop nice and clean vector emojis onto their project. It's grouped into many categories with a dropdown filter, and is searchable. Added new @emoji/ path prefix. Also supports custom emojis in /.openshot_qt/emojis/.

Emojis

**Models and Views**
Also, I've refactored our models, listview, and treeviews, so we have **1 primary model** owned by the MainWindow, and we use proxyModels for filtering. The listview and treeview share the same proxy model, and are always filtered identically. This also handles some custom sorting of transitions.

~Performance of the emojis is questionable, especially when filtering a category and searching text (at the same time). I have some ideas to improve this. However, with no category filter, the proxy model filters extremely fast, and is real-time. Once a category is selected though, it iterates and queries our model (in Python) too much, and is noticeably slow.~ Fixed the performance with adding the category name to the text, and create a custom regex for Qt to crunch on. Waaay faster!

**Transform**
A major refactor of the Transform tool in OpenShot, related to https://github.com/OpenShot/libopenshot/pull/496:
 - Ability to scale the video preview widget (zoom in/zoom out, restore zoom)
 - New custom cursor pixmaps (needed for transformation and rotation)
 - Added ability to rotate, shear_x, and shear_y
 - Added support for all gravity settings
 - Cursors now rotate and transform along with the canvas... a very subtle and cool effect. Many professional photo editing tools do this, and now we do!
 - Added border, and offset all handles (so they don't cover the edges)
 - Improved the update logic, so we only modify the current values, instead of clobbering them. Much smoother and more fluid transforms.
 - Update logic now takes into account the size of the item, so that 1 x height = 90 degrees rotation, and similar calculations, which just feel better and natural when dragging handles.
 - Fixed many, many bugs related to transform tool
Screenshot from 2020-04-10 14-52-41
Screenshot doesn't actually show the new cursors and cursor rotation, but it's cool! Trust me!

**Translations**
Emoji and Transition translations have been split off into their own POT template file. This is still a work-in-progress, but the idea is to de-prioritize these 2 templates as 'nice to have', but not really part of the 'core' openshot-qt translations. This reduces the # of 'core' translations for a good openshot-qt experience, but still allows for 100% translation for those who really want it.


SuslikV:
I'm disappointed. You want to store 3000+ of small files on user PC?

ferdnyc:
@jonoomph 

PR #3377 is a PR against _this branch_ (a PR against a PR, one of my favorite "Git is confusing and weird" tricks), with a bunch of tweaks I came up with. 

Part of the reason being that it's impossible to review this PR, because the GitHub web interface completely breaks down trying to show a diff for 6000 PNG and SVG files. IMNSHO, the commit(s) to add those in should really be broken out into a separate branch/PR, for sanity's sake. 

But the suggestions I had were also too involved to leave as review comments, it was easier to implement it and file the PR-PR once I had it working. However, I did place a review on _MY_ PR with a running commentary of my reasoning behind the suggested changes. Please do look #3377 over, if you have a chance.

ferdnyc:
> Part of the reason being that it's impossible to review this PR, because the GitHub web interface completely breaks down trying to show a diff for 6000 PNG and SVG files. IMNSHO, the commit(s) to add those in should really be broken out into a separate branch/PR, for sanity's sake.

This PR also contains at least two **major** change sets that are completely unrelated (the emoji feature and the improvements to Transform), plus I think I saw some fixes to the existing models. The latter of which, especially, are desperately needed and will be great to have, but really should be their own PR.

Some of the non-emoji changes appear to overlap with _other_ outstanding PRs, for one thing. So then everything depends on what order things merge in, and that's how we end up accidentally losing code. (Something that's happened several times already, when these massive-change branches are merged.)

ferdnyc:
> I'm disappointed. You want to store 3000+ of small files on user PC?
> 
> **Edit:** maybe it doesn't matter. Let it exist.

Honestly, that gives me pause as well, and now it's up to 6000 files with the cache.

(Which I have to say feels dubious to me.) QtSVG is _fast_ for rendering SVG icons, that's its entire reason for existence. I have a hard time believing cached PNG conversions really make that much of a performance difference. Maybe if the SVG thumbnails were being manipulated, or converted to PNG on the fly... but not if they were just being rendered as straight SVG loaded using `QFile()` or `QIcon()`.

...I mean, I wouldn't have thought, anyway. I'm willing to accept that there are unexpected issues that caused performance problems with the SVGs loaded directly. I just wonder if the slowness wasn't due to **other** things, like the earlier filtering code, rather than the image format.

ferdnyc:
But even putting aside the cache issue, individual files for every emoji is kind of insane. For one thing, storing that many tiny files is terribly inefficient. The total data size of `emojis/color/svg` is 11MB, but it takes up **eighteen** MB of disk space on Linux ext4 because the files are well below the filesystem block size.

There's got to be a way we can load them in as an indexed binary file, a zip file... _something_.

I looked at building a resource file, but unfortunately because the only option with PyQt5 is a text-format resource file, that actually takes up **MORE** space — 34MB, to be exact.

An uncompressed Qt resource file is even bigger at 47MB, but Qt also offers a binary resource file format that PyQt5 doesn't have. Packed into one of those with `rcc-qt5 --binary`, the file is only around 8MB, but of course it doesn't do us any good from Python code so it's really a moot point.

Python itself has gained the ability to load classes and data from zip files, as of recent releases. (Starting with 3.7.) There's a backport of that feature available as [importlib_resources](https://importlib-resources.readthedocs.io/en/latest/) (because the standard library module is [`importlib.resources`](https://docs.python.org/3.7/library/importlib.html#module-importlib.resources), which brings it back a couple of releases farther to Python 3.5. Problem is, we're still running 3.4 in the AppImage builds. :cry: 

Then there's the fact that, even with the PNG versions of the icons (if that really does make a difference), I just timed it and there is a **four-second** pause on my machine at "updating emoji model." That's a crazy amount of delay for one feature that someone may or may not even use.

I know I'm the chief crusader for persistent models, and for application-wide things I still firmly believe it makes sense. But given how heavy this particular model ends up being to construct, it might make sense to defer constructing it until the dock is **opened** for the first time after startup. That way, if someone just closes the emoji panel entirely, they don't have to pay that startup penalty — or the additional memory consumption — when building the model. (They'll instead have to pay it if/when they open the dock, of course.)

ferdnyc:
I just tested restarting OpenShot with the dock closed. That 4-second delay drops to about 1 second, which is significantly better (though not nothing). Clearly, a bunch of the time is taken up with rendering the view, not building the model. (So, I'm now more willing to entertain the idea that using SVG causes additional slowdowns.)

But, still, there's no reason even the _model_ needs to exist, if the view isn't created. Especially since it does carry something of a cost. On my Linux 5.5.13 x86_64 box with four cores and 6GB of RAM, measuring a few seconds after application startup with no project file loaded:

| Program | Resident RAM usage |
|-------------|-----------------------------|
| OpenShot 2.5.1 release | ~272 MB | 
| `emojis` branch w/dock open | ~355 MB |
| `emojis` branch w/dock closed | ~355 MB |

jonoomph:
All good points. I'm 100% in support of deferring things to when needed, if that's an easy change. Also, I'm fully supportive of reducing the # of emojis we ship with. Probably a few hundred of the best ones would still be really cool, and would reduce RAM, CPU, Files, and Translations. I can also split this out into a few different PRs if needed. 

As usual, this started as an Emoji branch, but once I realized the Transform tool was almost unusable when working with the Emojis, I decided to include that change as well, etc... :wink:

jonoomph:
But perhaps the "key" change needed is an improved "as needed" loading of views and models. By not loading on the launch of OpenShot, it would likely make for a much snappier launch.

jonoomph:
Emoji count is now down to 936, with optimized JSON metadata file, and updated the icon cache. Launching OpenShot is much faster now. Overtime, I want to introduce more usable assets baked right into the UI. Icons/Emojis are just the first part of many updates in progress. I also want 'borders', 'backgrounds', 'sound effects', 'music', etc... However, I imagine those will be in the "dozens" not "thousands" like emojis. :+1:

SuslikV:
@jonoomph 

> I also want...


You better focus on resolving existing bugs. The number exceeds 744. This is bad.

jonoomph:
> @jonoomph
> I'm disappointed. You want to store 3000+ of small files on user PC?
> You better focus on resolving existing bugs. The number exceeds 744. This is bad.

@SuslikV Your comments on this PR are neither helpful, nor constructive, nor accurate, and violate our [Code of Conduct](https://github.com/OpenShot/openshot-qt/wiki/Code-of-Conduct). We have over 300 enhancements marked as bugs, and many additional which are old, out-dated, duplicates, etc... It is unrealistic for 1 person (i.e. me) to manage the entire backlog of OpenShot issues. But together, as a community, and **following the code of conduct**, we can make big strides fixing the issues that affect the most people.

SuslikV:
@jonoomph Maybe I need a break. Current rate is 2 issues a day. And I got second warning from you.

jonoomph:
@ferdnyc I found the regression! https://github.com/OpenShot/openshot-qt/pull/3366/commits/b314151d24f2cc4d57f295c37a3ce9aa152e4ab1

SuslikV:
@jonoomph is this thread ready to merge? I tried some features and _Transform_ gives to me:
```
...
properties_model:INFO updating clip properties model.
  File "D:\Downloads\Openshot\{app}\windows\video_widget.py", line 199, in paintEvent
    origin_x = raw_properties.get('origin_x').get('value')
AttributeError: 'NoneType' object has no attribute 'get'
```

jonoomph:
@SuslikV Yes, this is about ready to merge. However, I don't see the error you are getting. What steps are you doing to get that message?

SuslikV:
Sure, I forgot this: https://github.com/OpenShot/libopenshot/pull/496 wasn't merged yet... Maybe mark "WIP" here too?

jonoomph:
I'm trying to decide if the panel should be called "Emojis", "Art", "Assets", "Extras", "Fun", etc.... If we keep the name of this feature more generic, we can add additional categories/filters, for things like:
 * backgrounds
 * borders
 * music
 * sound effects
 * etc...

I'm open to suggestions! :pray:

SuslikV:
@jonoomph just one question: how this feature differs from import image downloaded from the same internet source? In other words - what is new in this feature, so you making whole new model, dock etc. for it? All images is available online with the filters, SVG can be downloaded: https://iconify.design/icon-sets/openmoji/

I'm not fan of this change (moji itself). but name should point to meaning of the feature. I see it as additional set of images. It is more like templates for the mood expression. Like titles, it can be in the menu, not in the dock. In the Skype this feature lies next to the message typing (or short version shown directly over the camera screen). There it named "Smiles and Moji" if I understand it right.

ferdnyc:
@jonoomph See #3497, and keep in mind that it was _technically_ your idea in the first place.

(At least, that's how I'm characterizing https://github.com/OpenShot/openshot-qt/pull/3481#issuecomment-629018052, though my position admittedly relies on an _**extremely**_ broad interpretation of that exchange. :angel: )

jonoomph:
@SuslikV 
> how this feature differs from import image downloaded from the same internet source

1) This adds many high quality image assets into OpenShot by default, for students, young kids, beginner animators, fun overlays, call-outs, etc... There are tons of useful things to do with a high quality list of assets built-in to OpenShot
2) The list is searchable by more than file name (by description)
3) The list is filterable by category (meta data)
4) The list is translated, including categories and descriptions
 - Items 2 through 4 are all very different than just importing a bunch of random named emoji files into your project.

src/windows/models/transition_model.py
```diff
@@ -58,6 +60,37 @@ def mimeData(self, indexes):
...
+            # Return, if regExp match in displayed format.
+            return "common" == group_name and self.filterRegExp().indexIn(trans_name) >= 0
+
+        # Continue running built-in parent filter logic
+        return super(TransitionFilterProxyModel, self).filterAcceptsRow(sourceRow, sourceParent)
+
+    def lessThan(self, left, right):
+        """Sort with both group name and transition name"""
+        leftData = self.sourceModel().data(left)
+        leftGroup = self.sourceModel().data(left, 257) # Strange way to access 'type' column in model
```
ferdnyc, line 87: @jonoomph 
I have to say, this terrified the living crap out of me, when I came across it while working on these. I absolutely _do not_ understand how it works, even **sort of**. "Strange" doesn't begin to cover it. Where did that even come from?

If I'm understanding it, though, the equivalent in non-magical Qt API should be something like:
```python
leftGroup = self.sourceModel().sibling(left.row(), 2, left).data()
# or even possibly
leftGroup = left.sibling(left.row(), 2).data()
```

ferdnyc: I'm almost tempted to think that it works because it's overflowing an old-school, single-byte `char` field.
(`255 + 2 = 257` after all.) Thus reaching the second column that way.

But that's still crazy, possibly even crazier than whatever's _actually_ happening here.

ferdnyc:
@jonoomph 

> I'm trying to decide if the panel should be called "Emojis", "Art", "Stickers", "Assets", "Extras", "Fun", "Or just an icon... no visible name, like a Magic Wand, or an Oil Painting", etc.... If we keep the name of this feature more generic, we can add additional categories/filters, for things like:
> 
> * backgrounds
> * borders
> * overlays
> * music
> * sound effects
> * etc...
> 
> I'm open to suggestions!

I had meant to respond to this, and I never did.

I think, _for now_, "Emojis" is fine for the panel name. If it was going to expand to cover additional data sources, at that point I think it would be worth revisiting the question, and perhaps creating a set of **subpanels**, one per "category" — _one of which_ would be the current "Emojis" (and still named as such, so it wouldn't really cause any confusion).

In fact, one option for this going forward, especially if additional sources are added, would be to roll it all back into **Project Files** — which, instead of being "a list of files imported into the project", could become a metadirectory of _media source bins_, one or more of which represents user-imported files, with several others being application-defined (and perhaps configurable as to whether they show up at all). The whole "media library" thing is a pretty common pattern for NLE suites.

In fact, tonight I stumbled across another possibility for a media library source: [QtAwesome](https://github.com/spyder-ide/qtawesome). (Which, despite the name, is _actually_ PyQtAwesome. Except that it's PyQt5/PySide2 agnostic.) It's basically a PyQt wrapper around [FontAwesome](https://fontawesome.com/) and other types of clipart-ey libraries.

(In fact, it can even load external font libraries, so there's some possibility we could convert the OpenMoji SVG collection into a compatible type of font and leverage the QtAwesome library to do efficient loading.)

In the world I'm imagining, the user would be able to specify one or more disk locations as "library" folders that are listed as separate subtabs/sublists of Project Files, and are always scanned for all valid media files (silently ignoring anything that's not supported), with the contents made available for inclusion in any project they work on. Files loaded from those paths, just like files from the Emojis library or any other source, would still be imported into the project data, but only _at that point_, so the `"files": []` list in the `.osp` file would still contain only the media files actually used in the project.

(This also relates, indirectly, to an old request I just saved from being closed, because I hadn't seen it before and it's actually a _really_ good idea. As well as another very recent one that, in a different way (a better way, in terms of `.osp` data pollution) this would address.):

* #728 "feature request: warning during removal of used items in Project Files"
* #3492 "I can't drag folders to the playlist"

ferdnyc:
> @jonoomph 

> Files loaded from those paths, just like files from the Emojis library or any other source, would still be imported into the project data, but only _at that point_, so the `"files": []` list in the `.osp` file would still contain only the media files actually used in the project.
> 
> (This also relates, indirectly, to an old request I just saved from being closed, because I hadn't seen it before and it's actually a _really_ good idea. 
> 
> * #728 "feature request: warning during removal of used items in Project Files"

That could also play into an improvement in our recovery logic. Once we have something counting the actual uses of each "project file" item, when doing missing-file recovery any files we encounter that _aren't_ actually used in the project, we might as well just silently remove instead of bothering the user about them.

SuslikV:
@ferdnyc if you attempt to implement everything that users asks for - it will be unusable "nightmare" rather than nice video editor, believe me.

src/windows/models/transition_model.py
jonoomph: @ferdnyc Ahhh, yes, that makes sense. I found this using a debugger (almost positive about that, because how else would I find the index... lol)

ferdnyc:
> @ferdnyc if you attempt to implement everything that users asks for - it will be unusable "nightmare" rather than nice video editor, believe me.

That meaningless tautology aside, the fact is that SOME user requests are objectively good ideas. And showing an "Are you sure?" dialog before deleting a project file that's **actively in use** on the Timeline... that's just about the best example I can _imagine_ of one of those.

jonoomph:
@ferdnyc Okay, I think I'm ready to merge this beast, and then we can get things back to normal on `develop`. :stuck_out_tongue_winking_eye:.

src/windows/models/transition_model.py
ferdnyc: @jonoomph Maybe we fix just this one, before the merge? :rofl: I'll look at fixing it now, in fact.

jonoomph:
:boom:

src/windows/models/transition_model.py
ferdnyc: Fixed. For a little extra flexibility, I made it respect `sortRole()` for the main column (only), instead of forcing the default `Qt.DisplayRole`. So the sort role can be changed, if needed at some future point, and it'll still work.

ferdnyc: Oops, looks like I was too slow by less than 5 minutes. :laughing: I'll do a new PR, nbd.

jonoomph: Ooops, missed this comment.

ferdnyc:
@jonoomph 

I have one question: How do you _end_ the new transform mode? I'm finding myself rather stuck in it, since it's surprisingly resilient — even moving the playhead completely off the clip being transformed doesn't end it. (It shows the transform rectangle for that clip, overlaid on whatever _other_ clip is currently visible.)

Seems like there needs to be a way to say, "Okay, I'm done transforming now." Am I missing something?

jonoomph:
Okay, that makes sense. Unselecting the clip removes the transform from the screen. However, we could add an ESC and/or ENTER keypress that removes the transform (or unselects the clip, which would remove the transform).

SuslikV:
> ...However, we could add an ESC and/or ENTER keypress that removes the transform (or unselects the clip, which would remove the transform).

Take in mind that viewing Transform rectangle (boundaries of the Clip) is useful for positioning the Clips (even if Timeline's cursor placed out of the Clip).  See: https://github.com/OpenShot/openshot-qt/issues/2420#issuecomment-632317245

jonoomph:
Sure, the user can leave open the transform boundaries as long as needed. But for a user who is "done" transforming a clip, I think it might be an intuitive "guess" to hit ENTER when done, and then get the satisfaction/confirmation of the rectangle disappearing. But that's all the ENTER key would do, remove the selection. It's not required.

SuslikV:
> In particular, since moving the playhead while transform is active will create **more** keyframes at the destination frame...

I don't get it. What I see in my builds that the Transform Rotation cursor only shown after I **selected nothing** in the Timeline (after clip were animated). But no new keyframes are created and no Transform bounding box is shown in the Preview.

The "rolling stone" animation is much easier to do with this new Transform. I can see that the _Playback_ is may be active during Transform, but really - you still need to drag items in the Preview... This is expected. So what is wrong here?

ferdnyc:
> > In particular, since moving the playhead while transform is active will create **more** keyframes at the destination frame...
> 
> I don't get it. What I see in my builds that the Transform Rotation cursor only shown after I **selected nothing** in the Timeline (after clip were animated). But no new keyframes are created and no Transform bounding box is shown in the Preview.

Sorry, I want clear. _Once you've activated_ Transform, it **stays** active even if you move the playhead, which means it **can** still create new keyframes via mouse movements in the preview. Additional manual actions are still required, it's true.

But it does mean that you're always only one click away from accidentally modifying your properties. (As well as just having that blue rectangle in the preview, which is annoying if you don't need it.) Having to _deselect_ the clip to make it go away seems like an unintuitive interface; we should at least also accept standard keypresses like <kbd>Esc</kbd>.

> The "rolling stone" animation is much easier to do with this new Transform. 

I don't know what that is, but I don't _believe_ what I'm suggesting would change anything about how transform *can** be used. I only want users to have additional options for exiting the mode **if they choose**.

> I can see that the _Playback_ is may be active during Transform, but really - you still need to drag items in the Preview... This is expected. So what is wrong here?

Wrong? Nothing, per se. I'm just suggesting that things can be enhanced, without detracting from current behavior. Or maybe even improving it.

If transform were a toolbar mode, then when it's active, **every** click on a clip would activate transform immediately. Maybe that's what the new preference checkbox does, so really I'm just suggesting that the checkbox be moved out of Preferences and onto the toolbar -- and that, when it's deactivated, transform is exited if it was active.

If someone chooses to, and doesn't mind the blue rectangle, they can leave it active all the time and transform will be as accessible as possible. Whereas for other users who only occasionally need to activate it, it would provide a clear and simple way to exit the mode when they're finished. Best of both worlds, I hope. Unless I'm missing a use case?

(The one thing not covered in my current thinking is what should happen with multiple selection. I'm not even really sure what currently happens.)

SuslikV:
Usually, I clicking in the window where I see the object. In OpenShot I need to click on the Timeline instead (even if the clip only single object) to move something in the view. Of course, it is intuitive interface I simply don't understand its complexity...

SuslikV:
> If transform were a toolbar mode, then when it's active, **every** click on a clip would activate transform immediately...

It is interesting thought. Let's call it "Editing Mode" feature or "Animate" button. It would be strange if it only take control over the Transform and not over other edits...

rexdk:
@jonoomph I'm just following up #1040, where the transform changes here were hoped to have sorted out problems with a rotated clip.
Unfortuantely with the 28Jul20 daily build on Win10 the transform tool still does not work correctly on rotated clips. It is now almost useable, which is an improvement ;). I'll gladly give full details if you have no access to Win10.

With the build as a whole, my user experience (I'm quite new to this) was that the Christmas lights had arrived; I don't understand why the emojis have such prominence on the starting screen. Can you imagine what a new user will think - where their eye will focus? Even if someone actually wants to use this particular image library subset, it's too difficult to find the right image. Try a search for 'weather'.....

Sorry! It's one of those cases where 'more' != 'better'

rexdk:
There is another aspect about the OpenMoji library - it is Creatuve Commons licensed. GPL3 is listed as one way CC compatible, but according to the user guide Openshot is "GPL3 or later", which from my reading is not compatible.

Also, any derivative works must be CC licensed, according to my reading. That means the user must provide attribution with his video.
https://creativecommons.org/licenses/by-sa/4.0/

I could be wrong and I hope I am, but this does need serious thinking about. OpenMoji is not a public domain resource.

Sorry, again.

src/windows/models/transition_model.py
ferdnyc: SON OF A **BITCH**!!!!

@jonoomph 

Five _months_ later, I finally discovered the source of those phantom `257` data items!

It seems that, unlike most of Qt's pre-packaged Model/View classes, `QStandardItem.data()` and `QStandardItem.setData()` both [**DEFAULT TO `Qt.UserRole + 1`**](https://doc.qt.io/qt-5/qstandarditem.html#setData) when not given an explicit role:

### `QVariant QStandardItem::data(int role = Qt::UserRole + 1) const`
> Returns the item's data for the given role, or an invalid `QVariant` if there is no data for the role.
> 
> Note: The default implementation treats `Qt::EditRole` and `Qt::DisplayRole` as referring to the same data.
> 
> See also `setData()`.

<hr>

### `void QStandardItem::setData(const QVariant &value, int role = Qt::UserRole + 1)`
> Sets the item's data for the given role to the specified value.
> 
> If you subclass `QStandardItem` and reimplement this function, your reimplementation should call `emitDataChanged()` if you do not call the base implementation of `setData()`. This will ensure that e.g. views using the model are notified of the changes.
> 
> Note: The default implementation treats `Qt::EditRole` and `Qt::DisplayRole` as referring to the same data.
> 
> See also `Qt::ItemDataRole`, `data()`, and `setFlags()`.

I guess those handwaves about "The default implementation treats..." are their explanation for it, but it's still a pain in the ass. Classes like `QAbstractItemModel`— and, hell, evne `QStandardItemModel`! — return the _actual_ `Qt.DisplayRole` and `Qt.EditRole` for `data()` and `setData()` respectively.

So, the long and short of it is, **WE'RE** the ones shoving all of those role-257 data items into the data models, anytime we call `setData()` or `data()` on an instance of `QStandardItem`!

(And now that I know this, I can adjust my user-defined roles to start from _higher_ than `Qt.UserRole + 1`. Which would've been really nice to know like 5 or 6 hours ago, insead of spending them pointlessly fighting with disappearing model data. 😡)

ferdnyc: This is entirely @N3WWN 's fault. Stupid properties table keyframe-status code!

Issue #3325, opened 2020-03-28
Mit dem hellen Theme, Spuren sind trotzdem schwarz...


accorinti:
![screenShot](https://user-images.githubusercontent.com/10834155/77822906-4845fa00-70f7-11ea-852a-6f5fa01c6c95.jpg)
**Beschreiben Sie den Fehler**
Mit dem hellen Theme, Spuren sind trotzdem schwarz und die Knöpfe, um die Bereiche kleiner/größer zu machen unsichtbar

**Schritte zum Reproduzieren des Verhaltens**
1. Öffnen

**Erwartetes Verhalten**
S. oben.

**Systemdetails**
- OpenShot Version: 2.5.1
- libopenshot Version:  0.2.5
- Platform: Windows-10-10.0.18362
- Processor: Intel64 Family 6 Model 142 Stepping 12, GenuineIntel
- Machine: AMD64
- Python version: 3.7.4
- Qt5 version: 5.12.4
- PyQt5 version: 5.12.3
- Qt Detected Languages: ['de-DE', 'en-US']
- LANG Environment Variable: 
- LOCALE Environment Variable: 
- Daily Build Installed: http://github.com/OpenShot/openshot-qt/releases/download/OpenShot-v2.5.1-dev2-1585351843-166f6938-232f3f79-x86_64.exe

**Logdateien**
- http://logs.openshot.org/b8ryqu1igucj60aukgmhjrdvamdn49rm/openshot-qt.zip
- http://logs.openshot.org/b8ryqu1igucj60aukgmhjrdvamdn49rm/libopenshot.zip

**Ausnahme / Stapelverfolgung**
```No stacktrace found in log files```

**Bildschirmfotos** (Optional)
![screenShot](https://user-images.githubusercontent.com/10834155/77822914-614eab00-70f7-11ea-8821-d216e50f18b8.jpg)

SuslikV:
This is by design. No plans to change this (unfortunately).

accorinti:
It is by design that you cannot see the buttons for making smaller/bigger the sections???

SuslikV:
Absolutely.

accorinti:
I hope someone with common sense is reading this and has at least an idea for a workaround.
Anyway, thanks for this super software and happy easter.

SuslikV:
[...]

PR #3371, opened 2020-04-12
Add Fusion style to themes

SuslikV:
Inspired by: https://github.com/OpenShot/openshot-qt/issues/3325

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -178,7 +178,7 @@ class OpenShotApp(QApplication):
         self.context_menu_object = None
 
         # Set Font for any theme
-        if self.settings.get("theme") != "No Theme":
+        if self.settings.get("theme") != "No Theme" and self.settings.get("theme") != "Fusion":
             # Load embedded font
             try:
                 log.info("Setting font to %s" % os.path.join(info.IMAGES_PATH, "fonts", "Ubuntu-R.ttf"))
@@ -225,6 +225,8 @@ class OpenShotApp(QApplication):
 
             self.setPalette(darkPalette)
             self.setStyleSheet("QToolTip { color: #ffffff; background-color: #2a82da; border: 0px solid white; }")
+        elif self.settings.get("theme") == "Fusion":
+            self.setStyle(QStyleFactory.create("Fusion"))
 
         # Create main window
         from windows.main_window import MainWindow
--- a/src/settings/_default.settings
+++ b/src/settings/_default.settings
@@ -17,6 +17,10 @@
     "translate_values": true,
     "values": [
       {
+        "value": "Fusion",
+        "name": "Fusion"
+      },
+      {
         "value": "Humanity",
         "name": "Humanity"
       },


ferdnyc:
This makes total sense to me, in as much as it unifies our themes to display in the same style. But rather than adding a third theme option, I'm sort of more inclined to just fix "Humanity" (non-dark) so it also forces Fusion style the same way the dark version does. In fact, it feels like a bug/oversight that it **doesn't**. (Possibly a consequence of the light theme's overall neglect.)

On Linux with this change, "Humanity" and "Fusion" would be literally identical — er, fonts possibly excepted — because "Fusion" is the default style (which is why I couldn't see the #3325 issue originally). And on Windows, if the only difference between "Humanity" and "Fusion" right now is that one has broken/invisible splitters and the other doesn't, then why keep a broken "Humanity" around at all? It makes more sense to me that our themes always force the same style. Ditto macOS, at least to whatever extent "Humanity : Dark" is already successful in forcing its own app style on a Mac (a notoriously hard thing to do).

There's also the existing "No theme" option, which on Linux will again be the same as "Humanity" and/or "Fusion" except missing the font-size change. (So, even _more_ identical to "Fusion".) And right now on Windows, "No theme" is effectively "Humanity" without the font adjustment. So, if we change "Humanity" (non-dark) to force the Fusion style, then "No theme" will actually have a purpose on Windows, and "Humanity" will be consistent in light and dark versions.

SuslikV:
Difference in fonts size is far more noticeable then you think:

Humatity (100% DPI screen, 1280x1024)
Openshot Humanity

Fusion (100% DPI screen, 1280x1024)
Openshot Fusion

And fore sake of comparison Humanity with Fusion set same as for the Dark theme (main menu has smaller letters):

<details>

Humanity + Fusion (100% DPI screen, 1280x1024)
Openshot Humanity+Fusion

</details>

MBB232:
> Difference in fonts size is far more noticeable then you think:

Clear difference in your screenshots. Is this a _purpose_ of the themes, to offer a larger font? Should one in particular be recommended for people with bad eye sight or large resolution small screens?

SuslikV:
> Is this a _purpose_ of the themes, to offer a larger font? Should one in particular be recommended for people with bad eye sight or large resolution small screens?..

When I first time run the application it was my first thought too. But @ferdnyc said that it is just bug from the porting it to Windows. And fonts actually should be all same size (as No Theme), More over, there will be no visual difference between new Fusion and Humanity on linux :(

MBB232:
> > Is this a _purpose_ of the themes, to offer a larger font? Should one in particular be recommended for people with bad eye sight or large resolution small screens?..
> 
> When I first time run the application it was my first thought too. But @ferdnyc said that it is just bug from the porting it to Windows. And fonts actually should be all same size (as No Theme), More over, there will be no visual difference between new Fusion and Humanity on linux :(

ok, no real need to document it then. 
Nor to make a feature request out of if it won't work the same on all platforms.

SuslikV:
It is unknown if all Windows users already seeing this as a "feature". So, the linux variant of Humanity theme may be considered as bugged. Then PR is valid but linux has the bug...

jonoomph:
Closing PR. While this obviously needs some fixing, it doesn't seem like this PR solves the root issue. Also, when we revist the themes, it would be great to extend theme support into the timeline (conditional CSS include perhaps)

PR #3374, opened 2020-04-13
Linux/setup.py: install 128x128 icon

eszlari:


ferdnyc:
:+1: Yup.

Threw me for a second because I checked my local copy and this was already there, but it appears that's because I already fixed this locally, but hadn't gotten around to a PR yet. Thanks!

PR #3376, opened 2020-04-13
Titles: Re-establish user template location

### LINKED ISSUE #3375: User title templates not preserved in 2.5.0/2.5.1
Question: Why is it,if I save a custom footer/title template it is still in
the ~/honm/,openshot-qt/title folder even after program is closed,,,when I
open Openshot,all the saved templates are cleared out? I know I done this
in the past but i am thinking there may be like a 'custom' folder you have
to make within the openshot-qt folder.

_Originally posted by @brcisna in https://github.com/OpenShot/libopenshot/issues/474#issuecomment-611813193_

ferdnyc:
This PR adds an `info.USER_TITLES_PATH` to the standard set, pointing to `$HOME/.openshot_qt/title_templates`. This is the location for user-created title files, and its SVG contents will be included when building the titles model.

(Prior to 2.5.0 `info.TITLE_PATH` was that location, but when things got re-worked that became the generated title _output_ path.)

Fixes #3375

### DIFF
--- a/src/classes/info.py
+++ b/src/classes/info.py
@@ -56,22 +56,23 @@ RECOVERY_PATH = os.path.join(USER_PATH, "recovery")
 THUMBNAIL_PATH = os.path.join(USER_PATH, "thumbnail")
 CACHE_PATH = os.path.join(USER_PATH, "cache")
 BLENDER_PATH = os.path.join(USER_PATH, "blender")
-ASSETS_PATH = os.path.join(USER_PATH, "assets")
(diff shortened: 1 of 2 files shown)


ferdnyc:
(I also removed the `info.ASSETS_PATH`, which used to be the output location for generated titles but is no longer used.)

ferdnyc:
I'mma merge this to close the linked issue, it's a relatively minor change anyway and it restores (to the extent possible) old behavior that was inadvertently lost among some recent changes.

PR #3377, opened 2020-04-14
Emoji model/view tweaks

ferdnyc:
This PR is against the `emojis` branch of the project repo, with some suggestions I had. There's a bit too much to leave them all as suggested changes in a review on the emoji PR, but I'll explain my reasoning in change comments on _this_ PR.

cc: @jonoomph 
Affects #3366


src/windows/models/emoji_model.py
```diff
@@ -73,9 +74,10 @@ def update_model(self, clear=True):
         if clear:
             self.model_paths = {}
             self.model.clear()
+            self.emoji_groups.clear()
 
         # Add Headers
-        self.model.setHorizontalHeaderLabels([_("Thumb"), _("Name")])
+        self.model.setHorizontalHeaderLabels([_("Name")])
```
ferdnyc, line 80: There's no reason to maintain separate "Thumb" and "Name" columns. The Icon and the text for a column are set as different roles, meaning both items can occupy the same column. Actually, for reasons I'll get into it makes a lot of sense to put as _much_ as possible into a single column.

src/windows/models/emoji_model.py
```diff
@@ -153,68 +158,45 @@ def update_model(self, clear=True):
...
-                # Append Filename
+                # Set emoji data
                 col = QStandardItem("Name")
-                col.setData(emoji_name, Qt.DisplayRole)
+                col.setIcon(QIcon(thumb_path))
                 col.setText(emoji_name)
+                col.setToolTip(emoji_name)
+                col.setData(path)
                 col.setFlags(Qt.ItemIsSelectable | Qt.ItemIsEnabled | Qt.ItemIsUserCheckable | Qt.ItemIsDragEnabled)
                 row.append(col)
```
ferdnyc, line 168: Same thing here. This loads up the first column of the model with everything needed to define an item — the Name, the icon, and the filesystem path (in the UserRole as set by `setData()`).

By doing that, when we get QModelIndex entries for a selected or dragged item, we can query all that info directly from the individual item — there's no need to translate back to rows in the source model so we can match up siblings. All of the data is right where we need it.

src/windows/models/emoji_model.py
ferdnyc, line 80: There's no reason to maintain separate "Thumb" and "Name" columns. The Icon and the text for a column are set as different roles, meaning both items can occupy the same column. Actually, for reasons I'll get into it makes a lot of sense to put as _much_ as possible into a single column.

src/windows/models/emoji_model.py
ferdnyc, line 168: Same thing here. This loads up the first column of the model with everything needed to define an item — the Name, the icon, and the filesystem path (in the UserRole as set by `setData()`).

By doing that, when we get QModelIndex entries for a selected or dragged item, we can query all that info directly from the individual item — there's no need to translate back to rows in the source model so we can match up siblings. All of the data is right where we need it.

src/windows/models/emoji_model.py
```diff
@@ -153,68 +158,45 @@ def update_model(self, clear=True):
...
-                row.append(col)
-
-                # Append Path
-                col = QStandardItem("Path")
-                col.setData(path, Qt.DisplayRole)
-                col.setText(path)
-                col.setFlags(Qt.ItemIsSelectable | Qt.ItemIsEnabled | Qt.ItemIsUserCheckable | Qt.ItemIsDragEnabled)
+                # Append filterable group
+                col = QStandardItem(emoji_type)
                 row.append(col)
```
ferdnyc, line 172: The only other column I create is one that contains the `emoji_type` (group), because I used that for filtering with the dropdown selector. I'll go into that in more detail in a bit, too.

src/windows/models/emoji_model.py
```diff
@@ -153,68 +158,45 @@ def update_model(self, clear=True):
...
-        get_app().window.emojiFilterGroup.clear()
-        get_app().window.emojiFilterGroup.addItem(_("Show All"), "Show All")
-        dropdown_index = 1
-        for emoji_type in sorted(emoji_groups.keys()):
-            get_app().window.emojiFilterGroup.addItem(_(emoji_type.capitalize()), emoji_type)
-            if emoji_type == default_type:
-                # Initialize emoji filter group to settings
-                get_app().window.emojiFilterGroup.setCurrentIndex(dropdown_index)
-            dropdown_index += 1
-
```
ferdnyc, line 205: The dropdown list really shouldn't be populated from the model — that's a view function. I moved this to the view, and just moved to storing a list of strings in a class member `self.emoji_groups` that it references.

(By doing it in the view, and waiting until after the signals are connected, another advantage is that we don't have to worry about manually refreshing — the `setCurrentIndex()` on the dropdown box automatically triggers `refresh_view()`.)

src/windows/models/emoji_model.py
```diff
@@ -153,68 +158,45 @@ def update_model(self, clear=True):
...
+        self.group_model.setFilterKeyColumn(1)
 
-        # Create proxy model (for sorting and filtering)
         self.proxy_model = QSortFilterProxyModel()
         self.proxy_model.setDynamicSortFilter(False)
         self.proxy_model.setFilterCaseSensitivity(Qt.CaseInsensitive)
         self.proxy_model.setSortCaseSensitivity(Qt.CaseSensitive)
-        self.proxy_model.setSourceModel(self.model)
+        self.proxy_model.setSourceModel(self.group_model)
         self.proxy_model.setSortLocaleAware(True)
```
ferdnyc, line 202: This was a major one. It felt to me like, since the selection list is based on categories, it made more sense to have _two_ proxy filters, one feeding into the other.

(That's [fully supported by the Qt framework](https://doc.qt.io/qt-5/model-view-programming.html#using-proxy-models):)
> Since proxy models inherit from QAbstractItemModel, they can be connected to any kind of view, and can be shared between views. They can also be used to process the information obtained from other proxy models in a pipeline arrangement.

By doing that, a selection from the dropdown list instructs `group_model` to select rows from a certain group, and then that list is passed to the `proxy_model` which can further filter it using the text field. I'll come back to that in comments on the view code.

src/windows/views/emojis_listview.py
```diff
@@ -52,14 +52,11 @@ def startDrag(self, event):
...
-        selected_row = self.emojis_model.model.itemFromIndex(self.emojis_model.proxy_model.mapToSource(self.selectionModel().selectedIndexes()[0])).row()
-        icon = self.emojis_model.model.item(selected_row, 0).icon()
+        selected_index = self.selectedIndexes()
 
         # Start drag operation
         drag = QDrag(self)
-        drag.setMimeData(self.emojis_model.proxy_model.mimeData(self.selectionModel().selectedIndexes()))
-        drag.setPixmap(icon.pixmap(QSize(self.drag_item_size, self.drag_item_size)))
-        drag.setHotSpot(QPoint(self.drag_item_size / 2, self.drag_item_size / 2))
+        drag.setMimeData(self.model.mimeData(selected_index))
```
ferdnyc, line 59: All of this stuff was _crazy_ overcomplicated because model data wasn't encapsulated fully in the model items.

Proxy models are just that — they _proxy_ the source model, so if the data is arranged properly there's no reason to have to ever translate references all the way back to the source model. We can retrieve all of the data we need right from the item reference in the (final) proxy model — it will take care of going back through any other proxies in the chain, and ultimately to the source, to access the underlying model data.

(I realized I dropped the pixmap here, I'll push another commit momentarily to restore that.)

src/windows/views/emojis_listview.py
```diff
@@ -109,48 +106,49 @@ def add_file(self, filepath):
...
 
+    @pyqtSlot(int)
+    def group_changed(self, index=-1):
+        log.info("Group set to: {}".format(index))
+        item = get_app().window.emojiFilterGroup.itemData(index)
+        log.info("Filtering on {}".format(item))
+        self.group_model.setFilterFixedString(item)
+        self.group_model.setFilterKeyColumn(1)
+
+        self.refresh_view()
```
ferdnyc, line 117: And I also left my debug logging in, dammit. I'll fix that too.

This is the code that implements the group-selection, which like I said is now separate from text filtering. It makes it a lot simpler _and_ a lot more like what you'd expect, IMHO.

Because `group_changed` is connected to `currentIndexChanged`, it gets an index for an item in the dropdown box — so I just pull the string to match from the data element (UserRole) of that entry and filter the model with an exact string match (`setFilterFixedString`) on the selected item. (It has to be the data role, since getting the `.text()` would return the translated, reformatted display text which isn't what we want.)

src/windows/views/emojis_listview.py
```diff
@@ -109,48 +106,49 @@ def add_file(self, filepath):
...
+
     @pyqtSlot(str)
-    def filter_changed(self):
+    def filter_changed(self, filter_text=None):
+        """Filter emoji with proxy class"""
+        log.info("Filter text: {}".format(filter_text or ""))
+        self.model.setFilterRegExp(QRegExp(filter_text, Qt.CaseInsensitive))
+        self.model.setFilterKeyColumn(0)
```
ferdnyc, line 124: Making the `QRegExp` case insensitive automatically takes care of `setFilterCaseInsensitive()`. In my testing, the reverse was _not_ the case.

src/windows/views/emojis_listview.py
```diff
@@ -109,48 +106,49 @@ def add_file(self, filepath):
...
-            # It's much faster for Qt to do this regex than use a custom
-            # Python filter function
-            # FYI:    smiley.*alien.*|alien.*smiley.*
-            filter_text = "%s.*%s.*|%s.*%s.*" % (filter_text, current_emoji_group, current_emoji_group, filter_text)
-        self.emojis_model.proxy_model.setFilterRegExp(QRegExp(filter_text.replace(' ', '.*')))
-        self.emojis_model.proxy_model.setFilterCaseSensitivity(Qt.CaseInsensitive)
-        self.emojis_model.proxy_model.sort(Qt.AscendingOrder)
+    def refresh_view(self):
+        col = self.model.sortColumn()
+        self.model.sort(col)
```
ferdnyc, line 129: In addition to cutting out all of that complexity, `self.emojis_model.proxy_model.sort(Qt.AscendingOrder)` was actually not working, since `QAbstractItemModel.sort()` takes a column index as its first argument. It's just pure luck that `Qt.AscendingOrder` happens to be the 0 item in the enum, so it was sorting column 0 instead of crashing.

src/windows/views/emojis_listview.py
```diff
@@ -109,48 +106,49 @@ def add_file(self, filepath):
...
-        self.win = get_app().window
+        # Get external references
+        app = get_app()
+        _ = app._tr
+        self.win = app.window
 
         # Get Model data
         self.emojis_model = model
+        self.group_model = self.emojis_model.group_model
+        self.model = self.emojis_model.proxy_model
```
ferdnyc, line 143: Because we need to access the `group_model` for the filtering, I kept references to all of the models here. But under normal circumstances, the view should only have to concern itself with the final proxy model in the chain (the one I labeled `self.model` here, and changed all references to use `self.model` except the one that needs `self.group_model`. `self.emojis_model` is actually never directly referenced again, once the proxy models are connected up.)

If not for the group filtering, it'd be possible to instantiate the view class with the final proxy model, _instead of_ the source model, and have it only use that. Which is probably what we should be doing, and I should've moved the dropdown list code to `main_window.py` instead, since that's where the list lives. Then we wouldn't need so much ugly referencing of `self.win.emojiFilterGroup` here in the view. Oh well.

src/windows/views/emojis_listview.py
```diff
@@ -159,12 +157,25 @@ def __init__(self, model):
...
+        # Loop through emoji groups, and populate emoji filter drop-down
+        self.win.emojiFilterGroup.clear()
+        self.win.emojiFilterGroup.addItem(_("Show All"), "")
+        dropdown_index = -1
+        for index, emoji_type in enumerate(sorted(self.emojis_model.emoji_groups)):
+            self.win.emojiFilterGroup.addItem(_(emoji_type.capitalize()), emoji_type)
+            if emoji_type == default_type:
+                # Initialize emoji filter group to settings
+                dropdown_index = index
+        self.win.emojiFilterGroup.setCurrentIndex(dropdown_index)
```
ferdnyc, line 181: Like I think I mentioned earlier, doing this last ­— and in particular, _after_ the signal is connected — means that it takes care of refreshing the view once it's done.

src/windows/views/emojis_listview.py
ferdnyc: Fixed, so bumping this comment in the hopes it won't keep getting hidden as "Outdated" since it's not.

src/windows/views/emojis_listview.py
ferdnyc: Fixed, so bumping this comment in the hopes it won't keep getting hidden as "Outdated" since it's not.

src/windows/views/emojis_listview.py
ferdnyc: Also not outdated.

SuslikV:
There is no reason to create the whole model at once. Model can be extended any time later.

ferdnyc:
> There is no reason to create the whole model at once. Model can be extended any time later.

I'm not sure I see how that's true, or at least I don't immediately see how it would be implemented. I know that incremental models are possible in Qt, but I don't see how that would extend to UI elements like this. I mean, take the case of sorting — in order to _sort_ the list of emoji, you need to _have_ the **entire** list of emoji... no?

Or do you mean just not attaching the icon images to the rows, except when they're displayed? That I could see maybe working, if the source model is built without icons and they only get attached to the currently-visible rows (if any) in the proxy model.

src/windows/views/emojis_listview.py
ferdnyc: (A better design might be to have a method in `main_window` connected to the `emojiFilterGroup`'s `currentIndexChanged` signal, and have that function update the proxy model directly, relying on signals to update the corresponding view.

...It's tricky because proxying of a model is neither a view function nor a model function — which is kind of entirely the point, it's why proxies exist. Different views may very well share some proxy models, but a particular view may also bring along its own unique proxy models to interpolate the data in a certain way that only makes sense for that view. As long as the underlying _source_ model is persistent, that's not such a bad thing.

In the old Qt 4.6 [proxy model documentation](https://doc.qt.io/archives/4.6/model-view-proxy-models.html), Qt even spelled it out as such, in a direct, concise way that their current docs, while very comprehensive, sadly lack:
> Although it seems appropriate to perform sorting and filtering operations as internal functions of views, this approach does not allow multiple views to share the results of such potentially costly operations. The alternative approach, involving sorting within the model itself, leads to the similar problem where each view has to display items of data that are organized according to the most recent processing operation.
>
> To solve this problem, the model/view framework uses proxy models to manage the information supplied between individual models and views. **Proxy models are components that behave like ordinary models from the perspective of a view, and access data from source models on behalf of that view.** The signals and slots used by the model/view framework ensure that each view is updated appropriately no matter how many proxy models are placed between itself and the source model.

SuslikV:
You just said earlier that model creation on startup of the program takes too much time. I was thinking about small model (10 images) at start up and load everything else on demand.

src/windows/models/emoji_model.py
jonoomph: Very nice, never thought of that!

src/windows/views/emojis_listview.py
jonoomph: Nice catch!

jonoomph:
Merging these improvements into the emoji branch. Thanks @ferdnyc !

PR #3390, opened 2020-04-19
freeze.py: Add paths to moved Windows DLLs

ferdnyc:
This branch fixes the Windows builders' freezing of the application, after the DLL files are relocated from `lib` to `bin` under the install prefix. `bin/` is the correct install subdirectory for Windows DLLs (and every system DLL is in `C:\msys64\mingw64\bin`), so we would do better to follow convention.


ferdnyc:
(Note that OpenShot/libopenshot-audio#96 and OpenShot/libopenshot#501 make precisely such a change, so this is in anticipation of those PRs being merged. However, there's no harm in doing this either way, since it only adds to the possible search locations for files — no existing functionality is affected.)

PR #3394, opened 2020-04-22
Docs: Add Sphinx directive for youtube embeds

ferdnyc:
Add a simple directive `.. youtube`, implemented as a local sphinx extension, that takes one argument:
  - A YouTube video ID (e.g. `aqz-KE-bpKQ`)
  - A YouTube URL (e.g. `https://www.youtube.com/watch?v=aqz-KE-bpKQ`)

When generating HTML docs, any of these will be replaced with an embedded video player showing ["Big Buck Bunny"](https://www.youtube.com/watch?v=aqz-KE-bpKQ):
```rst
.. youtube:: aqz-KE-bpKQ

.. youtube:: https://www.youtube.com/watch?v=aqz-KE-bpKQ

.. youtube:: https://youtu.be/aqz-KE-bpKQ
````

Advanced customization features like start time (`?t=45`) are currently not supported.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -43,6 +43,15 @@ try:
 except ImportError:
     pass
 
+import sys
+sys.path.insert(0, '.')
+try:
+    # Load our YouTube directive
+    import youtube_directive
+    extensions.append('youtube_directive')
+except ImportError:
+    pass
+
 # External links mappings for extlinks
 # see: http://www.sphinx-doc.org/en/master/usage/extensions/extlinks.html
 extlinks = {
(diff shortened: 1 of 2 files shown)


ferdnyc:
Huh. Turns out (not at all surprisingly), someone had already written such a thing. Almost 10 years ago.

[sphinx-contrib: youtube extension](https://bitbucket.org/birkenfeld/sphinx-contrib/src/default/youtube/)

It offers more configurability for how the video is embedded (controlling the width/height, etc.) buuuut unlike my version it doesn't allow you to enter a video URL, you have to extract just the ID. 

It's also so ancient that it uses HTTP URLs, which wouldn't be acceptable since the rest of the guide is served over HTTPS and we'd end up triggering the browser's insecure-page-element blocking. So, guess we should stick with mine.

ferdnyc:
Oh, I'm way behind the times. The updated codebase lives here on GitHub as [sphinx-contrib/youtube](https://github.com/sphinx-contrib/youtube), and sports all the expected bells and whistles.

ferdnyc:
Merging this as it won't affect anything unless we use it. (And if we end up switching over to the version from the official extensions collection, we can just delete this one from the `doc/` dir.)

PR #3398, opened 2020-04-23
appdata: add <provides>

eszlari:
A search for "openshot"  in the software manager `plasma-discover` on distributions with a not up-to-date version of Openshot, returns two entries: one for the flatpak, the other for the (old) distro package, that uses the former `<id>`. This PR makes sure, the manager recognizes, they are the same application and you only get one entry (with the option to choose the source).


ferdnyc:
Works for me. I'd have already merged it, if Travis weren't blocking me and apparently out to lunch. :angry:

ferdnyc:
Finally succeeded in reviving Travis, merging. Thanks @eszlari !

PR #3400, opened 2020-04-23
Update the README copyright range to include 2020

ferdnyc:


ferdnyc:
Travis is finally un-f***d, merging this two-character README  change. :laughing:

Issue #3078, opened 2019-11-16
Copy & Paste / Drag & Drop of Effects


mkarg:
**Describe the new feature:**
Copy & Paste / Drag & Drop a customized effect from one clip to another.

**Describe the solution you'd like:**
CTRL-C copies the selected customized effect (i. e. the effect type with all its current custom settings including all key frames, hence the JSON) from a source clip into the clipboard. CTRL-V on a target clip applies the same effect type with the settings found in the clipboard (including the key frames). CTRL-X does the same as CTRL-C but additionally _removes_ the effect from the source clip.

Mouse can move a customized effect from source clip to target clip using drag & drop. Holding CTRL when releasing the mouse button will _copy_ instead of _move_.

**Describe alternatives you have considered:**
Manually copying setting-by-setting from source effect to target effect.

SuslikV:
Unfortunately, even available code is not working (Copy/Paste over the Clip makes not editable Effect that keeps reference to parent Clip).

PR #3402, opened 2020-04-23
[#1363] Keyframe navigation as marker

### LINKED ISSUE #1363: Shortcut for jumping to the next/previous keyframe
Hi,
coming from Blender VSE, I miss a small feature:
In Blender, I can let the marker jump to the previous or next keyframe of the selected clip. This way I can easily go through all keyframes of the clip.
It would be awesome if this could be added to OpenShot.
Maybe Shift+Right for the next keyframe and Shift+Left for the previous?

vincentdavoust:
A suggested in issue #1363, It would be nice to be able to navigate precisely between keyframes. 

This patch allows user to use arrow buttons or Ctrl-L/R to navigate to keyframes of selected clips as well as the normal markers and clip beginning/end.

Fixes #1363

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -1238,22 +1238,37 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
             fps = get_app().project.get("fps")
             fps_float = float(fps["num"]) / float(fps["den"])
 
-            clip_start_time=obj.data["position"]
-            clip_stop_time=obj.data["position"] + (obj.data["end"] - obj.data["start"])
+            clip_start_time = obj.data["position"]
+            clip_orig_time = clip_start_time - obj.data["start"]
+            clip_stop_time = clip_orig_time + obj.data["end"]
 
             # add clip boundaries
-            all_marker_positions.append(clip_start_time)
-            all_marker_positions.append(clip_stop_time)
+            positions.append(clip_start_time)
+            positions.append(clip_stop_time)
 
             # add all keyframes
             for property in obj.data :
                 try :
                     for point in obj.data[property]["Points"] :
-                        keyframe_time=(point["co"]["X"]-1)/fps_float - obj.data["start"] + obj.data["position"]
+                        keyframe_time = (point["co"]["X"]-1)/fps_float - obj.data["start"] + obj.data["position"]
                         if keyframe_time > clip_start_time and keyframe_time < clip_stop_time :
                             positions.append(keyframe_time)
                 except (TypeError, KeyError):
                     pass
+
+
+            # Add all Effect keyframes
+            if "effects" in obj.data:
+                for effect_data in obj.data["effects"]:
+                    for property in effect_data:
+                        try:
+                            for point in effect_data[property]["Points"]:
+                                keyframe_time = (point["co"]["X"]-1)/fps_float + clip_orig_time
+                                if keyframe_time > clip_start_time and keyframe_time < clip_stop_time:
+                                    positions.append(keyframe_time)
+                        except (TypeError, KeyError):
+                            pass
+
             return positions
 
         all_marker_positions = []
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -1227,13 +1227,29 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         marker.data = {"position": position, "icon": "blue.png"}
         marker.save()
 
-    def actionPreviousMarker_trigger(self, event):
-        log.info("actionPreviousMarker_trigger")
 
-        # Calculate current position (in seconds)
+    def findAllMarkerPositions(self):
+        """Build and return a list of all seekable locations for the currently-selected timeline elements"""
+
+        def getKeyframePositions(data, clip_start_time, clip_stop_time, fps_float):
+            """ Add all keyframes of a clip.data to all_marker_positions """
+            positions = []
+            for property in data :
+                try :
+                    for point in data[property]["Points"] :
+                        keyframe_time=(point["co"]["X"]-1)/fps_float - data["start"] + data["position"]
+                        if keyframe_time > clip_start_time and keyframe_time < clip_stop_time :
+                            positions.append(keyframe_time)
+                except TypeError:
+                    log.info("%s : %s : not itterable", property, data[property])
+                    pass
+                except KeyError:
+                    log.info("%s : %s : has no points", property, data[property])
+                    pass
+            return positions
+
         fps = get_app().project.get("fps")
         fps_float = float(fps["num"]) / float(fps["den"])
-        current_position = (self.preview_thread.current_frame - 1) / fps_float
         all_marker_positions = []
 
         # Get list of marker and important positions (like selected clip bounds)
@@ -1245,10 +1261,12 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
             # Get selected object
             selected_clip = Clip.get(id=clip_id)
             if selected_clip:
-                all_marker_positions.append(selected_clip.data["position"])
-                all_marker_positions.append(selected_clip.data["position"] + (selected_clip.data["end"] - selected_clip.data["start"]))
-                # add all keyframes. itterate on all properties of Data, look for each points
-                addKeyframesToMarkers(selected_clip.data)
+                clip_start_time=selected_clip.data["position"]
+                clip_stop_time=selected_clip.data["position"] + (selected_clip.data["end"] - selected_clip.data["start"])
+                all_marker_positions.append(clip_start_time)
+                all_marker_positions.append(clip_stop_time)
+                # add all keyframes
+                all_marker_positions.extend(getKeyframePositions(selected_clip.data, clip_start_time, clip_stop_time, fps_float))
 
         # Loop through selected transitions (and add key positions)
         for tran_id in self.selected_transitions:
@@ -1257,6 +1275,20 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
             if selected_tran:
                 all_marker_positions.append(selected_tran.data["position"])
                 all_marker_positions.append(selected_tran.data["position"] + (selected_tran.data["end"] - selected_tran.data["start"]))
+
+        # remove duplicates
+        all_marker_positions = list(set(all_marker_positions))
+
+        return all_marker_positions
+
+    def actionPreviousMarker_trigger(self, event):
+        log.info("actionPreviousMarker_trigger")
+
+        # Calculate current position (in seconds)
+        fps = get_app().project.get("fps")
+        fps_float = float(fps["num"]) / float(fps["den"])
+        current_position = (self.preview_thread.current_frame - 1) / fps_float
+        all_marker_positions = self.findAllMarkerPositions()
 
         # Loop through all markers, and find the closest one to the left
         closest_position = None
@@ -1289,28 +1321,7 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         fps = get_app().project.get("fps")
         fps_float = float(fps["num"]) / float(fps["den"])
         current_position = (self.preview_thread.current_frame - 1) / fps_float
-        all_marker_positions = []
-
-        # Get list of marker and important positions (like selected clip bounds)
-        for marker in Marker.filter():
-            all_marker_positions.append(marker.data["position"])
-
-        # Loop through selected clips (and add key positions)
-        for clip_id in self.selected_clips:
-            # Get selected object
-            selected_clip = Clip.get(id=clip_id)
-            if selected_clip:
-                all_marker_positions.append(selected_clip.data["position"])
-                all_marker_positions.append(selected_clip.data["position"] + (selected_clip.data["end"] - selected_clip.data["start"]))
-                # add all keyframes. itterate on all properties of Data, look for each points
-                addKeyframesToMarkers(selected_clip.data)
-        # Loop through selected transitions (and add key positions)
-        for tran_id in self.selected_transitions:
-            # Get selected object
-            selected_tran = Transition.get(id=tran_id)
-            if selected_tran:
-                all_marker_positions.append(selected_tran.data["position"])
-                all_marker_positions.append(selected_tran.data["position"] + (selected_tran.data["end"] - selected_tran.data["start"]))
+        all_marker_positions = self.findAllMarkerPositions()
 
         # Loop through all markers, and find the closest one to the right
         closest_position = None
@@ -1334,20 +1345,6 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
             # Update the preview and reselct current frame in properties
             get_app().window.refreshFrameSignal.emit()
             get_app().window.propertyTableView.select_frame(frame_to_seek)
-
-    def addKeyframesToMarkers(self, data):
-        for property in data :
-            try :
-                for point in data[property]["Points"] :
-                    keyframe=(point["co"]["X"]-1)/fps_float - data["start"] + data["position"]
-                    if keyframe > start and keyframe < stop :
-                        all_marker_positions.append(keyframe)
-            except TypeError:
-                log.info("%s : %s : not itterable", property, data[property])
-                pass
-            except KeyError:
-                log.info("%s : %s : has no points", property, data[property])
-                pass
 
     def actionCenterOnPlayhead_trigger(self, event):
         """ Center the timeline on the current playhead position """
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -1231,25 +1231,31 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
     def findAllMarkerPositions(self):
         """Build and return a list of all seekable locations for the currently-selected timeline elements"""
 
-        def getKeyframePositions(data, clip_start_time, clip_stop_time, fps_float):
-            """ Add all keyframes of a clip.data to all_marker_positions """
+        def getTimelineObjectPositions(obj):
+            """ Add boundaries & all keyframes of a timeline object (clip, transition...) to all_marker_positions """
             positions = []
-            for property in data :
+
+            fps = get_app().project.get("fps")
+            fps_float = float(fps["num"]) / float(fps["den"])
+
+            clip_start_time=obj.data["position"]
+            clip_stop_time=obj.data["position"] + (obj.data["end"] - obj.data["start"])
+
+            # add clip boundaries
+            all_marker_positions.append(clip_start_time)
+            all_marker_positions.append(clip_stop_time)
+
+            # add all keyframes
+            for property in obj.data :
                 try :
-                    for point in data[property]["Points"] :
-                        keyframe_time=(point["co"]["X"]-1)/fps_float - data["start"] + data["position"]
+                    for point in obj.data[property]["Points"] :
+                        keyframe_time=(point["co"]["X"]-1)/fps_float - obj.data["start"] + obj.data["position"]
                         if keyframe_time > clip_start_time and keyframe_time < clip_stop_time :
                             positions.append(keyframe_time)
-                except TypeError:
-                    log.info("%s : %s : not itterable", property, data[property])
-                    pass
-                except KeyError:
-                    log.info("%s : %s : has no points", property, data[property])
+                except (TypeError, KeyError):
                     pass
             return positions
 
-        fps = get_app().project.get("fps")
-        fps_float = float(fps["num"]) / float(fps["den"])
         all_marker_positions = []
 
         # Get list of marker and important positions (like selected clip bounds)
@@ -1261,20 +1267,14 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
             # Get selected object
             selected_clip = Clip.get(id=clip_id)
             if selected_clip:
-                clip_start_time=selected_clip.data["position"]
-                clip_stop_time=selected_clip.data["position"] + (selected_clip.data["end"] - selected_clip.data["start"])
-                all_marker_positions.append(clip_start_time)
-                all_marker_positions.append(clip_stop_time)
-                # add all keyframes
-                all_marker_positions.extend(getKeyframePositions(selected_clip.data, clip_start_time, clip_stop_time, fps_float))
+                all_marker_positions.extend(getTimelineObjectPositions(selected_clip))
 
         # Loop through selected transitions (and add key positions)
         for tran_id in self.selected_transitions:
             # Get selected object
             selected_tran = Transition.get(id=tran_id)
             if selected_tran:
-                all_marker_positions.append(selected_tran.data["position"])
-                all_marker_positions.append(selected_tran.data["position"] + (selected_tran.data["end"] - selected_tran.data["start"]))
+                all_marker_positions.extend(getTimelineObjectPositions(selected_tran))
 
         # remove duplicates
         all_marker_positions = list(set(all_marker_positions))
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -1247,7 +1247,7 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
             if selected_clip:
                 all_marker_positions.append(selected_clip.data["position"])
                 all_marker_positions.append(selected_clip.data["position"] + (selected_clip.data["end"] - selected_clip.data["start"]))
-                # add all keyframes. itterate on all properties of Data,, and for each every Points
+                # add all keyframes. itterate on all properties of Data, look for each points
                 for property in selected_clip.data :
                     try :
                         for point in selected_clip.data[property]["Points"] :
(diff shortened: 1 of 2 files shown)


src/windows/main_window.py
```diff
@@ -1332,6 +1335,20 @@ def actionNextMarker_trigger(self, event):
             get_app().window.refreshFrameSignal.emit()
             get_app().window.propertyTableView.select_frame(frame_to_seek)
 
+    def addKeyframesToMarkers(self, data):
+        for property in data :
+            try :
+                for point in data[property]["Points"] :
+                    keyframe=(point["co"]["X"]-1)/fps_float - data["start"] + data["position"]
+                    if keyframe > start and keyframe < stop :
+                        all_marker_positions.append(keyframe)
```
ferdnyc, line 1344: As with `fps_float`, `all_marker_positions` is undefined inside this function. You can't make references to variables from your caller.

The way I'd suggest to do this is, inside the function you should create a _list_ of the marker positions you want to add, and return that back to the caller. Then, you can add its results to the `all_marker_positions` list from the calling method. So, basically like this:

```python
# In your function -- which, again, needs to be defined BEFORE it's called — build and return a list:
# (I also renamed it to better indicate its purpose, since it can't directly add to the markers list)
def getKeyframePositions(data):
    positions = []
    for property in data:
        # (Do the work to compute a position...)
        positions.append(keyframe_time)
    return positions

# And then in either of the places where you call the function:
all_marker_positions.extend(getKeyframePositions(selected_clip.data))
```

Using `list.extend()` is crucial here, as it combines one list with the contents of another, rather than adding the additional list as a single element.
```python
# If you have a list like this:
>>> mylist = [1, 2, 3]
# the difference is:
>>> mylist.append([4, 5, 6])
# mylist = [1, 2, 3, [4, 5, 6]]
>>> mylist.extend([4, 5, 6])
# mylist = [1, 2, 3, 4, 5, 6]
```

src/windows/main_window.py
```diff
@@ -1247,6 +1247,8 @@ def actionPreviousMarker_trigger(self, event):
             if selected_clip:
                 all_marker_positions.append(selected_clip.data["position"])
                 all_marker_positions.append(selected_clip.data["position"] + (selected_clip.data["end"] - selected_clip.data["start"]))
+                # add all keyframes. itterate on all properties of Data, look for each points
```
ferdnyc, line 1250: Nit: One 't' in 'iterate', 'iterable'.

Also, now that the actual iteration is in a separate function, it might be better to leave out the description of _how_ this is done here, and move it to the function body instead.

src/windows/main_window.py
```diff
@@ -1332,6 +1335,20 @@ def actionNextMarker_trigger(self, event):
...
+                for point in data[property]["Points"] :
+                    keyframe=(point["co"]["X"]-1)/fps_float - data["start"] + data["position"]
+                    if keyframe > start and keyframe < stop :
+                        all_marker_positions.append(keyframe)
+            except TypeError:
+                log.info("%s : %s : not itterable", property, data[property])
+                pass
+            except KeyError:
+                log.info("%s : %s : has no points", property, data[property])
+                pass
```
ferdnyc, line 1350: I don't think there's any reason to _log_ either of these conditions. You won't actually get a `TypeError` for non-iterables, you'll hit it even earlier than that since non-keyframable properties like `data["duration"]` don't have a `Points` member, so the error will be that they're non-_subscriptable_. But that's expected and not an extraordinary condition.

Any properties without any keyframe points similarly don't need to be logged; that information is of no use to the user, or even to the devs in the context of the markers list. So the exceptions can simply be ignored.
```suggestion
            except (TypeError, KeyError):
                pass
```

src/windows/main_window.py
ferdnyc: I was going to say that writing this as a function means you _have_ to pass in `fps_float` from the caller, but I don't think that's right. `get_app()` should work from a function as well as from a method. So it's your call whether you pass it in, or compute it from the value of `get_app().project.get("fps")` inside your function.

ferdnyc:
@vincentdavoust 

(I updated your initial comment, just to add a [closing keyword](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) linking the PR to #1363.)

vincentdavoust:
@ferdnyc Thanks for spending time on this ! I'm sorry I made so many basic mistakes, it turns out when I tested the old vesion from openshot_qt/ was running not the one from src/ so obviously when I broke everything I didn't realise...

I'm making all the fixes you suggested (except the invisible keyframe issue). But I'm not sure where in the codebase I should define the getKeyframePositions() function. It seems unclean to me to leave it in mainwindow if it is no longer a method. Do you have a suggestion ?

ferdnyc:
> @ferdnyc Thanks for spending time on this ! I'm sorry I made so many basic mistakes, it turns out when I tested the old vesion from openshot_qt/ was running not the one from src/ so obviously when I broke everything I didn't realise...

Can't tell you how familiar that sounds. :-D It gets even more exciting when you need to test with a version of libopenshot that's not the same as the one installed on your system (say, because the current development code has a newer API) — as a hint in case you ever run into that, `PYTHONPATH=/where/you/built/libopenshot/src/bindings/python python3 launch.py` is your friend.

> I'm making all the fixes you suggested (except the invisible keyframe issue). 

Yeah, I wouldn't worry about that for now, that can be a "later" thing. I still have to think on it myself. Like I said, it may make more sense to change the _display_ logic, and if we do that then nothing will need to change in your code.

That actually did make me think of something else, though — you're going to end up with a _lot_ of duplicate positions in your keyframe list, since most properties will only have a single point at `X = 1`, and often when there are _other_ points, multiple properties get modified at the same frame. So, you're probably going to want to deduplicate the list that you return to the caller.

The common methods of deduplicating a list are to round-trip convert it using either a set or a dict as the intermediate form, since both require unique members. So either of these would work:
```python3
>>> t = [1, 2, 1, 1, 5, 1, 1, 1, 1]
>>> print(list(set(t)))
[1, 2, 5]
>>> print(list(dict.fromkeys(t)))
[1, 2, 5]
```

In fact, now that I think about it that should be done by the caller (either in addition to or instead of your code), because there may also be overlap between markers and keyframe points, and you'd want to consolidate those as well.

> But I'm not sure where in the codebase I should define the getKeyframePositions() function. It seems unclean to me to leave it in mainwindow if it is no longer a method. Do you have a suggestion ?

That's a fair question. _Personally_ I don't hate the idea of it being in `main_window.py`, but I can see the arguments against it.

Though, as I'm looking at it now what strikes me is that **ALL** of the code that assembles the marker list should really be abstracted out, rather than being cut-and-paste duplicated between `actionNextMarker_trigger` and `actionPreviousMarker_trigger`. (Don't get me started about all of the copy-paste code in OpenShot, I'll never stop ranting.)

If you were to abstract out the entirety of the list-building code, then it would absolutely have to be a method since it would need access to `self.selected_clips` and `self.selected_transitions`, so that handily avoids having to decide on an answer to that question.

ferdnyc:
Er, oh, right, the above doesn't actually address the issue of where to define the _algorithm_, since you'd still want a `getKeyframePositions()`.

But, here's a very Pythonic solution to that problem: If you've got things abstracted out to where there's only one method that needs to _call_ `getKeyframePoints()`, then the most natural place to define it is **in** that method, at the beginning before you start the overall list-building process. So,
```python3
def findAllMarkerPositions(self):
    """Build and return a list of all seekable locations for the
    currently-selected timeline elements"""
    
    # Function for extracting keyframe positions
    def getKeyframePositions(data, fps_float):
        # ... your code ...
        return positions

    # Calculate current position (in seconds)
    fps = get_app().project.get("fps")
    current_position = (self.preview_thread.current_frame - 1) / fps_float
    all_marker_positions = []

    # Get list of marker and important positions (like selected clip bounds)
    for marker in Marker.filter():
        all_marker_positions.append(marker.data["position"])

    # Loop through selected clips (and add key positions)
    for clip_id in self.selected_clips:
        # Get selected object
        selected_clip = Clip.get(id=clip_id)
        if selected_clip:
            all_marker_positions.append(selected_clip.data["position"])
            all_marker_positions.append(selected_clip.data["position"] + (selected_clip.data["end"] - selected_clip.data["start"]))
            # add all keyframes for clip
            all_marker_positions.extend(getKeyframePositions(selected_clip.data, fps_float))
    # and so on...
```

vincentdavoust:
I've now pushed your recommended changes. 
2 additional notes :
- Maybe I should add the same logic for transition keyframes as well ? I didn't do it yet because I never encountered the need to have that as a user, so i'm not sure it's relevant. 
- Ctrl+click visually appears to add a selected clip, but the additional clips are not added to self.selected_clips, so for navigation between keyframes of multiple clips they all need to be selected at once. I will investigate a bit more and open a separate issue.

SuslikV:
> * Ctrl+click visually appears to add a selected clip, but the additional clips are not added to self.selected_clips, so for navigation between keyframes of multiple clips they all need to be selected at once. I will investigate a bit more and open a separate issue...

This part has PR: https://github.com/OpenShot/openshot-qt/pull/3317 (p.2)

src/windows/main_window.py
```diff
@@ -1245,8 +1261,12 @@ def actionPreviousMarker_trigger(self, event):
...
             selected_clip = Clip.get(id=clip_id)
             if selected_clip:
-                all_marker_positions.append(selected_clip.data["position"])
-                all_marker_positions.append(selected_clip.data["position"] + (selected_clip.data["end"] - selected_clip.data["start"]))
+                clip_start_time=selected_clip.data["position"]
+                clip_stop_time=selected_clip.data["position"] + (selected_clip.data["end"] - selected_clip.data["start"])
+                all_marker_positions.append(clip_start_time)
+                all_marker_positions.append(clip_stop_time)
+                # add all keyframes
+                all_marker_positions.extend(getKeyframePositions(selected_clip.data, clip_start_time, clip_stop_time, fps_float))
```
ferdnyc, line 1269: Can't you move this logic into `getKeyframePositions()` as well? It will be the same whether you've got a transition or a clip selected, either way you're adding two positions to the list:
```python3
positions.append(data["position"])
positions.append(data["position"] + (data["end"] - data["start"]))
```
If you do that _inside_ the function, then
1. You don't have to duplicate that code in the `if selected_clip:` and `if selected_tran:` sections
2. You don't need to pass `start_time` and `stop_time` as arguments to `getKeyframePositions()`. You already have everything you need in the `data` dictionary to just compute them inside the function.

Then you should only need a single call per clip/transition:
```python
all_marker_positions.extend(getKeyframePositions(selected_clip.data, fps_float))
# or
all_marker_positions.extend(getKeyframePositions(selected_tran.data, fps_float))
```

src/windows/main_window.py
```diff
@@ -1227,13 +1227,29 @@ def actionAddMarker_trigger(self, event):
...
 
-    def actionPreviousMarker_trigger(self, event):
-        log.info("actionPreviousMarker_trigger")
 
-        # Calculate current position (in seconds)
+    def findAllMarkerPositions(self):
+        """Build and return a list of all seekable locations for the currently-selected timeline elements"""
+
+        def getKeyframePositions(data, clip_start_time, clip_stop_time, fps_float):
+            """ Add all keyframes of a clip.data to all_marker_positions """
```
ferdnyc, line 1235: I wouldn't use "clip" here. It could confuse someone reading the code in the future, since it's actually called for both clips and transitions. Something like "timeline element" is suitably generic.

src/windows/main_window.py
```diff
@@ -1256,6 +1276,20 @@ def actionPreviousMarker_trigger(self, event):
                 all_marker_positions.append(selected_tran.data["position"])
                 all_marker_positions.append(selected_tran.data["position"] + (selected_tran.data["end"] - selected_tran.data["start"]))
 
+        # remove duplicates
+        all_marker_positions = list(set(all_marker_positions))
```
ferdnyc, line 1280: I realized — one thing we should watch for in testing in case it's an issue, though I don't think it's anything that needs to be addressed until that time: We could end up with some duplicates remaining on the list, because these are floating-point values so "equal" points won't always come out exactly equal.

In practice I don't think it'll matter 99% of the time, since even if there are multiples they should all translate back to the same single frame number for seeking purposes. Though I suppose if the user's current position is bordered by other not-quite-equal positions, worst case they might have to skip an extra time or two to get off that position and jump to the actual next point.

ferdnyc:
> * Maybe I should add the same logic for transition keyframes as well ? I didn't do it yet because I never encountered the need to have that as a user, so i'm not sure it's relevant.

Do you mean transitions (which you already do), or _effects_? Those are the one part that's so far unaddressed here.

In testing this with a project that does use effects on some clips, it **is** sort of noticeable that _some_ of the green marks on a clip can be seeked to (because they're property keyframes), but then _other_ green marks will just be ignored (because they're effect keyframes). So, there's that confusion aspect to consider. Plus, I feel like users will probably want to be able to seek to those keyframes as well, the same way they do with clip properties.

It shouldn't take that much to add, either, I wouldn't think. Just a matter of checking for the existence of a `data["effects"]` member, and then iterating over its contents the same way you do for the outer clip/transition properties list.

src/windows/main_window.py
```diff
@@ -1227,13 +1227,29 @@ def actionAddMarker_trigger(self, event):
...
+                    for point in data[property]["Points"] :
+                        keyframe_time=(point["co"]["X"]-1)/fps_float - data["start"] + data["position"]
+                        if keyframe_time > clip_start_time and keyframe_time < clip_stop_time :
+                            positions.append(keyframe_time)
+                except TypeError:
+                    log.info("%s : %s : not itterable", property, data[property])
+                    pass
+                except KeyError:
+                    log.info("%s : %s : has no points", property, data[property])
+                    pass
```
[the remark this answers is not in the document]
ferdnyc: Hmm, that logging also points out that the current code gets tripped up by color keyframes, since they have a slightly different format. IIRC the rest of the code typically looks for a `red` key as an indicator that it should interpret the property as a *group* of color-channel points, instead of just a single list of points.

src/windows/main_window.py
ferdnyc: (And yes, I missed that myself when I was whipping up my function-placement example earlier, so I'm partly at fault here.)

ferdnyc: Also, I supppose it would make `getKeyframePositions()` an inaccurate name for the function. Maybe something like `getSeekablePositions()` is better.

src/windows/main_window.py
vincentdavoust: I haven't investigated why, but in practice even without this duplicate removal I found no case of having to iterate several times on the same point, even with multiple elements on one point. I'm guessing because we navigate one right or one left searching for nearest non equal. But as you say it could possibly happen also with floating points.

ferdnyc: Oh, yeah, the duplicate removal isn't meant to affect seeking — you'll always be looking for the _next_ lower/higher time value to your current position. So it wouldn't matter if there are 1 or 27 duplicates of that next position — or even your current position — on the list. There'll still be a next adjacent position.

The duplicate removal is just to make the list shorter, because otherwise you'd end up sorting/searching a list filled with literally two dozen `0`s for **each** selected clip. :laughing: 

If the floating-point thing does prove to be an issue, duplicate removal isn't going to solve it anyway, since the whole thing there is that "equal" values may _not_ be **equal**-equal. But I'm really not worried about that, unless it actually comes up. For now I'm content to file it away as something to watch out for, nothing more.

ferdnyc: (Where "`0`" is the clip's `data["position"]` on the timeline, not necessarily `0`.)

SuslikV:
Effect's keys support (for selected clips only):

```
from classes.query import File, Clip, Transition, Effect, Marker, Track

...

        def getTimelineObjectPositions(obj, is_effect=False):
            """ Add boundaries & all keyframes of a timeline object (clip, transition...) to all_marker_positions """
            positions = []

            fps = get_app().project.get("fps")

            if is_effect:
                # Get the effect's correct starting position
                obj.data["position"] = obj.parent["position"]
                obj.data["start"] = obj.parent["start"]
                obj.data["end"] = obj.parent["end"]

...

        # Loop through selected clips (and add key positions)
        for clip_id in self.selected_clips:
            # Get selected object
            selected_clip = Clip.get(id=clip_id)
            if selected_clip:
                all_marker_positions.extend(getTimelineObjectPositions(selected_clip))
                # Add all effects of the selected clip too
                for e in selected_clip.data["effects"]:
                    effect = Effect.get(id=e["id"])
                    if not effect:
                        continue
                    all_marker_positions.extend(getTimelineObjectPositions(effect, True))

...
```

src/windows/main_window.py
```diff
@@ -1227,13 +1227,35 @@ def actionAddMarker_trigger(self, event):
...
+            positions = []
+
+            fps = get_app().project.get("fps")
+            fps_float = float(fps["num"]) / float(fps["den"])
+
+            clip_start_time=obj.data["position"]
+            clip_stop_time=obj.data["position"] + (obj.data["end"] - obj.data["start"])
+
+            # add clip boundaries
+            all_marker_positions.append(clip_start_time)
```
SuslikV, line 1245:
```suggestion
            positions.append(clip_start_time)
```

Better to make function `getTimelineObjectPositions(obj)` outside the method and class (before _MainWindow_). Just illogical - you modifying some unknown object and later returning other object, and both actions are in one function.

SuslikV:
_ToolTip_ and _Text_ of the marker's action requires correction too, like:

```
   <property name="text">
    <string>Previous Marker</string>
   </property>
   <property name="toolTip">
    <string>Previous Marker</string>
   </property>
```
to 
```
   <property name="text">
    <string>Previous Marker/Keyframe</string>
   </property>
   <property name="toolTip">
    <string>Previous Marker/Keyframe</string>
   </property>
```

and Next Maker in similar way.

SuslikV:
Only bug I found - is the copy/paste bug of the Clip: https://github.com/OpenShot/openshot-qt/issues/3078#issuecomment-619598784 , but this is not about the navigation itself (despite it becomes useless in this case). And this will affect the Curve Editor too...

src/windows/main_window.py
vincentdavoust: I do agree about changing `all_marker_positions` to `positions` in `getTimelineObjectPositions(obj)`, however I'm not sure about the placement of the function. Unless it is used by something else than `findAllMarkerPositions()`, I think it is cleanest to leave it inside. And otherwise, maybe finding another place than `main_window.py`, if it is not inside the `MainWindow` class

SuslikV: Nested functions is hard to read. This usually introduces mistakes in the code.

Performance hit:
https://stackoverflow.com/questions/14055843/what-is-the-performance-overhead-of-nested-functions

ferdnyc: ^ Not a meaningful one. As the conclusion to that answer says,
> Remember that in this trivial case creating and calling functions is all the interpreter does. In any real usage the difference will be much less.

The difference, in terms of OpenShot's total CPU cycles, will be negligible at worst, most likely undetectable.

Locally-declared worker functions are an accepted pattern in high-level languages like Python and JavaScript. Heck, they're even starting to gain some traction in C++, with STL classes like `std::sort()` that take comparison functors. Those usually get declared anonymously right before they're passed in to the library class.

ferdnyc:
> Only bug I found - is the copy/paste bug of the Clip: https://github.com/OpenShot/openshot-qt/issues/3078#issuecomment-619598784 , but this is not about the navigation itself

Ouch! Good find. Yeah that'll need to get fixed.

But as you say, it's a separate issue from this PR. Until copy/paste is fixed for effects, using it will cause problems for this code... but only in ways it wouldn't be expected to deal with. Fixing the bug at its source will solve the problem here and anywhere else it impacts as well.

(My kingdom, though, for a generic "timeline element" superclass. A way to handle Clips, Effects, and Transitions all at once, generically and interchangeably in cases where they _are_ interchangeable (like selection, or positioning), would simplify a _lot_ of code. On **both** the Python and JavaScript sides. 

Oh well. Maybe for 3.0.)

src/windows/main_window.py
ferdnyc: > however I'm not sure about the placement of the function. Unless it is used by something else than `findAllMarkerPositions()`, I think it is cleanest to leave it inside. 

That's my thinking, as you know.

> And otherwise, maybe finding another place than `main_window.py`, if it is not inside the `MainWindow` class

Because it deals actively — _inter_actively, in fact — with selection and UI navigation, it's not really appropriate for any of the `src/classes/` modules where helper functions typically reside. So it has to live _somewhere_ in the `src/windows/` tree. The only other reasonable place would be somewhere in `src/windows/views/timeline_webview.py`.

But IMHO it's already right where it belongs.

SuslikV: @ferdnyc I see no arguments in your words but only feelings. It slower. It is harder to read. So, what benefits of "nesting" here?

vincentdavoust: @SuslikV I agree with @ferdnyc. 
Performance is the only truly objective metric here, and yes, nesting the function is a slight performance loss, but it's objectively negligible. (number of additional cpu cycles would be orders of magnitude lower than total) 
Code readability however, is subjective but important. And I think it's much much easier to read with the function being nested than anywhere else. It would make no sense to me to have the main_window file start with a function that has no obvious role for the mainwindow. In the calling method however, it makes sense. From that method, we know exactly what the relevance of the callee is for the caller, it's close by in the code so no scrolling up and down far, and as the nested function is never called by anything else than it's parent method, so there is no use or sense exposing it to a larger scope.

If performance was the only important aspect, we would be using a purely imperative language, with no objects, no functions... they are all convenient overheads that sacrifice performance for code clarity. I think we should make the same choice here.

vincentdavoust: I would actually be ok with changing `getTimelineObjectPositions` to a method of MainWindow, and declare it just above `findAllMarkerPositions`. Less nesting so better performance and geographically close to the caller so it's easy to read. But it's still uselessly exposed to the MainWindow scope, and it's also unnecessary for it to be a method, it does not need any access to the instance

ferdnyc:
> _ToolTip_ and _Text_ of the marker's action requires correction too, like:
> 
> ```
>    <property name="text">
>     <string>Previous Marker</string>
>    </property>
>    <property name="toolTip">
>     <string>Previous Marker</string>
>    </property>
> ```
> 
> to
> 
> ```
>    <property name="text">
>     <string>Previous Marker/Keyframe</string>
>    </property>
>    <property name="toolTip">
>     <string>Previous Marker/Keyframe</string>
>    </property>
> ```
> 
> and Next Maker in similar way.

Hmm. Debatable since even that isn't _entirely_ correct. Prev/Next Marker have always stopped and will continue to stop at clip edges, as well. They weren't _exclusively_ "move to marker" actions even before this change.

So I figure we can either continue to accept the use of "Marker" as an increasingly-broad term that refers to any and all positions where the playhead will stop when those actions are triggered (only _some_ of which are manually-placed ruler markers), **or** we should pick new, concise terminology that covers everything.

(Maybe even something completely different like "Seek backward" / "Seek forward".)

src/windows/main_window.py
ferdnyc: > If performance was the only important aspect, we would be using a purely imperative language, with no objects, no functions... 

This. If we were being miserly with every CPU cycle, for starters we'd _never_ write `try`/`except` blocks — the penalty for handling even a _single_ exception is **orders of magnitude** higher than the alternative ugly forest of conditionals it could be replaced with. But even though standard conditionals are much, _much_ faster, programmers accept the performance penalty for exception handling. As long as you avoid writing tight loops that have to handle thousands of exceptions per second, the additional cost is a relatively small price to pay.

>they are all convenient overheads that sacrifice performance for code clarity.

I wouldn't even say "sacrifice", since we're talking about a function that's only called in response to an explicit user action. The limiting factor on performance in those cases is reaction time — first the interface's, then the _user's_. As long as the UI remains fully responsive, performance is at its effective maximum level.

I mean, during each second you only get 60 chances to update the position of the playhead. Within each of those 60ths of a second, things can't possibly happen any faster no matter how many cycles you shave off the algorithm.

@SuslikV 
> I see no arguments in your words but only feelings.

They're not "feelings", they're opinions. Just like this is an opinion:

> Nested functions is hard to read. 

Not only do I disagree, I don't really understand the objection. Every class method is already a nested function, effectively. People somehow manage.

The simple fact is, I personally find local worker functions to be _more_readable than classes stuffed with utility methods that aren't really part of their API. The function is right there in the method where it's used, and it's only exposed to the one piece of code that's meant to be calling it. So we'll have to agree to disagree on this point, since readability is an opinion.

SuslikV: @ferdnyc So, why it is here?!! WHY??? For what purpose nesting is needed here? For what? I don't understand it!

SuslikV: @ferdnyc , It was wrong suggestion to use nesting... and you know it.

SuslikV:
> ...we should pick new, concise terminology that covers everything.

_Next Key Point_
_Previous Key Point_

Where _Key Points_ are: keyframes, clip boundaries, markers etc.

vincentdavoust:
> > ...we should pick new, concise terminology that covers everything.
> 
> _Next Key Point_
> _Previous Key Point_
> 
> Where _Key Points_ are: keyframes, clip boundaries, markers etc.

Sounds good to me. Updating it now

jonoomph:
@vincentdavoust This looks great so far! Is the PR in a good state for merging? Or are there still additional changes you want to make first? :+1:

jonoomph:
In general, this will be a welcome addition for many people who animate in OpenShot regularly. In the future though, we need much more control over the the keyframes, and the ability to visualize them, drag them, duplicate them, etc... But that is a problem for our future selves, lol. For now, this will do great!

vincentdavoust:
> @vincentdavoust This looks great so far! Is the PR in a good state for merging? Or are there still additional changes you want to make first? +1

IMO it's ready to merge !

jonoomph:
Thanks! LGTM!

PR #3413, opened 2020-04-27
Set locale for blender process stdout

ferdnyc:
The `blender` subprocess will run with its streams using the detected default encoding, which may be `'ascii'`. That will cause it to throw a UnicodeEncodeError if any Blender output is _non_-ascii (including file paths).

In Python 3.5+ we could pass `encoding='utf-8'` to `subprocess.Popen()`, but because the AppImage uses Python 3.4 we have to reconfigure the stream manually after it's opened.

Fixes #3408

PR #3416, opened 2020-04-28
Fix Blender subprocess pipe decoding

### LINKED ISSUE #3408: Openshotv2.5.1-dev2 blender-2.82a crash "can't use animated title"
**Describe the bug**
OpenShot-v2.5.1-dev2-1587624156-15c3efb7-26b7db94-x86_64.AppImage blender-2.82a-linux64 AnimatedTitle crash
Can't use blender for animated Titled with the last version of OpenShot
Generate button is greyed
I use the daily build of Openshot because the last stable build is not working on my debian machine.
I download the last blender binary (2.82a) from the official site because the version in the stable debian package manager is too old (2.79)
I reproduced the same bug on Ubuntu 18.04 with the same package.  

1. Select the blender binary in Preferences
2. Go to Title-> Animated Title
3. Select one of the title "snow effect for example"
4. The generate button is greyed

**Expected behavior**
Generation of my animated title with blender

**System Details**
- libopenshot Version:  0.2.5-dev2
- Platform: Linux-4.19.0-8-amd64-x86_64-with-debian-10.3
- Processor: 
- Machine: x86_64
- Python version: 3.4.3
- Qt5 version: 5.2.1
- PyQt5 version: 5.2.1
- Qt Detected Languages: ['fr-FR']
- LANG Environment Variable: fr_FR.UTF-8
- Daily Build Installed: http://github.com/OpenShot/openshot-qt/releases/download/OpenShot-v2.5.1-dev2-1587624156-15c3efb7-26b7db94-x86_64.AppImage

**Log Files**
- http://logs.openshot.org/k2bi7jrvpni9eqk02t470kscf4d5o9yt/openshot-qt.zip
- http://logs.openshot.org/k2bi7jrvpni9eqk02t470kscf4d5o9yt/libopenshot.zip

**Exception / Stacktrace**
```No stacktrace found in log files```

ferdnyc:
Second attempt at a PR that fixes #3408. 

My first attempt (#3413) tried to work around the fact that we can't set `encoding` for the `subprocess.PIPE` streams until Python 3.6, by using `.reconfigure()` on them after opening. Problem is, `.reconfigure()` isn't supported until Python **3.7**.

So, it's back to bytes streams for the subprocesses, because text mode in Python < 3.6 is always beholden to whatever's detected as the default locale — and in the case of the reporter's system from #3408, that turned out to be `'ascii'`. Now, we read as bytes and decode to `utf-8` unconditionally.

I also:
* Fixed `BlenderListView.__init__()`, which is subclassed from `QListView` but was calling `QTreeView.__init__()`! We need to start using `super()` more to avoid that kind of thing (as `BlenderListView.__init__()` now does).
* Made sure we aren't silently discarding errors from the Blender process runs, by logging and/or raising exceptions when warranted. (Something that would have revealed the problem with my #3413 attempt instantly, instead of making me hunt for it.)
* Streamlined the XML parsing heavily, and added an `unlink()` after processing to free up the memory consumed ([highly recommended](https://docs.python.org/3.8/library/xml.dom.minidom.html?highlight=xml%20dom%20minidom#xml.dom.minidom.Node.unlink) in the docs) — I'm going to check the rest of our XML-parsing routines now and do the same there, if needed.
* Cleaned up the `blender_listview.py` imports, getting rid of any `*` imports and anything unused
* Demoted a lot of unnecessary class variables in the render worker to be just local variables

### DIFF
--- a/src/windows/views/blender_listview.py
+++ b/src/windows/views/blender_listview.py
@@ -745,11 +745,12 @@ class Worker(QObject):
             # Run real command to render Blender project
             self.process = subprocess.Popen(command_render, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, startupinfo=startupinfo, universal_newlines=True)
 
-        except:
+        except Exception as ex:
             # Error running command.  Most likely the blender executable path in
             # the settings is incorrect, or is not a supported Blender version
             self.is_running = False
             self.blender_error_nodata.emit()
+            log.error("Could not execute Blender: {}".format(ex))
             return
 
         while self.is_running and self.process.poll() is None:
--- a/src/windows/views/blender_listview.py
+++ b/src/windows/views/blender_listview.py
@@ -363,61 +363,39 @@ class BlenderListView(QListView):
 
         # Get list of params
         animation = {"title": animation_title, "path": xml_path, "service": service, "params": []}
-        xml_params = xmldoc.getElementsByTagName("param")
 
         # Loop through params
-        for param in xml_params:
-            param_item = {}
+        for param in xmldoc.getElementsByTagName("param"):
+            # Set up item dict, "default" key is required
+            param_item = {"default": ""}
 
             # Get details of param
-            if param.attributes["title"]:
-                param_item["title"] = param.attributes["title"].value
+            for att in ["title", "description", "name", "type"]:
+                if param.attributes[att]:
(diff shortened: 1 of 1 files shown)


ferdnyc:
Reported as working  by @SebO-L in #3408 (and in my own testing), merging this to squash that bug.

PR #3419, opened 2020-04-28
Free xml.dom.mindom parsers when finished

ferdnyc:
As I mentioned in #3416, after doing this in `blender_listview.py` (part of that PR) I scanned over the rest of the code to find more places that could benefit from the same fix:
- A strongly recommended practice (see Python official docs) that we haven't been following, the objects returned by `xml.dom.minidom.parser()` should be `unlink()`ed as soon as they're no longer needed
- Also cleaned up some * imports, logging, formatting in affected files


jonoomph:
LGTM!

PR #3432, opened 2020-05-01
Update launch-win.bat

### LINKED ISSUE #3431: launch-win.bat doesn't work if you call it from a cmd/powershell prompt that's not in the OpenShot install directory.
**Describe the bug:**
When launching the `launch-win.bat` file from a terminal, it should launch OpenShot as expected.

1. open `cmd` or `powershell` terminal
2. launch `C:\Program Files\OpenShot Video Editor\launch-win.bat`
3. Receive error at the end: `'openshot-qt.exe' is not recognized as an internal or external command, operable program or batch file.`
4. There is also the side effect that the terminal window does not close if you launch it by double-clicking the file...

**Expected behavior:**
OpenShot should open regardless of where you launch it from.
Terminal window should close when OpenShot launches.

**System Details:**
- OpenShot Version [e.g. 2.4.3]: 2.5.1

corbob:
Fixes #3431 

- Add start command to last line
  - This continues execution of the script after launching the exe.
- Add %~dp0 to the launch of exe
  - This is batch shorthand for Drive and Path of argument 0 (file being called) equivalent to $PSScriptRoot in PowerShell

### DIFF
--- a/installer/launch-win.bat
+++ b/installer/launch-win.bat
@@ -12,4 +12,4 @@ SET QT_PLUGIN_PATH=%cd%
 SET QT_DEBUG_PLUGINS=1
 
 :: Launch application
-openshot-qt.exe
+start "" "%~dp0openshot-qt.exe"


corbob:
We'll just call this an issue where @corbob doesn't know what he's talking about 😬

Issue #449, opened 2016-12-29
Temporarily hide a track


Symbai:
In other video editors we can hide a track temporarily from the preview. Or we can "solo" a track so only this track appears in the preview.

This is very useful when cutting and using several tracks as otherwise I have to move a track outside of all others just to find the location where to cut (or move it at the top of alll tracks).

JtLively:
You can hide a track by selecting the **eye** icon on the left side of the timeline. For a specific section within the timeline, select **eye** icon on it underneath its individual title name or within its properties.

You can solo a track by right-clicking it from the timeline, selecting _Properties_, and the _Play_ button on the left.

dvogel:
@JtLively is this enhancement in 2.4.3? I don't see any eye icon [in the UI when I run it on my machine](https://i.imgur.com/yHccFZr.png). I also don't see an eye icon in [this screenshot from the documentation](https://www.openshot.org/static/files/user-guide/_images/tracks.jpg). When I right click a track, no menu appears. If I right click on a clip I do get a menu with a _Properties_ option but there is no _Play_ button. In that properties area I can either set _Enable Video_ to zero or I can set _Alpha_ to zero. Both settings only affect that clip and not the entire track. That can get pretty tedious with a track that has a few dozen cuts. Is there a way to mute or solo an entire track? When I search the repo for "hide" or "eye" or "solo" I don't get many obvious results. What am I misunderstanding?

qlerebours:
I can't see it either.
@JtLively are we missing something ?

ferdnyc:
@qlerebours @dvogel 

This feature is not currently part of OpenShot 2.x. @JtLively was describing features from the retired OpenShot 1.x interface/codebase.

Innitinnuit:
Wow, someone suggested this essential and basic feature 3 years ago and it hasn't been implemented? What? Thank god for the new emojis though. We all needed that.

yogifred:
Please re-add this feature!

NathanChristie:
Amazed this still hasn't been implemented. This would be very useful.

Colorjet3:
We hear you and it is in the queue.  Unfortunately no ETA!

seanwang0209:
The eye button really useful.

XavierBit:
Any progress on that topic? I mean, that's really an essential function....

Colorjet3:
Not at this time, unfortunately!

x59a:
One thing you can do to accomplish this:

[...]

PR #3438, opened 2020-05-02
Add controls to temporary disable Audio or Video per Track

SuslikV:
Fixes: https://github.com/OpenShot/openshot-qt/issues/449
...and many related requests.

Skipping also works for export.

OpenShot skip track menu


SuslikV:
Shifting clip one track up/down doesn't updates skipping yet. I will see if simple fix is possible.

SuslikV:
Clips can be moved from Track to Track now. Skipping streams will update accordingly.

SuslikV:

SuslikV:
Updated on top of the current _develop_ branch. Should work now.

SuslikV:

SuslikV:
Updated on top of the current _develop_.

jonoomph:
Closing PR. This could be accomplished with much less code, and probably icons on the track and clip are all we need. But this PR is too complex for my taste. This might be a good PR to model a future change after though, as some of it works fine.

PR #3447, opened 2020-05-04
Slice shortcuts

### LINKED ISSUE #1764: Keyboard shotcut to slice just the selected clip

ferdnyc:
This PR adds slight variations on the "Slice All: (mode)" actions, named "Slice Selected: (mode)", which just like "Slice All" use the playhead as the slice point. However, where "Slice All..." acts on all intersecting clips, "Slice Selected" slices _only_ the intersecting clips which are also currently selected.

As requested, "Slice Selected: Keep Both Sides" is bound to the <kbd>s</kbd> key by default (modifiable in the Preferences). For symmetry with the "Slice All" bindings, the other two modes are therefore bound to <kbd>a</kbd> and <kbd>d</kbd>.

These actions are currently keyboard-shortcut-only, they do not appear on any context menu. (TBH I'm just not sure where they _could_ go, having context menu choices that operate on the active selection — even multi-selection, possibly — _and_ also depend on the playhead positioning... it just feels like it could get pretty messy.)

Fixes #1764

### DIFF
--- a/src/settings/_default.settings
+++ b/src/settings/_default.settings
@@ -797,6 +797,30 @@
   },
   {
     "category": "Keyboard",
+    "title": "Slice Selected: Keep Both Sides",
+    "restart": true,
+    "setting": "sliceSelectedKeepBothSides",
+    "value": "s",
+    "type": "text"
+  },
+  {
+    "category": "Keyboard",
+    "title": "Slice Selected: Keep Left Side",
+    "restart": true,
+    "setting": "sliceSelectedKeepLeftSide",
+    "value": "d",
+    "type": "text"
+  },
+  {
+    "category": "Keyboard",
+    "title": "Slice Selected: Keep Right Side",
+    "restart": true,
+    "setting": "sliceSelectedKeepRightSide",
+    "value": "a",
+    "type": "text"
+  },
+  {
+    "category": "Keyboard",
     "title": "Copy",
     "restart": true,
     "setting": "copyAll",
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -1529,7 +1529,7 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
             if intersecting_clips or intersecting_trans:
                 # Get list of clip ids
                 clip_ids = [c.id for c in intersecting_clips if c.id in self.selected_clips]
-                trans_ids = [t.id for t in intersecting_trans if t.id in self.selected_transitons]
+                trans_ids = [t.id for t in intersecting_trans if t.id in self.selected_transitions]
                 self.timeline.Slice_Triggered(0, clip_ids, trans_ids, playhead_position)
         elif key.matches(self.getShortcutByName("sliceSelectedKeepLeftSide")) == QKeySequence.ExactMatch:
             intersecting_clips = Clip.filter(intersect=playhead_position)
@@ -1537,7 +1537,7 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
             if intersecting_clips or intersecting_trans:
                 # Get list of clip ids
                 clip_ids = [c.id for c in intersecting_clips if c.id in self.selected_clips]
-                trans_ids = [t.id for t in intersecting_trans if t.id in self.selected_transitons]
+                trans_ids = [t.id for t in intersecting_trans if t.id in self.selected_transitions]
                 self.timeline.Slice_Triggered(1, clip_ids, trans_ids, playhead_position)
         elif key.matches(self.getShortcutByName("sliceSelectedKeepRightSide")) == QKeySequence.ExactMatch:
             intersecting_clips = Clip.filter(intersect=playhead_position)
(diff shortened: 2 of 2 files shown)


SuslikV:
> not sure where they _could_ go...

Menu entries already exist - they should work for single selected file as for multiply selections. What's wrong here? It is just bug that it doesn't slices all selected files.
Of course, selections itself (like <kbd>Ctrl</kbd> + mouse) has additional PR (https://github.com/OpenShot/openshot-qt/pull/3317)

SuslikV:
...and what is:
```
self.selected_transitons
```
?

ferdnyc:
> Menu entries already exist - they should work for single selected file as for multiply selections. What's wrong here? It is just bug that it doesn't slices all selected files.

That's a fair point, the Align actions already do operate on multiple selected clips so there's some precedent there. Of course, whether that works correctly right now will depend on whether the selection list is accurate, but that's equally true of the keyboard shortcuts.

I wonder whether just changing the Clip context menu "Slice" to operate on multiple selected items would be OK, or if it risks screwing up users expecting the existing functionality. _Arguably_ it's just making Slice function in a way that's closer to what a typical user would expect. (After all, why would you select multiple clips if you only want to slice one of them?)

But if there _are_ users who regularly slice a single clip while having multiple selected, having come to expect that Slice works that way, then to avoid screwing up those users these new Slice options should have a separate context menu subsection that, like Align, only shows up when two or more items have been selected.

From a keyboard

> Of course, selections itself (like Ctrl + mouse) has additional PR (#3317)

Aha, that's right. I mentioned in #1764 that there might already be a fix pending for that.

> ...and what is:
> 
> ```
> self.selected_transitons
> ```
> 
> ?

...I guess, proof that I KAN SPEL? I'll fix that, thanks.

SuslikV:
> if there _are_ users who regularly slice a single clip while having multiple selected, having come to expect that Slice works that way...

Unfortunately, right now they are clicking twice and more times the same option just to complete the intended slicing task.

ferdnyc:
> Unfortunately, right now they are clicking twice and more times the same option just to complete the intended slicing task.

Depends how they select. "Draw-a-box" group selection works fine, and there are definitely some users who do all of their multi-selection that way.

SuslikV:
No, I mean slicing: click clip - slice one. Multiply selections doesn't works in any way (rectangle selections or not). Only _Slice All_ is doing its job right now, but that may slice clip at background...

ferdnyc:
@SuslikV Aha! Yeah, gotcha.

SuslikV:
Your code adapted to `timeline_webview.py` _ShowClipMenu()_ section, similar code should apply to the _ShowTransitionMenu()_
```
...
        # Get list of intercepting clips with position (if any)
        intersecting_clips = Clip.filter(intersect=playhead_position)
        intersecting_trans = Transition.filter(intersect=playhead_position)

        if intersecting_clips or intersecting_trans:
            # Get list of clip ids
            clip_ids = [c.id for c in intersecting_clips if c.id in get_app().window.selected_clips]
            trans_ids = [t.id for t in intersecting_trans if t.id in get_app().window.selected_transitions]

            # Add split clip menu
            Slice_Menu = QMenu(_("Slice"), self)
            Slice_Keep_Both = Slice_Menu.addAction(_("Keep Both Sides"))
            Slice_Keep_Both.setShortcut(QKeySequence(self.window.getShortcutByName("sliceSelectedKeepBothSides")))
            Slice_Keep_Both.triggered.connect(partial(self.Slice_Triggered, MENU_SLICE_KEEP_BOTH, clip_ids, trans_ids, playhead_position))
            Slice_Keep_Left = Slice_Menu.addAction(_("Keep Left Side"))
            Slice_Keep_Left.setShortcut(QKeySequence(self.window.getShortcutByName("sliceSelectedKeepLeftSide")))
            Slice_Keep_Left.triggered.connect(partial(self.Slice_Triggered, MENU_SLICE_KEEP_LEFT, clip_ids, trans_ids, playhead_position))
            Slice_Keep_Right = Slice_Menu.addAction(_("Keep Right Side"))
            #Slice_Keep_Right.setShortcut(QKeySequence(self.window.getShortcutByName("sliceSelectedKeepRightSide")))
            Slice_Keep_Right.triggered.connect(partial(self.Slice_Triggered, MENU_SLICE_KEEP_RIGHT, clip_ids, trans_ids, playhead_position))
            menu.addMenu(Slice_Menu)
...
```

SuslikV:
Razor Tool works for single item only. This may be leaved unchanged.

ferdnyc:
> Razor Tool works for single item only. This may be leaved unchanged.

Agreed, not even considering a trip down that road. If the user wants to slice multiple clips at once, they need to use the playhead and Slice All or Slice Selected.

In fact, nearly every function of the Razor Tool now has a viable-if-not-superior alternative using the playhead "razor" functions. If not for the tool's ability to quickly slice single <b>UN-</b>selected clips, I'd say there's no longer any reason for it to even exist.

But, it still is the quickest/best way to slice a single clip without having to worry about selection or playhead position, so I guess it lives to slice another day.

jonoomph:
> so I guess it lives to slice another day.

LOL :laughing:

jonoomph:
LGTM! Thanks for this one!

ferdnyc:
Hmm. We never resolved the menu-selection issue. I'm sort of leaning more towards @SuslikV 's suggestion that the Clip/Transition context-menu Slice functions become SliceSelected, replacing the current slice options. (If they only have one selected item, SliceSelected will be equivalent to the current single-item slice anyway. The only thing it'll affect is what happens when you activate Slice on a clip while you also have additional clips selected. And that just doesn't feel like a case we should be losing sleep over.)

...Still, I can slap that change into a separate PR, this works just fine for now in its current keyboard-only mode.

PR #3454, opened 2020-05-05
Use Github Actions to auto-generate user guide HTML

ferdnyc:
We're running the docs build at the end of each Travis run right now, which is fine for CI purposes, but not very convenient.

Using Github Actions, instead, lets the results of the run be exported as an Artifact from the job, so that the generated HTML tree can be downloaded as a ZIP file right from the Actions status page. Handy for documentation contributors, so they don't have to go through the hassle of setting up to run Sphinx on their own PCs.

A tweak to `src/classes/info.py` makes it handle PyQt5 being unavailable (as it will be, in the Actions environment), and the `requirements.txt` file added to `doc/` is a standard config defining the Python module dependencies for our Sphinx config. The Actions task will ensure that all dependencies are met before starting the build.

### DIFF
--- a/.github/workflows/sphinx.yml
+++ b/.github/workflows/sphinx.yml
@@ -1,0 +1,32 @@
+# Github Actions workflow to generate documentation
+# Uses the following shared task definitions:
+# - (checkout, upload artifact) from Github
+# - sphinx-action maintained by @ammaraskar
+name: Sphinx build
+
+# Controls when the action will run.
+# Triggers the workflow on push or pull request events.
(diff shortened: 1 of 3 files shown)


ferdnyc:
> We're running the docs build at the end of each Travis run right now,

Actually, I take it back, I guess we're _not_ doing that! Well, even better then.

ferdnyc:
Merging as this is primarily new configs, and the `classes.info` change just wraps a `try:` block around part of the code.

PR #3463, opened 2020-05-08
WIP: Proposal for meta documentation

MBB232:
Proposal for description of how documentation files can be structured + template (based on existing documentation)
meta-art like numbers, arrows and fonts for documentation pictures can be saved to this directory,


AboutDoc/Documentation_About.rst
```diff
@@ -0,0 +1,250 @@
...
+
+- Copyright notice
+- Openshot description
+- Openshot disclaimer
+- License notice
+- Referral anchor for the title
+
+Followed by the title with double underlining (with == )
+Then a short description of what the chapter will be about, and why the reader should care (or skip). 
+Subtitles are single underlined ( -- )
```
ferdnyc, line 129:
```suggestion
The content starts with a chapter title, double-underlined using equals signs (``=====``).
This is followed by a short introduction describing what will be covered in the chapter.

A chapter may be divided into sections and subsections, each beginning with an underlined heading.
Like the chapter title, section headings are double-underlined using equals signs.
Subsection headings are single-underlined using hyphens (``----``).

.. code-block:: ReST

    Chapter title
    =============

    Introduction paragraph.
    
    Section heading
    ===============
    
    Subsection heading
    ------------------
```

AboutDoc/Documentation_About.rst
```diff
@@ -0,0 +1,250 @@
...
+.. OpenShot Video Editor is free software:
+ you can redistribute it and/or modify it under the terms of the GNU General Public License as published by  the Free Software Foundation, 
+ either version 3 of the License,
+ or (at your option) any later version.
+
+.. OpenShot Video Editor is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ See the GNU General Public License for more details.
+
+.. You should have received a copy of the GNU General Public License
```
ferdnyc, line 15: The standard heading shouldn't be reformatted, the copyright years can be updated but other than that it shouldn't really be messed with as it's in part a license statement. Consistency with other files / history is more important than consistency with the rest of the file's formatting.

```suggestion
.. Copyright (c) 2008-2020 OpenShot Studios, LLC
 (http://www.openshotstudios.com). This file is part of
 OpenShot Video Editor (http://www.openshot.org), an open-source project
 dedicated to delivering high quality video editing and animation solutions
 to the world.

.. OpenShot Video Editor is free software: you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
 the Free Software Foundation, either version 3 of the License, or
 (at your option) any later version.

.. OpenShot Video Editor is distributed in the hope that it will be useful,
 but WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 GNU General Public License for more details.

.. You should have received a copy of the GNU General Public License
 along with OpenShot Library.  If not, see <http://www.gnu.org/licenses/>.
```

AboutDoc/Documentation_About.rst
```diff
@@ -0,0 +1,250 @@
...
+
+File structure
+--------------
+
+Every file starts with 5 hidden segments (See `template <Template.rst>`_ )
+
+- Copyright notice
+- Openshot description
+- Openshot disclaimer
+- License notice
```
ferdnyc, line 124: These are really all just one four-paragraph block, our standard header. (It serves as all of those things, but they shouldn't be considered separate things.)

AboutDoc/Documentation_About.rst
```diff
@@ -0,0 +1,250 @@
...
+or before starting an inline markup command.
+It is a guideline, not a rule.
+This tends to be a pretty good fit for any sort of written prose, when it is in a markup language like ReST and managed in a version control system.
+
+There are four reasons for this:
+
+- Writing that way, there is no worrying about line length or when to wrap. 
+- It encourages shorter, simpler sentences which is a good thing when writing docs. 
+- The diffs when changes are submitted also tend to be more readable and focused. 
+- Lines are easier to translate and less likely to be changed.
```
ferdnyc, line 143: Again, nix all mentions of translation please. Especially since this one also isn't true, or at least doesn't matter. When Sphinx outputs translation files, it flows sentence-per-line blocks back together and exports the entire paragraph as a single translation string.

AboutDoc/Documentation_About.rst
```diff
@@ -0,0 +1,250 @@
...
+There may be translation notes hidden in the documentation, blocked out with \.. TRANSLATION NOTE: 
+
+Files for translation will be hosted at `Launchpad <https://translations.launchpad.net/openshot/2.0/+translations>`_.
+
+When translating numbers referencing a screenshot in non-western languages, please make sure to update the screenshot too. 
+If available, images of the translation should be saved in their subdirectory *(to be decided)* 
+
+.. TODO: Add subdirectory
+
+.. TRANSLATION NOTE: After translating tables, make sure that the underlining of table rows stay the same length as the new words.
```
ferdnyc, line 162: Nope to all of this, too. It should probably go. :disappointed:

MBB232: >Drawing a distinction between "the online HTML and the offline manual" is a bit dubious here, since they're exactly the same HTML. A copy of what is otherwise the "offline manual" gets uploaded to the openshot.org site after it's generated by the build servers. (So, certainly the sentence as-is isn't wrong... other than the capitalization.)

Hey, that is on you, for explaining Sphinx too well. 

"Plus, WYSIWYG to some extent implies a particular layout, whereas docs can be generated from reST using a variety of styles and formats, so the same docs will look very different depending how it's processed."
If it _could_, I thought it _would_. So I presumed that the offline doc would be PDF. (And possibly Epub, Winhelp/CHM/MShelp2, RTF, Writer and whatnot versions could exist.)

If they do not exist, then that line can indeed be shortened. 

>> Or in *this* reddit thread. 
> Ah. Maybe remove the previous line, and jut make the TODO about adding Reddit info?

**It's a proposal that I did not feel up to me to decide. **
**It also depends on if topics on Reddit can be sticky to the top, how often you expect changes to the Help be needed, and if you have someone to regularly check it. **
I thought it could be a low-level way for people to make small suggestions. But if it is on top it may get abused, and if it is not on top it will be unable to find. 
Since the meta-documentation is on github, it is probably fair to asume that the readers have an account. But if there are other places that suggestions for the docs are made, they should be listed so you can check them to see if something needs to be added. 

>I don't know that it's really worth belaboring all the details here. 

(Re license info) I agree that it could be phrased shorter, but I was intentionally *going* for completeness of the details. 
In five years the docs may need another large overhaul and someone is bound to ask; 'why is this not under CC license?' (Like I did). And you may not be here to explain. 

>The 'question' tag is actually auto-applied to any issue opened using the Question

Every 'question' is a question, but not all questions are the same. 
Questions that get asked a lot should be in the FAQ, even (especially) if the answere is no, the asker does not stay around or it is already in the docs. 

Questions with a good/complex answer may (partially) be used to update the docs or be in the FAQ, even if they are also an enhancement or bug. 

So **yes, I am proposing the creation of a FAQ-tag and more rigorous use of the Doc-tag.** (Every time a new feature is added, it should be documented or screenshots may need updating. 
For example, I am waiting for https://github.com/OpenShot/openshot-qt/pull/3346 to make it into the dailys so the framenumber will be included in the new screenshots.

Reddit also seems to support tags, though I have not figured out how yet. If you have a regular Reddit-watcher, (s)he may want to add tag too. 

> Again, nix all mentions of translation please.

Yes, you mentioned that at the end of the other topic, was leaving it in for completion. (And maybe someone has an alternative)
Disappointing that it won't work in Sphinx. However, shorter lines may remain relevant if you process the Rest source files in other translation programs. OTOH, translating entire paragraphs can have benefits as some languages prefer longer sentences. 
So I propose to hide it, and leave it till a process for translation is decided upon.

ferdnyc:
In addition to the comments on specific details that I left above, a couple of general questions/notes:

1. I notice that in this PR, `AboutDoc` is in the root of the project, rather than inside the `doc/` directory. Any reason for that change? I'm not sure about adding another directory to the project root, inside the `doc/` direcory seemed like the right place to me.
1. In fact if we wanted to, we could rename/repurpose this slightly into the first chapter of a broader overall "Contributors Handbook" or "Contributors Guidelines". Something that could potentially cover more than just the documentation (things like coding style, testing methodology, etc.) Then it could live at `doc/Guidelines/` or `doc/Handbook/` or something like that.
    1. (I'm mostly bringing this up because it would affect how various settings would need to be defined, in things like the sphinx config file or the Actions task to auto-generate the doc.)
1. As an alternative to ad-hoc comments in the documentation, Sphinx has a plugin that enables support for a more formal work-in-progress annotation via a `todo` directive. And it has a couple of nice advantages:
    1. Ther's an accompanying `todolist` directive that will insert a collected listing of _all_ of the `todo`s across the entire set of documentation, displaying each note along with a link back to its location in the original context.
    1. `todo` directives have _configurable_ visibility — if they're configured non-visible, they act like any other comment and don't show up in the rendered output.
     1. But if they're configured _visible_ when generating the docs, then each one will be visible as an admonition box. Which opens up the possibility of even using them for notes inside the User Guide content as well. We'd normally configure it so they're hidden, but it then gives us the option to generate "editorial versions" of the Guide with all of the `todo` notes visible _and_ indexed via `todolist`.

To embed a `todo`, it's written just like any other directive:
```rst
.. todo::

    Some note that should normally be invisible, but can be made visible.

.. todo:: For short notes, the entire thing can be on one line.

```
(The directive doesn't _have_ to be all-lowercase, though it's more typical. But any capitalization is technically legal.)

And then (after converting an existing TODO comment into a `todo` directive and enabling the plugin), when the docs are configured for visible todos you get this:

image

AboutDoc/Documentation_About.rst
ferdnyc: You can reply to individual comments directly, in context, BTW. Instead of combining them all together. It helps a lot, in keeping track of the various points raised; each of the individual conversations can be marked "Resolved" when that item has been dealt with.

AboutDoc/Documentation_About.rst
```diff
@@ -0,0 +1,250 @@
...
+
+License
+-------
+We use the GPL-3 license (see above) for documentation, see the header.
+This is for simplicity because it is the same license as the project code.
+And because the documentation gets parsed in other tools before it reaches its final form.
+
+Github
+------
+In the issue tracker, subjects that contain explanations that should probably be included in the documentation can be labeled ** `docs. <https://github.com/OpenShot/openshot-qt/labels/docs>`_ **
```
ferdnyc, line 47: According to [the docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#gotchas) nested inline markup is simply not possible. There's no way to bold a hyperlink, or anything of that nature.
```suggestion
In the issue tracker, subjects that contain explanations that should probably be included in the documentation can be labeled `docs <https://github.com/OpenShot/openshot-qt/labels/docs>`_\ .
```

MBB232:
> I notice that in this PR, `AboutDoc` is in the root of the project, rather than inside the `doc/` directory. Any reason for that change?

Github. It must have jumped ba k to the root again when creating the new branch or when I switched or something. It was **intended** to be a subdirectoy of docs, as then it would be easily found by people wanting to edit the docs. 
Can the location be easily fixed during the merge, or does it need to be re-pulled again?

ferdnyc:
> Can the location be easily fixed during the merge, or does it need to be re-pulled again? 

It's easy enough to fix, we can just add a commit to the PR that moves the directory. (Renames all of the files inside it, effectively.)

ferdnyc:
@MBB232: You asked, elsewhere,

> I noticed a 'commit suggestion' button under some of your comments. Is that to accept them into my request?
>  Should I click on that to add them for the ones I agree with?

Yeah, exactly. Any you want to incorporate into the PR, you can do that right from the suggestion.

In fact, when suggestions are related (particularly in the same file), you can use "Add to batch" to queue them up and apply them all as a single commit.

MBB232:
test. Comment keeps disappearing

MBB232:
Added most other suggestions in new branch, not sure what would happen if I made the changes directly in my pull request branch. Although I saw that your commits were included in it now. 
 https://github.com/MBB232/openshot-qt/blob/MBB232-AboutDoc-patch-1/doc/AboutDoc/Documentation_About.rst

jonoomph:
@MBB232 You can keep committing and pushing to this PR branch. It won't break anything. Worst case, it might mark some comments as no longer relevant, but it doesn't delete any history on this PR. Go for it!

ferdnyc:
> Should the copyright notice in the template be redone in the old format too? As the headers are hidden, they are not processed by Spinx so the single-line does not really matter.

If it's going to serve as a template, then yeah. It should contain the header as we'd want it to be used in any new files, which would be an exact copy of the header from any of the existing files (possibly with the year(s) modified).

> Added most other suggestions in new branch, not sure what would happen if I made the changes directly in my pull request branch.

You should definitely make the changes on your `MBB232-AboutDoc` branch, so they're incorporated into the "living" PR. And as you noticed, you'll also see changes committed from our end too, either automatically by GitHub (as it synchronizes with the target branch, `develop`) or manually, like the changes I just pushed to fix some sphinx configuration.

ferdnyc:
@MBB232 I pushed in the lion's share of my CSS changes, as well as some documentation changes that make use of them... which I'm now realizing I probably shouldn't have, since you mentioned wanting to keep those separate. So I'll probably back those out.

Do you have an in-progress branch for changes to the User Guide files? If you open a PR from it I can push the changes there, instead.

jonoomph:
We have made some major changes to our documentation recently, and I am closing this PR until conflicts and merges can be resolved.

PR #3464, opened 2020-05-08
Fix Qt warning message. UI code cleanup

SuslikV:
1. Fixes "QMainWindow::addDockWidget: invalid 'area' argument" message from Qt.
In _QMainWindow_ any dock should has valid _dockWidgetArea_ when added (even if areas where the dock widget may be placed excludes everything).
Message itself noticeable when running `openshot-qt-cli` application.

2. Removing doubles of some properties.


ferdnyc:
:+1:

Issue #3444, opened 2020-05-03
missing file even in case of good path


snoopy8612:
I have created a openshot video file but when I try to open it, there is always the same error " file misisng". I haven't changed anything on the path of the file. I have tried to delete ".openshot_qt" and opened again the file but still the same error message. Something is strange because the name of the files (from the error message) are not the same than in the folder (I changed nothing when I imported them, for example "museliereNoodle" in the folder became "u00e8reNoodle" in the error message). 

I really don't understand what is happening, do you have an idea? 

I am sorry, I hope this will be clear (I am french...)

Thanks for your help because I am really not an IT expert...

**System Details:**
OpenShot-v2.5.1-x86_64 

![image](https://user-images.githubusercontent.com/64739124/80916910-e0b94500-8d5b-11ea-8246-40a49708cbfd.png)

SuslikV:
This is project file saved in v2.5.0 or similar build, where non-English letters in paths not saved correctly. It may be caused by auto-saving enabled in _Preferences_ of OpenShot - I recommend to disable it. When it enabled, it's just enough to open the project file once in the v2.5.0 to get it damaged...
See for details: https://github.com/OpenShot/openshot-qt/issues/3227

Try to reinstall the 2.5.1 application, start new project and make backups manually time to time.

ferdnyc:
@snoopy8612 I agree this is clearly caused by the 2.5.0 pathname issues, though it's disappointing that the repair code built into OpenShot 2.5.1 wasn't able to correct the problems. (It's _supposed_ to fix bad names automatically, though of course nothing is foolproof.)

It's also strange that "museliereNoodle" would be corrupted, as it doesn't contain any non-ASCII characters. Are you sure the filename wasn't "muselièreNoodle"?

Regardless, if you have significant work invested in the project and don't want to lose it, repairing your project file "by hand" is likely to succeed where the automated repair code failed. If you'd like me to try, just create a ZIP file containing the OpenShot `.osp` project file and attach it to a comment on this report, and I'll take a look.

Only projects that were saved from OpenShot 2.5.0 are affected by this issue, fortunately, and for the most part they seem to be detected and fixed by the auto-repair code. Once the data is repaired (manually, in the case of your project), there shouldn't be any problems working with it using OpenShot 2.5.1.

snoopy8612:
thanks a lot @ferdnyc fir your help

[...]

PR #3479, opened 2020-05-11
json_data: Fix path_regex to support all valid JSON files

ferdnyc:
See #3444.


SuslikV:
I vote for the
```
path_regex = re.compile(r'"(image|path)"\s*:\s*"([^"]+)"')
```
instead of proposed changes. It takes less steps and skips empty path strings (second group is different).

jonoomph:
LGTM

PR #3481, opened 2020-05-11
files views: Fix updateSelection()

ferdnyc:
Boneheaded bug in my #3312 code.

### DIFF
--- a/src/windows/views/files_listview.py
+++ b/src/windows/views/files_listview.py
@@ -118,7 +118,6 @@ class FilesListView(QListView):
         # Start drag operation
         drag = QDrag(self)
         drag.setMimeData(self.files_model.model.mimeData(self.selectionModel().selectedIndexes()))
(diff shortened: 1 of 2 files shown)


jonoomph:
@ferdnyc Some of this will likely be reverted with my changes to models and view (in the Emoji branch). It's difficult to reconcile the changes, so I just wanted to give you a heads up, that we might need to "re-implement" a few of these changes you've made to the file model and views after I'm done merging things. Or better yet, feel free to merge the relevant changes into the `emoji` branch, instead of `develop`. Sorry for the confusion!

ferdnyc:
@jonoomph Totally understand, and with improved files model code hopefully more of the view stuff becomes unnecessary anyway. (I've always thought the amount of duplicated code between the two views feels like a sign of things that really belong either in the model code, or somewhere else entirely. Selection-handling logic, in particular, can hopefully live with a selection model that's shared between the two views.)

I'll take a quick look now at the current `emoji` HEAD and see what's different from `develop`, figure out how/if the changes here and in the other PRs fit in.

Issue #3485, opened 2020-05-12
Fullscreen does not remember if OpenShot window was maximized


MBB232:
**Describe the bug:**
When using the  button, to go and return from fullscreen, the window size drops down to medium (AKA  RestoreDown) size regardless if it was maximized before. 

**Priority**
Low. Just discovered the  fullscreen modus, and found it odd. (May suitable as a 'beginner issue?' )

**Steps to reproduce the behavior:**
0.  Start with Openshot as loose window (drag it somewhere)
1. Go to 'maximize window' (right top corner) [See window fill screen but contain title bar]
2. Click on 'Fullscreen' (Openshot bar) [Openshot fills screen without title bars
3. Click on 'Fullscreen' (Openshot bar)' [Openshot moves back to window with title bars]
4. See loose window 

**Expected behavior:**
See window maximized again, as it started out before it went fullscreen. 
So it should remember if it was maximized or not when going fullscreen

**System Details:**
- OpenShot Version   2.5.1
- Windows 10,

**Log Files:** 
I can add them if they would help? 

**Screenshots:** 

<img width="840" alt="1-Window" src="https://user-images.githubusercontent.com/62191951/81753792-dbbf6880-94b4-11ea-94b6-ac4a23132632.png">
<img width="840" alt="2 Maximize" src="https://user-images.githubusercontent.com/62191951/81753796-dcf09580-94b4-11ea-82ae-14165364284a.png">
<img width="840" alt="3 Fullscreen" src="https://user-images.githubusercontent.com/62191951/81753797-dcf09580-94b4-11ea-92e0-dd60b2fb441a.png">
<img width="840" alt="4-ExitFullscreen_RestoreDown" src="https://user-images.githubusercontent.com/62191951/81753798-dcf09580-94b4-11ea-9d70-dafabb2334e9.png">

ferdnyc:
> **Log Files:**
> I can add them if they would help?

No need.

> (May suitable as a 'beginner issue?' )

Agreed.

(For anyone considering working on this): Qt windows have separate properties for `maximized` and `fullscreen` (among other states). And `maximized` will stay **_`true`_** even after calling `showFullscreen()`, if the window was previously maximized. So, that's a restore-logic path. Buuut, those properties are read-only, so it's not a restore _interface_ per se.

[...]

PR #3488, opened 2020-05-14
Remember and restore window state on mode switch (full screen)

SuslikV:
Fixes: https://github.com/OpenShot/openshot-qt/issues/3485

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -1874,9 +1874,10 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
     def actionFullscreen_trigger(self, event):
         # Toggle fullscreen mode
         if not self.isFullScreen():
+            self.last_window_state = self.windowState()
             self.showFullScreen()
         else:
-            self.showNormal()
+            self.setWindowState(self.last_window_state & ~Qt.WindowFullScreen)
 
     def actionFile_Properties_trigger(self, event):
         log.info("Show file properties")
@@ -2781,6 +2782,9 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
 
         # Save settings
         s.save()
+
+        # Last window state to restore from full screen
+        self.last_window_state = self.windowState()
 
         # Refresh frame
         QTimer.singleShot(100, self.refreshFrameSignal.emit)


src/windows/main_window.py
```diff
@@ -2782,6 +2783,9 @@ def __init__(self, mode=None):
         # Save settings
         s.save()
 
+        # Last window state to restore from full screen
+        self.last_window_state = self.windowState()
+
```
ferdnyc, line 2788: The problem with storing this in the application is, it can get out of sync with reality pretty easily. Unless you record *every* state change that can occur in the application both while it's fullscreened and while it's not.
```suggestion
```

src/windows/main_window.py
```diff
@@ -1874,9 +1874,10 @@ def actionTimelineZoomOut_trigger(self, event):
     def actionFullscreen_trigger(self, event):
         # Toggle fullscreen mode
         if not self.isFullScreen():
+            self.last_window_state = self.windowState()
             self.showFullScreen()
         else:
-            self.showNormal()
+            self.setWindowState(self.last_window_state & ~Qt.WindowFullScreen)
```
ferdnyc, line 1880: But that's OK, because the _entire_ `actionFullScreen_trigger()` method can be replaced with [Qt's suggested](https://doc.qt.io/qt-5/qwidget.html#setWindowState) XOR logic:
```python
def actionFullScreen_trigger(self, checked=False):
    # Toggle fullscreen state (current state mask XOR WindowFullScreen)
    self.setWindowState(self.windowState() ^ Qt.WindowFullScreen)
```

Written that way, it really does work as a fullscreen toggle — any other state parameters, like being maximized or snapped to a window border or whatever — are preserved, and only the fullscreen bit is flipped each time.

(I can't submit a proper 'suggestion' if it needs to change any lines outside of the original PR diff; annoying.)

src/windows/main_window.py
SuslikV: It intended to store only window state before going into fullscreen (to return normally). No need to store other states, because they all saved in settings already.

SuslikV:
@ferdnyc has better solution.

ferdnyc:
I was kind of hoping you'd just update your PR, @SuslikV ... but that's cool, I just submitted one.

PR #3497, opened 2020-05-17
Overhaul of models, views, data management

### LINKED ISSUE #1288: Issue : Can not add pictures files on project if picture's title is a number ?

### LINKED ISSUE #2031: openshot wont update image sequence video name and wont allow a 2nd image sequence video
Hello everyone, i discovered a small bug tonight, i made an image sequence into a video perfectly
and i was able to rename the image sequence video in the project video however.
It will not update on the time line video the new name.

Also once you have completed the image sequence video, if you attempt to bring a new image sequence
openshot wont allow you do this , even after deleting the image sequence individual png's, and
wanting to start a new image sequence video . 

All openshot does is to import the pngs only and 
does not offer to import as image sequence.

heres a picture to show what im talking about
cheers bob

image sequence wont update name

### LINKED ISSUE #2341: can't import image sequence starting with 0326.png
win10, made in blender. recent version, no idea which one.  try it with the file 0326 (top image)

### LINKED ISSUE #2557: Importing sequence of images takes too much time
Openshot Daily Build
Windows 10 64 Bits i5 / SSD 8Go RAM
Importing png images as a sequence takes a long time ( 250 image takes almost 5 mn)

### LINKED ISSUE #3372: problems for importing image sequences. The program stops runnig. (Give your bug report a descriptive title)
**Describe the bug**
problems for importing image sequences. The program stops runnig.

**Expected behavior**

**System Details**
- libopenshot Version:  0.2.5-dev2
- Platform: Windows-10-10.0.18362
- Processor: Intel64 Family 6 Model 142 Stepping 10, GenuineIntel
- Machine: AMD64
- Python version: 3.7.4
- Qt5 version: 5.12.4
- PyQt5 version: 5.12.3
- LANG Environment Variable: 
- Daily Build Installed: http://github.com/OpenShot/openshot-qt/releases/download/OpenShot-v2.5.1-dev2-1586173838-166f6938-232f3f79-x86_64.exe

**Log Files**
- http://logs.openshot.org/4owqsersrok58evrks0ud3kwtcrsv6ko/openshot-qt.zip
- http://logs.openshot.org/4owqsersrok58evrks0ud3kwtcrsv6ko/libopenshot.zip

**Exception / Stacktrace**
```No stacktrace found in log files```

**Screenshots** (Optional)
If applicable, add screenshots to help explain your problem. You can include screenshots by
copy/pasting them on GitHub or dragging-and-dropping into the GitHub page. All images are public,
so please don't post screenshots containing personal information.

### PR DESCRIPTION
**Pull request by ferdnyc**
As invited by @jonoomph (careful what you wish for), this PR against the in-progress `emojis` branch builds on the model/view changes there, extending some nice performance improvements into a fairly extensive overhaul of the entire way data is managed for project asset lists.

### TODO for merge
- [x] remove the `add_file()` implementations from `title_editor.py`, `animated_title.py`, that won't work anymore

There are probably some bugs, still, but I've finally got it to a point where it doesn't keel over as soon as I start testing it, so I wanted to get it out there in the world. The major benefits are:

#### Speed

I just imported a set of (what I subsequently discovered was) 776 image files, mostly PNGs, into a new project as Project Media. It wasn't _instantaneous_, but it took... maybe a minute? A bit more? The way the OpenShot code through 2.5.1 implemented that sort of thing, it would've quickly ground to a halt and taken possibly an hour or more, with each individual file causing a several-minutes-long rebuild of the full model on each addition. With the new code, not only did the import run at ~ 10 files/second right through to the end, but I was able to _delete_ a random 7 of them instantaneously.

#### Reduced <strike>data</strike> code duplication

`add_file()` and `get_image_sequence_data()` or whatever have moved out of the two view classes and into the `files_model`. Not only does that elmiinate duplication, but it allows for some strreamlining — for example, the method is now called `add_files()`, and it takes an entire list of files (like from a File Open dialog) instead of having to be spun over them one-by-one

### Improved user experience

My focus was also on eliminating some long-standing pain points for the users

#### Selection is now preserved in all lists, <strike>always</strike> almost always

There's a persistent `QItemSelectionModel` that's created with each **MODEL** class, and applied to both of the views (shared between them) when they're set up. You can switch modes and etc.,  you still won't lose your selected items.

(I notice that selections _are_ lost if the selected items are filtered out of the list, then restored to visibility again. So, that's a bug that I'd like to fix, if possible.)

#### Image Sequence import is less brain-dead

One of the advantages to `add_files()` is that it can be smarter about groups of files, when discovering for image sequences. Now, when you import a group of images that could be a sequence:
1. OpenShot will prompt you ONCE to import that sequence
1. If you say No, the code tracks which import files fall under the same sequence pattern, and won't ask you again about any of the rest of them.
1. If you say _yes_, the code now **removes** the rest of the matching images from your import list, so that if you happen to try to import a 30-image sequence by selecting all 30 images (which users do **all the time**), you'll end up with only one imported image sequence item, with the other 29 images automatically and _silently_ excluded.

There's a bunch more stuff, too — see the commit messages, and beyond that, the code.

I'm _hopeful_ that, among many others, this...
fixes #3372, fixes #2557, fixes #2341, fixes #2031, fixes #1288

### DIFF
--- a/src/windows/models/files_model.py
+++ b/src/windows/models/files_model.py
@@ -27,41 +27,26 @@
  """
 
 import os
+import json
+import re
+import glob
 
-from PyQt5.QtCore import QMimeData, Qt, pyqtSignal, QSortFilterProxyModel, QEventLoop
-from PyQt5.QtGui import *
-from PyQt5.QtWidgets import QMessageBox
-import openshot  # Python module for libopenshot (required video editing module installed separately)
-
+from PyQt5.QtCore import (
+    QMimeData, Qt, pyqtSignal, QEventLoop, QObject,
+    QSortFilterProxyModel, QItemSelectionModel, QPersistentModelIndex,
+)
+from PyQt5.QtGui import (
+    QIcon, QStandardItem, QStandardItemModel
+)
 from classes import updates
 from classes import info
+from classes.image_types import is_image
 from classes.query import File
 from classes.logger import log
 from classes.app import get_app
 from requests import get
 
-import json
-
-class FileStandardItemModel(QStandardItemModel):
-    ModelRefreshed = pyqtSignal()
-
-    def __init__(self, parent=None):
-        QStandardItemModel.__init__(self)
-
-    def mimeData(self, indexes):
-        # Create MimeData for drag operation
-        data = QMimeData()
-
-        # Get list of all selected file ids
-        files = []
-        for item in indexes:
-            selected_row = self.itemFromIndex(item).row()
-            files.append(self.item(selected_row, 5).text())
-        data.setText(json.dumps(files))
-        data.setHtml("clip")
-
-        # Return Mimedata
-        return data
+import openshot
 
 
 class FileFilterProxyModel(QSortFilterProxyModel):
@@ -98,10 +83,32 @@ class FileFilterProxyModel(QSortFilterProxyModel):
             return self.filterRegExp().indexIn(file_name) >= 0 or self.filterRegExp().indexIn(tags) >= 0
 
         # Continue running built-in parent filter logic
-        return super(FileFilterProxyModel, self).filterAcceptsRow(sourceRow, sourceParent)
+        return super().filterAcceptsRow(sourceRow, sourceParent)
+
+    def mimeData(self, indexes):
+        # Create MimeData for drag operation
+        data = QMimeData()
+
+        # Get list of all selected file ids
+        ids = self.parent.selected_file_ids()
+        data.setText(json.dumps(ids))
+        data.setHtml("clip")
+
+        # Return Mimedata
+        return data
+
+    def __init__(self, **kwargs):
+        if "parent" in kwargs:
+            self.parent = kwargs["parent"]
+            kwargs.pop("parent")
+
+        # Call base class implementation
+        super().__init__(**kwargs)
 
 
-class FilesModel(updates.UpdateInterface):
+class FilesModel(QObject, updates.UpdateInterface):
+    ModelRefreshed = pyqtSignal()
+
     # This method is invoked by the UpdateManager each time a change happens (i.e UpdateInterface)
     def changed(self, action):
 
@@ -246,7 +253,7 @@ class FilesModel(updates.UpdateInterface):
         self.ignore_updates = False
 
         # Emit signal when model is updated
-        self.model.ModelRefreshed.emit()
+        self.ModelRefreshed.emit()
 
     def get_thumb_path(self, file_id, thumbnail_frame, clear_cache=False):
         """Get thumbnail path by invoking HTTP thumbnail request"""
@@ -287,31 +294,39 @@ class FilesModel(updates.UpdateInterface):
                     item = self.model.itemFromIndex(thumb_index)
                     item.setIcon(QIcon(thumb_path))
                     item.setText(name)
+            # Emit signal when model is updated
+            self.ModelRefreshed.emit()
 
-                    # Emit signal when model is updated
-                    self.model.ModelRefreshed.emit()
-                    break
         self.ignore_updates = False
 
     def __init__(self, *args):
 
-        # Add self as listener to project data updates (undo/redo, as well as normal actions handled within this class all update the tree model)
+        # Add self as listener to project data updates
+        # (undo/redo, as well as normal actions handled within this class all update the model)
         app = get_app()
         app.updates.add_listener(self)
 
         # Create standard model
-        self.model = FileStandardItemModel()
+        self.model = QStandardItemModel()
         self.model.setColumnCount(6)
         self.model_ids = {}
         self.ignore_updates = False
 
+        self.ignore_image_sequence_paths = []
+
         # Create proxy model (for sorting and filtering)
-        self.proxy_model = FileFilterProxyModel()
+        self.proxy_model = FileFilterProxyModel(parent=self)
         self.proxy_model.setDynamicSortFilter(True)
         self.proxy_model.setFilterCaseSensitivity(Qt.CaseInsensitive)
         self.proxy_model.setSortCaseSensitivity(Qt.CaseSensitive)
         self.proxy_model.setSourceModel(self.model)
         self.proxy_model.setSortLocaleAware(True)
 
+        # Create selection model to share between views
+        self.selection_model = QItemSelectionModel(self.proxy_model)
+
         # Connect signal
         app.window.FileUpdated.connect(self.update_file_thumbnail)
+
+        # Call init for superclass QObject
+        super(QObject, FilesModel).__init__(self, *args)
--- a/src/windows/models/files_model.py
+++ b/src/windows/models/files_model.py
@@ -249,7 +249,7 @@ class FilesModel(QObject, updates.UpdateInterface):
         # Emit signal when model is updated
         self.ModelRefreshed.emit()
 
-    def add_files(self, files):
+    def add_files(self, files, image_seq_details=None):
         # Access translations
         app = get_app()
         _ = app._tr
@@ -268,11 +268,11 @@ class FilesModel(QObject, updates.UpdateInterface):
             if file:
                 return
 
-            # Load filepath in libopenshot clip object (which will try multiple readers to open it)
-            clip = openshot.Clip(filepath)
-
-            # Get the JSON for the clip's internal reader
             try:
+                # Load filepath in libopenshot clip object (which will try multiple readers to open it)
+                clip = openshot.Clip(filepath)
+
+                # Get the JSON for the clip's internal reader
                 reader = clip.Reader()
                 file_data = json.loads(reader.Json())
 
@@ -291,8 +291,11 @@ class FilesModel(QObject, updates.UpdateInterface):
                 file = File()
                 file.data = file_data
 
-                # Is this file an image sequence / animation?
-                image_seq_details = self.get_image_sequence_details(filepath)
+                if not image_seq_details:
+                    # Try to discover image sequence, if not supplied
+                    image_seq_details = self.get_image_sequence_details(filepath)
+
+                # Is this an image sequence / animation?
                 if image_seq_details:
                     # Update file with correct path
                     folder_path = image_seq_details["folder_path"]
@@ -401,7 +404,6 @@ class FilesModel(QObject, updates.UpdateInterface):
 
             # Yes, import image sequence
             parameters = {
-                "file_path": file_path,
                 "folder_path": dirName,
                 "base_name": base_name,
                 "fixlen": fixlen,
@@ -451,8 +453,8 @@ class FilesModel(QObject, updates.UpdateInterface):
                 return
 
             # Update thumb for file
-            thumb_index = id_index.sibling(id_index.row(), 0, Qt.DecorationRole)
             thumb_path = self.get_thumb_path(file_id, 1, clear_cache=True)
+            thumb_index = id_index.sibling(id_index.row(), 0)
             item = m.itemFromIndex(thumb_index)
             item.setIcon(QIcon(thumb_path))
             item.setText(name)
--- a/src/windows/models/files_model.py
+++ b/src/windows/models/files_model.py
@@ -60,14 +60,14 @@ class FileFilterProxyModel(QSortFilterProxyModel):
                 or get_app().window.filesFilter.text():
             # Fetch the file name
             index = self.sourceModel().index(sourceRow, 0, sourceParent)
-            file_name = self.sourceModel().data(index) # file name (i.e. MyVideo.mp4)
+            file_name = self.sourceModel().data(index)  # file name (i.e. MyVideo.mp4)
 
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

**Comment by ferdnyc** [comment]
> ### There are probably some bugs, still,

(I just wanted to reiterate that in big, flashing letters on the side of a mountain, so to speak.)

This is a _big_ set of changes, and I'm not going to pretend that I think every line is bug-free, nor will I lie and say that I've personally tested every possible scenario myself before submitting. I haven't. Not even remotely. That's _why_ I opened this PR, to get other eyes on the code / other testing experiences, so hopefully the remaining bugs can be shaken out.)

**Comment by jonoomph** [comment]
@ferdnyc Wow! Nice work on this! I'll grab this branch and do some testing today for sure! I'll let you know what I find shortly.

**Comment by ferdnyc** [comment]
If there's one thing that makes me the most nervous, it's the multiple inheritance in `files_model.py` that makes `FilesModel` derive from _both_ `updates.UpdateInterface` _and_ `QObject`.

But it was necessary to transfer the  `ModelRefreshed` signal there (the other model classes now singly-inherit from `QObject` for the same reason), and I really think it's the right way to go. By just moving that signal,  along with promoting the `mimeData()` implementation to live in the outermost proxy class instead (which is where it really needs to be called by the view), I was able to do away with all of the unnecessary chasing of proxy class `ModelIndex`es back to the source model (when the proxy already has all of the data we need), and I eliminated the custom base model classes _**entirely**_ for both Files and Transitions.

Swapping those inner model classes over to just the stock `QStandardItemModel` made me feel like I was on the right track; our data models are fairly uncomplicated (as they should be), no reason they can't store what they need Qt's pre-built model classes.

I still want to make another pass through the table structure for the base models, in fact, and experiment with defining some User Roles so that more data can be moved from hidden columns over to the first ("main") cell in the row — like I did with the Emoji model.

Unless we're either: (1) displaying, or (2) sorting by, a piece of data, there's no reason for it to have a separate column of its own — the lookups get _far_ simpler if it's all stored in the same table cell (under different Roles). Then it can all be accessed using the same `ModelIndex`, without having to do a lot of `row()` / `sibling()` lateral lookups.

**Comment by ferdnyc** [comment]
> If there's one thing that makes me the most nervous, it's the multiple inheritance in `files_model.py` that makes `FilesModel` derive from _both_ `updates.UpdateInterface` _and_ `QObject`.
> 
> But it was necessary to transfer the `ModelRefreshed` signal there

...And it seems to work just fine, is the real point. I said it makes me _nervous_, not that it causes any actual problems. Even though I kept half-expecting to hit some. But despite my misgivings, it all seems to work out as it should.

(Though, that new Qt aspect to the outer `FilesModel` custom class **IS** the reason I moved the `QMessageBox`-displaying pieces out of its `add_files()` and `get_image_sequence_info()` methods, when I moved them from the view code. The simplest approach I could think of was to make them little helper methods provided by `main_window`, instead — so that's what I did. Qt didn't really seem too keen on a `QObject` subclass popping up gui elements all by its lonesome. And it wouldn't have made any sense to make `FilesModel` a `QWidget`-derived class.)

**Comment by SuslikV** [comment]
@ferdnyc don't forget that proxy models can be sorted by name, date etc.

**Comment by jonoomph** [comment]
Testing Feedback (so far):
 * Error on Effects Details View (dragging an effect onto a clip, but only drag from the non-icon columns): 
  ```
main_window:INFO Switch to Details View
  File "/home/jonathan/Downloads/openshot-qt-emojis-mods2/src/windows/views/effects_treeview.py", line 70, in startDrag
    drag.setPixmap(icon.pixmap(QSize(self.drag_item_size, self.drag_item_size)))
AttributeError: 'NoneType' object has no attribute 'pixma
  ```

**Comment by jonoomph** [comment]
`Edit Title` breaks (on both Details and Thumbnails) views with the following error:
```
 main_window:INFO title editor add confirmed
  File "/home/jonathan/Downloads/openshot-qt-emojis-mods2/src/windows/main_window.py", line 403, in actionEditTitle_trigger
    index = self.filesView.indexAt(event.pos())
AttributeError: 'bool' object has no attribute 'pos'
```

File: src/windows/models/files_model.py [hunk 1]
```diff
@@ -277,41 +442,84 @@ def update_file_thumbnail(self, file_id):
...
+        return files
+
+    def current_file_id(self):
+        """ Get the file ID of the current files-view item, or the first selection """
+        cur = self.selection_model.currentIndex()
+
+        if not cur or not cur.isValid() and self.selection_model.hasSelection():
+            cur = self.selection_model.selectedIndexes()[0]
+
+        if cur and cur.isValid():
```
jonoomph, line 486: @ferdnyc What if this doesn't find a valid ID, and thus does not return a value?

src/windows/models/files_model.py
```diff
@@ -277,41 +442,84 @@ def update_file_thumbnail(self, file_id):
...
+            return cur.sibling(cur.row(), 5).data()
+
+    def current_file(self):
+        """ Get the File object for the current files-view item, or the first selection """
+        cur_id = self.current_file_id()
```
jonoomph, line 491: Same question, what if this doesn't get a value returned? And what if current_file() doesn't return a File object?

SuslikV:
disadvantages:
1. Progress of loading isn't visible. UI remains is in "not responding" state until task complete.
2. There are two identical FilesModel added to listeners that maintains separately.

All 3 published solutions that modifies files model: https://github.com/OpenShot/openshot-qt/pull/3497, https://github.com/OpenShot/openshot-qt/pull/3366, https://github.com/OpenShot/openshot-qt/pull/3264 has bottle-neck in thumbnails creation. This is most resource demanding task. Only one solution uses asynchronous update of generated thumbnails (total time required to draw all new icons is the same). Other two waits for thumbnails during model creation.

ferdnyc:
> @ferdnyc don't forget that proxy models can be sorted by name, date etc. Indexes wouldn't be preserved. Proxy needed only for sorting, nothing else. Unsorted is also an option.


Well, exactly. That's why everything has to be looked up on the fly, by pulling up the `row()` of an index in the proxy model and querying for its `sibling()` columns for other data. The row numbers will change as things are sorted and filtered, but e.g. the ID of a given file will always be in column 5 of its `row()` — in any of the models. (Assuming that row exists at all, since filtered proxy models won't contain all possible rows.)

But it would be even easier if, when we got the `ModelIndex` for a file entry, the ID lived at that same index under a User Role. Then, getting it is just a matter of `index.data(self.ID_ROLE)`.

(The reason I store a  `QPersistentModelIndex` on the base model at row creation is, that index is guaranteed to remain valid throughout the lifetime of the row. So, later on we can go back and use it to quickly find and delete the same row. Propagation of changes to the base model is automatically enabled for proxy models.)

ferdnyc:
> disadvantages:
> 1. Progress of loading isn't visible. UI remains in "not responding" state until task complete.

Agreed, and while it's much faster now (and loading 766 PNG files isn't something a sane person should be doing anyway), you are left wondering what's happening. Some sort of progress would be good — even if it's just that the Project Files list keeps itself scrolled down to the last item added, so you can watch it populating.

> 2. There are two identical FilesModel added to listeners that maintains separately.

Good catch, I missed/forgot that properties was creating its own files model. It needs to not.

IIRC **both** the Title Editor and the Animated Title Editor have their own `add_file` implementations (I think my fix for that in the blender view was the one I had to cancel?), for this to work they'll need to switch to using `add_files()` in the model. I think that explains the error @jonoomph saw with the title editor.

> 3. When emoji drag canceled - new file object still appears in the Project Files list.

I noticed that one too, however with the new persistent model indexing it should be trivially fixable. It just needs to request deletion of the new entry on drag abort.

ferdnyc:
I also feel like "Add to Timeline" (which right now constructs its own model out of the selected files) should be able to work directly off the `SelectionModel` for the main model, instead. That'll make its code a lot simpler, it won't need a model implementation of its own at all. It'll just be another type of view on the central files model.

ferdnyc:
> `Edit Title` breaks (on both Details and Thumbnails) views with the following error:
> ```
>  main_window:INFO title editor add confirmed
>   exceptions:ERROR Unhandled Exception
> Traceback (most recent call last):
>   File "/home/jonathan/Downloads/openshot-qt-emojis-mods2/src/windows/main_window.py", line 403, in actionEditTitle_trigger
>     index = self.filesView.indexAt(event.pos())
> AttributeError: 'bool' object has no attribute 'pos'
> ```

Oh! No, that's my fault. I know how to fix it, but it ultimately comes down to an issue with our action triggers: All of the functions are specced wrong. Every single one.

They all have the form `foo_triggered(self, event)`. But there's no `event` arg to a `QAction:: triggered`. The `event` on all of those functions should be `checked`, instead. That's why it's a Boolean, and why you can't get the `pos()` of a ` triggered ()`. (In an event handler, like the context menu event of a view, you can use `indexAt()` to look up the index of the item under the mouse pointer.)

I'll fix the title editor, and fix all of our triggered functions to `foo_triggered(self, checked)` as well.

ferdnyc:
> and fix all of our triggered functions to `foo_triggered(self, checked)` as well.

... Maybe AFTER the emoji branch is merged. If I do it here, it'll make it even harder to merge with `develop`. If I do it on `develop`, ditto.

ferdnyc:
The two commits I just pushed should fix both Edit Title, and dragging from non-icon columns for all views. (Had _that_ one half-fixed in my source tree, turns out; just forgot to commit it! And I decided to change all of the view files to use the same code; cleaner — even if it is still copy-pastey.)

src/windows/models/files_model.py
ferdnyc: Up to the callers. if it's being called in an assignment context, (`f = self.current_file_id()`), then `f` will be `None`. I generally test it with an `if not f:` after.

If it's being passed to some other function (`do_something_with(self.current_file_id())`, then that function needs to be able to deal with a possible `None` argument.

ferdnyc:
I was correct in thinking that both title editors had their own `add_file()` implementations; those are going to have to come out. So I've set this WIP and added an outstanding TODO item for at _least_ that change.

ferdnyc:
### Properties table migrated

I just finished getting the Properties table off of its own `FilesModel()` and `TransitionsModel()` copies, and over to using the shared `self.win.files_model` / `self.win.transition_model`. A connection to both models' `ModelRefreshed` signal takes the place of the previous `update_model()` calls, and means the context menus will only refresh when _necessary_.

### QAbstractItemModelTester
The other commit I just pushed is a new `launch.py` command-line option (`--test-models`) which, when enabled, will link Qt's [QAbstractItemModelTester](https://doc.qt.io/qt-5/qabstractitemmodeltester.html) onto each of the base and proxy models of the Files, Emoji, Effects, and Transitions lists. This will only work with Qt 5.11 or higher, and will silently do nothing useful with earlier versions. The model tester automatically monitors model transactions for access violations, unsafe index use, etc.

By default, if the environment variable `QT_LOGGING_RULES` is **not** already set, when `--test-models` is used the launcher will set it to `"qt.modeltest.debug=true"` which causes _quite_ a lot of console output. (Multiple lines for every single transaction on every single model, as well as the proxies they bubble to.) So it's advisable to apply some less-verbose value, like `QT_LOGGING_RULES="qt.modeltest.warning=true"`, before launching OpenShot with `--test-models`.

ferdnyc:
Ohh-kay. I now have both title editors using `add_files()` from the files model, instead of their own implementations.

Along the way, I changed the naming pattern for title SVG files from "TitleFileName-N.svg" (where N is a counter) to the very Windows-ey "TitleFileName (N).svg". This is for extremely pragmatic reasons: With the previous naming, once you had more than one title `add_files()` would offer to import them as an image sequence :rofl: 

And I incorporated _some_ of my changes from #3357 (but most of it is still on hold)

ferdnyc:
@jonoomph At this point I believe I've got things here working well enough that I'd suggest we merge this to at least the _`emojis` branch_ quickly — as soon as you're comfortable with it, basically. 

There are still a few more things I should do, and even more things I'd _like_ to do, but they're relatively small and can just as easily be done directly on that branch.

The `emojis` branch is going to be a bear to merge into `develop` no matter what... probably best not to let this two-stage craziness live on any longer than we have to.

jonoomph:
@ferdnyc I've been merging `develop` into `emojis` quite frequently, to minimize the crazy. Okay, let me take a quick look at this one again! Thanks!

PR #3507, opened 2020-05-22
translations: Use OpenShot_locale.qm file names

### LINKED ISSUE #418: Nomenclature of .qm translation files
See https://bugzilla.rpmfusion.org/show_bug.cgi?id=4358
and https://wiki.qt.io/How_to_create_a_multi_language_application#File_system_structure_of_the_application

The .qm files are not presented in a standard nomenclature, therefore it's difficult to pick them using standard tools.

ferdnyc:
While names of the form <code>OpenShot.<var>locale</var>.qm</code> are supported by Qt, many packaging tools expect translations to use _ separators in the file naming, so our filenames cause some difficulties.

This PR renames all `OpenShot.$LANG_$LOC.qm` files to `Openshot_$LANG_$LOC.qm`, and makes the necessary changes to `launch.py`, `classes.info`, `classes.language`, the `openshot_lang.qrc` file, and etc. to support the updated names.

(External scripts — like anything that manages the Launchpad translations — may also have to be corrected, though. If it's not part of this repo, I haven't fixed it.)

We've been using a custom `find_lang.sh` script [in the RPM Fusion packaging](https://bugzilla.rpmfusion.org/show_bug.cgi?id=4358) for several years now, due to this naming issue. Issue #418 was the initial report of that problem, and while we worked around it, the naming was something I failed to completely correct in #1795.

Fixes #418 (which should give you an idea of just _how_ old this issue is)


jonoomph:
LGTM! I'm fixing the launchpad translation system right now, so it will produce the correct *.qrc file, and the correct file naming. I'll test this out in a moment.

PR #3511, opened 2020-05-25
transitions model: Look up groups via sibling()

ferdnyc:
(And respect a `sortRole()` if defined for the model.)

### DIFF
--- a/src/windows/models/transition_model.py
+++ b/src/windows/models/transition_model.py
@@ -65,10 +65,10 @@ class TransitionFilterProxyModel(QSortFilterProxyModel):
 
     def lessThan(self, left, right):
         """Sort with both group name and transition name"""
-        leftData = self.sourceModel().data(left)
-        leftGroup = self.sourceModel().data(left, 257)  # Strange way to access 'type' column in model
-        rightData = self.sourceModel().data(right)
-        rightGroup = self.sourceModel().data(right, 257)
+        leftData = left.data(self.sortRole())  # Transition name (or other role data)
+        rightData = right.data(self.sortRole())
+        leftGroup = left.sibling(left.row(), 2).data()  # group column
+        rightGroup = right.sibling(right.row(), 2).data()
 
         return leftGroup < rightGroup and leftData < rightData
 


ferdnyc:
@jonoomph Hey, looks like the travis-ci.com configs came online — now it's doing _both_. (And since the `travis-ci.org` builds are still marked as required, but still not coming through, that probably needs to be shut off.)

jonoomph:
Working on it

ferdnyc:
@jonoomph Sent you a couple of email ideas, too.

src/windows/models/transition_model.py
```diff
@@ -65,10 +65,10 @@ def filterAcceptsRow(self, sourceRow, sourceParent):
 
     def lessThan(self, left, right):
         """Sort with both group name and transition name"""
-        leftData = self.sourceModel().data(left)
-        leftGroup = self.sourceModel().data(left, 257)  # Strange way to access 'type' column in model
-        rightData = self.sourceModel().data(right)
-        rightGroup = self.sourceModel().data(right, 257)
+        leftData = left.data(self.sortRole())  # Transition name (or other role data)
+        rightData = right.data(self.sortRole())
+        leftGroup = left.sibling(left.row(), 2).data()  # group column
```
jonoomph, line 70: 2 is not as cool as 257, but otherwise, looks good!

jonoomph:
I unchecked the "Branch" status check for now, so this can be merged. I'm not clear on if the "Branch" status is really needed. Also, I don't have the "Pull Request" status on our other 2 repos (it's just not there). Hmmm...

Screenshot from 2020-05-25 18-21-44

src/windows/models/transition_model.py
ferdnyc: I mean, if it was `42` I'd _absolutely_ have kept it.

ferdnyc:
@jonoomph 
>  Also, I don't have the "Pull Request" status on our other 2 repos (it's just not there). Hmmm...

Probably, the other two repos haven't gotten a pull request from a non-repo branch since switching Traviseseses, since it claims to list only the checks that it's seen in the past week. I bet it wouldn't have shown up here, either, if nobody'd opened a PR from a fork recently.

It sounds like "Branch" might not be necessary, as long as "Pull Request" checking is enabled on all of the repos. Any commits to (project repo) branches that _aren't_ the subject of a PR don't really need to be checked — they'll get checked when the PR is opened.

In the past Travis has always been double-scanning PRs from project repo branches, since it would end up running _both_ checks at once on the PR branch. Hopefully the Newer, Dot-Commier Travis® will put a stop to that.

PR #3514, opened 2020-05-26
Fix background color of the Preview widget

SuslikV:
The widget itself occupies whole Preview window and for bright UI themes (and for any) it should be painted into the window color first. Later, the painter will fill the actual video area.

Before (16:9 preview. Theme: _No Theme_):

OpenShot Preview background color Before

After (16:9 preview. Theme: _No Theme_):

OpenShot Preview background color After

### DIFF
--- a/src/windows/video_widget.py
+++ b/src/windows/video_widget.py
@@ -81,7 +81,7 @@ class VideoWidget(QWidget, updates.UpdateInterface):
         painter.setRenderHints(QPainter.Antialiasing | QPainter.SmoothPixmapTransform | QPainter.TextAntialiasing, True)
 
         # Fill the whole widget with the solid color
-        painter.fillRect(event.rect(), QColor("#191919"))
+        painter.fillRect(event.rect(), self.palette().window())
 
         # Find centered viewport
         viewport_rect = self.centeredViewport(self.width(), self.height())


SuslikV:
Too many of "the" was in the description.

ferdnyc:
Makes sense to me, though before I merge I'm just curious if @jonoomph had some reason for changing it, during the transform updates? @jonoomph ?

src/windows/video_widget.py
```diff
@@ -81,7 +81,7 @@ def paintEvent(self, event, *args):
         painter.setRenderHints(QPainter.Antialiasing | QPainter.SmoothPixmapTransform | QPainter.TextAntialiasing, True)
 
         # Fill the whole widget with the solid color
-        painter.fillRect(event.rect(), QColor("#191919"))
+        painter.fillRect(event.rect(), self.palette().window())
```
ferdnyc, line 84: I'm guessing the change was intentional — wanting to show the widget area better in the context of the zooming feature, probably. But even in that case, there's still a palette color that should be used:
```suggestion
        painter.fillRect(event.rect(), self.palette().base())
```
That'll use the same background color as the View widgets: `rgb(53, 53, 53)` for the dark theme, and `rgb(255, 255, 255)` for the light theme.

SuslikV: > and `rgb(255, 255, 255)` for the light theme...

LOL @ferdnyc  this an eyesore color. Are you really suggesting this change?

SuslikV: Window is window. There is nothing to reinvent.

jonoomph:
Closing PR, happy with current color. It's slightly lighter than black, to show the bounds of the actual video vs the preview widget itself. Makes aligning things during transform much easier.

PR #3516, opened 2020-05-26
Clear Recent Projects menu

jonoomph:
Adding "Clear Recent Projects" to recent project list, and a new "No Recent Projects" menu (and updating translations).

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -2296,7 +2296,7 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         # Add recent projects to menu
         # Show just a placeholder menu, if we have no recent projects list
         if not recent_projects:
-            self.recent_menu.addAction(_("No Recent Projects"))
+            self.recent_menu.addAction(_("No Recent Projects")).setDisabled(True)
             return
 
         for file_path in reversed(recent_projects):
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -2294,12 +2294,21 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
             self.recent_menu.clear()
 
         # Add recent projects to menu
-        for file_path in reversed(recent_projects):
-            new_action = self.recent_menu.addAction(file_path)
-            new_action.triggered.connect(functools.partial(self.recent_project_clicked, file_path))
+        if recent_projects:
+            for file_path in reversed(recent_projects):
+                # Add each recent project
+                new_action = self.recent_menu.addAction(file_path)
+                new_action.triggered.connect(functools.partial(self.recent_project_clicked, file_path))
 
-    # Remove a project from the Recent menu if OpenShot can't find it
+            # Add 'Clear Recent Projects' menu to bottom of list
+            self.recent_menu.addSeparator()
+            self.recent_menu.addAction(self.actionClearRecents)
+            self.actionClearRecents.triggered.connect(self.clear_recents_clicked)
+        else:
+            self.recent_menu.addAction(_("No Recent Projects"))
+
     def remove_recent_project(self, file_path):
+        """Remove a project from the Recent menu if OpenShot can't find it"""
         s = settings.get_settings()
         recent_projects = s.get("recent_projects")
         if file_path in recent_projects:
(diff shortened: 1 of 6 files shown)


src/windows/main_window.py
```diff
@@ -2294,12 +2294,21 @@ def load_recent_menu(self):
...
+                # Add each recent project
+                new_action = self.recent_menu.addAction(file_path)
+                new_action.triggered.connect(functools.partial(self.recent_project_clicked, file_path))
+
+            # Add 'Clear Recent Projects' menu to bottom of list
+            self.recent_menu.addSeparator()
+            self.recent_menu.addAction(self.actionClearRecents)
+            self.actionClearRecents.triggered.connect(self.clear_recents_clicked)
+        else:
+            self.recent_menu.addAction(_("No Recent Projects"))
```
ferdnyc, line 2308: Just in the interest of keeping the indentation down...
```suggestion
        # Show just a placeholder menu, if we have no recent projects list
        if not recent_projects:
            self.recent_menu.addAction(_("No Recent Projects"))
            return

        for file_path in reversed(recent_projects):
            # Add each recent project
            new_action = self.recent_menu.addAction(file_path)
            new_action.triggered.connect(functools.partial(self.recent_project_clicked, file_path))

        # Add 'Clear Recent Projects' menu to bottom of list
        self.recent_menu.addSeparator()
        self.recent_menu.addAction(self.actionClearRecents)
        self.actionClearRecents.triggered.connect(self.clear_recents_clicked)
```

ferdnyc: OK, I guess maybe it's only 9 lines, _technically_. :roll_eyes:

src/windows/ui/main-window.ui
```diff
@@ -1655,6 +1655,18 @@
     <string>Export XML (Final Cut Pro)</string>
    </property>
   </action>
+  <action name="actionClearRecents">
+   <property name="icon">
+    <iconset theme="edit-clear" resource="../../images/openshot.qrc">
+     <normaloff>:/icons/Humanity/actions/16/edit-clear.svg</normaloff>:/icons/Humanity/actions/16/edit-clear.svg</iconset>
```
ferdnyc, line 1661: Based on the icon used in Fedora (Adwaita theme), I'd say `edit-clear` is probably not a universally ideal choice.

Screenshot from 2020-05-27 11-04-20


Looking at the icon set, either `edit-clear-all` or `edit-delete` would be preferable:

image

ferdnyc: Hmm. Looks like Humanity doesn't have an `edit-clear-all`. So I guess `edit-delete` wins by default?

ferdnyc: (Humanity's `edit-delete` is this, which is "fine":)

edit-delete

ferdnyc:
@jonoomph 

Oh, hey, if you're going to be "sneaking in" :wink: translation updates, could you take a look at #3507 ? I covered all of the in-repo bases, I _believe_, but I'm reluctant to merge it without knowing whether it affects your processes external to the repo (launchpad and whatnot).

src/windows/main_window.py
```diff
@@ -2294,12 +2294,23 @@ def load_recent_menu(self):
             self.recent_menu.clear()
 
         # Add recent projects to menu
+        # Show just a placeholder menu, if we have no recent projects list
+        if not recent_projects:
+            self.recent_menu.addAction(_("No Recent Projects"))
```
ferdnyc, line 2299: I was also thinking that this should be a _disabled_ menu item, maybe? (I have to remember how to do that...)

jonoomph: I've added no handler for this action, so it just does nothing when clicked. We could disable it also, but I'm not sure it matters, IMO. :smile:

jonoomph:
This PR was made in response to a user who contacted me, concerned that he thought the **number** of recent projects listed was unusable. It gave him anxiety with each new project that appeared in the recent list. So, in his computer inexperience, he figured out that deleting his *.osp project files would also remove them from the `Recent` list. But of course, now all his projects are deleted, so that's no good. :grimacing:  Anyway, although I disagree with 11 projects being unusable, I kind of liked the idea of giving users an easy way to clear the list.

src/windows/main_window.py
ferdnyc, line 2299: Oh, right.
```suggestion
            self.recent_menu.addAction(_("No Recent Projects")).setDisabled(True)
```

ferdnyc: I tried it, it looks a bit cleaner disabled IMHO. It's a visual thing, disabled it's _clearly_ not meant to be clicked.

jonoomph: Thanks

ferdnyc: (Sounds like it has that user's name written all over it, too.) :wink:

PR #1726, opened 2018-06-23
WIL*: New tutorial system, better child window management

ferdnyc:
#### * WIL == "Work In Limbo"

At the moment the status of this is "on hold indefinitely", until I can revisit its status more concretely, and figure out a plan for revisiting the PR _itself_.

#### Previous Update

To my very great surprise, to be perfectly honest, initial tests of this code in Win7 (after development on Linux) seemed to indicate that it requires no changes to function equally well on that platform. This is something I was genuinely not expecting, and apparently means that the code may be more finished than I thought. So, in honor of that surprise development, I'm creating this PR to open the code up to wider review and testing.

The set of commits in this PR do two basic things:
1. Re-work the OpenShot main window's communication with its child windows
1. Completely restructure the way tutorial popups are created and managed

You can see the individual commits for further details, but the main takeaways are:
1. There is a new Tutorial dock widget defined in `main-window.ui`, initially hidden and permanently floating and non-dockable.
1. All Tutorial popups are now displayed inside of that (same) dock widget, which has all of its decoration removed so that it is completely invisible (aside from the tutorial popup itself).
1. Any move/resize events on the main window are communicated to the Tutorial dock, and it updates its position accordingly (just as before), but because it's a QDockWidget, Qt makes sure it stays on top of the main window. Also, because it's a dock, not a window, it doesn't appear in the window list.
1. Show/hide events from the main window are now propagated to **ALL** floating dock widgets (not _just_ the tutorial, though this does apply to the tutorial as well). Which means that if you minimize OpenShot, floating docks will minimize with it, and they'll return when you restore it. No more "the tutorial disappeared when I minimized OpenShot, but Properties didn't, and when I restored OpenShot the tutorial was gone" type issues.
1. The expensive event filters on both the main window and the tutorial widget are gone, as they're no longer needed.
1. Calls to `show()` on both the main window and the tutorial widget are now tightly controlled to avoid unnecessary window updates and repaints, which I am very hopeful will alleviate the flickering and unresponsiveness seen by some users (see #1080, #1264, #1555, #1623, etc.)

There are some unresolved issues/questions, some of them stemming from this change, some not.

[...]

PR #3517, opened 2020-05-26
Adjust position of tutorial message (for smaller screens)

jonoomph:
Adjust position of tutorial message for smaller screens, or when OpenShot is moved past the edge of the screen. For example:
Screenshot from 2020-05-26 18-40-02

### DIFF
--- a/src/windows/views/tutorial.py
+++ b/src/windows/views/tutorial.py
@@ -77,7 +77,30 @@ class TutorialDialog(QWidget):
             # Disable metric sending
             s.set("send_metrics", False)
 
-    def __init__(self, id, text, arrow, *args):
+    def keyPressEvent(self, event):
+        """ Process key press events and match with known shortcuts"""
+        # Detect the current KeySequence pressed (including modifier keys)
+        key_value = event.key()
+        modifiers = int(event.modifiers())
+
+        # Abort handling if the key sequence is invalid
+        if (key_value <= 0 or key_value in
+           [Qt.Key_Shift, Qt.Key_Alt, Qt.Key_Control, Qt.Key_Meta]):
+            return
+
+        # A valid keysequence was detected
+        key = QKeySequence(modifiers + key_value)
+        if key == Qt.Key_Escape:
+            # User hit Escape key, hide tutorial permanently
+            self.manager.hide_tips(self.id, True)
+
+    def mouseReleaseEvent(self, event):
+        """Process click events on tutorial. Especially useful when tutorial messages are partially
+        obscured or offscreen (i.e. on small screens). Just click any part of the tutorial, and it will
+        move on to the next one."""
+        self.manager.next_tip(self.id)
+
+    def __init__(self, id, text, arrow, manager, *args):
         # Invoke parent init
         QWidget.__init__(self, *args)
 
@@ -88,6 +111,7 @@ class TutorialDialog(QWidget):
         # Keep track of widget to position next to
         self.id = id
         self.arrow = arrow
+        self.manager = manager
 
         # Create vertical box
         vbox = QVBoxLayout()
@@ -99,6 +123,7 @@ class TutorialDialog(QWidget):
         self.label.setTextFormat(Qt.RichText)
         self.label.setWordWrap(True)
         self.label.setStyleSheet("margin-left: 20px;")
+        self.label.setAttribute(Qt.WA_TransparentForMouseEvents)
         vbox.addWidget(self.label)
 
         # Add error and anonymous metrics checkbox (for ID=0) tooltip
@@ -138,6 +163,7 @@ class TutorialDialog(QWidget):
         self.setCursor(Qt.ArrowCursor)
         self.setMinimumWidth(350)
         self.setMinimumHeight(100)
+        self.setFocusPolicy(Qt.ClickFocus)
 
         # Make transparent
         self.setAttribute(Qt.WA_NoSystemBackground, True)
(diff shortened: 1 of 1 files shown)


jonoomph:
@ferdnyc When dragging the screen, we continue to call `re_position_dialog()`, which continues to call `self.dock.move(QPoint(x, y))`. For some reason (possibly calling move(), when Qt is already moving the widget), we get a very jumpy effect (flickering from where Qt thinks the widget should be moved, and where we want to put it). Try this PR out, and drag OpenShot off the bottom of the screen (or the right side of the screen) when the tutorial is visible. You'll see what I mean. :+1: 

Just curious if you know any magic way to solve this. Ultimately, I am trying to prevent the tutorials from appearing off-screen, even if it messes up our positining.

ferdnyc:
@jonoomph 
> You'll see what I mean. 

I'm afraid I don't, though. That's been part of the problem all along — I was _aware_ that there are positioning issues on Windows, but they appear to be unique to Windows. And some other Linux users have reported issues with flickering and the like, but those appear to be isolated cases from people using somewhat "fringe" distros / desktops.

Here's a screencast running the code from this PR, in GNOME Shell 3.36. The current `develop` code behaves identically. What little glitching you _do_ see is attributable to OBS' screen capture, and wasn't present live.

https://youtu.be/VaFXjQj25o8

> Just curious if you know any magic way to solve this. Ultimately, I am trying to prevent the tutorials from appearing off-screen, even if it messes up our positining.

When this [came up in the past](https://github.com/OpenShot/openshot-qt/pull/1726#issuecomment-403429846), my take was that the simplest solution would just be to give the users control:
> Also, there's an issue reported with tutorial windows being placed partially offscreen for some users (which I hadn't seen, because Gnome Shell prohibits applications from locating windows outside the visible display area). That's equally an issue with the old code, but still isn't fixed in the new code.
> 
> As a result, I think it would be best to make the tutorial popups draggable. There's no reason to prevent the user from moving the tutorial popups, if they CHOOSE to. And it would prevent more "lockout" problems where users can't advance the tutorial because the "Next Tutorial" button is inaccessible off the bottom of their screen.

I urged caution about any sort of smart placement, because (a) _decent_ window managers already handle that, and it would suck to break things for the rest of us in an attempt to unbreak Windows' shitty positioning logic; and (b) there will always be more corner cases.

Heck, we once had an issue (can't find it now, of course — I hate you, Github issues search) where a user couldn't see the entire contents of the profiles dropdown in the Choose Profile dialog, because it was larger than his display. Normally Qt would be smart about that, but he happened to be using a dual-screen display, with the two screens stacked vertically, and the wider (top) screen happened to have positioned the dialog in the _overlap_ area where the bottom screen didn't reach.. Stuff like that will always screw attempts at "intelligent" positioning.

jonoomph:
@ferdnyc Thanks. I decided to remove the "offscreen" logic for the tutorial (a bit too glitchy). Instead, I made the entire tutorial dialog box clickable, to move on to the next one. So, even a tiny corner can now be clicked, and the box will progress to the next one. Also, added ESC key to dismiss all tutorials.

jonoomph:
I've also simplified the placement logic with smarter coordinate mapping, and I'm feeling pretty good about this PR now. Feels much more polished, and less hard-coded.

src/windows/views/tutorial.py
```diff
@@ -208,13 +232,20 @@ def get_object(self, object_id):
...
             return self.win.transitionsView
         elif object_id == "effectsView":
             return self.win.effectsView
+        elif object_id == "emojisView":
+            return self.win.emojiListView
+        elif object_id == "actionPlay":
+            play_button = None
+            for toolbutton in self.win.videoToolbar.children():
+                if type(toolbutton) == QToolButton and toolbutton.defaultAction() and toolbutton.defaultAction().objectName() == "actionPlay":
+                    return toolbutton
```
ferdnyc, line 248: This could actually be...
```suggestion
            for w in self.win.actionPlay.associatedWidgets():
                if type(w) == QToolButton:
                    return w
```

ferdnyc: (The list it returns will be `[the Qtoolbar, the QToolButton]`.)

src/windows/views/tutorial.py
```diff
@@ -311,56 +355,63 @@ def __init__(self, win):
         # Add all possible tutorials
         self.tutorial_objects = [
             {"id": "0",
-             "x": 400,
+             "x": 0,
              "y": 0,
-             "object_id": "filesView",
+             "object_id": "dockVideoContents",
```
ferdnyc, line 360:
```suggestion
             "object_id": "dockVideo",
```
(And in the string conditions above.) Just for clarity, since it's ultimately returning `self.win.dockVideo` now.

src/windows/views/tutorial.py
```diff
@@ -208,13 +232,20 @@ def get_object(self, object_id):
         elif object_id == "timeline":
             return self.win.timeline
         elif object_id == "dockVideoContents":
```
ferdnyc, line 234:
```suggestion
        elif object_id == "dockVideo":
```
As previous.

jonoomph:
@ferdnyc I used GitHub to "batch" up your suggested edits and commit them. Pretty slick! I haven't done that before. :+1:

ferdnyc:
> @ferdnyc I used GitHub to "batch" up your suggested edits and commit them. Pretty slick! I haven't done that before.

@jonoomph One of my favorite features! Helps keep the git history cleaner and leaner.

They have a bunch of nice features for PRs, actually. Check this out (you'll have to scroll down a little bit): https://github.com/OpenShot/openshot-qt/pull/3463/files#diff-6b28d8a432d605b35f1930b43f84a4d6

I had no idea Actions workflow results would even be _integrated_ into the change view — that's kind of extra-awesome. I'm hopeful Travis might possibly do the same, or at least could be taught to do it, now that we've migrated over to their Actions implementation. Just have to see if it knows how to parse GCC output.

(I keep meaning to make a test PR branch where I can introduce some errors, both non-fatal and fatal, to see how it reacts to them now.)

ferdnyc:
I filed #3519 with one last suggested tweak, which also has the happy effect of making the <kbd>Esc</kbd>-key handling _global_ so you don't have to focus the tutorial popup first.

PR #3519, opened 2020-05-27
Tutorial: Process Esc key as a QAction

ferdnyc:
By setting a temporary `QAction` with an application-wide `QShortcut`, we can skip the expensive key-event handler. And <kbd>Esc</kbd> works anywhere, even if the popup doesn't have focus.

### DIFF
--- a/src/windows/views/tutorial.py
+++ b/src/windows/views/tutorial.py
@@ -29,7 +29,7 @@ import functools
 
 from PyQt5.QtCore import Qt, QPoint, QRectF, QEvent
 from PyQt5.QtGui import *
-from PyQt5.QtWidgets import QLabel, QWidget, QDockWidget, QVBoxLayout, QHBoxLayout, QPushButton, QToolButton, QCheckBox
+from PyQt5.QtWidgets import QLabel, QWidget, QDockWidget, QVBoxLayout, QHBoxLayout, QPushButton, QToolButton, QCheckBox, QAction
 
 from classes.logger import log
 from classes.settings import get_settings
@@ -76,23 +76,6 @@ class TutorialDialog(QWidget):
 
             # Disable metric sending
             s.set("send_metrics", False)
-
-    def keyPressEvent(self, event):
-        """ Process key press events and match with known shortcuts"""
-        # Detect the current KeySequence pressed (including modifier keys)
-        key_value = event.key()
-        modifiers = int(event.modifiers())
-
-        # Abort handling if the key sequence is invalid
-        if (key_value <= 0 or key_value in
-           [Qt.Key_Shift, Qt.Key_Alt, Qt.Key_Control, Qt.Key_Meta]):
-            return
-
-        # A valid keysequence was detected
-        key = QKeySequence(modifiers + key_value)
-        if key == Qt.Key_Escape:
-            # User hit Escape key, hide tutorial permanently
-            self.manager.hide_tips(self.id, True)
 
     def mouseReleaseEvent(self, event):
         """Process click events on tutorial. Especially useful when tutorial messages are partially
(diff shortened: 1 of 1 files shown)


ferdnyc:
@jonoomph: One last optimization I hit on, see what you think.

jonoomph:
@ferdnyc Yes, I love this. Much cleaner and more 'Qt'ish! Thanks!

jonoomph:
@ferdnyc Oops, might have smashed the merge button too soon:
```
  File "/home/jonathan/apps/openshot-qt-git/src/windows/main_window.py", line 2130, in actionTutorial_trigger
    self.tutorial_manager = TutorialManager(self)
  File "/home/jonathan/apps/openshot-qt-git/src/windows/views/tutorial.py", line 436, in __init__
    self.process()
  File "/home/jonathan/apps/openshot-qt-git/src/windows/views/tutorial.py", line 198, in process
    tutorial_dialog = TutorialDialog(tutorial_id, tutorial_details["text"], tutorial_details["arrow"], self)
  File "/home/jonathan/apps/openshot-qt-git/src/windows/views/tutorial.py", line 135, in __init__
    self.close_action = QAction(_("Hide Tutorial"))
TypeError: arguments did not match any overloaded call:
  QAction(QObject): argument 1 has unexpected type 'str'
  QAction(str, QObject): not enough arguments
  QAction(QIcon, str, QObject): argument 1 has unexpected type 'str'
```

src/windows/views/tutorial.py
```diff
@@ -148,12 +131,20 @@ def __init__(self, id, text, arrow, manager, *args):
         hbox = QHBoxLayout()
         hbox.setContentsMargins(20, 10, 0, 0)
 
+        # Close action
+        self.close_action = QAction(_("Hide Tutorial"))
```
jonoomph, line 135: self.close_action = QAction(_("Hide Tutorial"), self)

jonoomph: Probably due to my older version of Qt, I'm guessing

jonoomph:
Also, this PR only seems to work when the tutorial window has focus. So, in my testing, only the very first tutorial window we show accepts the ESC key. Once you click Next, the ESC stops working (at least on my system). Perhaps this was an issue of the `keyPressEvent` solution as well... but I could swear it worked on all the tutorials. Hmmm... strange.

ferdnyc:
@jonoomph Hmmm, I'll try it now, I'm pretty sure it worked for all of them, for me.

Yep, works on any tutorial step, whether the widget has focus or not. In Qt 5.14, anyway.

src/windows/views/tutorial.py
ferdnyc: Yeah, I guess this note applies to _both_ of [these constructors](https://doc.qt.io/qt-5/qaction.html#QAction) (not that they make that remotely clear.)
> **Note:** The _parent_ argument is optional since Qt 5.7.

ferdnyc:
Perhaps setting the `parent` to `get_app().window`, instead, will make it work everywhere on older Qt?

ferdnyc:
Experimentally tried that locally, didn't _break_ anything in Qt 5.13.2. So that's a good sign.

jonoomph:
Okay, I've verified that both solutions (QAction and keyPressEvent) both have the same issue on my local system. I'd be curious how it works on our AppImage though. Also, the get_app().window did not change the behavior for me. 

However, before I implemented the "click to continue" functionality, I was able to click (i.e. and focus the window) before testing my ESC logic. So, I bet that is why it seemed to work before.

ferdnyc:
/me embraces his hover-focus mode gratefully. ("Sloppy", natch, so windows don't lose focus when my pointer wanders out of them and onto the desktop.) :wink: 

Pity the poor Windows users, though — they have to click on things just to be able to _scroll_ them!

PR #3520, opened 2020-05-27
Titles: Don't pollute temp dir with title previews

### LINKED ISSUE #3518: titles leave lot of tmp* files in /tmp
in /tmp i've been left with lot of files with a name that starts with "tmp".

their content shows that they are "previews" for the titles in my openshot project.

for each title there is a `tmp88b89d4_` and a `tmp88b89d4_.png` file.

could you please add a unique prefix (`/tmp/openshot-88b89d4_`) that would allow me to delete all those files in one go? (or put them in a subdirectory of `/tmp`?

(i had 858 files left there by openshot...)

### LINKED ISSUE #3581: Error Saving Project: expected str, bytes or os.PATH like object, not NoneType
- First and foremost, **thank you** for adding creating this amazing software!

I have an issue with it which I believe is a bug and I have no idea how to solve and as I can see, nobody has it or have given the title of the issue as I have here... The tile of this issue is the exact text that appears interrupting any process of video or music creating and editing. OpenShot is able to process and create videos, for instance, but it is interrupted by this error message and then I can close it and it continues doing its job. It is basically a nuisance and inconvenient. Anyone has any idea how to solve it? 

- Thanks in advance for your attention!

ferdnyc:
Fixes #3518

### DIFF
--- a/src/windows/title_editor.py
+++ b/src/windows/title_editor.py
@@ -151,8 +151,8 @@ class TitleEditor(QDialog):
 
     def display_svg(self):
         # Create a temp file for this thumbnail image
-        new_file, tmp_filename = tempfile.mkstemp()
-        tmp_filename = "%s.png" % tmp_filename
+        new_file, tmp_filename = tempfile.mkstemp(suffix=".png")
+        os.close(new_file)
 
         # Create a clip object and get the reader
         clip = openshot.Clip(self.filename)
@@ -161,7 +161,7 @@ class TitleEditor(QDialog):
         # Open reader
         reader.Open()
 
-        # Save thumbnail image and close readers
+        # Overwrite temp file with thumbnail image and close readers
         reader.GetFrame(1).Thumbnail(
             tmp_filename,
             self.graphicsView.width(),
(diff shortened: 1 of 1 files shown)


jonoomph:
LGTM. I really thought we were using our QtImageReader to rasterize the title previews, so it would use our rsvg library (if compiled). Hmmm, I guess that is for another day. :wink:

ferdnyc:
> LGTM. I really thought we were using our QtImageReader to rasterize the title previews, so it would use our rsvg library (if compiled). Hmmm, I guess that is for another day.

Oh, no, we are. It uses `reader.GetFrame().Thumbnail()` to write the preview out to the `tmp_filename`, then reads it back in via `QPixmap().load()`. Then it adds it to the `QGraphicsScene`. (The variable name `svg` for the QPixmap is, arguably, slightly confusing perhaps.)

https://github.com/OpenShot/openshot-qt/blob/0d3da87caf1b5976482e69460662cef4aa12aa9d/src/windows/title_editor.py#L157-L177

PR #3522, opened 2020-05-27
Add new OSP icon, install & use

ferdnyc:
I finally noticed that Linux wasn't showing any sort of customized icon, for `.osp` files:

image

That's because, as it turns out, the name was specified wrong in the mimetypes file. (It was listed as just `openshot`, when our icon name is `openshot-qt`.)

I changed it to `openshot-qt`, then realized that looked terrible because suddenly OSP files looked like the OpenShot application, which they're not.

So, I made an OSP mimetype icon, `openshot-qt-doc.svg`, and updated the appropriate bits of things to install and make use of it. (I changed the tag from `icon` to `generic-icon` in the XML, because that's what almost every other package on my system uses. And it makes sense, as it _is_ a generic icon for OSP files.)

Best of all, since this is based on a Humanity document icon, it's already covered by our existing licensed use of the Humanity collection. Now OSPs will look like this, on Linux. (Windows already does something similar, automatically.)

image


jonoomph:
Awesome, LGTM! Thanks

PR #3526, opened 2020-05-28
Fix hide tutorial QAction syntax for older Qt versions

jonoomph:
Related fix for https://github.com/OpenShot/openshot-qt/pull/3519


jonoomph:
@ferdnyc Adding the **parent** argument for the QAction. All works great on my system with this syntax.

PR #3530, opened 2020-05-29
Update models only once on init

SuslikV:
Because view models has same origin, there is no sense to update models from the each view during initialization.

Now log-file shows that same models are updated twice on init (one for each View). This PR removes this excessive update on View init.

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -2677,6 +2677,7 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
         self.files_model = FilesModel()
         self.filesTreeView = FilesTreeView(self.files_model)
         self.filesListView = FilesListView(self.files_model)
+        self.files_model.update_model()
         self.tabFiles.layout().insertWidget(-1, self.filesTreeView)
         self.tabFiles.layout().insertWidget(-1, self.filesListView)
         if s.get("file_view") == "details":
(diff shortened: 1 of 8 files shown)


ferdnyc:
I'm on my phone right now and can't get a REALLY good look at it, but makes total sense to me. 👍

We probably don't even need the call to `self.refresh_columns()` in the view `__init__()`s, since by then it'll be connected to the signal that the model will emit after refresh.

ferdnyc:
In fact, now I'm wondering if the model refresh shouldn't be delayed until after the project file is loaded (when one is being loaded) -- if it isn't already. (Again, phone. Can't remember.) Before then, there's nothing to refresh.

ferdnyc:
Hrm. So, I was _going_ to merge this (sorry for the delay, lost track of it), and then during a final quick test I actually spotted a difference from the current `develop` code.

In `develop`, if I have the Emojis filter selector set on one of the groups (e.g. 'Smileys'), then when I bring up OpenShot that group will be selected, since it's stored in my user settings.

But with this branch, the Emojis list always comes up showing all items — it doesn't respect the saved filter settings. Not sure why.

It _might_ be because the refresh is being done before the dock is visible. If it's done _after_ that point, the _view_ refresh should happen automatically, because the signal's been connected. Might work to just move the `update_model()` calls down a few lines each, to after the `show() calls for the enabled View. (Or, alternately, that may be backwards. It may have to happen before the view objects are _created_.)

ferdnyc:
Oh, yeah, moving the `update_model()` to before `EmojisListView()` is created fixes it, for that one.  Which makes total sense come to think of it, because the View code makes use of the model data, when it's creating that dropdown list.

jonoomph:
LGTM, seems to work in my testing

PR #3536, opened 2020-06-01
Code Clean-up flagged by Codacy

jonoomph:
Fixing some flaws identified in Codacy. Using this PR as a test to see how the GitHub integration works with Codacy.

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -102,7 +102,7 @@ class OpenShotApp(QApplication):
             log.info("qt5 version: %s" % QT_VERSION_STR)
             log.info("pyqt5 version: %s" % PYQT_VERSION_STR)
         except Exception:
-            log.warning("Failed to read version info from dependencies and platform")
+            log.warning("Error displaying dependency/system details", exc_info=1)
 
         # Setup application
         self.setApplicationName('openshot')
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -102,7 +102,7 @@ class OpenShotApp(QApplication):
             log.info("qt5 version: %s" % QT_VERSION_STR)
             log.info("pyqt5 version: %s" % PYQT_VERSION_STR)
         except Exception:
-            pass
+            log.warning("Failed to read version info from dependencies and platform")
 
         # Setup application
         self.setApplicationName('openshot')
@@ -271,12 +271,12 @@ class OpenShotApp(QApplication):
 @atexit.register
 def onLogTheEnd():
     """ Log when the primary Qt event loop ends """
+    from classes.logger import log
+    import time
 
     try:
-        from classes.logger import log
-        import time
         log.info('OpenShot\'s session ended'.center(48))
         log.info(time.asctime().center(48))
         log.info("================================================")
     except Exception:
-        pass
+        log.warning('Failed to write session ended log')
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -271,12 +271,12 @@ class OpenShotApp(QApplication):
 @atexit.register
 def onLogTheEnd():
     """ Log when the primary Qt event loop ends """
-    from classes.logger import log
-    import time
-
     try:
+        from classes.logger import log
+        import time
         log.info('OpenShot\'s session ended'.center(48))
         log.info(time.asctime().center(48))
         log.info("================================================")
     except Exception:
+        from classes.logger import log
         log.warning('Failed to write session ended log')
--- a/src/classes/thumbnail.py
+++ b/src/classes/thumbnail.py
@@ -62,7 +62,7 @@ def GenerateThumbnail(file_path, thumb_path, thumbnail_frame, width, height, mas
         if reader.info.metadata.count("rotate"):
             rotate = float(reader.info.metadata.find("rotate").value()[1])
     except Exception:
-        pass
+        log.warning('Failed to parse rotation data as float')
 
     # Create thumbnail folder (if needed)
     parent_path = os.path.dirname(thumb_path)
--- a/src/classes/thumbnail.py
+++ b/src/classes/thumbnail.py
@@ -60,9 +60,12 @@ def GenerateThumbnail(file_path, thumb_path, thumbnail_frame, width, height, mas
     rotate = 0.0
     try:
         if reader.info.metadata.count("rotate"):
-            rotate = float(reader.info.metadata.find("rotate").value()[1])
+            rotate_data = reader.info.metadata.find("rotate").value()[1]
+            rotate = float(rotate_data)
+    except ValueError as ex:
+        log.warning("Could not parse rotation value {}: {}".format(rotate_data, ex))
     except Exception:
-        log.warning('Failed to parse rotation data as float')
+        log.warning("Error reading rotation metadata from {}".format(file_path), exc_info=1)
 
     # Create thumbnail folder (if needed)
     parent_path = os.path.dirname(thumb_path)
--- a/src/windows/about.py
+++ b/src/windows/about.py
@@ -75,8 +75,7 @@ class About(QDialog):
                                 self.btnchangelog.setVisible(True)
                                 break
                     except:
-                        # Ignore decoding errors
-                        pass
+                        log.warning('Failed to parse log file %s with encoding %s' % (changelog_path, encoding_name))
 
         create_text = _('Create &amp; Edit Amazing Videos and Movies')
         description_text = _('OpenShot Video Editor 2.x is the next generation of the award-winning <br/>OpenShot video editing platform.')
@@ -270,8 +269,7 @@ class Changelog(QDialog):
                                                    'subject': line[45:].strip() })
                         break
                 except:
-                    # Ignore decoding errors
-                    pass
+                    log.warning('Failed to parse log file %s with encoding %s' % (changelog_path, encoding_name))
         self.openshot_qt_ListView = ChangelogTreeView(commits=changelog_list, commit_url="https://github.com/OpenShot/openshot-qt/commit/%s/")
         self.vbox_openshot_qt.addWidget(self.openshot_qt_ListView)
         self.txtChangeLogFilter_openshot_qt.textChanged.connect(partial(self.Filter_Triggered, self.txtChangeLogFilter_openshot_qt, self.openshot_qt_ListView))
@@ -291,8 +289,7 @@ class Changelog(QDialog):
                                                    'subject': line[45:].strip() })
                         break
                 except:
-                    # Ignore decoding errors
-                    pass
+                    log.warning('Failed to parse log file %s with encoding %s' % (changelog_path, encoding_name))
         self.libopenshot_ListView = ChangelogTreeView(commits=changelog_list, commit_url="https://github.com/OpenShot/libopenshot/commit/%s/")
         self.vbox_libopenshot.addWidget(self.libopenshot_ListView)
         self.txtChangeLogFilter_libopenshot.textChanged.connect(partial(self.Filter_Triggered, self.txtChangeLogFilter_libopenshot, self.libopenshot_ListView))
@@ -312,8 +309,7 @@ class Changelog(QDialog):
                                                    'subject': line[45:].strip() })
                         break
                 except:
-                    # Ignore decoding errors
-                    pass
+                    log.warning('Failed to parse log file %s with encoding %s' % (changelog_path, encoding_name))
         self.libopenshot_audio_ListView = ChangelogTreeView(commits=changelog_list, commit_url="https://github.com/OpenShot/libopenshot-audio/commit/%s/")
         self.vbox_libopenshot_audio.addWidget(self.libopenshot_audio_ListView)
         self.txtChangeLogFilter_libopenshot_audio.textChanged.connect(partial(self.Filter_Triggered, self.txtChangeLogFilter_libopenshot_audio, self.libopenshot_audio_ListView))
--- a/src/windows/animated_title.py
+++ b/src/windows/animated_title.py
@@ -132,4 +132,4 @@ class AnimatedTitle(QDialog):
                 self.settingsContainer.layout().removeWidget(child)
                 child.deleteLater()
             except:
-                pass
+                log.warning('Failed to remove child widget for effect controls')
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -292,6 +292,7 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
                     f.write(lock_value)
                 break
             except Exception:
+                log.warning('Failed to write lock file (attempt: %s)' % attempts)
                 attempts -= 1
                 sleep(0.25)
 
@@ -306,6 +307,7 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
                 os.remove(lock_path)
                 break
             except Exception:
+                log.warning('Failed to destroy lock file (attempt: %s)' % attempts)
                 attempts -= 1
                 sleep(0.25)
 
@@ -2589,8 +2591,7 @@ class MainWindow(QMainWindow, updates.UpdateWatcher):
             self.unity_launcher.set_property("progress", current_frame / (end_frame - start_frame))
             self.unity_launcher.set_property("progress_visible", True)
         except Exception:
-            # Just ignore
-            pass
+            log.warning('Failed to notify unity launcher of export progress. Frame: %s' % current_frame)
 
     def ExportFinished(self, path):
         """Show completion in Unity Launcher (if connected)"""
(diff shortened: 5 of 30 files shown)


src/classes/thumbnail.py
```diff
@@ -62,7 +62,7 @@ def GenerateThumbnail(file_path, thumb_path, thumbnail_frame, width, height, mas
         if reader.info.metadata.count("rotate"):
             rotate = float(reader.info.metadata.find("rotate").value()[1])
     except Exception:
-        pass
+        log.warning('Failed to parse rotation data as float')
```
ferdnyc, line 65: We could be even more informative, with something like,
```python

    try:
        if reader.info.metadata.has_key("rotate"):
            rotate_data = reader.info.metadata.find("rotate").value()[1]
            rotate = float(rotate_data)
    except ValueError as ex:
            log.warning("Could not parse rotation value {}: {}".format(rotate_data, ex))
    except Exception:
            log.warning("Error reading rotation metadata from {}".format(file_path), exc_info=1)
```
(That'll cover the bases, whether the exception comes from the `float()` or the previous `.find().value()[1]`)

src/windows/about.py
```diff
@@ -75,8 +75,7 @@ def __init__(self):
                                 self.btnchangelog.setVisible(True)
                                 break
                     except:
-                        # Ignore decoding errors
-                        pass
+                        log.warning('Failed to parse log file %s with encoding %s' % (changelog_path, encoding_name))
```
ferdnyc, line 78: `log.info()` at _most_, preferably `log.debug() even (so the users won't see it)`. There is nothing the user can do about that, and no reason they should care.

src/windows/animated_title.py
```diff
@@ -132,4 +132,4 @@ def clear_effect_controls(self):
                 self.settingsContainer.layout().removeWidget(child)
                 child.deleteLater()
             except:
-                pass
+                log.warning('Failed to remove child widget for effect controls')
```
ferdnyc, line 135: Codacy (and any Python linter) is even more pissy about bare `except:` than it is about `try/except/pass'! This will never fly unless it's at least:
```python
except Exception:
    log.something(something something dark side something exception)
```
But I would argue that the exception is, again, nothing the user can do anything about, so "warning" doesn't make sense. My preference would be:
```python3
except Exception as ex:
    log.debug("Failure removing Blender parameter widget: {}".format(ex))
```

src/windows/main_window.py
```diff
@@ -292,6 +292,7 @@ def create_lock_file(self):
                     f.write(lock_value)
                 break
             except Exception:
+                log.warning('Failed to write lock file (attempt: %s)' % attempts)
```
ferdnyc, line 295:
```suggestion
                log.debug('Failed to write lock file (attempt: %s)' % attempts)
```

src/windows/main_window.py
```diff
@@ -306,6 +307,7 @@ def destroy_lock_file(self):
                 os.remove(lock_path)
                 break
             except Exception:
+                log.warning('Failed to destroy lock file (attempt: %s)' % attempts)
```
ferdnyc, line 310:
```suggestion
                log.debug('Failed to destroy lock file (attempt: %s)' % attempts)
```

src/windows/main_window.py
```diff
@@ -2589,8 +2591,7 @@ def FrameExported(self, title_message, start_frame, end_frame, current_frame):
             self.unity_launcher.set_property("progress", current_frame / (end_frame - start_frame))
             self.unity_launcher.set_property("progress_visible", True)
         except Exception:
-            # Just ignore
-            pass
+            log.warning('Failed to notify unity launcher of export progress. Frame: %s' % current_frame)
```
ferdnyc, line 2594:
```suggestion
            log.debug('Failed to notify unity launcher of export progress. Frame: %s' % current_frame)
```
Especially since this will get logged _every_ frame, if there's a general connectivity problem!

src/windows/main_window.py
```diff
@@ -2599,7 +2600,7 @@ def ExportFinished(self, path):
             self.unity_launcher.set_property("progress", 0.0)
             self.unity_launcher.set_property("progress_visible", False)
         except Exception:
-            pass
+            log.warning('Failed to notify unity launcher of export progress. Completed.')
```
ferdnyc, line 2603:
```suggestion
            log.debug('Failed to notify unity launcher of export progress. Completed.')
```

src/windows/preferences.py
```diff
@@ -498,7 +498,7 @@ def text_value_changed(self, widget, param, value=None):
             if not value:
                 value = widget.toPlainText()
         except:
-            pass
+            log.warning('Failed to get plain text from widget')
```
ferdnyc, line 501:
```suggestion
            log.debug('Failed to get plain text from widget')
```

src/windows/title_editor.py
```diff
@@ -222,7 +222,7 @@ def load_svg_template(self):
                 self.settingsContainer.layout().removeWidget(child)
                 child.deleteLater()
             except Exception:
-                pass
+                log.warning('Failed to delete child settings widget')
```
ferdnyc, line 225:
```suggestion
                log.debug('Failed to delete child settings widget')
```

src/windows/title_editor.py
```diff
@@ -441,7 +441,7 @@ def update_font_color_button(self):
                 color = txt[5:]
             except Exception:
                 # If the color was in an invalid format, try the next text element
-                continue
+                log.warning('Failed to parse color value from text element')
```
ferdnyc, line 444:
```suggestion
                log.debug('Failed to parse {} as color value'.format(txt))
```

src/windows/title_editor.py
```diff
@@ -450,7 +450,7 @@ def update_font_color_button(self):
                 txt = ar[opacity]
                 opacity = float(txt[8:])
             except Exception:
-                pass
+                log.warning('Failed to parse opacity value from text element')
```
ferdnyc, line 453:
```suggestion
                log.debug('Failed to parse {} as opacity value'.format(txt))
```

src/windows/views/blender_listview.py
```diff
@@ -210,6 +210,7 @@ def text_value_changed(self, widget, param, value=None):
             if not value:
                 value = widget.toPlainText()
         except Exception:
+            log.warning('Failed to read plain text value from widget')
```
ferdnyc, line 213:
```suggestion
            log.debug('Failed to read plain text value from widget')
```

src/windows/views/properties_tableview.py
```diff
@@ -181,7 +181,7 @@ def mouseMoveEvent(self, event):
             except Exception:
                 # If item is deleted during this drag... an exception can occur
                 # Just ignore, since this is harmless
-                return
+                log.warning('Failed to access data on selected label widget')
```
ferdnyc, line 184:
```suggestion
                log.debug('Failed to access data on selected label widget')
```
Although, since the comment gives a perfectly good reason for this happening normally, logging it at all doesn't really seem necessary. There's a reason Codacy allows its notations to be Ignored on either an individual or repo-wide basis, some "issues" really aren't.

src/windows/views/tutorial.py
```diff
@@ -297,8 +297,7 @@ def exit_manager(self):
             self.win.dockProperties.visibilityChanged.disconnect()
             self.win.dockVideo.visibilityChanged.disconnect()
         except Exception:
-            # Ignore errors from this
-            pass
+            log.warning('Failed to properly disconnect from dock signals')
```
ferdnyc, line 300:
```suggestion
            log.debug('Failed to properly disconnect from dock signals', exc_info=1)
```

src/windows/views/timeline_webview.py
```diff
@@ -229,7 +229,7 @@ def update_clip_data(self, clip_json, only_basic_props=True, ignore_reader=False
                 clip_data = clip_json
         except Exception:
             # Failed to parse json, do nothing
-            return
+            log.warning('Failed to parse clip JSON data')
```
ferdnyc, line 232:
```suggestion
            log.warning('Failed to parse clip JSON data', exc_info=1)
```

src/windows/about.py
ferdnyc: Oh, and this is another one where it needs to be `except Exception:` at a minimum.

jonoomph:
@ferdnyc Thanks for the feedback. I agree about the .debug() being much better for these types of things. I also like the idea of having some ability to see Exceptions if they are happening. The idea of things failing, a `pass` in the exception handler, seems to not help anyone. I would love to make OpenShot less chatty in general, but allow for more verbose modes when troubleshooting/debugging.

ferdnyc:
> @ferdnyc Thanks for the feedback. I agree about the .debug() being much better for these types of things. I also like the idea of having some ability to see Exceptions if they are happening. The idea of things failing, a `pass` in the exception handler, seems to not help anyone. I would love to make OpenShot less chatty in general, but allow for more verbose modes when troubleshooting/debugging.

A few times, in testing, I've added `log.debug()` statements to OpenShot (or made use of the couple that were already there), and just edited `classes.logger` to lower the minimum level to `logging.DEBUG`. I could _prooooooobably_ be persuaded to turn that into a command-line `--debug` flag.

The other option would be wiring it up to the Preferences debug switch. But that has two down sides:
1. It stays on unless you turn it off, whereas running `openshot-qt --debug` only lasts for the session
2. It also enables libopenshot debugging, even if you don't want that. (And you often don't, in fact having the users turn it on for even UI issues is kind of a support headache, when we know there can't possibly be any useful info in the libopenshot log.)

src/classes/thumbnail.py
jonoomph: reader.info.metadata is not a dict, but otherwise I love it. I'll adapt this shortly.

src/classes/app.py
```diff
@@ -271,12 +271,12 @@ def run(self):
 @atexit.register
 def onLogTheEnd():
     """ Log when the primary Qt event loop ends """
+    from classes.logger import log
+    import time
 
     try:
-        from classes.logger import log
-        import time
```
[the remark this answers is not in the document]
jonoomph: Okay, reverting this change.

jonoomph:
I'm just playing around with this PR, to better get a handle on how Codacy works. It's pretty great! Of course, if I can improve the code quality a bit in the process, that's great also!

jonoomph:
It's interesting, some issues are flagged as "New Issues" in Codacy, even though they were not modified. Although, I did change the indent and TABS to spaces... so I guess I touched every line on app.js, which is why it's triggering all these other failures now. 

Also, on some files, it will list "1 New Issue". So, I fix it, and then it says "2 New Issues" on the same file. Strange it wouldn't just list all issues that are found. It seems like it stops analyzing when it hits certain issues perhaps. Just interesting...

jonoomph:
I'm gonna really dig into the timeline JS files, and see what it will take for Codacy to be happy with them.

jonoomph:
Wow, our timeline JS files are a huge mess. Mixed indents (tab and spaces), lots and lots of lint issues, unused code, etc... This might take a few tries to get it right, but it's already looking way better.

SuslikV:
You will spend a year to fix all styling issues and some will be still broken. Don't waste your time on this. This always slows down development if made in the middle.

ferdnyc:
@jonoomph I believe I tackled every issue flagged in `tutorial.py` except [the "method-could-be-function" one](https://app.codacy.com/gh/OpenShot/openshot-qt/file/45246587004/issues/source?bid=18413121&fileBranchId=18414098#l64)

ferdnyc:
I kind of have a love/hate relationship with Codacy, and static analysis in general.

It has its _place_ and can definitely catch simple boner moves that are otherwise easy to miss. (Like the classic `if (sampleRate = 44100)` in libopenshot, the bug that inspired me to initially start running scans on the code in the first place.)

But at the same time, these tools all encourage their perceived worth to be based on how many supposed "problems" they can flag, and the signal:noise ratio is poor. Codacy _tries_ to do slightly better by grouping flags into categories like "Security", "Code Style", etc. — which is better than the traditional "everything's an issue" false-equivalency model of analysis reporting — but it still means you have to dig through a lot of crap whining about non-issues.

The "OMG shut the f--- up about this, already!" button is definitely an important component of static analysis response.

(And I've smacked it a couple of times already in the Codacy config, you'll notice:)
Screenshot from 2020-06-02 07-58-23

Because, no, it's really _not_ a problem that our CSS doesn't include Firefox- and Opera-compatible gradients for its exclusively-WebKit-rendered HTML. And I'm just kind of OK with the fact that our docstrings don't end with sentence punctuation and **_OMFG why is that enabled by DEFAULT!?!_**

> It's interesting, some issues are flagged as "New Issues" in Codacy, even though they were not modified. Although, I did change the indent and TABS to spaces... so I guess I touched every line on app.js, which is why it's triggering all these other failures now.
> 
> Also, on some files, it will list "1 New Issue". So, I fix it, and then it says "2 New Issues" on the same file. Strange it wouldn't just list all issues that are found. It seems like it stops analyzing when it hits certain issues perhaps. Just interesting...

Codacy does two types of analysis: Branch analysis, and PR analysis. The branch analysis will list all of the issues in the committed code. But in PR analysis, it only looks at the code that's part of the PR diff, ignoring the surrounding code. The idea is to (a) perform quicker analysis, and (b) keep _new_ issues out of the codebase, regardless of the general state of the code prior to the PR's changes.

That's why it sometimes comes up with issues being "added" only because that line was touched for other reasons.

(Really, "added" and "new" are both bad labels. It should say that the code in the PR "contains" N issues — they may very well have been pre-existing, but the stance is that if you touched the line and _didn't_ fix the issue, now you own it. Which... has some merit, honestly, at least in a world where code is checked regularly and expected _not_ to have any existing issues. It's just a bit unrealistic on an already-dirty codebase.)

ferdnyc:
> (Like the classic `if (sampleRate = 44100)` in libopenshot, the bug that inspired me to initially start running scans on the code in the first place.)

Wait, no! It was even sneakier than that, it was something like...
```c++
if (condition) {
    somecode();
    sampleRate == 44100;
    moreCode();
}
```
Which is even _easier_ to miss.

SuslikV:
To be honest, the ` '%s' %` strings format in python is outdated too. The fixes requires more fixes...

ferdnyc:
> To be honest, the ` '%s' %` strings format in python is outdated too. The fixes requires more fixes...

It _is_, though I've since concluded that format's actually preferable in one case: Strings that will be translated via gettext, since it uses the same escapes. So, inside `_(...)` I've actually come back around to `%s`, `%d`, and friends. (And I'm regretting a few of the spots where I replaced those strings with `.format()` and `{}` placeholders — sorry translators!)

Other than that, well, hell — if we were exclusively running under Python 3.6+ we could use f-strings, which kick the ass of both `%` and `.format()`. Hard to beat this for clarity:
```python
log.warning(f"Error parsing {item_data} as a float: {exception}")
```
(Yes, it would strongly encourage us to stop using shorthand variable names like `ex` for the exception message. One nice side-effect about f-strings, you're really forced to think more about how you name your variables.)

ferdnyc:
The threatened `--debug` command line flag(s) are now PR #3539. Presuming we agree that's a good approach and decide to merge, we can/should start making more liberal use of `log.debug()` alongside the higher log levels, when outputting messages that aren't intended for endusers.

jonoomph:
@ferdnyc Okay, I've got this down to 1 issue on Codacy, and I don't want to disable the code pattern for this one (if possible). I'm at a loss, or maybe I need to let my brain sleep some. Maybe you can find a solution for this 1 line/issue? Thx!

ferdnyc:
@jonoomph I'll take a look. I haven't been playing along at home for all of this, so it may take me a bit to catch up on the eslint stuff. (Complicated by the fact that Chrome is _refusing_ to load the eslint.org site, for some reason... so I'm flying without documentation. Bother.)

ferdnyc:
A Medium article I'm still digesting: [A simple intro to Javascript imports and exports
](https://medium.com/@thejasonfile/a-simple-intro-to-javascript-imports-and-exports-389dd53c3fac). It covers both ES5 syntax (which I suspect we're stuck with, in the AppImage's ancient JS?) and ES6, though not Angular per se.

I have a feeling that wrapping `functions.js` into a class (maybe `class Tools`), and making each function a class member, might be the way to go there. Then we can export `Tools` from that file and import it into each of the other JS files, rather than including it in the HTML file (and polluting the global namespace). But that's first-impressions, I haven't tried any of it yet.

ferdnyc:
Oh, wait... that might not work, though, because I believe we also use some of the `functions.js` functions in the HTML as well. Hrm.

jonoomph:
@ferdnyc I'm going to disable the Codacy status check, and make it optional. For example, I've fixed 334 issues on this PR, but 1 "fake" issue is still reported. That shouldn't keep us from merging. So, for now, I think we can use the Codacy status check as a way to call attention to potential quality issues. And if they are serious/real, we can wait for them to be resolved. But I like the flexibility to override them once we determine it's wasting our time. :wink:

jonoomph:
Screenshot from 2020-06-04 17-39-50

jonoomph:
Progress!

PR #3542, opened 2020-06-04
Improved new codec handling

eisneinechse:
The presets for the libaom-av1 codec were improved and the title of the export window now shows a message requesting the user to wait when 100% are reached but the clip is not finished due to finalising work to be done by the codec.

### DIFF
--- a/src/windows/export.py
+++ b/src/windows/export.py
@@ -913,7 +913,10 @@ class Export(QDialog):
                     if ((( frame - start_frame_export ) != 0) & (( end_time_export - start_time_export ) != 0)):
                         seconds_left = round(( start_time_export - end_time_export )*( frame - end_frame_export )/( frame - start_frame_export ))
                         fps_encode = ((frame - start_frame_export)/(end_time_export-start_time_export))
-                        title_message = _("%(hours)d:%(minutes)02d:%(seconds)02d Remaining (%(fps)5.2f FPS)") % {
+                        if frame == end_frame_export:
+                            title_message = _("FINALIZING please wait")
+                        else:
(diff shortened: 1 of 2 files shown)


src/windows/export.py
```diff
@@ -913,7 +913,10 @@ def accept(self):
                     if ((( frame - start_frame_export ) != 0) & (( end_time_export - start_time_export ) != 0)):
                         seconds_left = round(( start_time_export - end_time_export )*( frame - end_frame_export )/( frame - start_frame_export ))
                         fps_encode = ((frame - start_frame_export)/(end_time_export-start_time_export))
-                        title_message = _("%(hours)d:%(minutes)02d:%(seconds)02d Remaining (%(fps)5.2f FPS)") % {
+                        if frame == end_frame_export:
+                            title_message = _("FINALIZING please wait")
```
jonoomph, line 917:
```suggestion
                            title_message = _("Finalizing video export, please wait...")
```

src/windows/export.py
```diff
@@ -1008,6 +1011,16 @@ def accept(self):
             # Reveal done button
             self.close_button.setVisible(True)
 
+            # Restore windows title to show elapsed time
```
jonoomph, line 1014: Perhaps we can create a new function, to prevent duplication of this string formatting?
 `title_message = generate_title_message(seconds, fps)`

eisneinechse: Yes. The whole way this is done should be simplified. It's just I included this in this unrelated PR and I absolutely didn't want to mix things up.

jonoomph:
LGTM!

Issue #3546, opened 2020-06-06
Adding Transitions or Effects in Windows Version


USATechDude:
**Describe the bug:**
When I try the first time to add a transition to the timeline, I can only apply common transitions.  All other transitions are denied.  I can keep adding common transitions to the timeline until I select an extra transition.  After selecting an extra transition I cannot add any transition or effect to the timeline.

**Steps to reproduce the behavior:**

1. Create a new project 
2. Import three clips to the Project Files list
3. Add a clip to the timeline 
4. Move the playhead to the end of the clip 
5. Add another clip to the timeline
6. Move the playhead back one second on the first clip
7. Drag the second clip to the playhead overlaying the end of first clip
8. Select the default Fade transition and remove it from the timeline
9. Click on the Transitions tab of the Project Files window
10. Click one of the common transitions and drag it to the timeline
11. Edit the Properties of the transition
12. Click on the Effects tab of the Project Files window
13. Click and drag one of the Effects to the clip on the timeline
14. Edit the Properties of the Effect
15. Move the playhead to the end of the second clip
16. Add the last clip to the timeline
17. Move the playhead back one second on the first clip
18. Drag the second clip to the playhead to create the default Fade transition
19. Select the Fade transition and remove it from the timeline
20. Click on the Transitions tab of the Project Files window
21. Click on a transition other than the common transitions and drag it to the timeline
22. ***ERROR*** OpenShot will no longer add any transition or any effect to the timeline

**Expected behavior:**
I should be able to select any of the transitions and add them to my timeline, not just the common transitions.  After selecting an extra transition, I should be able to add any transition or effect to the clips on the timeline.
These issues do not occur in the Linux AppImage on Ubuntu 20.04 LTS
These issues do not occur in the stable version for Windows 

**System Details:**
- OpenShot Version: OpenShot-v2.5.1-dev2-1591067614-782d764c-7cec1822-x86_64.exe
- Operating System / Distribution: Windows 10 Version 2004

nanohtpc:
similar to bug 3401, and it has almost 2 months old!!!!

USATechDude:
I can convert an extra transition (i.e.: dissolve.jpg) to an SVG file then copy it to the 
`/OpenShot Video Editor/transitions/common `
folder and use it in OpenShot and not affect other transitions, effects or emojis.

USATechDude:
After reading Issue #3401 I edited the file:

[...]

Issue #3466, opened 2020-05-08
Can't create or load anything with a random transition on it...


nanohtpc:
Can't create or load anything with a random transition on it...

winx64

latest daily...

8((((

SuslikV:
Duplicate of: https://github.com/OpenShot/openshot-qt/issues/3401

nanohtpc:
probkem REMAINS, even editing the hidden files....

jonoomph:
This is fixed in the latest daily build: https://www.openshot.org/download#daily

Issue #3401, opened 2020-04-23
Give your bug Adding a Transition = hangs the program


nanohtpc:
when adding a video to the timeline, choosing a RANDOM transition... hangs the program...

Win x64

SuslikV:
Not enough info to track the issue. I cannot reproduce this. Can you share OpenShot's log-file (from the session when trouble is happened)?

nanohtpc:
OpenShot-v2.5.1-dev2-1586151841-166f6938-232f3f79-x86_64

Was the last version that doesnt have problems, but beyond that, if i tried to add a random transition OR try to load a OSP that have a ramdom one, it just hangs/crash the program....

![Screenshot (243)](https://user-images.githubusercontent.com/34278444/80128058-c6f95000-855a-11ea-8bb7-44b9564a01ad.png)

win x64

it happens with all kind of videos...

nanohtpc:
that was a perfect OSP with older version, but hangs with the last 8 new ones...

also happens trying to add new files... only with random transitions....

SuslikV:
Hmm... It's far deeper issue.

nanohtpc:
OpenShot-v2.5.1-dev2-1586151841-166f6938-232f3f79-x86_64

was the last version that works...

SuslikV:
Issue caused by next line of code:
https://github.com/OpenShot/openshot-qt/blob/38be10411756cbc76447afb0a1d1b202ededddde/src/classes/app.py#L110

Try to remove it like:
```
        self.setApplicationVersion(info.SETUP['version'])
        self.setWindowIcon(QIcon(":/openshot.svg"))
        try:
```
after
```
        self.setApplicationVersion(info.SETUP['version'])
        try:
```
Use _notepad++_ application (minimalist version is enough for your needs). Save file, restart application, try again.

-------------

PR in question: https://github.com/OpenShot/openshot-qt/pull/3354

SuslikV:
Usually, when error occurs, openshot log entry has next info:
```
...
add_to_timeline:INFO accept
  exceptions:ERROR Unhandled Exception
Traceback (most recent call last):
  File "I:\Temp\OpenShot-dev\01\{app}\windows\add_to_timeline.py", line 328, in accept
    transition_reader = openshot.QtImageReader(transition_path)
  File "C:\Users\Administrator\builds\7546b651\0\OpenShot\openshot-qt\openshot.py", line 2501, in __init__
RuntimeError: File could not be opened.
[...]

Issue #3383, opened 2020-04-15
Add "remove all keyframes" to property editor


fwolfst:
**Describe the new feature:**
Related to https://github.com/OpenShot/openshot-qt/issues/2420 , sometimes I messed up with the keyframe thing (seen the PR for the curve editor, awesome! https://github.com/OpenShot/openshot-qt/pull/3084 ) and I want to delete all keyframes or set a value for the whole clip

**Describe the solution you'd like:**
A right mouse button on the value allows me to "remove all keyframes" (which to my understanding internally will create a keyframe at frame 1 with value 0 which is not always the greatest choice).

**Describe alternatives you have considered:**
Another helpful addition/solution would be to have the option (also right mouse click on property value) ~"make this value the only keyframe", which would remove all keyframes and add one at the current position with the current value.

Thanks for all your great work. I'd love to contribute differently but lack the (time and other resources-) budget.

SuslikV:
Maybe "Reset to defaults"?

fwolfst:
That sounds like a interesting option, too. Although then if you want to
edit the property for the whole clip, you would again have to go to frame 0
(which might be not representative, e.g. because of a fade-in, a logo, etc)
or delete the default keyframe.

SuslikV:
Right now only _Animate > No Animation_ is present in right-click menu over the entire clip in the _Timeline_. It emulates similar "reset" for some of the options.

MBB232:
Removing *all* animations is a bit crude if you have multiple animations on separate parts of the video (or combined)
The same thing happens when applying default animations to already modified properties. Like if you already set a location a bit out of the middle. 'transition from edge to center' replaces the end location to the center, rather then stopping on originally set position when sliding in. 
Where are the default transition effects stored, can they be modified for projects like other effects?

In the introduction description the was a menu option to choose the keyframe, but I do not see it in the current version
https://www.openshot.org/blog/2009/07/14/key-frame-awesome/
Clip properties>layout>Keyframe>end of clip


The green points are a bit too small to select, and sometimes keyframes are very close together. (Especially if you tried to change one but was a bit off, and accidentally created a new keyframe. ) 

Maybe an option can be made to "list all keyframes for clip/animation", so you can easily modify them and set some values the same as the previous while setting different values for other properties.
You would get a list of all keyframes on a clip in a row. Then you can apply changes on all those that you select, like removal, copy or change values. 

[...]

PR #3547, opened 2020-06-06
Removing setWindowIcon() from our main window

jonoomph:
Removing setWindowIcon() from our main window, as it has a crazy side-effect on Windows and prevents our QtImageReader from being able to open certain types of files, such as JPEG. Or at least, that is what appears to be happening. No JPEG transitions of effects can be dropped on the timeline. They both fail when trying to construct the QtImageReader. I've also verified the path is correct going into the QtImageReader constructor. 

Fixes:
https://github.com/OpenShot/openshot-qt/issues/3546
https://github.com/OpenShot/openshot-qt/issues/3401
https://github.com/OpenShot/openshot-qt/issues/3466

Here is the error when dropping a transition on the timeline when `self.setWindowIcon(QIcon(":/openshot.svg"))` is present in `app.py`:
```
timeline_webview:INFO addTransition...
  File "C:\Program Files (x86)\OpenShot Video Editor\windows\views\timeline_webview.py", line 2802, in dragEnterEvent
    self.addTransition(data, pos)
  File "C:\Program Files (x86)\OpenShot Video Editor\windows\views\timeline_webview.py", line 2907, in addTransition
    transition_reader = openshot.QtImageReader(file_ids[0])
  File "C:\Users\Administrator\builds\7546b651\1\OpenShot\openshot-qt\openshot.py", line 2503, in __init__

```

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -107,7 +107,6 @@ class OpenShotApp(QApplication):
         # Setup application
         self.setApplicationName('openshot')
         self.setApplicationVersion(info.SETUP['version'])
-        self.setWindowIcon(QIcon(":/openshot.svg"))
         try:
             # Qt 5.7+ only
             self.setDesktopFile("org.openshot.OpenShot")


jonoomph:
@ferdnyc Merging this for now, so it unblocks lots of people getting stuck on the daily Windows build. But this is sooooo weird, and seemingly unrelated to our QtImageReader. My brain hurts to think how this connects, lol. Maybe you can figure it out, or maybe it's not worth the time to investigate, since I'm unclear if we needed the command in the first place. Looking forward to your thoughts though!

ferdnyc:
I'll take a quick look, I suspect I know what's going on here. It only happens when dragging files in from the operating system, right? Not internal drags?

jonoomph:
Nope, happens on internal drags. (I didn't test on external drags yet).
1) Launch daily build for Windows
2) Drag a non-common transition onto the timeline. It fails to drop it.
3) The error above will appear in your openshot-qt.log

However, removing the setWindowIcon line from app.py, and everything works.

ferdnyc:
Yeah, sorry, I misread that traceback. But I think I was on the right track.

I suspect what's happening is, the window icon is probably getting bunlded into the QDrag MIME data, in some or all cases. It's probably used as a default/fallback value.

If the window icon is unset, or a filesystem path, there's no problem because worst case, QtImageReader gets the icon itself. (Like that old bug we had, where you could drag marker icons onto the Timeline and they'd become new Clips! @DylanC fixed that one.) ...But if the icon that's in there is a path based on a QResource reference, no part of libopenshot has any possible way of making use of that.

Just means we need to sanitize the MIME payload better.

The reason the `setWindowIcon()` is there is so that certain Linux desktops use the correct image in the window / taskbar decorations. Without it, the window won't have an icon. It was requested by a user a few months back. But I agree it makes sense to drop it until we figure this out.

jonoomph:
Yeah, that makes sense, and it sounds about right! However, I added debug line right above the failing QtImageReader constructor line, and output the `file_ids[0]` variable being passed in. It was the same on both versions (working and non-working). That's when I gave up, haha.

jonoomph:
I believe this code, inside QtImageReader, was failing during the Open() method, and throwing the `InvalidFile` exception.
```
	if (!loaded) {
		// Attempt to open file using Qt's build in image processing capabilities
		image = std::shared_ptr<QImage>(new QImage());
		success = image->load(path);
	}
	if (!success) {
		// raise exception
		throw InvalidFile("File could not be opened.", path.toStdString());
	}
```

SuslikV:
I what to make attention that the root of the issue is quite different, I think. Because my own experience points to the position of the `setWindowIcon()` call in daily builds:

https://github.com/OpenShot/openshot-qt/issues/3401#issuecomment-626169148

And this `setWindowIcon()` call has no influence at all on my own builds (freeze) of OpenShot. You may give it a try (maybe I miss something?..): https://github.com/OpenShot/openshot-qt/issues/3383#issuecomment-638020348
By the way, In my builds - I'm not modifying the final `.exe` file by third party tool to set the info data as daily builds do.

jonoomph:
Another idea, we could just exclude the `setWindowIcon()` method from Windows, but still call it on Linux/Mac. That might be a simple fix.

SuslikV:
@jonoomph can you make test build without version overriding (verpatch.exe), just to test my assumption?

jonoomph:
@SuslikV Sure, I'll give that a try shortly.

SuslikV:
@jonoomph Any news on this?

PR #3560, opened 2020-06-13
Ability to not store the history in the project

rabits:
If save history is set to 0 - it will not work properly and store the entire history in the project. Not sure if openshot should store history by default in the project (maybe only in backups?), but this is useful if I don't want to store megabytes of history.


jonoomph:
Looks good to me! Thanks!

Issue #2922, opened 2019-08-07
Interlaced video import or export not supported (displayed/encoded as Progressive)


lucaetere:
![adva](https://user-images.githubusercontent.com/53849689/62596139-9c576080-b913-11e9-81be-17a8d4345b0a.png)
advanced tab property bug

export file 29.97 hd interlaced (59.94I),mpeg2
but the result file is Progressive!!!

even if I clicked on interlaced =YES on adavnced tab of the export file window

**System Details**
- OpenShot Version: 2.4.4

- Platform: Windows2016 std

**Log Files**
included and uploaded

lucaetere:
moreover

it did not embed the CC eia-708 from the source file

ferdnyc:
At least on the face of it, confirmed. Exporting a project set up with the "HD 1080i 25fps (1920x1080)" profile — set in both the project and the Export dialog's Simple tab — and using the "MP4 (h.264)" target, results in an output video with Progressive, non-interlaced frames, when checked with MediaInfo.

Caption embedding, I don't believe we support, sorry. But the field order thing is most definitely a bug.

ferdnyc:
Tried switching to "AVI (mpeg4)" and "AVI (mpeg2)" as the target, no change. Still a Progressive-scan output video.

ferdnyc:
Looping in @jonoomph and @eisneinechse, because after a quick grep through the `libopenshot` code, I am not finding _any_ code that would seem to handle interlaced output in `FFmpegWriter`. 

Like, **at _all_**. 

...Can that possibly be right?

ferdnyc:
@lucaetere also provided this screenshot, showing an imported video with interlaced fields being interpreted as Progressive-scan by OpenShot.

![62604176-10046800-b92a-11e9-9688-19e83fafa9ac](https://user-images.githubusercontent.com/538020/62635506-d02b8980-b905-11e9-91c5-2e6a53ff5710.png)

ferdnyc:
To be honest, @lucaetere , I suspect it's been so long since anyone's even tried to use interlaced video with OpenShot, that at some point support for it broke and you're just the first one to notice.

lucaetere:
I see @ferdnyc , interlaced video are a standard in TV broadcast. and this is the usage I would do of this software.
Hope this may be fixed soon
Thanks

ferdnyc:
> I see @ferdnyc , interlaced video are a standard in TV broadcast.

[...]

PR #3570, opened 2020-06-17
Implement interlace export

eisneinechse:
So far only an export in an mov container of an mpeg2 encoded clip works.
Further work is needed.


SuslikV:
For what?

eisneinechse:
This should fix 
https://github.com/OpenShot/openshot-qt/issues/2922

SuslikV:
Isn't better to remove this feature completely? Someone said that there is 20 years of XXI century already gone.

eisneinechse:
Short answer: No.
Longer:
There is even someone that wants it and we said it is there. We are no multinational company that screws people by selling them things and then delivering a subset.
We could of course streamline everything a little more, only export mpeg2 in a mov container with 720p at 24 fps ; that would make thing so much easier. And while we're at it let's ban the wheel, old technology.

eisneinechse:
@ferdnyc This would be the interlace support we promised a long time ago.

jonoomph:
LGTM

PR #3606, opened 2020-07-14
timeline css: Fix bad color value

ferdnyc:
This has been bugging me for a while — one of the CSS colors in the `.transition` styling has a missing hex digit. Fortunately CSS is forgiving enough that it doesn't break the world if we write:
```css
.transition { background-color: #0192c; }
```
...but it's still a bad value.

It works just fine like this, though:
```css
.transition { background-color: #0192c1; }
```


ferdnyc:
Ah, yes, now that cx_Freeze is back down to a compatible version and Travis is unfucked, I can merge this 1-character change.

PR #3607, opened 2020-07-15
Travis: Downgrade to cx_Freeze 6.1

ferdnyc:
It looks like Travis builds are failing because something about the newly-released cx_Freeze 6.2 (which `pip` is now installing during the setup phase) doesn't like our `freeze.py` code. Until we can get the problem sorted out, Travis can just stay locked on the previous release with `pip install cx_Freeze==6.1`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ matrix:
       python: 3.6
 
 install:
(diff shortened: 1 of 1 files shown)


ferdnyc:
Bingo. Merging.

PR #3617, opened 2020-07-21
Timeline UI: Fix sizing/placement, highlighting, hover effects for timeline nodes

ferdnyc:
This branch reworks the HTML template code and CSS structure for the DOM nodes that represent each Clip or Transition, and to my knowledge addresses _all_ of the interface's twitchy sizing issues. With the code I have here:

#### Clips never magically resize themselves.
Dragging the entire clip has no effect at all on its size. Dragging either of its left/right edges moves only that edge, and nothing else.

#### Clips never "shift around" vertically in the track lanes
With the new code, the Y coordinate that each track snaps to when being moved is the same Y coordinate it will be placed at when dropped.

#### UI feedback for hover and selection is implemented entirely in CSS
I was able to drop the `mouseover()` event handlers in the `Clip` directive, since I reimplemented those effects entirely through `:hover` styling, in concert with the existing jQueryUI `.ui-selecting` and `.ui-selected` class assignments.

#### UI feedback is consistent: clips or transitions, selected or unselected or being-selected, every object placed on the timeline responds the same way regardless of type or state
* Selecting something now _always_ turns its border red. That applies to transitions as well as clips, so you can actually see what you have selected at all times
* Hovering over something now _always_ causes any visible border styling to brighten. Again, that now applies to selected item borders (red), not just unselected item borders (blue). 

* An unselected clip will show a medium blue border that glows brighter when it's hovered over, same as before. (However, with fewer JS event handlers messing with the styling, the UI won't get into those loops where it keeps adding and removing border styles like it can't figure out which one it's supposed to be using.)
* A selected clip will now show a medium red border, which becomes a brighter red border when it's hovered over, same as for unselected clips.
* A selected transition will now show a medium red border, which becomes a brighter red border when it's hovered over, same as for a selected clip.
* Transitions and clips all get bright yellow borders when they're being lasso-selected, same as before. Same as before, this is the only mode that doesn't have a "dim" non-hover version of the styling: jQueryUI considers any node that's captured within the selection lasso to be "hovered", so only the bright yellow styling is applicable for that mode. (You can't have a node that's _in the process_ of being selected, without it also being currently hovered, IOW. The "hovered" property is extended to encompass the entire area inside the selection-rectangle, it's not confined to the single pixel directly under the mouse pointer.)


ferdnyc:
I consolidated the three different `.something_menu` CSS stanzas into a shared set of `.menu_button` styles, and the `menu_button` class is now applied to the menu buttons in all three contexts.

ferdnyc:
OK, so now conflict resolution with the webengine branch forced me to create a separate _BRANCH_ for that PR. Definitely closing this one, so I can delete this branch and reduce the confusion. It's gonna get merged to webengine anyway.

ferdnyc:
Basically, it translates from C++ to Python pretty directly. If the C++ is:
```c++
using namespace openshot;

Clip myclip1("/path/to/file.mp4");
myclip1.Open();
std::shared_ptr<Frame> frame1 = myclip1.GetFrame(1);
frame1.Save("/tmp/frame1.png", 1.0);  // (path, scale[, format="PNG", quality="100%"])
myclip1.Close();
```
then the Python is:
```Python
import openshot

myclip1 = openshot.Clip("/path/to/file.mp4")
myclip1.Open()
frame1 = myclip1.GetFrame(1)
frame1.Save("/tmp/frame1.png", 1.0)
myclip1.Close()
```

One wrinkle: all enum values are hoisted to the root of the library, so an `openshot::ChannelLayout` value of `LAYOUT_STEREO` is just `openshot.LAYOUT_STEREO`, `openshot::InterpolationType::LINEAR` is just `openshot.LINEAR`, etc. (So the contents of all enums have to be globally unique; no conflicts.)

ferdnyc:
> It's gonna get merged to webengine anyway.

It did not.

PR #3618, opened 2020-07-21
Merge new box-model code into QtWebEngine conversion

ferdnyc:
See #3617 for PR details. This is the same source branch, but targeting the WebEngine PR branch instead of `develop`.


src/timeline/index.html
```diff
@@ -75,13 +75,14 @@
...
+				<div ng-repeat="layer in project.layers.slice().reverse()" id="track_{{layer.number}}" ng-right-click="showTimelineMenu($event, layer.number)" class="{{getTrackStyle(layer.lock)}}">
 				</div>
-				
+
 				<!-- CLIPS -->
-				<div ng-hide tl-clip ng-repeat="clip in project.clips" id="clip_{{clip.id}}" ng-click="selectClip(clip.id, true, $event)" ng-right-click="showClipMenu(clip.id, $event)" class="clip droppable" ng-class="getClipStyle(clip)" style="width:{{(clip.end - clip.start) * pixelsPerSecond}}px; left:{{clip.position * pixelsPerSecond}}px; top:{{getTrackTop(clip.layer)}}px;z-index:{{1000 + $index}};">
+				<div ng-hide tl-clip ng-repeat="clip in project.clips" id="clip_{{clip.id}}" ng-click="selectClip(clip.id, true, $event)" ng-right-click="showClipMenu(clip.id, $event)" class="clip timeline_obj droppable" ng-class="getClipStyle(clip)" style="width:{{(clip.end - clip.start) * pixelsPerSecond}}px; left:{{clip.position * pixelsPerSecond}}px; top:{{getTrackTop(clip.layer)}}px;z-index:{{1000 + $index}};">
+				<div id="clip_border_{{clip.id}}" class="clip_border tl-border">
 					<div class="clip_top">
```
ferdnyc, line 84: As small a thing as it is, inserting the extra `<div class="tl-border">` is the big thing that makes all the difference.

By moving the borders to that `<div>` and unsetting `box-sizing:border-box` on the outer `<div tl-clip>`, the borders no longer alter the dimensions/placement of the main `<div tl-clip>`, which means all of the sizing/placement math becomes predictable and repeatable.

src/timeline/index.html
```diff
@@ -75,13 +75,14 @@
...
-						<div tl-clip-menu class="clip_menu" ng-show="!enable_razor" ng-mousedown="showClipMenu(clip.id, $event)" tooltip-enable="!enable_razor" tooltip="{{clip.title}}" tooltip-placement="bottom" tooltip-popup-delay="400"></div>
+						<div tl-clip-menu class="clip_menu menu_button" ng-show="!enable_razor" ng-mousedown="showClipMenu(clip.id, $event)" tooltip-enable="!enable_razor" tooltip="{{clip.title}}" tooltip-placement="bottom" tooltip-popup-delay="400"></div>
```
ferdnyc, line 85: All of the menus were using the same styling, so I created a shared `.menu_button { }` CSS stanza and applied it to all of them, eliminating the duplicated CSS.

src/timeline/js/directives/clip.js
```diff
@@ -211,21 +211,6 @@ App.directive("tlClip", function ($timeout) {
...
-            element.addClass("highlight_clip", 200, "easeInOutCubic");
-          }
-        },
-        function () {
-          if (!dragging) {
-            element.removeClass("highlight_clip", 200, "easeInOutCubic");
-          }
-        }
-      );
-
```
ferdnyc, line 228: I knew I really wanted to get rid of this, because it was fighting with the rest of the CSS rules. You could even see it happening in the UI, clips would get into a state where mousing between them would cause borders to flash, or disappear momentarily and reappear — all sorts of issues. Plus, `:hover` styling is at this point a basic CSS feature, there's no reason it couldn't be used to create the hover effects without needing to add any extra JavaScript code.

src/timeline/js/functions.js
```diff
@@ -298,10 +298,12 @@ function moveBoundingBox(scope, previous_x, previous_y, x_offset, y_offset, left
...
-  var bounding_box_padding = 3; // not sure why this is needed, but it helps line everything up
-  var results = scope.getNearbyPosition([bounding_box.left, bounding_box.right + bounding_box_padding], 10.0, bounding_box.selected_ids);
-  var nearby_offset = results[0];
-  var snapline_position = results[1];
+    var [nearby_offset, snapline_position]
+      = scope.getNearbyPosition(
+        [bounding_box.left, bounding_box.right],
+        10.0, bounding_box.selected_ids);
+  // var nearby_offset = results[0];
+  // var snapline_position = results[1];
```
ferdnyc, line 306: Heh, forgot I even did that. Our JS code creates _way_ too many temporary variables, as it's tossing around data. This eliminates one temporary array, which is something.

src/timeline/js/functions.js
```diff
@@ -298,10 +298,12 @@ function moveBoundingBox(scope, previous_x, previous_y, x_offset, y_offset, left
   }
 
   // Find closest nearby object, if any (for snapping)
-  var bounding_box_padding = 3; // not sure why this is needed, but it helps line everything up
-  var results = scope.getNearbyPosition([bounding_box.left, bounding_box.right + bounding_box_padding], 10.0, bounding_box.selected_ids);
```
ferdnyc, line 302: Doing away with `box-sizing: border-box` let me eliminate the `bounding_box_padding` adjustment to the lasso processing, which was only there (whether that comment knew it or not) to correct for the borders being included as part of the object dimensions.

src/timeline/index.html
```diff
@@ -75,13 +75,14 @@
 		<div tl-scrollable-tracks id="scrolling_tracks">
 			<div id="track-container" tl-track tl-multi-selectable style="width: {{getTimelineWidth(1024)}}px; padding-bottom: 2px;">
 				<!-- TRACKS -->
-				<div ng-repeat="layer in project.layers.slice().reverse()" id="track_{{layer.number}}" ng-right-click="showTimelineMenu($event, layer.number)"  class="{{getTrackStyle(layer.lock)}}" style="width:{{getTimelineWidth(1024)}}px;">
+				<div ng-repeat="layer in project.layers.slice().reverse()" id="track_{{layer.number}}" ng-right-click="showTimelineMenu($event, layer.number)" class="{{getTrackStyle(layer.lock)}}">
```
ferdnyc, line 78: Another unexpected optimization: I noticed we were forcing the `width` of every track `DIV` with inline styling. But they're all in a container that's _also_ forced with **the exact same width**. So I just added `width: 100%` to the track CSS, removed the inline styling, and let the container dictate the width of the tracks.

src/timeline/media/css/main.css
```diff
@@ -1,174 +1,542 @@
...
+  background: -webkit-gradient(linear, left top, left bottom,
+    color-stop(0%, rgba(50, 50, 50, 1)),
+    color-stop(100%, rgba(6, 6, 6, 1)));
+  border-top: 1px solid #4b92ad;
+  border-bottom: 1px solid #4b92ad;
+  border-right: 1px solid #4b92ad;
+  border-top-right-radius: 8px;
+  border-bottom-right-radius: 8px;
+  box-shadow: 0 0 10px #000;
+  box-sizing: border-box;
```
ferdnyc, line 179: Ironically, the problem with vertical placement of the clips/transitions was the _lack_ of `border-box` sizing on the track `DIV`s, which weren't the same size we expected them to be due to their borders pushing them out of place. We don't actually _care_ about the size of the tracks themselves, we just care that they're in the positions we expect them to be. So, using `box-sizing: border-box` ensures that the (bordered) track containers will be the same size as the (now-unbordered) draggables, and they line up correctly.

src/timeline/media/css/main.css
```diff
@@ -1,174 +1,542 @@
...
+}
+
+/* Selected item border, semi-transparent unless hovered */
+.ui-selected>.tl-border {
+  border-color: #ff0000a8;
+}
+
+.ui-selected:hover>.tl-border {
+  border-color: #ff0000ff;
+}
```
ferdnyc, line 470: Because the `<div tl-track>` is a container for `<div class="tl-border">`, which is in turn a container for all of the rest of its contents, my initial attempts at dicking around with setting `opacity` to create hover effects all failed. I couldn't change the opacity of the border `<div>` without everything inside it also becoming semi-transparent. So, I finally hit on the idea of just assigning semi-transparent _colors_ to the borders, in the non-hovered state.

PR #3621, opened 2020-07-23
Travis: Add Ubuntu Focal (Python 3.8) build

ferdnyc:
We've been building `libopenshot` and `libopenshot-audio` on Ubuntu Focal for a little while now, since our FFmpeg 4 backport PPA went bye-bye. For some reason, I never thought to add it to OpenShot's build matrix. But it makes sense now, especially as Xenial will have to go away when WebEngine gets merged (its Qt is too old).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,10 @@ matrix:
       os: linux
       dist: bionic
       python: 3.6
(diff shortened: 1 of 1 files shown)


ferdnyc:
Merging for more comprehensive CI.

Issue #3620, opened 2020-07-23
My project files tab is missing


TheBigone0:
The first time I opened open shot the project tiles tab was there but when I went to chrome, when I came back the tab was gone! Anyone know what I did to make it disappear?

ferdnyc:
@TheBigone0 Hmm, strange, not sure what could've caused that. You can bring it back by right-clicking on any of the other dock tabs or titlebars. You'll see a menu listing all of the docks, with checkmarks to enable/disable each one.

ferdnyc:
@TheBigone0 Sorry, I was mistaken. The dock tabs are one of the few areas you _can't_ click, to get the context menu. But if you click anywhere on the toolbar, the status bar — the "window surface" in general, basically — of the OpenShot window, you'll get a popup menu that lets you hide/show the various docks and toolbars.

I also just submitted #3622, which will add a **Docks** submenu to the main **View** menu, since that's a natural place people would generally look for such a list. Should've done that ages ago.

PR #3622, opened 2020-07-23
Add Docks submenu to View menu, to easily toggle dock visibility

ferdnyc:
The menu will contain all of the closable docks, with checkboxes showing which are visible/invisible, so they can be toggled conveniently.

_Hopefully_ the list will be sorted in localized order. (I'm using `dock.windowTitle()` as the sort key, which should be the translated name, but I'm having Python do the sorting so I'm not sure it'll DTRT for the current interface language if it doesn't use English-language sorting rules.)

Also: A bunch of code cleanup/formatting in `main_window.py`, including the elimination of all `*` imports.

Fixes: #3620


jonoomph:
Looks great! I've resolved 2 conflicts, but I think it should be good to merge now!

PR #3629, opened 2020-07-28
Prev/Next Marker jumps to 0 (and end, w/o selection)

ferdnyc:
- Always add the timeline's first frame to the marker list
- When nothing is selected, also add the last frame (right edge of latest clip)

Along the way I created `conversion.getMaxTime()` which takes a list of either `openshot.Clip()` or `classes.query.Clip` objects, and finds the end of the latest time on that list. That function replaces three copies of the same code in `actionPlay_trigger`, `actionJumpEnd_trigger`, and `export.updateProgressBar`, as well as the fourth copy I would've had to add to `findAllMarkerPositions`.

And I did a bunch of formatting/cleanup in export, and a bit in main_window


ferdnyc:
This is outdated, with the new libopenshot APIs. I'll work on a new version.

PR #3637, opened 2020-07-31
title editor: Fix name-duplication regex

ferdnyc:
The previous version was treating the parentheses as non-optional, and could come out with no match at all if the input name lacked them, leading to a traceback instead of the expected title editor window.

The new code will guarantee that `match.group(1)` always has some contents, even if — in the very degenerate case — the filename is something insane like `(1).svg`.

(In which case, `match.group(1) == '(1)'` and the rest of the match groups are empty. This would result in the output filename for the copy being generated as `(1) (1).svg`. Which I would argue is actually more correct than / preferable to `(2).svg`, in that admittedly pathological case.)

### DIFF
--- a/src/windows/title_editor.py
+++ b/src/windows/title_editor.py
@@ -249,13 +249,19 @@ class TitleEditor(QDialog):
             if self.duplicate and self.edit_file_path:
                 # Re-use current name
                 name = os.path.basename(self.edit_file_path)
-                # Splits the filename into "[base-part][optional space]([number]).svg
-                match = re.match(r"^(.*?)(\s*)\(([0-9]*)\)\.svg$", name)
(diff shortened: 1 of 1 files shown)


ferdnyc:
Merging this, also an unhandled failure condition observed in the wild. (And a pretty localized change, only impacting the Title Editor.)

ferdnyc:
(Or, well, I _would_, if Travis would get its cranium out of its posterior...)

ferdnyc:
I don't get why Travis just decides to hate me sometimes. :rage:

ferdnyc:
OK, finally convinced Travis to sort itself out. Merging quick before it breaks again. :weary:

PR #3648, opened 2020-08-04
Fix lockfile logic

ferdnyc:
The code to delete the lockfile was treating every exception as a retry-able failure, including `FileNotFoundError`. Obviously, if the file's not there, attempting four more times to delete it is counterproductive, so now we take that exception as a reason to immediately give up trying.


ferdnyc:
Merging this as it fixes a well-understood failure condition which has been observed in the wild multiple times. (It's only visible when running wirh `--debug`, which is why we hadn't noticed it until now.)

PR #3653, opened 2020-08-07
About dialog: Enhance changelog parsing, add support for new format

ferdnyc:
I finally came up with the perfect solution to OpenShot/libopenshot#486 and OpenShot/libopenshot-audio#92 (my proposed new Git changelog formats), which have been in WIP status pending compatibility with the About dialog's changelog listing.

This PR updates `about.py` to:
- Rework the existing parsing into reusable function `parse_changelog()`, called once per project
- Add a second, alternative `parse_new_changelog()` to read the new-format files
- Check for _both_ file types, the old-style files in <code>settings/<var>project</var>.log</code>, the _new_ files in <code>resources/<var>project</var>.log</code>.

If it finds a new-style file, it'll read it. If it can't, or it's not found, it'll try for an old-style file. If _neither_ can be found/read, it'll remove that individual tab from the dialog, so that there aren't any empty listings. (If none of the 6 files is present, the Changelog button is removed from the About dialog completely, same as always.)

That means that this can (and should) be merged _before_ any format changes. It'll keep on supporting the old format, and then at whatever point we choose we can switch over to generating _new_ format files and placing them in `resources/` instead.


src/windows/about.py
```diff
@@ -43,6 +44,54 @@
...
+        return None
+    changelog_list = None
+    for encoding_name in ('utf_8', 'utf_16'):
+        try:
+            with codecs.open(changelog_path, 'r', encoding=encoding_name) as changelog_file:
+                # Generate match object with fields from all matching lines
+                matches = re.findall(
+                    r"^-\s?([0-9a-f]{40})\s(\d{4,4}-\d{2,2}-\d{2,2})\s(.*)\s\[(.*)\]\s*$",
+                    changelog_file.read(), re.MULTILINE)
+                log.debug("Parsed {} changelog lines from {}".format(len(matches), changelog_file))
```
ferdnyc, line 80: Dummy.

```suggestion
                log.debug("Parsed {} changelog lines from {}".format(len(matches), changelog_[ath))
```

jonoomph:
@ferdnyc What about the `for encoding_name in ('utf_8', 'utf_16'):` support for Windows changelog files. Is this no longer needed? I found it necessary due to Windows' git changelog was (for some reason) in the wrong format.

jonoomph:
I resolved the conflicts on this one, so we can merge it. If we need to handle the different UTF formats for Windows, we can always make a new PR. I'll test this one shortly.

ferdnyc:
@jonoomph Sorry, missed that question until now. 

The different encoding attempts are still there. I moved that code into a reusable function `parse_changelog` (at the start of `about.py`), so it can be called for each of the files. But the code there is pretty much the same as before, just DRYer.

PR #3662, opened 2020-08-10
Use delayed-action QTimers properly

ferdnyc:
Our code makes liberal use of `QTimer`s to take action after a delay, which is good and a sensible reason to use them.

However, it uses them all _wrong_, creating them as repeating timers, then explicitly `start()`ing and `stop()ping` them to make sure the callback only fires once. That's... unnecessary.

If a timer is simply meant to introduce a delay before some action, it can be `.setSingleShot(True)`. Then:
  * `start()` will cause it to run for `interval()` ms, trigger once, and then auto-reset to `stop()`ped.
  * Calls to `start()` WHILE it's running will reset the timeout to `interval()` ms. (Technically it's stopped-and-restarted.)
  * It will only ever run its `timeout()` callback ONCE per `start()` from a stopped state, even if it's reset multiple times.
  * It never needs to be explicitly `stop()`ped except to cancel it.

This PR updates all of the timers used in this fashion to be `setSingleShot(True)`, and removes the explicit `stop()`s from all of their own callbacks.

Also: Cleanup in aisle `properties_tableview.py`


jonoomph:
My brain can't follow this logic... guess it's time for bed. For example, the "display property" timer is used to prevent spamming the loading of properties... It's not really a singleshot scenario, but maybe I'm just not reading it correctly. I'll take a look in the morning. :+1:

jonoomph:
@ferdnyc I resolved conflicts, but found 1 instance of incorrect indention. So perhaps you can take a look, and make sure this PR still looks good to you.

src/windows/main_window.py
```diff
@@ -2897,8 +2896,8 @@ def __init__(self, mode=None):
         self.show_property_type = None
         self.show_property_timer = QTimer()
         self.show_property_timer.setInterval(100)
```
jonoomph, line 2898: So, with setSingleShot(True), will the interval still work? For example, if 500 starts of this timer, will it keep canceling the current 100 ms timers? Or will it actually create 500 calls to `show_property_timeout`)... which is not want anyone wants of course.

ferdnyc:
@jonoomph 

> My brain can't follow this logic... guess it's time for bed. For example, the "display property" timer is used to prevent spamming the loading of properties... It's not really a singleshot scenario, but maybe I'm just not reading it correctly. For example, if 500 rapid "display properties" events happen, perhaps we only want to draw the properties 1 time. We don't want 500 singleshot timers. 

Right, no, no, definitely not. And Qt single shot timers actively work to **prevent** that happening. (Permanent single shot timers, that is, which are different from ad-hoc one-off ones.)

When you `start()` a permanent Qt singleshot timer, it starts counting down. If you `start()` it AGAIN before it times out, it just resets the countdown. No matter how many times it's (re-)started, it will only time out **once** when the countdown is finally able to complete without interruption. 

IOW, it does _exaaaactly_ what we want, automatically. And we never have to bother with explicitly stopping them, unless we need to abort a countdown for some reason.

I'll take a look at the commits later, I'm on my phone right now.

src/windows/main_window.py
ferdnyc: Each additional `start()` while the timer's already running will reset it back to 100ms, but leave it running. It will only issue a single `timeout` signal, after it's allowed to run for 100ms uninterrupted (un-reset, IOW) — at which point it'll automatically revert back to stopped state, until it's `start()`ed again.

src/windows/models/properties_model.py
```diff
@@ -320,14 +331,18 @@ def color_update(self, item, new_color, interpolation=-1, interpolation_details=
                                     point["handle_left"]["X"] = interpolation_details[2]
                                     point["handle_left"]["Y"] = interpolation_details[3]
 
-                                log.debug("updating interpolation mode point: co.X = %s to %s" % (point["co"]["X"], interpolation))
-                                log.debug("use interpolation preset: %s" % str(interpolation_details))
+                                log.info(
+                                    "updating interpolation mode point: co.X = %s to %s"
+                                    % (point["co"]["X"], interpolation))
+                                log.info("use interpolation preset: %s" % str(interpolation_details))
```
ferdnyc, line 337:
```suggestion
                                log.debug(
                                    "updating interpolation mode point: co.X = %d to %d",
                                    point["co"]["X"], interpolation)
                                log.debug("use interpolation preset: %s", str(interpolation_details))
```

src/windows/models/properties_model.py
```diff
@@ -307,8 +316,10 @@ def color_update(self, item, new_color, interpolation=-1, interpolation_details=
                                     point["handle_right"]["X"] = interpolation_details[0]
                                     point["handle_right"]["Y"] = interpolation_details[1]
 
-                                log.debug("updating interpolation mode point: co.X = %s to %s" % (point["co"]["X"], interpolation))
-                                log.debug("use interpolation preset: %s" % str(interpolation_details))
+                                log.info(
+                                    "updating interpolation mode point: co.X = %s to %s"
+                                    % (point["co"]["X"], interpolation))
+                                log.info("use interpolation preset: %s" % str(interpolation_details))
```
ferdnyc, line 322:
```suggestion
                                log.debug(
                                    "updating interpolation mode point: co.X = %d to %d",
                                    point["co"]["X"], interpolation)
                                log.debug("use interpolation preset: %s", str(interpolation_details))
```

src/windows/models/properties_model.py
```diff
@@ -294,7 +301,9 @@ def color_update(self, item, new_color, interpolation=-1, interpolation_details=
                                 clip_updated = True
                                 # Update point
                                 point["co"]["Y"] = new_value
-                                log.debug("updating point: co.X = %s to value: %s" % (point["co"]["X"], float(new_value)))
+                                log.info(
+                                    "updating point: co.X = %s to value: %s"
+                                    % (point["co"]["X"], float(new_value)))
```
ferdnyc, line 306:
```suggestion
                                log.debug(
                                    "updating point: co.X = %d to value: %.3f"
                                    point["co"]["X"], float(new_value))
```

src/windows/models/properties_model.py
```diff
@@ -30,16 +30,20 @@
...
-from PyQt5.QtGui import *
+from PyQt5.QtGui import (
+    QStandardItemModel, QStandardItem,
+    QPixmap, QColor,
+    )
 
-from classes import info, updates, openshot_rc
-from classes.query import Clip, Transition, Effect, File
+from classes import info, updates
+from classes.query import Clip, Transition, Effect
```
ferdnyc, line 39:
```suggestion
from classes import info, updates
from classes import openshot_rc  # noqa
from classes.query import Clip, Transition, Effect
```

jonoomph:
Awesome, thanks!

Issue #3624, opened 2020-07-26
Openshot version 2.5.1 not able to find assets with foreign characters when opening a file from Openshot 2.4.x


giray:
**Describe the bug:**
OpenShot 2.5.1 (OpenShot-v2.5.1-x86_64.AppImage) is not able to find assets of an .OSP file created in 2,4,3 (Ubuntu) with assets using foreign characters.

**Steps to reproduce the behavior:**
1. Create project using 'openshot-2.4.3'
2. Add Assets with names such as
- `intubated yaşlı.jpeg`
- `Hemşirenin Gizli Kamera Çekimleri.mp4`
3. Use assets and save project
4. Run OpenShot-v2.5.1
5. Open osp file
6. Get "`Missing File - intubated ya\u015fl\u0131.jpeg cannot be found`" message

Pointing to the directory where the asset resides does not resolve the issue

**Expected behavior:**
Expected to find assets even though they 
**System Details:**
- OpenShot Version : Ubuntu 2.4.3 to OpenShot-v2.5.1-x86_64.AppImage
- Operating System / Distro: Tested on ubuntu 20.04 and ubuntu 19.10

**Log Files:**
```
Loaded modules from current directory: /tmp/.mount_8XFKMA/usr/bin
         app:INFO ------------------------------------------------
         app:INFO             Sun Jul 26 09:42:16 2020            
         app:INFO               Starting new session              
         app:INFO ------------------------------------------------
         app:INFO             OpenShot (version 2.5.1)            
         app:INFO ------------------------------------------------
         app:INFO openshot-qt version: 2.5.1
         app:INFO libopenshot version: 0.2.5
         app:INFO platform: Linux-5.4.0-40-generic-x86_64-with-Ubuntu-20.04-focal
         app:INFO processor: x86_64
         app:INFO machine: x86_64
         app:INFO python version: 3.4.3
         app:INFO qt5 version: 5.2.1
         app:INFO pyqt5 version: 5.2.1
    language:INFO Qt Detected Languages: ['en-US']
    language:INFO LANG Environment Variable: en_US.UTF-8
    language:INFO LOCALE Environment Variable: 
    language:INFO OpenShot Preference Language: Default
project_data:INFO Setting default profile to HD 720p 30 fps
         app:INFO Setting font to /tmp/.mount_8XFKMA/usr/bin/images/fonts/Ubuntu-R.ttf
logger_libopenshot:INFO Connecting to libopenshot with debug port: 5556
[...]

PR #3664, opened 2020-08-11
Fix loading Unicode paths from 2.4 (#3624)

MarcelWaldvogel:
This is the hotfix approach from #3624. It would be cleaner to not mess with the string representation of the file.

Coded adapted from the hotwiring I did for the 2.5.1 files, no in-branch testing, no test cases. Feel free to complain ;-)

### DIFF
--- a/src/classes/json_data.py
+++ b/src/classes/json_data.py
@@ -39,7 +39,7 @@ from classes import info
 from classes.app import get_app
 
 # Compiled path regex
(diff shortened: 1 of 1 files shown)


jonoomph:
Looks solid to me! Thanks, let's merge it.

jonoomph:
Okay, this break. I should have headed the warnings, lol
`"path": "apps/openshot-qt-git/src/\"/home/jonathan/Videos/TearsOfSteel.mp4\"",`

PR #3681, opened 2020-08-25
timeline-mixins: Enable WebEngine JS logging

ferdnyc:
NOTE: I've added a version of this code to #3684, so if we go that route this PR is unnecessary.

Getting logging working via the QtWebEngine required some extra finessing. But, on the plus side, now they have three logging levels! So, we map those to `logging.log` severity levels when passing along the logs.


ferdnyc:
Still no movement on #3684, so merging this to get logging working under QtWebEngine.

PR #3700, opened 2020-08-31
Fix "Add track" action (toolbar icon, keyboard shortcut)

### LINKED ISSUE #3696: "Add Track" Timeline menu icon has no effect
2.5.1-dev2 Win10
Above the Timeline are some icons; the left one is a plus, which says "Add Track" when hovered over. It's next to "Snapping Enable."

It does not add a track, or do anything else.

I've spent 45 mins searching for an existing issue on this - sorry if I've missed it. These icons are undocumented, so I guess it's possible this works in an unexpected way, but I can't imagine what it is, if so.

ferdnyc:
Seems I tried to get too clever with Python iterators.

Also squashes a bug I hit in the JS `updateLayerIndex` implementation.

Fixes #3696

### DIFF
--- a/src/classes/logger_libopenshot.py
+++ b/src/classes/logger_libopenshot.py
@@ -34,6 +34,9 @@ import zmq
 
 
 class LoggerLibOpenShot(Thread):
+    def __init__(self):
+        super().__init__()
+        self.daemon = True
 
     def kill(self):
         self.running = False
--- a/src/classes/metrics.py
+++ b/src/classes/metrics.py
@@ -99,6 +99,7 @@ def track_metric_screen(screen_name):
     metric_params["cid"] = s.get("unique_install_id")
 
     t = threading.Thread(target=send_metric, args=[metric_params])
+    t.daemon = True
     t.start()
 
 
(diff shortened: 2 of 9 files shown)


ferdnyc:
Hrm. Travis is doing that thing again where one of the two builds for the PR goes out to lunch and never comes back.

If I visit the Travis CI website, the job is completed and successful. But the Github Actions interface never notices, and there's no way I can see to tell it, "Check again." (In fact, if I forcibly restart the job, the same thing will happen again.)

Adding a commit to the PR nearly always fixes it, but that's a pain and we shouldn't have to. I need to open a bug report with them.

ferdnyc:
I only just today noticed a critical data point about this problem, and have [documented it to the Travis support forum](https://travis-ci.community/t/duplicate-job-attempts-for-pr-builds-causing-queued-build-created-problems/9776). Now we wait...

ferdnyc:
I also noticed that my fork still had an old Webhook from travis-ci.org installed on it, so I removed that. I'm wondering if maybe the same thing might be true on the OpenShot repo end. Time to email @jonoomph ...

ferdnyc:
Stuffed in a totally unrelated, but good-to-have, change to unstick the build.

jonoomph:
LGTM! Merge away!

PR #3701, opened 2020-08-31
Travis: Re-enable QtWebKit for Xenial builds

ferdnyc:
Since we ended up retaining QtWebKit support, we can use that when running CI builds on Xenial.


ferdnyc:
Hmm, I suppose this means Xenial should also no longer be "allowed to fail".

ferdnyc:
Done, and merging. Welcome back, Ubuntu Xenial.

PR #3702, opened 2020-08-31
timeline: Fix HTML source processing, load WebChannel from Qt resource system

ferdnyc:
- Don't leave file handles dangling open, when reading HTML file
- Change mixin placeholder to an HTML comment, so the source is valid with or without replacement

Fixes this error spit out when Travis ran the unit tests:
```
logger:ERROR /home/travis/build/OpenShot/openshot-qt/src/windows/views/timeline_mixins.py:124:
 ResourceWarning: unclosed file
 <_io.TextIOWrapper name='/home/travis/build/OpenShot/openshot-qt/src/timeline/index.html' mode='r' encoding='utf-8'>
logger:ERROR   html = open(self.html_path, 'r', encoding='utf-8').read()
```

Also fixes a few `setInterval()` calls we were making using float arguments, leading to:
```
preview_thread:INFO initPlayer
logger:ERROR /home/travis/build/OpenShot/openshot-qt/src/windows/main_window.py:2787:
 DeprecationWarning: an integer is required (got type float).
  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
logger:ERROR   self.auto_save_timer.setInterval(s.get("autosave-interval") * 1000 * 60)
```

- Added a clever trick I found somewhere for handling `qwebchannel.js`: That file is automatically present in Qt builds that support it via the Qt resource system. `QtWebEnginePage` supports accessing Qt resources from HTML as `qrc:` URLs. So instead of keeping `qwebchannel.js` on the filesystem and conditionally inserting code to load it, we can just place a `<script>` tag in the HTML that loads the source as `qrc:/qtwebchannel/qwebchannel.js`. If the code isn't running in a QtWebEngine process, that URL will be invalid and the script tag will be silently ignored.


jonoomph:
Nice! LGTM

PR #3709, opened 2020-09-07
A few small timeline fixes

ferdnyc:
This PR contains a few small tweaks to things that affect how OpenShot operates very minimally, if at all.

* Added one more that does technically change functionality, though the change _should_ be invisible to users. (And I've been wanting to do it for a while now.) By styling the `div.playhead-top` element that holds the playhead "handle" with a `margin-left: -12px`, and tweaking some of the other playhead component positioning to match, the playhead's "anchor" point is moved to its center, instead of being at its left edge. This means that all of the `$scope.playheadOffset` adjustments the code makes become completely unnecessary. So, this PR applies that negative margin change, and removes `$scope.playheadOffset` from all of the calculations where it was being used (as well as removing the variable from the scope itself), making for slightly simpler, cleaner code.
* A small fix to the WebKit backend, which has been injecting the Qt `MainWindow` object into the JavaScript environment (as `mainWindow`). That's a _lot_ of data to inject, because `main_window.py` creates a big, heavy object full of bindings and local storage.
    But then `mainWindow` is never once used for anything on the JS side, making it totally unnecessary. Gone now.
* A few references to `src/images/AudioThumbnail.png` were all changed to reference `src/images/AudioThumbnail.svg`, as the PNG version now only exists in the resource file. This mostly only **affects** the Timeline debug mode when it's running in a web browser, though technically most of the references appeared in the Python code. I'm reasonably sure those were only used as fallback/legacy paths, though, from before the thumbnail server.
* I removed several files from `src/timeline/media/images/` that aren't used for anything. (For the most part, they never really were. These were early development versions, dating back to 2014 with no changes since, of things that went in a different direction. Like a first-draft set of effect icons, before that design morphed into the HTML-based "Skittles" we use today.)


ferdnyc:
Merging, as these are all nice-to-haves and should hopefully be pretty unobjectionable.

PR #3711, opened 2020-09-08
Fix large canvas rendering (at closest zoom levels), with WebEngine

ferdnyc:
While running some tests with the most recent AppImage Daily Build, I loaded a project file and noticed that the ruler _completely_ disappeared at 1-second zoom ­— it wasn't merely cut off, the whole thing was gone.

Turned out, the project had a duration of 644 seconds, which at 1-second zoom made the total timeline width `64450` pixels, well in excess of the 32Kpixel max supported width for HTML `<canvas>` elements. When I zoomed out to the 3-second level (21000-and-change pixels wide), the ruler reappeared.

It appears that the WebEngine layout engine, in contrast to WebKit, will _completely_ fail to render a canvas element that exceeds the max supported dimensions. With WebKit, the canvas was merely cut off at the maximum width, but WebEngine is clearly more fragile.

This PR adds a `canvasMaxWidth()` constraint function to the `$scope` in `controller.js`, which takes a single numerical argument (`desired_width`) and executes one statement: `return Math.min(32767, desired_width);` All code for computing canvas width in both the HTML and the JavaScript now wraps all of those computations in `canvasMaxWidth()` to ensure that none of the canvas elements exceed their maximum allowable dimensions.

The audio-waveform rendering appears a bit "off" in this code, placed oddly and sticking out onto the border of the Clip area slightly. But that's identical to how it renders with the current `develop` branch code _before_ this PR, so **#NOTMYFAULT**.

Might fix some open Issues, not sure. I haven't checked in a few days.

### DIFF
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -467,8 +467,8 @@ App.controller("TimelineCtrl", function ($scope) {
       var stop_second = parseFloat(progress[p]["end"]) / fps;
 
       //figure out the actual pixel position, constrained by max width
-      var start_pixel = canvasMaxWidth(start_second * $scope.pixelsPerSecond);
-      var stop_pixel = canvasMaxWidth(stop_second * $scope.pixelsPerSecond);
+      var start_pixel = $scope.canvasMaxWidth(start_second * $scope.pixelsPerSecond);
+      var stop_pixel = $scope.canvasMaxWidth(stop_second * $scope.pixelsPerSecond);
       var rect_length = stop_pixel - start_pixel;
       if (rect_length < 1) {
         break;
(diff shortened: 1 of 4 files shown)


src/timeline/js/controllers.js
```diff
@@ -619,6 +621,12 @@ App.controller("TimelineCtrl", function ($scope) {
     }
   };
 
+  // Constrain canvas width values to under 32Kpixels
+  $scope.canvasMaxWidth = function (desired_width) {
+    return Math.min(32767, desired_width);
+  };
+
+
```
ferdnyc, line 629:
```suggestion
```

ferdnyc:
I'mma merge this one, since having no ruler _at all_ on larger projects is kind of a pain, and this change at least brings things back to the way they were pre-WebEngine.

PR #3715, opened 2020-09-11
Allow CTRL to multi-select timeline items (clips and transitions)

jonoomph:
Adapting from another PR (https://github.com/OpenShot/openshot-qt/pull/3317), and making some changes. This adds the ability to hold CTRL to select and unselect items (clips and transitions).

### DIFF
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -517,10 +517,12 @@ App.controller("TimelineCtrl", function ($scope) {
     // Trim clip_id
     var id = clip_id.replace("clip_", "");
 
-    // Clear transitions also (if needed)
-    if (id !== "" && clear_selections) {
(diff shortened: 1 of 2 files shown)


jonoomph:
I accidentally included a commit from another PR. Oops. In theory, after my previous PR is merged, this should be resolved I think.

ferdnyc:
On this subject, I recently filed #3716 — seems WebEngine broke <kbd>Ctrl</kbd>+<kbd>A</kbd> functionality on the Timeline.

ferdnyc:
@jonoomph 

Hmm, I seem to have hit a bug in this (separate from #3716) — now if you right-click on an unselected Timeline object, it is _added_ to the current selection. So you can accidentally keep growing the selection just by right-clicking one Clip/Transition at a time.

PR #3721, opened 2020-09-13
Teach Prev/Next Marker to jump to 0, end

ferdnyc:
- Prev Marker will always jump to the first frame on the Timeline
- If nothing is selected, Next Marker will also jump to the very end

##### Unrelated change I snuck in (tsk)
- Change a lot of our `actionFoo_Trigger(self, event)` signatures to `actionFoo_trigger(self, checked=True)`. simply because the action triggers **don't take event objects as arguments**. Only event handlers do that.


jonoomph:
Looks good, thanks!

PR #3722, opened 2020-09-13
Change of name of library

eisneinechse:
With the official inclusion of SVT-AV1 the library changed from svt_av1 to svtav1. This takes this into account

### DIFF
--- a/src/presets/format_mp4_libsvtav1.xml
+++ b/src/presets/format_mp4_libsvtav1.xml
@@ -4,7 +4,7 @@
 	<type translatable="True">All Formats</type>
 	<title translatable="True">MP4 (AV1 svt)</title>
 	<videoformat>mp4</videoformat>
-	<videocodec>libsvt_av1</videocodec>
+	<videocodec>libsvtav1</videocodec>
 	<audiocodec>libvorbis</audiocodec>
 	<audiochannels>2</audiochannels>
 	<audiochannellayout>3</audiochannellayout>


jonoomph:
Not sure why Travis completed successfully (when I check travis-ci.com), but it shows "queued" here. @ferdnyc Any ideas? Travis just seems very unstable recently.

ferdnyc:
There's actually _three_ copies of the same job listed, on this one. So, I suspect @eisneinechse is in the same state I was, with Travis Webhooks.

@eisneinechse : Can you visit https://github.com/eisneinechse/openshot-qt/settings/hooks and see if you have _anything_ listed for Travis there? If so, you should delete those hooks, they're outdated infrastructure from the old, pre-Github Actions Travis setup.

ferdnyc:
@eisneinechse Also:
1. If you visit  https://github.com/eisneinechse/openshot-qt/settings/installations, there should only be one Travis CI listed there. (It's unlikely that'll be anything different.)
1. You may also want to visit https://travis-ci.org/account/repositories and check that your OpenShot repos are listed as being migrated to Travis-ci.com — or at least, that they're _not_ showing as having Legacy Services Integration enabled at the .org site. (The switch being switched "on" would indicate legacy services are activated; if they're migrated, you'll see this instead of any controls:

image

ferdnyc:
However, apart from that, pushing **any** random commit of any sort to the PR branch should unstick this — checks are only duplicated on the initial PR creation, in my experience. Not sure why things are bleeding over from forks to the upstream like this, but it seems to be a one-time thing when it happens. (Once per PR, that is.)

ferdnyc:
Oh, never mind, it's already merged. :laughing: Well, my suggestions for why that happened still stand. It's likely old travis-ci.org configuration in @eisneinechse 's fork.

Issue #3708, opened 2020-09-06
When are logging warnings used?


MartinThoma:
I'm currently writing an article about logging in Python. I'm looking at several big projects to see what the best practices are. One part that I'm not certain about is when exactly which log levels are used and what is logged.

This is how I used it so far:

* **log.debug**: If there are bugs that happen often enough that one should deal with them, but all other tries so far to pin them down failed. Kind of "debugging in production". Often, I remove those once I've found the issue.
* **log.info**: This is what I use most of the time. Key parts of the application are logged here. If something is more time-consuming, I tend to log it more thoroughly.
* **log.warning**: Something that could become a problem or indicate a problem, e.g. free disk space or a timeout when trying to reach a website. 
* **log.error**: Something failed which is user-facing, e.g. the user wanted to save a file and the location was not writable
* **log.critical**: I think I've never used it :thinking: 
* **warnings.warn**: Deprecations or missing environment options that are non-critical (e.g. a BLAS is missing)

I've copied the lines where logging is used in OpenShot below. It would be awesome if you could tell me what the best practices are :pray: 

## log.warning

This is where openshot uses logging.warning:

```
freeze.py:268:            log.warning("Skipping {}: {}".format(mod_name, ex))
src/classes/language.py:148:            log.warning('Failed to parse language for %s' % locale_name)
src/classes/app.py:118:            log.warning("Error displaying dependency/system details", exc_info=1)
src/classes/app.py:294:        log.warning('Failed to write session ended log')
src/classes/project_data.py:71:            log.warning("Cannot get empty key.")
src/classes/project_data.py:152:            log.warning("_set() key must be a list. key: {}".format(key))
src/classes/project_data.py:155:            log.warning("Cannot set empty key.")
src/classes/project_data.py:259:                log.warning("Unable to load user project defaults from {}: {}".format(info.USER_DEFAULT_PROJECT, ex))
src/classes/importers/final_cut_pro.py:145:                        log.warning('Error building File object for %s' % clip_path, exc_info=1)
src/classes/importers/edl.py:99:            log.warning('Error building File object for %s' % clip_path, exc_info=1)
src/classes/json_data.py:200:        log.warning(msg)
src/classes/thumbnail.py:66:        log.warning("Could not parse rotation value {}: {}".format(rotate_data, ex))
src/classes/thumbnail.py:68:        log.warning("Error reading rotation metadata from {}".format(file_path), exc_info=1)
[...]

PR #3724, opened 2020-09-16
Adjust log levels

MartinThoma:
Related to issue #3708

### DIFF
--- a/src/classes/metrics.py
+++ b/src/classes/metrics.py
@@ -200,6 +200,5 @@ def send_exception(stacktrace, source):
         try:
             r = requests.post(url, data=data, headers={"user-agent": user_agent, "content-type": "application/x-www-form-urlencoded"}, verify=False)
             log.info("Track exception: [%s] %s | %s", r.status_code, r.url, r.text)
-
-        except Exception as ex:
+        except Exception:
             log.warning("Failed to track exception", exc_info=1)
--- a/src/classes/metrics.py
+++ b/src/classes/metrics.py
@@ -64,8 +64,8 @@ try:
         # Get the distro name and version (if any)
         linux_distro = "-".join(platform.linux_distribution())
 
-except Exception as Ex:
-    log.error("Error determining OS version in metrics.py")
+except Exception:
+    log.debug("Error determining OS version in metrics.py")
 
 # Build user-agent
 user_agent = "Mozilla/5.0 (%s) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/37.0.2062.120 Safari/537.36" % os_version
@@ -173,8 +173,8 @@ def send_metric(params):
                 r = requests.get(url, headers={"user-agent": user_agent}, verify=False)
                 log.info("Track metric: [%s] %s | (%s bytes)" % (r.status_code, r.url, len(r.content)))
 
-            except Exception as Ex:
-                log.error("Failed to Track metric: %s" % (Ex))
+            except Exception as ex:
+                log.warning("Failed to Track metric: %s", ex)
 
             # Wait a moment, so we don't spam the requests
             time.sleep(0.25)
@@ -199,7 +199,7 @@ def send_exception(stacktrace, source):
         # Send exception HTTP data
         try:
             r = requests.post(url, data=data, headers={"user-agent": user_agent, "content-type": "application/x-www-form-urlencoded"}, verify=False)
-            log.info("Track exception: [%s] %s | %s" % (r.status_code, r.url, r.text))
+            log.info("Track exception: [%s] %s | %s", r.status_code, r.url, r.text)
 
-        except Exception as Ex:
-            log.error("Failed to Track exception: %s" % (Ex))
+        except Exception as ex:
+            log.warning("Failed to Track exception: %s", ex)
--- a/src/classes/metrics.py
+++ b/src/classes/metrics.py
@@ -202,4 +202,4 @@ def send_exception(stacktrace, source):
             log.info("Track exception: [%s] %s | %s", r.status_code, r.url, r.text)
 
         except Exception as ex:
-            log.warning("Failed to Track exception: %s", ex)
+            log.warning("Failed to track exception", exc_info=1)
--- a/src/classes/metrics.py
+++ b/src/classes/metrics.py
@@ -65,7 +65,7 @@ try:
         linux_distro = "-".join(platform.linux_distribution())
 
 except Exception:
-    log.debug("Error determining OS version in metrics.py")
+    log.debug("Error determining OS version", exc_info=1)
 
 # Build user-agent
 user_agent = "Mozilla/5.0 (%s) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/37.0.2062.120 Safari/537.36" % os_version
--- a/src/classes/metrics.py
+++ b/src/classes/metrics.py
@@ -174,7 +174,7 @@ def send_metric(params):
                 log.info("Track metric: [%s] %s | (%s bytes)" % (r.status_code, r.url, len(r.content)))
 
             except Exception as ex:
-                log.warning("Failed to Track metric: %s", ex)
+                log.warning("Failed to track metric", exc_info=1)
 
             # Wait a moment, so we don't spam the requests
             time.sleep(0.25)
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -68,7 +68,7 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
 
         # Verify key is valid type
         if not key:
-            log.debug("ProjectDataStore cannot get empty key (%s)", key)
+            log.warning("ProjectDataStore cannot get empty key.")
             return None
         if not isinstance(key, list):
             key = [key]
@@ -149,10 +149,10 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
 
         # Verify key is valid type
         if not isinstance(key, list):
-            log.debug("_set() key must be a list. key=%s", key)
+            log.warning("_set() key must be a list. key=%s", key)
             return None
         if not key:
-            log.debug("Cannot set empty key (key=%s)", key)
+            log.warning("Cannot set empty key (key=%s)", key)
             return None
 
         # Get reference to internal data structure
@@ -256,7 +256,7 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
             try:
                 self._data = self.read_from_file(info.USER_DEFAULT_PROJECT)
             except (FileNotFoundError, PermissionError) as ex:
-                log.debug("Unable to load user project defaults from %s: %s", info.USER_DEFAULT_PROJECT, ex)
+                log.warning("Unable to load user project defaults from %s", info.USER_DEFAULT_PROJECT, exc_info=1)
             except Exception:
                 raise
             else:
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -693,9 +693,9 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
                         # Increment track counter
                         track_counter += 1
 
-            except Exception as ex:
+            except Exception:
                 # Error parsing legacy contents
-                msg = "Failed to load project file %(path)s: %(error)s" % {"path": file_path, "error": ex}
+                msg = "Failed to load legacy project file %(path)s" % {"path": file_path}
                 log.error(msg, exc_info=1)
                 raise RuntimeError(msg) from ex
 
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -60,15 +60,15 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
         self.new()
 
     def needs_save(self):
-        """Returns if project data Has unsaved changes"""
+        """Returns if project data has unsaved changes"""
         return self.has_unsaved_changes
 
     def get(self, key):
-        """ Get copied value of a given key in data store """
+        """Get copied value of a given key in data store"""
 
         # Verify key is valid type
         if not key:
-            log.warning("Cannot get empty key.")
+            log.debug("ProjectDataStore cannot get empty key (%s)", key)
             return None
         if not isinstance(key, list):
             key = [key]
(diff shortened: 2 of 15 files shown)


MartinThoma:
I've also seen a couple of print statements where I wonder if they should be log messages:

```
blender/scripts/earth.py:179:print("EmptyPointA Transform Rotation: Y= %f Z= %f" % (point_a["lat"], point_a["lon_Z"]))
blender/scripts/earth.py:180:print("EmptyPointB Transform Rotation: Y= %f Z= %f" % (point_b["lat"], point_b["lon_Z"]))
blender/scripts/earth.py:181:print("EmptyPointC Transform Rotation: Y= %f Z= %f" % (point_c["lat"], point_c["lon"]))
blender/scripts/earth.py:182:print("EmptyPointD Transform Rotation: Y= %f Z= %f" % (point_d["lat"], point_d["lon"]))
blender/scripts/earth.py:183:print("EmptyPointC.001 Transform Location: X= %f" % location_CD)
blender/scripts/earth.py:184:print("EmptyPointD.001 Transform Location: X= %f" % location_CD)
blender/scripts/earth.py:185:print("EmptyCam Frame 20 ->Transform Rotation: Y= %f Z= %f  And press I key" % (point_a["lat"], point_a["lon_Z"]))
blender/scripts/earth.py:186:print("EmptyCam Frame 80 ->Transform Rotation: Y= %f Z= %f  And press I key" % (point_b["lat"], point_b["lon_Z"]))
blender/scripts/dissolve.py:272:print(material_object)
blender/scripts/dissolve.py:275:print(material_object)
classes/info.py:105:    print("Compiled translation resources missing!")
classes/info.py:106:    print("Loading translations from: {}".format(language_path))
emojis/optimize-emojis.py:57:        print('Removed emoji: %s' % emoji_path)
emojis/optimize-emojis.py:64:print('Emojis Optimized (%s removed)!' % emoji_removed_count)
language/show_translations.py:47:print("Scanning {} strings in all translation files...".format(len(all_strings)))
language/show_translations.py:59:        print("\n=================================================")
language/show_translations.py:60:        print("Showing translations for {}".format(filename))
language/show_translations.py:61:        print("=================================================")
language/show_translations.py:66:                print('  {} => {}'.format(source_string,translated_string))
language/test_translations.py:52:    print("Testing {} strings in {}...".format(len(all_strings), template_name))
language/test_translations.py:64:            print(red, '%s-%s' % (success, lang_code), endc)
language/test_translations.py:71:                    print(red, '\tInvalid string replacement found: "%s" vs "%s" [%s]' %
launch.py:86:        print("OpenShot version %s" % info.SETUP['version'])
launch.py:97:        print("Supported Languages:")
launch.py:99:            print("  {:>12}  {}".format(lang[0],lang[1]))
launch.py:107:                    print("Added {} to PYTHONPATH".format(os.path.realpath(p)))
launch.py:109:                    print("{} does not exist".format(os.path.realpath(p)))
launch.py:111:                    print("Bad path {}: {}".format(p, ex))
launch.py:124:            print("Unsupported language '{}'! (See --list-languages)".format(args.lang))
launch.py:128:    print("Loaded modules from: %s" % info.PATH)
windows/views/timeline_webview.py:2001:            print('has_audio_data: %s' % has_audio_data)
windows/main_window.py:1629:            print("actionInsertKeyframe")
```

src/classes/project_data.py
```diff
@@ -60,15 +60,15 @@ def __init__(self):
...
         return self.has_unsaved_changes
 
     def get(self, key):
-        """ Get copied value of a given key in data store """
+        """Get copied value of a given key in data store"""
 
         # Verify key is valid type
         if not key:
-            log.warning("Cannot get empty key.")
+            log.debug("ProjectDataStore cannot get empty key (%s)", key)
```
ferdnyc, line 71: There's not a lot of point in logging `key` here — we know it's empty.

Also, I would _not_ lower this one to debug, personally — that code path that should never be reached except through bad user data or programmer error, so it's one of those "could signify an error resulting in potential user data loss" ones. Anytime the code tries to query an empty key, we want to shout about it.

src/classes/project_data.py
```diff
@@ -149,10 +149,10 @@ def _set(self, key, values=None, add=False, partial_update=False, remove=False):
 
         # Verify key is valid type
         if not isinstance(key, list):
-            log.warning("_set() key must be a list. key: {}".format(key))
+            log.debug("_set() key must be a list. key=%s", key)
             return None
         if not key:
-            log.warning("Cannot set empty key.")
+            log.debug("Cannot set empty key (key=%s)", key)
```
ferdnyc, line 155: Ditto both of these. Definitely not debug-only.

src/classes/project_data.py
```diff
@@ -256,7 +256,7 @@ def new(self):
             try:
                 self._data = self.read_from_file(info.USER_DEFAULT_PROJECT)
             except (FileNotFoundError, PermissionError) as ex:
-                log.warning("Unable to load user project defaults from {}: {}".format(info.USER_DEFAULT_PROJECT, ex))
+                log.debug("Unable to load user project defaults from %s: %s", info.USER_DEFAULT_PROJECT, ex)
```
ferdnyc, line 259: **Definitely** `log.warning`: This is a user-supplied file, **and** we've already checked that it exists. If we're now having trouble reading it, the issue is on their end (wrong file permissions, something like that). They need to know about it. However, I'd again go with `exc_info=1` for the exception details.

src/windows/about.py
```diff
@@ -75,7 +75,7 @@ def __init__(self):
                                 self.btnchangelog.setVisible(True)
                                 break
                     except:
```
ferdnyc, line 77:
```suggestion
                    except Exception:
```
 **\#NoBareExcepts**

src/windows/about.py
```diff
@@ -269,7 +269,7 @@ def __init__(self):
                                                    'subject': line[45:].strip() })
                         break
                 except:
```
ferdnyc, line 271:
```suggestion
                except Exception:
```

src/windows/about.py
```diff
@@ -289,7 +289,7 @@ def __init__(self):
...
```
ferdnyc, line 291:
```suggestion
                except Exception:
```

src/windows/about.py
```diff
@@ -311,7 +311,7 @@ def __init__(self):
...
```
ferdnyc, line 313:
```suggestion
                except Exception:
```

src/windows/animated_title.py
```diff
@@ -131,4 +131,4 @@ def clear_effect_controls(self):
                 self.settingsContainer.layout().removeWidget(child)
                 child.deleteLater()
             except:
```
ferdnyc, line 133:
```suggestion
            except Exception:
```

src/windows/views/effects_listview.py
```diff
@@ -59,7 +59,7 @@ def startDrag(self, event):
 
         if not current.isValid():
             # We can't find anything to drag
-            log.warning("No draggable items found in model!")
+            log.warning("No draggable items found in EffectsListView model!")
```
ferdnyc, line 62: Eh. This'll display with the module name prepended to it, so the log message will look like:
```
effects_listview:WARNING: No draggable items found in model!
```
Naming the class is kind of redundant. And it's the kind of thing that screws us up later, when someone cut-and-pastes that code to another module without updating the class name in the log message. (Yes, we end up coding defensively against the _stupidest_ potential problems sometimes...)

src/windows/views/files_listview.py
```diff
@@ -103,7 +103,7 @@ def startDrag(self, supportedActions):
 
         if not current.isValid():
             # We can't find anything to drag
-            log.warning("No draggable items found in model!")
+            log.warning("No draggable items found in FilesListView model!")
```
ferdnyc, line 106: Ditto

src/windows/views/files_treeview.py
```diff
@@ -104,7 +104,7 @@ def startDrag(self, supportedActions):
 
         if not current.isValid():
             # We can't find anything to drag
-            log.warning("No draggable items found in model!")
+            log.warning("No draggable items found in FilesTreeView model!")
```
ferdnyc, line 107: Ditto.

src/windows/views/transitions_listview.py
```diff
@@ -62,7 +62,7 @@ def startDrag(self, event):
 
         if not current.isValid():
             # We can't find anything to drag
-            log.warning("No draggable items found in model!")
+            log.warning("No draggable items found in TransitionsListView model!")
```
ferdnyc, line 65: And again.

src/windows/views/transitions_treeview.py
```diff
@@ -59,7 +59,7 @@ def startDrag(self, event):
 
         if not current.isValid():
             # We can't find anything to drag
-            log.warning("No draggable items found in model!")
+            log.warning("No draggable items found in TransitionsTreeView model!")
```
ferdnyc, line 62: Here too, I'm afraid. :wink:

src/classes/metrics.py
```diff
@@ -173,8 +173,8 @@ def send_metric(params):
                 r = requests.get(url, headers={"user-agent": user_agent}, verify=False)
                 log.info("Track metric: [%s] %s | (%s bytes)" % (r.status_code, r.url, len(r.content)))
 
-            except Exception as Ex:
-                log.error("Failed to Track metric: %s" % (Ex))
+            except Exception as ex:
+                log.warning("Failed to Track metric: %s", ex)
```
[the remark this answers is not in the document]
ferdnyc: (Technically when using `exc_info=1` you don't even need access to the exception object itself, so the `as ex` is superfluous. But I don't think it does any harm, either.)

src/classes/metrics.py
```diff
@@ -64,8 +64,8 @@
         # Get the distro name and version (if any)
         linux_distro = "-".join(platform.linux_distribution())
 
-except Exception as Ex:
-    log.error("Error determining OS version in metrics.py")
+except Exception:
+    log.debug("Error determining OS version in metrics.py")
```
ferdnyc, line 68: I have no idea why this names its file. It shouldn't. Like I said, that info is automatically included.
```suggestion
    log.debug("Error determining OS version", exc_info=1)
```

src/classes/project_data.py
ferdnyc: (If I mistakenly gave a different impression in #3708, my apologies.)

src/windows/views/transitions_treeview.py
ferdnyc: (In case you needed evidence that cut-and-paste code duplication is a major concern, here it is. Most of these would be logging with the _wrong_ class name, if the original had included the name of the class in the message. Nobody ever fixes those things up after copying the code.)

ferdnyc (changes requested):
Thanks for putting this together! It's much appreciated. I suggested a few tweaks inline, and tried to provide my reasoning on them as well.

ferdnyc:
> I've also seen a couple of print statements where I wonder if they should be log messages:

For the most part, no. The ones in the Blender code run inside their Python interpreter, not our process. The code in `classes/info.py`, `language/show_translations.py`, and `launch.py` will run during initial commandline processing in the pre-startup phase, before logging is initialized. (Some of those statements _are_ the commandline processing, in fact.) And  `language/test_translations.py` is a utility script.

These two, however, should indeed be logging statements. Thanks for catching those.
```
windows/views/timeline_webview.py:2001:            print('has_audio_data: %s' % has_audio_data)
windows/main_window.py:1629:            print("actionInsertKeyframe")
```

src/classes/project_data.py
MartinThoma: I undid this, sorry for my missunderstanding.

As a sidenote: `boo(key)` could (without knowing the application) mean many things:

* key = ""
* key = None
* key = []
* key = False
* key = 0

src/windows/views/files_listview.py
MartinThoma: What do you mean?

src/windows/views/effects_listview.py
MartinThoma: Ah, I see. I'm sorry for causing more work. I'll undo it and all of the others :-)

ferdnyc: Not at all! Like I said, it's a _dumb_ problem to have to structure the code around. Hardly the sort of thing you'd expect to have to consider.

But, for better or for worse, copy-paste code duplication is also undeniably a major characteristic of our existing codebase. (I have, in my more cynical periods, [been known](https://github.com/OpenShot/openshot-qt/pull/2512#issuecomment-453385547) to hold it up as the _defining feature_ of our codebase.) :laughing:

src/windows/views/transitions_treeview.py
```diff
@@ -58,7 +58,6 @@ def startDrag(self, event):
             current = selected[0]
 
         if not current.isValid():
-            # We can't find anything to drag
```
ferdnyc, line 61: :rofl: Thank you for helping delivering us, at least in part, from the tedium of narrative comments. These make my teeth grind. There was one I saw recently, it was something very close to this:
```python
# Emit SomeSignal
self.SomeSignal.emit()
```
Which just makes me want to scream.

ferdnyc (changes requested):
This is looking good now, thanks! Only one _required_ change, a small typo on one of the exception classes. But with that fixed, this looks good to go from my POV.

src/windows/views/transitions_treeview.py
ferdnyc: (Naturally, having said this, a quick `git blame` informs me the commit that added the "We can't find anything to drag" comments came from... hmm... some guy named @ferdnyc. Typical. He's the worst.)

MartinThoma:
@ferdnyc Thank you so much for all of the feedback! I think I added all of the requested changes, but it still shows me "1 change requested". I don't know why.

ferdnyc:
@MartinThoma Thanks! I _think_ that _might_ men simply that I requested changes, and haven't come back to "accept" the new commits. (IOW, it's more like " reviewer requested changes". GitHub Is Confusing™.

No matter, reviews don't block merges in the repo, so it could go in regardless. I'll take a look this evening when I'm back at my computer, and we should be all set. Thanks again!

ferdnyc (approved):

ferdnyc:
Yup, as I suspected the "1 change" is dumb and was just counting the number of reviewers who selected "Approve" or "Request Changes" instead of "Comment", it wasn't actually tracking the changes that needed to be made. All looks good now, merging — thanks again!

Issue #2305, opened 2018-10-30
OpenShot on Windows requires undocumented network access


russellvt:
**Describe the bug**
Following installation, the Windows installer complains about requiring network access from launch.exe. Efforts to find documentation supporting this requirement have (so far) proven fruitless.

**System Details:**
 - Windows 7
 - OpenShot Version v2.4.3 for Windows

**To Reproduce**
1. Install OpenShot on Windows 7 from https://www.openshot.org/
2. Once setup is complete, choose "Launch"
3. Observe the Windows Security Alert from "C:\program files\openshot video editor\launch.exe"
4. Notice the "Publisher" for the Windows Security Alert is "Unknown" (read: Unsigned)
5. See screenshot, below.

**Expected behavior**
Network access requirements should be documented, or contain some sort of warning. Windows binaries should be signed by a verifiable publisher.

Current web searches only show typical responses for "trojan'd" software (generally from other scanner output listing OpenShot in the Windows Firewall whitelist, ironically enough).

**Screenshots**
![image](https://user-images.githubusercontent.com/915922/47692085-d68c1880-dbb1-11e8-9e14-7cb70789d6ab.png)

**Logs**
N/A

**Additional context**
Hopefully, there's at least a general "README" to be had for non-UNIX users. No apparent requirement in launch.py for network, though that's only a top-level script for python modules, obviously.

It would likely be a good exercise to launch the Linux version and observe the network requirements (which I have yet to do). Unfortunately, the low-level troubleshooting on Windows 7 (and others) often leaves much to be desired.

Assuming this *might* have something related to "Automatically send errors and metrics to OpenShot" (which defaults as "Yes, I would like to improve OpenShot!")... though obviously this action is taken *prior* to actual approval from the user.

peanutbutterandcrackers:
It is for the error metrics collection (google analytic). Open-shot does not need any other network use (except, maybe check for updates; OpenShot notifies users of available updates when a new stable version is out, so perhaps that).

Yes, we could have done things a little better. But we are short on active volunteers. But hopefully this will get addressed in the future releases. Thank you for notifying us.

DylanC:
@peanutbutterandcrackers @ferdnyc - I feel maybe we should add this to documentation but where is the question? The GitHub readme.md?

ferdnyc:
[...]

PR #3732, opened 2020-09-22
Update windows-installer.iss

dscotese:
The description of the firewall exception now explains itself.  This is a proposed fix for #2305 .

### DIFF
--- a/installer/windows-installer.iss
+++ b/installer/windows-installer.iss
@@ -112,7 +112,7 @@ Name: "slovak"; MessagesFile: "compiler:Languages\Slovak.isl"
 [Tasks]
 Name: "desktopicon"; Description: "{cm:CreateDesktopIcon}"; GroupDescription: "{cm:AdditionalIcons}"; Flags: unchecked
 Name: "fileassoc"; Description: "{cm:AssocFileExtension,{#MyAppName},.osp}"; GroupDescription: "{cm:AdditionalIcons}";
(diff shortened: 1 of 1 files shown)


ferdnyc:
@dscotese Hmmm. The only thing I'm wondering is if we should explicitly say that the anonymous data submission is "optional".

I mean, no amount of... ... ... _user_-proofing will prevent the more determined types from finding a way to feel we've somehow violated them, no matter how much we bend over backwards to avoid sneezing without first submitting a written request for violent nasal expulsion. But that being said, we may want to avoid leaving room for the interpretation that simply installing that firewall exception enables reporting. They still _also_ have to check the "Yes, I would like to improve OpenShot" box in the tutorial.

(TBH, if we were Windows-only we probably _could_ use the firewall checkbox to opt them in to reporting, and not have to effectively ask a second time in the tutorial. But, that wouldn't work for Linux and Mac users, and it's really not worth the special-casing to do it that way only on Windows.)

ferdnyc:
? @dscotese For the record, I was only suggesting that the word "optional" be added to your updated text. (That could've been done within this PR, it's possible to update a PR branch after submitting it. But if you just closed this to resubmit with different wording, that's great too.)

ferdnyc:
Oooh, hey, it seems Windows Firewall rules can have a _description_ field:

image

We should figure out how to set that, too.

image

PR #3735, opened 2020-09-23
Alias pyQtSignal/pyQtSlot to Signal and Slot

ferdnyc:
This PR updates all Python files using `pyqtSignal` or `pyqtSlot`, so that they now:
```python3
# Import the names aliased
from PyQt5.QtCore import (
    pyqtSignal as Signal,
    pyqtSlot as Slot,
)

# Use only the aliases after that point
class BlenderListView(QListView):    

    @Slot()
    def enable_interface(self):
        """ Disable all controls on interface """
        self.win.btnRefresh.setEnabled(True)
        self.win.sliderPreview.setEnabled(True)
        self.win.btnRender.setEnabled(True)

    def __init__(self, *args):
         # Hook up signals to Background Worker
        self.worker.enable_interface.connect(self.enable_interface)

class Worker(QObject):
    enable_interface = Signal()
```

This makes our code potentially much better compatible with Qt for Python (aka PySide2) and wrapper libraries like qtpy and AnyQt, which almost all use `QtCore.Signal` and `QtCore.Slot`.

I tested, and if we wanted to make the code maximally portable, it would even work to do this...
(Just for the Signals/Slots stuff! Not claiming it would magically make all the rest of our code PySide2-compatible.)

```python3
if some_wrapper_condition:
    from PyQt5 import QtCore
    from PyQt5.QtCore import (
        pyqtSlot as Slot,
        pyqtSignal as Signal,
    )
    QtCore.Slot = Slot
    QtCore.Signal = Signal
else:
    from PySide2 import QtCore

class BlenderListView(QListView):    

    @QtCore.Slot()
    def enable_interface(self):
        """ Disable all controls on interface """
        self.win.btnRefresh.setEnabled(True)
        self.win.sliderPreview.setEnabled(True)
        self.win.btnRender.setEnabled(True)

    def __init__(self, *args):
         # Hook up signals to Background Worker
        self.worker.enable_interface.connect(self.enable_interface)

class Worker(QObject):
    enable_interface = QtCore.Signal()
```


ferdnyc:
Dumb typo fixed.

ferdnyc:
> I tested, and if we wanted to make the code maximally portable, it would even work to do this...
> (Just for the Signals/Slots stuff! Not claiming it would magically make all the rest of our code PySide2-compatible.)
> 
> ```python
> if some_wrapper_condition:
>     from PyQt5 import QtCore
>     from PyQt5.QtCore import (
>         pyqtSlot as Slot,
>         pyqtSignal as Signal,
>     )
>     QtCore.Slot = Slot
>     QtCore.Signal = Signal
> else:
>     from PySide2 import QtCore
> ```

As would this, though somehow this feels irrationally more naughty, when really it's precisely the same thing just without the local namespace pollution:
```python3
from PyQt5 import QtCore
QtCore.Slot = QtCore.pyqtSlot
QtCore.Signal = QtCore.pyqtSignal
```

jonoomph:
@ferdnyc This also looks good (minus the conflicts). Merge away!

jonoomph:
@ferdnyc I briefly tried to resolve these conflicts, but ultimately backed out. I'll let you have fun with this one. :smile:

ferdnyc:
Yeah, resolving that was no fun at all. But, done.

jonoomph:
Resovled conflicts again. I am good with this merge, unless @ferdnyc wants to change the 3 moved signals as well (which I believe was part of a previous PR).

ferdnyc:
I'm going to close this one, while the _idea_ of it has some appeal I think I've come down on the side of it being more complicated than it's worth, especially given the maintenance burden. If we had a single wrapper library that translated the interfaces, then maybe, but as things stand anywhere we wanted to use signals or slots we'd have to remember to alias the symbols when importing them.

PR #3738, opened 2020-09-24
STY: Simplify the code

MartinThoma:
I used my tool [flake8-simplify](https://pypi.org/project/flake8-simplify/0.4.1/) (version 0.4.1) to find those.
I didn't use [contextlib.suppress](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) as I wasn't certain what people think about that.

### DIFF
--- a/src/windows/models/properties_model.py
+++ b/src/windows/models/properties_model.py
@@ -219,57 +219,55 @@ class PropertiesModel(updates.UpdateInterface):
             # Get effect object
             c = Effect.get(id=clip_id)
 
-        if c:
-            # Update clip attribute
-            if property_key in c.data:
-                log.info("remove keyframe: %s" % c.data)
+        if c and property_key in c.data:  # Update clip attribute
+            log.info("remove keyframe: %s" % c.data)
 
-                # Determine type of keyframe (normal or color)
-                keyframe_list = []
-                if property_type == "color":
-                    keyframe_list = [c.data[property_key]["red"], c.data[property_key]["blue"], c.data[property_key]["green"]]
-                else:
-                    keyframe_list = [c.data[property_key]]
+            # Determine type of keyframe (normal or color)
+            keyframe_list = []
+            if property_type == "color":
+                keyframe_list = [c.data[property_key]["red"], c.data[property_key]["blue"], c.data[property_key]["green"]]
+            else:
+                keyframe_list = [c.data[property_key]]
 
-                # Loop through each keyframe (red, blue, and green)
-                for keyframe in keyframe_list:
+            # Loop through each keyframe (red, blue, and green)
+            for keyframe in keyframe_list:
 
-                    # Keyframe
-                    # Loop through points, find a matching points on this frame
-                    closest_point = None
-                    point_to_delete = None
-                    for point in keyframe["Points"]:
-                        if point["co"]["X"] == self.frame_number:
-                            # Found point, Update value
-                            clip_updated = True
-                            point_to_delete = point
-                            break
-                        if point["co"]["X"] == closest_point_x:
-                            closest_point = point
-
-                    # If no point found, use closest point x
-                    if not point_to_delete:
-                        point_to_delete = closest_point
-
-                    # Delete point (if needed)
-                    if point_to_delete:
+                # Keyframe
+                # Loop through points, find a matching points on this frame
+                closest_point = None
+                point_to_delete = None
+                for point in keyframe["Points"]:
+                    if point["co"]["X"] == self.frame_number:
+                        # Found point, Update value
                         clip_updated = True
-                        log.info("Found point to delete at X=%s" % point_to_delete["co"]["X"])
-                        keyframe["Points"].remove(point_to_delete)
+                        point_to_delete = point
+                        break
+                    if point["co"]["X"] == closest_point_x:
+                        closest_point = point
 
-                # Reduce # of clip properties we are saving (performance boost)
-                c.data = {property_key: c.data[property_key]}
+                # If no point found, use closest point x
+                if not point_to_delete:
+                    point_to_delete = closest_point
 
-                # Save changes
-                if clip_updated:
-                    # Save
-                    c.save()
+                # Delete point (if needed)
+                if point_to_delete:
+                    clip_updated = True
+                    log.info("Found point to delete at X=%s" % point_to_delete["co"]["X"])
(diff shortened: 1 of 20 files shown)


src/windows/views/timeline_webview.py
```diff
@@ -2367,13 +2372,12 @@ def Time_Triggered(self, action, clip_ids, speed="1X", playhead_position=0.0):
                     clip.data["reader"]["video_length"] = self.round_to_multiple(
                         float(clip.data["reader"]["video_length"]) / speed_factor, even_multiple)
 
-                if action == MENU_TIME_NONE:
+                if action == MENU_TIME_NONE and "original_data" in clip.data.keys():
```
ferdnyc, line 2375: Or just,
```suggestion
                if action == MENU_TIME_NONE and "original_data" in clip.data:
```

src/windows/views/timeline_webview.py
```diff
@@ -2367,13 +2372,12 @@ def Time_Triggered(self, action, clip_ids, speed="1X", playhead_position=0.0):
...
                     # Reset original end & duration (if available)
-                    if "original_data" in clip.data.keys():
-                        clip.data["end"] = clip.data["original_data"]["end"]
-                        clip.data["duration"] = clip.data["original_data"]["duration"]
-                        clip.data["reader"]["video_length"] = clip.data["original_data"]["video_length"]
-                        clip.data.pop("original_data")
+                    clip.data["end"] = clip.data["original_data"]["end"]
+                    clip.data["duration"] = clip.data["original_data"]["duration"]
+                    clip.data["reader"]["video_length"] = clip.data["original_data"]["video_length"]
+                    clip.data.pop("original_data")
```
ferdnyc, line 2380: This could even be,
```suggestion
                    orig = clip.data.pop("original_data")
                    clip.data.update({
                        "end": orig["end"],
                        "duration": orig["duration"],
                    })
                    clip.data["reader"]["video_length"] = orig["video_length"]
```
(Though I honestly can't fathom how that last line can be correct, regardless. Holy hell.)

src/windows/views/timeline_webview.py
```diff
@@ -434,16 +434,17 @@ def ShowTimelineMenu(self, position, layer_id):
...
-            if len(clipboard_clip_ids) + len(clipboard_tran_ids) > 0:
-                menu = QMenu(self)
-                Paste_Clip = menu.addAction(_("Paste"))
-                Paste_Clip.setShortcut(QKeySequence(self.window.getShortcutByName("pasteAll")))
-                Paste_Clip.triggered.connect(
-                    partial(self.Paste_Triggered, MENU_PASTE, float(position), int(layer_id), [], [])
-                )
+        if (self.copy_clipboard or self.copy_transition_clipboard) and (
+            len(clipboard_clip_ids) + len(clipboard_tran_ids) > 0
+        ):
```
ferdnyc, line 439: I kind of want this to be an early return as well, but it hurts my brain too much just trying to work out how the condition should be specified.

ferdnyc (approved):
Looks good to me! I've been wanting to hit a lot of these for a _while_ now.

I added some discussion/suggestions, but all purely discretionary and entirely open to debate.

src/windows/models/properties_model.py
```diff
@@ -219,57 +219,55 @@ def remove_keyframe(self, item):
...
-            if property_key in c.data:
-                log.info("remove keyframe: %s" % c.data)
-
-                # Determine type of keyframe (normal or color)
-                keyframe_list = []
-                if property_type == "color":
-                    keyframe_list = [c.data[property_key]["red"], c.data[property_key]["blue"], c.data[property_key]["green"]]
-                else:
-                    keyframe_list = [c.data[property_key]]
+        if c and property_key in c.data:  # Update clip attribute
```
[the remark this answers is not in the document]
MartinThoma: I like the change, but are you certain it's correct. Does the code have side-effects?

src/windows/views/timeline_webview.py
MartinThoma: Simply a `not [what is currently inside the if]` would do it. Should I do it?

MartinThoma: I would probably add another variable for the current if-condition. Is there a reasonable name?

src/windows/models/properties_model.py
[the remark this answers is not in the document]
ferdnyc: I had to triple-check, but yeah it should be safe. The **entire** rest of the method body is inside the `if`, so none of it will ever be executed if the condition fails.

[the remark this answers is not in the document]
ferdnyc: (Applying it would require more than just accepting the suggestion, as the rest of the body would _have_ to be outdented to keep Python from complaining about unexpected indentation / unreachable-code-after-`return`.)

src/windows/views/timeline_webview.py
ferdnyc: Something like `have_clipboards` should serve well enough, I think. That's really all it's testing: that there are contents (of any sort) on **both** the instance variables for the clipboard lists (self.copy_*_clipboard`) as well as the lists passed in to the function (clipboard_clip_ids`, `clipboard_tran_ids`) — though, no, I can't say that I immediately follow why there would be _**four**_ distinct lists of timeline elements that can be involved in a menu-construction operation, TBH.

src/windows/models/properties_model.py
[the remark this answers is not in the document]
ferdnyc: Actually, Python wouldn't complain about that at all, it's far too forgiving and would just assume we meant to do that (despite it being an absurd thing to do). Fortunately, Codacy would take issue with it.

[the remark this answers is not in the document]
MartinThoma: Ah, right. I looked at the wrong code piece. Thank you for mentioning it; I think I will integrate this in flake8-simplify: https://github.com/MartinThoma/flake8-simplify/issues/8

src/windows/views/timeline_webview.py
MartinThoma: Ok :-) I've pushed it :-)

ferdnyc:
Well, it all looks great to me, and some quick testing showed that Context menus and cut-and-paste still work as expected on the Timeline, so let's `git` 'er done. Thanks again @MartinThoma .

PR #3739, opened 2020-09-24
Make use of new Timeline lookup APIs

ferdnyc:
This follows on from OpenShot/libopenshot#563, by actually making use of the more efficient new APIs in OpenShot.

(This means that the `devel` branch requires a recent development version of libopenshot, built within the past two weeks since 563 merged. The `develop` branch is now completely incompatible with libopenshot 0.2.5. Fair warning.)

### DIFF
--- a/src/windows/export.py
+++ b/src/windows/export.py
@@ -359,17 +359,7 @@ class Export(QDialog):
         self.delayed_fps_timer.start()
 
         # Determine max frame (based on clips)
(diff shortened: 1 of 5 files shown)


ferdnyc:
I'mma merge this — it's now been over a month since OpenShot/libopenshot#563 merged, which hopefully is enough time for anyone who's testing `develop` code to have updated their library build.

PR #3755, opened 2020-10-11
Correction to the title template directory.

gdecaso:
Correction to template directory, see #3587

### DIFF
--- a/doc/titles.rst
+++ b/doc/titles.rst
@@ -44,7 +44,7 @@ Overview
 Custom Titles
 -------------
 OpenShot can use any vector SVG image file as a title template. Just add an SVG image file to your 
-*/.openshot_qt/title_templates/* folder, and it will appear in OpenShot's title editor after you restart OpenShot. You can
+*.openshot_qt/title_templates/* folder, and it will appear the next time you open the Title Editor interface. You can
 also right click on any SVG files in your **Project Files** panel, and choose **Edit Title** or **Duplicate Title**.
 
 3D Animated Titles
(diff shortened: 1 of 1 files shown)


doc/titles.rst
```diff
@@ -43,9 +43,9 @@ Overview
 
 Custom Titles
 -------------
-OpenShot can use any vector SVG image file as a title template. Just add an SVG image file to your */.openshot_qt/titles/*
-folder, and it will appear in OpenShot's title editor after you restart OpenShot. You can also right click on any SVG
-files in your **Project Files** panel, and choose **Edit Title** or **Duplicate Title**.
+OpenShot can use any vector SVG image file as a title template. Just add an SVG image file to your 
+*/.openshot_qt/title_templates/* folder, and it will appear in OpenShot's title editor after you restart OpenShot. You can
```
ferdnyc, line 47:
```suggestion
*.openshot_qt/title_templates/* folder, and it will appear the next time you open the Title Editor interface. You can
```

I don't know what that leading slash is about, but it shouldn't be there. (If we were a Linux-only app it would make sense to say `$HOME/.openshot_qt/title_templates/`, but that's not right for Windows.)

Also, a quick test tells me that you don't even need to restart OpenShot — the list in the Title Editor dialog is rebuilt each time it's opened, so any additions to the directory will appear as soon as it's opened again.

ferdnyc (approved):
One small tweak to the path formatting, and one correction to the surrounding text (that, as it turns out, wasn't entirely accurate either). But, looks good, thanks @gdecaso !

ferdnyc:
I'll merge as soon as Travis finishes with my follow-up commit.

PR #3756, opened 2020-10-12
Simplify code by combining 'or' statements

MartinThoma:


src/windows/views/blender_listview.py
```diff
@@ -843,9 +843,8 @@ def Render(self):
 
         _ = get_app()._tr
 
-        if not self.version:
-            if not self.blender_version_check():
-                return
+        if not self.version and not self.blender_version_check():
+            return
```
ferdnyc, line 847: Heh. I suspect I wrote this like that because (looking at it now) `self.blender_version_check()` can raise an `Exception`, something I should probably be handling here but am not. (Clearly, writing it "wrong" was a poor reminder, as it didn't stop me from merging it without addressing that. So a `TODO` or `XXX` comment would've been a better choice.)

Anyway, point is, this should really have a `try:` wrapped around the inner condition, so it may end up getting changed some more — but that's unrelated to **_this_** change. Plus, addressing the exception issue may involve having `blender_version_check()` just return `False` instead of raising the exception in the first place.

Point is, outside issues aside, for the current code combining them is the right call.

src/windows/views/timeline_webview.py
```diff
@@ -1183,11 +1183,11 @@ def Layout_Triggered(self, action, clip_ids):
...
-            if action == MENU_LAYOUT_CENTER or \
-               action == MENU_LAYOUT_TOP_LEFT or \
-               action == MENU_LAYOUT_TOP_RIGHT or \
-               action == MENU_LAYOUT_BOTTOM_LEFT or \
-               action == MENU_LAYOUT_BOTTOM_RIGHT:
+            if action in [MENU_LAYOUT_CENTER,
+                          MENU_LAYOUT_TOP_LEFT,
+                          MENU_LAYOUT_TOP_RIGHT,
+                          MENU_LAYOUT_BOTTOM_LEFT,
+                          MENU_LAYOUT_BOTTOM_RIGHT]:
```
ferdnyc, line 1190: I think `black` would format this as,
```python3
if action in [
        MENU_LAYOUT_CENTER,
        MENU_LAYOUT_TOP_LEFT,
        MENU_LAYOUT_TOP_RIGHT,
        MENU_LAYOUT_BOTTOM_LEFT,
        MENU_LAYOUT_BOTTOM_RIGHT,
        ]:
#\-----/ (note: EIGHT spaces)
```
(I'm not 100% sure where it would put the `]:`, it would either be there or aligned under the `if`.)

...I have a _soft_ preference for that. Not for the usual reasons of enabling cleaner diffs / easier code reviews (the contents of **these** lists just aren't going to change), but more because we frequently end up with conditions so deeply nested and so long that they lead to horrors like,
```python3
                if really_long_variable_name and really_long_variable_name in [MENU_LAYOUT_CENTER,
                                                                               MENU_LAYOUT_TOP_LEFT,
                                                                               MENU_LAYOUT_TOP_RIGHT,
                                                                               MENU_LAYOUT_BOTTOM_LEFT,
                                                                               MENU_LAYOUT_BOTTOM_RIGHT]:
```
That's just the _worst_, and unquestionably (in my mind) should be:
```python3
                if really_long_variable_name and really_long_variable_name in [
                        MENU_LAYOUT_CENTER,
                        MENU_LAYOUT_TOP_LEFT,
                        MENU_LAYOUT_TOP_RIGHT,
                        MENU_LAYOUT_BOTTOM_LEFT,
                        MENU_LAYOUT_BOTTOM_RIGHT,
                ]:
```
or whatever.

But, for **these** that's not a concern, so it really doesn't matter and I wouldn't bother changing it — the aligned-to-the-first-line form is just fine.

ferdnyc (approved):
All sensible fixes, LGTM.

ferdnyc:
Had to resolve a conflict I just created with a few _other_ PR merges, so this is back in Travis' teeth. But as soon as it gets tired of gnawing on the code, I'll chuck this one onto the merge pile. Thanks again!

PR #3780, opened 2020-10-20
Revert "Fix loading Unicode paths from 2.4 (#3624)"

jonoomph:
Reverts OpenShot/openshot-qt#3664


jonoomph:
This seems to break our file paths on save: `"path": "apps/openshot-qt-git/src/\"/home/jonathan/Videos/TearsOfSteel.mp4\"",`

PR #3781, opened 2020-10-21
project_data: New keyframe scaler implementation

ferdnyc:
This PR replaces the `rescale_keyframes` method of `project_data.py` with a new implementation:
- Move the bulk of `rescale_keyframes` to a separate file, and implement it as a factory class `KeyframeScaler`. Instances of the class are callables which have been assigned a scaling factor. Calling the scaler instance on a project data dict will apply the factor.
- Break up the scaling code into multiple private methods, optimized to be as DRY as possible. Avoid deep nesting and repetition.

The previous code was a diagonal cascade of tests and loops, and repeated the exact same code three times for each of the lists it acted on.

The new code breaks the algorithm up into four methods, each one representing the algorithm for one level of that cascade, and each outer method repeatedly calls the next innermost one as they process all of the objects, keyframe properties, points lists, and individual points.

Beyond the unspooling of the algorithms, the new methods use a couple of somewhat daunting list comprehensions to efficiently process the contents of the various lists and dicts being modified. All I can say is, I've run a bunch of project data structures through it, and so far it's always done the right thing — even when I've added non-standard data that was specifically _trying_ to trick it.

(For example, placing a `"Points"` key directly on a `"clips"` list member, or placing a `"co"` or `"X"` key in something that's not a `"Points"` list, won't fool it. It also won't touch the keyframe-related contents of `"history"` objects, or any of the other structures aside from a specific type of key found inside an object from the list stored in either `"clips"` or `"effects"`.)

But you, too, can easily test the algorithm, because it's all self-contained in `src/classes/keyframe_scaler.py` and has no `PyQt` or `openshot` dependencies.

Running a project file through it is as easy as:

```python3
import copy
import json
from keyframe_scaler import KeyframeScaler
# Load the file in to a Python dict
with open("myfile.osp", "r") as f:
    project_data = json.load(f)
# Create a callable scaler instance which doubles all X coordinates > 1
double_scaler = KeyframeScaler(factor=2)
# Apply the scaling factor to a copy of a project data dictioary
new_dict = double_scaler(copy.deepcopy(project_data))
# And if you want to see what it did...
from pprint import pprint
pprint(new_dict)
```

(The KeyframeScaler class won't copy the dict, so unless it's passed a structure that's been run through `copy.deepcopy()` it will modify the source data in place. Perhaps I should change that, and move the `deepcopy()` into the class?)

### DIFF
--- a/src/classes/keyframe_scaler.py
+++ b/src/classes/keyframe_scaler.py
@@ -1,9 +1,7 @@
 """
  @file
  @brief Process project data, scaling keyframe X coordinates by the given factor
- @author Noah Figg <eggmunkee@hotmail.com>
  @author Jonathan Thomas <jonathan@openshot.org>
- @author Olivier Girard <eolinwen@gmail.com>
  @author FeRD (Frank Dana) <ferdnyc@gmail.com>
 
  @section LICENSE
@@ -36,7 +34,7 @@ class KeyframeScaler:
     multiplied by the scaling factor, except X=1 (because the first
     frame never changes)"""
 
-    def _scale_value(self, value: float) -> int:
+    def _scale_x_value(self, value: float) -> int:
         """Scale value by some factor, except for 1 (leave that alone)"""
         if value == 1.0:
             return value
@@ -55,7 +53,7 @@ class KeyframeScaler:
         for k in keyframes:
             # Scale the X coordinate (frame #) by the stored factor
             [point["co"].update({
-                "X": self._scale_value(point["co"].get("X", 0.0))
+                "X": self._scale_x_value(point["co"].get("X", 0.0))
                 })
                 for point in k if "co" in point]
 
--- a/src/classes/keyframe_scaler.py
+++ b/src/classes/keyframe_scaler.py
@@ -1,0 +1,87 @@
+"""
+ @file
+ @brief Process project data, scaling keyframe X coordinates by the given factor
+ @author Noah Figg <eggmunkee@hotmail.com>
+ @author Jonathan Thomas <jonathan@openshot.org>
+ @author Olivier Girard <eolinwen@gmail.com>
+ @author FeRD (Frank Dana) <ferdnyc@gmail.com>
+
+ @section LICENSE
+
+ Copyright (c) 2008-2020 OpenShot Studios, LLC
+ (http://www.openshotstudios.com). This file is part of
+ OpenShot Video Editor (http://www.openshot.org), an open-source project
+ dedicated to delivering high quality video editing and animation solutions
+ to the world.
+
+ OpenShot Video Editor is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ OpenShot Video Editor is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with OpenShot Library.  If not, see <http://www.gnu.org/licenses/>.
+ """
+
+
+class KeyframeScaler:
+    """This factory class produces scaler objects which, when called,
+    will apply the assigned scaling factor to the keyframe points
+    in a project data dictionary. Keyframe X coordinate values are
+    multiplied by the scaling factor, except X=1 (because the first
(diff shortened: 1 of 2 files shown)


src/classes/keyframe_scaler.py
```diff
@@ -0,0 +1,88 @@
+"""
+ @file
+ @brief This file listens to changes, and updates the primary project data
```
ferdnyc, line 3: Aaand, copy-pasting bites us again; forgot to change the file @brief from the original project_files source. Oops. I'll have to fix that.

(The list of authors, though, was intentional. I have no idea who wrote the `rescale_keyframes` function this is based on, so I decided the best approach was to just credit everyone listed in the original file.)

ferdnyc:
The code also uses type annotations wherever possible on the method arguments and return values. That requires a minimum Python version of 3.5, but I think that's safe at this point. Python 3.4 is already end-of-life, and IIRC all of our official packages are built with Python 3.6 or greater these days.

src/classes/keyframe_scaler.py
ferdnyc: Update: Fixed.

src/classes/keyframe_scaler.py
```diff
@@ -0,0 +1,87 @@
+"""
+ @file
+ @brief Process project data, scaling keyframe X coordinates by the given factor
+ @author Noah Figg <eggmunkee@hotmail.com>
```
jonoomph, line 4: I think only Myself and you need to be listed here

src/classes/keyframe_scaler.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class KeyframeScaler:
+    """This factory class produces scaler objects which, when called,
+    will apply the assigned scaling factor to the keyframe points
+    in a project data dictionary. Keyframe X coordinate values are
+    multiplied by the scaling factor, except X=1 (because the first
+    frame never changes)"""
+
+    def _scale_value(self, value: float) -> int:
```
jonoomph, line 39: Interesting, I have not seen type hints before in Python. Pretty cool. So, in this case, we are passing in a float, but returning an int? Is that by design?

jonoomph: Oh I see, this is only called with X values (which are of type INT), representing frame #'s

jonoomph: Should we rename this to be more specific to INT or FRAME NUMBER or X?

src/classes/project_data.py
```diff
@@ -406,53 +408,13 @@ def scale_keyframe_value(self, original_value, scale_factor):
...
-                if type(effect.get(attribute)) == dict and "red" in effect.get(attribute):
-                    for color in effect.get(attribute):
-                        for point in effect.get(attribute).get(color).get("Points"):
-                            if "co" in point:
-                                point["co"]["X"] = self.scale_keyframe_value(point["co"].get("X", 0.0), scale_factor)
-
-        # return the copied and scaled project data
-        return data
+        #
+        log.info('Scale all keyframes by a factor of %s', scale_factor)
```
jonoomph, line 412: This is cool! Nice work!

src/classes/keyframe_scaler.py
ferdnyc: That's actually a good point. The old code produced int values because it did the same thing:
```py
return round(original_value * scale_factor)
```
...But even X coordinate values _technically_ are floats, so maybe the function should convert the new rounded integer values to float, and return those instead.

As far as the name, yeah `_scale_x_value()` could work, sure... The keyframe scaler's entire function is to scale X values, but that's no reason _not_ to make it explicit, and therefore clearer, still.

src/classes/project_data.py
ferdnyc: Thanks! I should check if I can remove the old `scale_keyframe_value` method as well, or if it's still used somewhere else in the code.

src/classes/keyframe_scaler.py
ferdnyc: Fixed.

src/classes/keyframe_scaler.py
ferdnyc: I changed it to `_scale_x_value()`, but left the `int` return since it matches the old code.

src/classes/project_data.py
ferdnyc: It wasn't; removed.

jonoomph:
LGTM! Merge away if this is ready. It's been sitting here a while, so I would rather @ferdnyc  merge it for safety reasons. 👍

jonoomph:
Got too excited to merge this. :+1:

PR #3784, opened 2020-10-21
Input handling cleanup, new multi-file import feedback, context menu improvements

### LINKED ISSUE #909: Unexpected behaviour of Rewind and Fast Forward controls
**System Details:**
* Operating System / Distro: Linux Mint 18.2 64bit
* OpenShot Version: 2.4.0

**Issue Description and steps to reproduce:*

1. When the **Rewind** control (in **Video Preview**) is clicked, while the play-head is stationary, the play-head starts moving forward rather than backward (same as **Play**). Only when clicked again the playhead starts moving backward as intended. Proposed solution: Make it move backward right away.

2. The feature of increased speed on repeatedly clicking **Rewind** or **Fast Forward** is nice. However slowing down Fast Forward by clicking Rewind (and vice versa) is imho counter intuitive. Proposed solution: While playing fast foward, **Rewind** reverses the playing direction right away, whatever the current fast forward speed is. Same applies for **Fast Forward** while playing in rewind mode.

3. The expected behaviour of Rewind is moving the playhead towards the start of the track (clip) without showing the contents, which actually is the behaviour of the **Jump To Start** control. Proposed solution: Rename '**Rewind**' '**Play Backward**' or '**Play Reverse**' or any more suitable phrase.

4. The expected behaviour of **Fast Forward** is to move the play-head to the end of the track (clip) without showing the contents. A more unambiguous phrase would be **Play Fast Forward**. Because the Play control plays the clip at normal speed, the **Play Fast Forward** control should start at double speed on the first click.

### LINKED ISSUE #3782: Drag-and-drop Image Sequence import issues under MacOS
@ferdnyc I downloaded and installed [OpenShot-v2.5.1-dev2-1603218028-07a447c9-78db4b94-x86_64.dmg](https://github.com/OpenShot/openshot-qt/releases/download/daily/OpenShot-v2.5.1-dev2-1603218028-07a447c9-78db4b94-x86_64.dmg) (October 20, 2020) on my MacBook Pro laptop.  I tested the import folder function by dragging-and-dropping the Media Test Files folder to the Project List.  

Now, I can drag-and-drop single and multiple images, audio and video but not image sequences.  When I drag-and-drop an image sequence, I am prompted whether or not I want to import the file as a sequence, which I answer yes.  However, I cannot use the mouse to do so, I have to use the keyboard to tab to the Yes button then press Return.

When I use the File Import dialog the image sequence imports to the Project Files list and I can preview the Snow%04d.png file using the right-click menu.  When I add the image sequence to the timeline then press play under the Preview window, OpenShot either freezes at frame 13 or crashes.  If OpenShot freezes, I have to Force Quit the application.

Windows and Linux tested successfully without issues.

_Originally posted by @USATechDude in https://github.com/OpenShot/openshot-qt/issues/3492#issuecomment-713127089_

### LINKED ISSUE #3783: Properties context menu generation improvements
@jonoomph asked via email about the menu-locking functionality in the Properties context menu. It doesn't work particularly well. As he noted, "it seems to be blocking every other call to the context menu (on the properties table view)"

I responded:

> Yeah, I've noticed that it's not working perfectly, I keep meaning to look at that.
> 
> The original purpose (the PR I added it in is #2997: Properties: Ignore overlapping context menu events) was to address bug #2996, an issue with the context menus for things like the Source property of Mask transitions, and other properties that listed project files as selections. As I wrote in the PR:
> 
> > Because "reader" type properties require a fairly lengthy context menu that can take some time to display, this PR adds a lock variable to protect against repeated invocations of the event handler from being queued up during menu construction, and subsequently trampling all over each other.
> 
> Basically, if you right-clicked on a Source property, the menu takes some hundreds of milliseconds or even multiple seconds to actually come up. Giving plenty of opportunity to right-click again during that period, and then things would get totally confused. So, the lock was intended as a way of debouncing the context-menu request, by having the event handler ignore additional events if the menu is already in the process of being constructed. But, it seems not to limit itself to those situations as well as I'd hoped.
>  
> I think part of the solution might be to actually use a persistent single-shot timer, rather than the one-off ones I had been using, and to have that call a different slot to release the lock rather than relying on re-invoking the event handler (which seems not to work very well). My original idea was that, by re-using the handler, the unlocking would get priority over any other events waiting in the queue. But I think it really needs to be a separate signal, instead.
> 
> The other thing it could do is only set the lock when the request is for a reader-type property, since the other menu types come up quickly enough that they don't really need to be debounced.

Then, after thinking about it some more:

> Hm. Another option altogether, come to think of it — I added that locking (shortly) before we re-did all of the data models to make things persistent instead of constantly blowing away and re-creating them. The context menus still work the same way the old models used to do, they build themselves completely from scratch every time. That's why it takes so long for the Source context menu to build. It's also why we're actually bleeding menu objects, something I recently noticed in GammaRay. Things aren't getting cleaned up from all those single-use menu objects we build.
> 
> Now that the data models for the files and transitions and the like are more persistent, we could probably build a "property source proxy model" that combines both of those models together and provides the available source items in a form that's ready-made for the context menu to consume (icons already generated and etc.) Done that way, it might be quick enough to display that we wouldn't need to worry about debouncing at all.

ferdnyc:
#### Update

OK, this grew into quite a few things. This PR includes the following changes:

- A lot (hopefully all?) of the event handlers now `accept()` their input early in the process, to make sure the system stops propagating it and frees up the GUI for other tasks
- A lot of the callback functions that _aren't_ even handlers have had their `event` parameter removed, because they never accepted `event`s as arguments. (The only standard argument an action's `triggered()` callback takes is `checked`, a boolean that's passed with checkable actions to indicate their current status.)
- Drag-and-drop file processing (actually, all `add_files` processing) now regularly triggers event loop processing, to avoid UI freezes
- Imports of more than 15 files at once will now show a localized "Imported M / N" counter in the status bar of the main window as they're running through the imports. The total count may adjust itself downward, if image sequence redundancies are removed from the list during the middle of the run.
- Context menus now use `menu.popup()` instead of `menu.exec()`, as `popup()` runs asynchronously and keeps the menu from freezing the rest of the UI while it's displayed. (You can even right-click things while OpenShot's preview is playing, though it's rarely very useful to do so.)

#### Unrelated or only loosely-related
- I removed the locking mechanism from the Properties context menu and cleaned up that menu code a bit, to address #3783
- I fixed a longstanding bug where hitting the Reverse/Rewind button while the player is Paused would make it start playing _forward_ instead of backwards
- I improved the console/logfile output from the preview player, to make it more useful and less noisy

Fixes #3782, fixes #3783, fixes #909

##### Original summary

The code for the drop events in the two file view classes never sat right with me, as it was only calling `event.accept()` _after_ processing all of the dropped URLs. (The idea being that it would only accept on a successful import. But that's not how Qt events work. An accept just means you're _handling_ the event, which we are.)

@USATechDude reported input-handling issues on MacOS that I think are likely caused by the event not being accepted, so this PR is an attempt to correct that.

I tested this on Linux and it works fine, but then again the old code seemed to work fine on Linux. Still, even if it doesn't fix the issue, I'm pretty sure the code's more correct this way.

### DIFF
--- a/src/windows/models/files_model.py
+++ b/src/windows/models/files_model.py
@@ -443,14 +443,12 @@ class FilesModel(QObject, updates.UpdateInterface):
                     log.warning("Directory recursion failed", exc_info=1)
             elif os.path.isfile(filepath):
                 media_paths.append(filepath)
-
+        if not media_paths:
+            return
         # Import all new media files
-        if media_paths:
-            media_paths.sort()
-            log.debug("Importing file list: {}".format(media_paths))
-            return self.add_files(media_paths, quiet=import_quietly)
-        else:
-            return False
+        media_paths.sort()
+        log.debug("Importing file list: {}".format(media_paths))
+        self.add_files(media_paths, quiet=import_quietly)
 
     def get_thumb_path(
             self, file_id, thumbnail_frame, clear_cache=False):
--- a/src/windows/models/files_model.py
+++ b/src/windows/models/files_model.py
@@ -189,43 +189,43 @@ class FilesModel(QObject, updates.UpdateInterface):
                     thumbnail_frame = round(float(file.data['start']) * fps_float) + 1
 
                 # Get thumb path
-                thumb_path = self.get_thumb_path(file.id, thumbnail_frame)
+                thumb_icon = QIcon(self.get_thumb_path(file.id, thumbnail_frame))
             else:
                 # Audio file
-                thumb_path = os.path.join(info.PATH, "images", "AudioThumbnail.svg")
+                thumb_icon = QIcon(os.path.join(info.PATH, "images", "AudioThumbnail.svg"))
 
             row = []
+            flags = Qt.ItemIsSelectable | Qt.ItemIsEnabled | Qt.ItemIsDragEnabled | Qt. ItemNeverHasChildren
 
             # Append thumbnail
-            col = QStandardItem(name)
-            col.setIcon(QIcon(thumb_path))
+            col = QStandardItem(thumb_icon, name)
             col.setToolTip(filename)
-            col.setFlags(Qt.ItemIsSelectable | Qt.ItemIsEnabled | Qt.ItemIsDragEnabled)
+            col.setFlags(flags)
             row.append(col)
 
             # Append Filename
             col = QStandardItem(name)
-            col.setFlags(Qt.ItemIsSelectable | Qt.ItemIsEnabled | Qt.ItemIsDragEnabled | Qt.ItemIsEditable)
+            col.setFlags(flags | Qt.ItemIsEditable)
             row.append(col)
 
             # Append Tags
             col = QStandardItem(tags)
-            col.setFlags(Qt.ItemIsSelectable | Qt.ItemIsEnabled | Qt.ItemIsDragEnabled | Qt.ItemIsEditable)
+            col.setFlags(flags | Qt.ItemIsEditable)
             row.append(col)
 
             # Append Media Type
             col = QStandardItem(media_type)
-            col.setFlags(Qt.ItemIsSelectable | Qt.ItemIsEnabled | Qt.ItemIsDragEnabled | Qt.ItemIsEditable)
+            col.setFlags(flags)
             row.append(col)
 
             # Append Path
             col = QStandardItem(path)
-            col.setFlags(Qt.ItemIsSelectable | Qt.ItemIsEnabled | Qt.ItemIsUserCheckable | Qt.ItemIsDragEnabled)
+            col.setFlags(flags)
             row.append(col)
 
             # Append ID
             col = QStandardItem(id)
-            col.setFlags(Qt.ItemIsSelectable | Qt.ItemIsEnabled | Qt.ItemIsUserCheckable | Qt.ItemIsDragEnabled)
+            col.setFlags(flags | Qt.ItemIsUserCheckable)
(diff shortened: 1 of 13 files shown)


src/windows/models/files_model.py
```diff
@@ -189,43 +189,43 @@ def update_model(self, clear=True, delete_file_id=None):
...
                 # Get thumb path
-                thumb_path = self.get_thumb_path(file.id, thumbnail_frame)
+                thumb_icon = QIcon(self.get_thumb_path(file.id, thumbnail_frame))
             else:
                 # Audio file
-                thumb_path = os.path.join(info.PATH, "images", "AudioThumbnail.svg")
+                thumb_icon = QIcon(os.path.join(info.PATH, "images", "AudioThumbnail.svg"))
 
             row = []
+            flags = Qt.ItemIsSelectable | Qt.ItemIsEnabled | Qt.ItemIsDragEnabled | Qt. ItemNeverHasChildren
```
jonoomph, line 198: :+1:

jonoomph:
@ferdnyc This is all great stuff! Thanks!

jonoomph:
LGTM, I'm gonna merge this one

ferdnyc:
Oh, cool. I was gonna replace my status bar progress text with a generalized version of the new progress bar from the Blender interface... But that still isn't finished, and I can just as easily do it in another PR.

PR #3787, opened 2020-10-24
Housekeeping: Eliminate all star imports, etc.

ferdnyc:
A few days back, our Codacy problem count suddenly skyrocketed when I went through and enabled a whole bunch of checks that we previously weren't getting flagged on in the code. Things like (most significantly) dependency problems like missing imports. Codacy will now flag all unknown symbols it encounters in the code.

In order to do that, though, the code can't be using any star imports (`from foo import *`), because obviously then there's no way to know whether a symbol is valid or not unless you have the imported module to examine.

So, this PR removes _all_ of the remaining star imports in the source tree. It also:
- Removes most unused variables
- Reformats sections of code to reduce overlong lines


src/classes/app.py
```diff
@@ -142,8 +141,11 @@ def __init__(self, *args, mode=None):
         libopenshot_version = openshot.OPENSHOT_VERSION_FULL
         if mode != "unittest" and libopenshot_version < info.MINIMUM_LIBOPENSHOT_VERSION:
             QMessageBox.warning(
-                None, _("Wrong Version of libopenshot Detected"),
-                _("<b>Version %(minimum_version)s is required</b>, but %(current_version)s was detected. Please update libopenshot or download our latest installer.") % {
+                None,
```
jonoomph, line 144: This will change our translation text...

src/classes/app.py
```diff
@@ -142,8 +141,11 @@ def __init__(self, *args, mode=None):
...
+                _("Wrong Version of libopenshot Detected"),
+                _("<b>Version %(minimum_version)s is required</b>, but %(current_version)s was detected. \
```
jonoomph, line 146: Another wrapped translation

jonoomph:
LGTM, other than some of the wrapped translations. I'm not sure if those are really issues, but it would be good to test the POT generation, and make sure these strings still match exactly as before. :+1:

src/classes/app.py
ferdnyc: @jonoomph 
Tried to be careful about that, but I wasn't careful enough. Fixed so the translation master generates the same way.

jonoomph:
LGTM

PR #3305, opened 2020-03-20
New AppRun binary for linux AppImages

ferdnyc:
To add an environment variable to the AppImage sandbox, turns out it needs to be compiled into the `AppRun` executable itself. So, I built a new binary with `OPENSSL_CONF=/dev/null` compiled in, and added it to the `installer` dir so that the `build-server.py` script (with appropriate modifications) can pick it up from there.

Hopefully, for real this time, this
Fixes #3242

Issue #3293, opened 2020-03-12
OpenShot-2.5.x wont run Debian 11 Bullseye


brcisna:
Hello All,

 A fresh install of Debian 11 / Bullseye ,,only OpenShot-2.4.x will run . 
The versions including nightly builds of OS-2.5.x fails with the following.

The below terminal output only happens after trying to load a saved project,,,which I have many of,,all created with OS-2.4.x.

140469257038720:error:25066067:DSO support routines:DLFCN_LOAD:could not load the shared library:dso_dlfcn.c:185:filename(libssl_conf.so): libssl_conf.so: cannot open shared object file: No such file or directory
140469257038720:error:25070067:DSO support routines:DSO_load:could not load the shared library:dso_lib.c:244:
140469257038720:error:0E07506E:configuration file routines:MODULE_LOAD_DSO:error loading dso:conf_mod.c:285:module=ssl_conf, path=ssl_conf
140469257038720:error:0E076071:configuration file routines:MODULE_RUN:unknown module name:conf_mod.c:222:module=ssl_conf
QMutex: destroying locked mutex

They all end in the module=ssl_conf.

Not sure what that means.

This machine uses Intel HD graphics  but same results on a  machine with an old Matrix G200ex chip as well.

Thank You

SuslikV:
I think your issue is variant of: https://github.com/OpenShot/openshot-qt/issues/3242
That itself has fix: https://github.com/OpenShot/openshot-qt/pull/3283

ferdnyc:
Yeah, that reminds me, I need to merge that. It should be fairly harmless, except for affected systems.

ferdnyc:
If you're interested, @brcisna , you can try downloading the next Daily Build AppImage package (which should be ready in a few minutes) from https://openshot.org/download/ and see if that does indeed fix the problem internally.

ferdnyc:
(Please feel free to reopen if it _doesn't_.)

ferdnyc:
My first fix (first two, actually) for this issue and #3242 was unsuccessful. However, the evidence is strong that #3305 solves this For Reals™, finally. That fix is included in the most recent AppImage Daily Builds.

0xcrypto:
[...]

PR #3283, opened 2020-03-08
AppImage: Fix a crash issue running on systems with newer OpenSSL

ferdnyc:
On systems with newer versions of openssl, the installed file `/etc/ssl/openssl.cnf` is incompatible with our bundled version of `libssl.so`. But it will be loaded by default, leading to a crash, unless `OPENSSL_CONF` is set to `/dev/null` (or some other invalid path) in the AppImage environment.

Fixes #3242

Issue #3242, opened 2020-02-23
Cannot open project file created with Openshot 2.4.4 on Debian 10


cometchaserde:
**Describe the bug**
Cannot open Openshot.osp created with Openshot 2.4.4 on Debian 10

To check if it is a problem, I created an new project with a video + music file with Openshot 2.4.4 and saved it. Next I tried to open it with Openshot 2.5.0.

Openshot doesn't open the file. Openshot 2.5.0 just crashes. It happens also with the latest build.

**Expected behavior**
Open the project file

**System Details**
- OpenShot Version: 2.5.0
- libopenshot Version:  0.2.4
- Platform: Linux-4.19.0-6-amd64-x86_64-with-debian-10.2
- Processor: 
- Machine: x86_64
- Python version: 3.4.3
- Qt5 version: 5.2.1
- PyQt5 version: 5.2.1
- Qt Detected Languages: ['de-DE']
- LANG Environment Variable: de_DE.UTF-8
- LOCALE Environment Variable: 
- Daily Build Installed: http://github.com/OpenShot/openshot-qt/releases/download/OpenShot-v2.5.0-dev1-1582091950-fe8ea216-836e49fa-x86_64.AppImage

**Log Files**
- http://logs.openshot.org/lcgdtynpdkibq06yguwirjn9457pblug/openshot-qt.zip
- http://logs.openshot.org/lcgdtynpdkibq06yguwirjn9457pblug/libopenshot.zip

**Exception / Stacktrace**
```No stacktrace found in log files```

**Screenshots** (Optional)
If applicable, add screenshots to help explain your problem. You can include screenshots by
copy/pasting them on GitHub or dragging-and-dropping into the GitHub page. All images are public,
so please don't post screenshots containing personal information.

SuslikV:
Can you still open it in v2.4.4?

cometchaserde:
Hi,

yes I can still open it in 2.4.4

SuslikV:
Can you share a copy of the `.osp` file that you are unable to open with the OpenShot?

cometchaserde:
[TestOpenshot-Projekt.zip](https://github.com/OpenShot/openshot-qt/files/4246752/TestOpenshot-Projekt.zip)

Hope you can use it.

SuslikV:
I didn't  found issues with the project itself. Are you sure that you have issues exactly with this project (1 video clip + 1 audio clip)?

cometchaserde:
Hi,
[...]

PR #3793, opened 2020-10-26
Updating AppImage logic to support desktopintegration again

jonoomph:
Fixing a regression which did not use the official AppRun executable... and thus our openshot-qt.wrapper was never called, which is what enables desktopintegration. I know this must still have some issues, since I'm having to do some of my own bootstrapping (which I should need to do, in order for libopenshot to be found at runtime). But this seems to work on all the distros I've tested on so far.

**The basic flow:**
1 - AppImage is mounted and executes `AppRun` in the root of our image. This is the official version of AppRun executable, and is a binary file.
2 - AppRun searches our `*.desktop file`, and locates the `openshot-qt-launch.wrapper` line, and then searches for that file, and executes it. This file is the `desktopintegration` file packaged along with AppImage code (renamed as openshot-qt-launch). It prompts the user to create a launcher, and then finally, removes the `.wrapper` and calls `openshot-qt-launch`. 
3 - `openshot-qt-launch` is a renamed version of `installer/launch-linux.sh`. This exports `LD_LIBRARY_PATH` and `QT_PLUGIN_PATH` to the current directory (i.e. the `usr/bin` folder in our AppImage), and then finally, executes `usr/bin/openshot-qt`, which is a binary created by cx_freeze (and is the official entry point to our source code). Without this script, libopenshot and Qt fail to be found inside the AppImage... which is strange and incorrect. But it's a simple work-around for now.

Finally, I also enabled auto-transform, which enables transform for any selected clip. To make it disappear, simple unselect all clips. I think it really improves visibility on transforming elements during video editing. It can be disabled in the Preferences.

### DIFF
--- a/installer/build-server.py
+++ b/installer/build-server.py
@@ -354,22 +354,18 @@ try:
         with open(desk_in, "r") as inf, open(desk_out, "w") as outf:
             for line in inf:
                 if line.startswith("Exec="):
-                    outf.write("Exec=openshot-qt.wrapper %F\n")
+                    outf.write("Exec=openshot-qt-launch.wrapper %F\n")
                 else:
                     outf.write(line)
 
         # Copy desktop integration wrapper (prompts users to install shortcut)
         launcher_path = os.path.join(app_dir_path, "usr", "bin", "openshot-qt-launch")
         os.rename(os.path.join(app_dir_path, "usr", "bin", "launch-linux.sh"), launcher_path)
-        desktop_wrapper = os.path.join(app_dir_path, "usr", "bin", "openshot-qt.wrapper")
+        desktop_wrapper = os.path.join(app_dir_path, "usr", "bin", "openshot-qt-launch.wrapper")
         shutil.copyfile("/home/ubuntu/apps/AppImageKit/desktopintegration", desktop_wrapper)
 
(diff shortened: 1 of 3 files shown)


ferdnyc:
@jonoomph 

The reason we were using our own AppRun binary was #3283 & #3305. (Well, #3242 / #3293, which they fixed).

I compiled it from the sources the official one was built from, though, with the only change being the addition of that environment variable.

If `/home/ubuntu/apps/AppImageKit/AppRun` file is still the same ancient one compiled back in 2014, then without that fix, the AppImages will likely go back to crashing on launch, on Debian 11.

But if it's been rebuilt (as part of the Bionic builder update?) to use a newer `libssl.so`, then that should be fine, the workaround wouldn't be needed anymore.

ferdnyc:
(Actually, I don't think the AppRun itself was the problem, it was something bundled into the AppImage — possibly the `cx_Freeze` binary launcher, because the only way to get `OPENSSL_CONF=/dev/null` into the environment soon enough to _prevent_ the crash was to build it into the AppRun binary itself.

But if that old-incompatible-OpenSSL dependency has been upgraded, then like I said no longer a problem. We should check that this didn't re-break Debian 11 support in the AppImage, though.

PR #3795, opened 2020-10-27
Profile: Seek to first frame on change

### LINKED ISSUE #3119: Changing project profile moves playhead back to start of Timeline, but doesn't update position
**Describe the bug:**
Changing the profile resets the interface, but the playhead's position isn't updated. As a result, attempting to adjust property values will apply keyframes in the wrong position (the playhead's former position, _not_ the first frame).

1. With a clip on the Timeline, select it and move the playhead to some frame past the first
2. Open Choose Profile and change the selected project profile
3. Click Close
4. With the playhead now seemingly on the first frame, re-select the clip
5. Change the Alpha value to 0.5
6. Note that a keyframe has been added at the _former_ position of the playhead.

**Expected behavior:**
Setting a keyframe after changing the profile applies the value at the correct position on the Timeline.

**System Details:**
- OpenShot Version [e.g. 2.4.3]: 2.4.4-dev2 (latest `develop` branch)
- Operating System / Distro: Linux, Fedora 31 x86_64 

**Screenshots:**
(Note position of green keyframe mark)

image

CC: @jonoomph , maybe this is what you were referring to last week, about keyframe issues?

ferdnyc:
Visually, changing the project profile resets the playhead marker back to the start of the timeline. But the preview player was retaining the old position, causing subsequent operations to be performed in the wrong place. Now, we `Seek()` the player back to frame 1 as well.

Fixes #3119

### DIFF
--- a/src/windows/profile.py
+++ b/src/windows/profile.py
@@ -83,7 +83,8 @@ class Profile(QDialog):
                     self.profile_paths[profile_name] = profile_path
 
                 except RuntimeError as e:
-                    # This exception occurs when there's a problem parsing the Profile file - display a message and continue
+                    # This exception occurs when there's a problem parsing
+                    # the Profile file - display a message and continue
                     log.error("Failed to parse file '%s' as a profile: %s" % (profile_path, e))
 
         # Sort list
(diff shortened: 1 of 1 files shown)


jonoomph:
LGTM!

PR #3796, opened 2020-10-27
Fix string formatting on yml stale message 

jonoomph:
Fixing YML string for Stale plugin so it doesn't mess up our string formatting (instead of folded style, use a literal string)

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -10,7 +10,7 @@ exemptLabels:
 # Label to use when marking an issue as stale
 staleLabel: stale
 # Comment to post when marking an issue as stale. Set to `false` to disable
-markComment: >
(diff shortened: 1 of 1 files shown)


ferdnyc:
You're killing me, Stale.

Screenshot from 2020-11-06 15-14-36

I would pay real money for a way to watch the repo _without_ email notifications for all of Stale's activity.

(I know in the long run it'll improve the situation, but _man_ the road to reach that point is a rough one. It's like this **every day**, lately — several times a day, in fact. **#AreWeThereYet?**)

jonoomph:
I know... same feeling here.

*-Jonathan Thomas*
OpenShot Video Editor
http://www.openshot.org

PR #3802, opened 2020-10-28
More AppImage Icon Fixes/Experiments

jonoomph:
Apparently, AppImage desktopintegration assumes the icon is PNG and not SVG, which causes a failure to load the AppImage icon for desktop launcher on some distros (i.e. Chrome OS)


jonoomph:
Nevermind, this is not good... icon is grainy and needs to be SVG. I'm going in the wrong direction. Gonna try and fix the desktopintegration script instead, to support SVG icons

PR #3803, opened 2020-10-29
Fixing AppImage Icon for Chrome OS

jonoomph:
I've done a few new things to resolve the AppImage icon
- During the launch script in the AppImage, we now manually copy a 'normal' and 'large' thumbnail into ~/.cache/thumbnails/
- I fixed the internal AppImage location of all icons: `usr/share/icons/hicolor/128x128/apps/openshot-qt.png` (for example)
- We no longer use an SVG for our desktop icon, since that does not seem to be supported with .cache (as far as I can tell)
- I've tested this on Ubuntu 20.04 and the newest version of Chrome OS, but work, and include desktop launchers with an icon

I'm happy to take further PRs to improve our AppImage support, especially with respect to icons and launchers, but for now, this seems much better than before.

### DIFF
--- a/installer/build-server.py
+++ b/installer/build-server.py
@@ -349,6 +349,9 @@ try:
         # See: https://docs.appimage.org/reference/appdir.html
         shutil.copyfile(icons[3][1], os.path.join(app_dir_path, ".DirIcon"))
 
+        # Install program icon
+        shutil.copyfile(icons[0][1], os.path.join(app_dir_path, "openshot-qt.svg"))
+
         dest = os.path.join(app_dir_path, "usr", "share", "pixmaps")
         os.makedirs(dest, exist_ok=True)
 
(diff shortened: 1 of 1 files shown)


installer/build-server.py
```diff
@@ -327,35 +327,51 @@ def parse_version_info(version_path):
...
+        shutil.copyfile(icons[0][1], os.path.join(dest, "openshot-qt.svg"))
+        shutil.copyfile(icons[1][1], os.path.join(dest, "openshot-qt.png"))
 
         # Install MIME handler
         dest = os.path.join(app_dir_path, "usr", "share", "mime", "packages")
         os.makedirs(dest, exist_ok=True)
 
         shutil.copyfile(os.path.join(PATH, "xdg", "org.openshot.OpenShot.xml"),
-                        os.path.join(dest, "org.openshot.OpenShot.xml"))
+                        os.path.join(dest, "openshot-qt.xml"))
```
jonoomph, line 366: AppImage interprets the .desktop file name as the APP name, and uses it to install icons, etc... So it must match.

PR #3805, opened 2020-10-29
Blender: Make color-picker dialog non-modal

### LINKED ISSUE #3799: Color picker blocks the render view
**Describe the bug:**
The dialog picker pops up in the middle of the app window and moves the window with it, it's always blocking the render view. That makes it impossible to color select anything in that region (see example screenshot with purple outline). 

**Steps to reproduce the behavior click behind behavior:**
1. Loading OpenShot
2. Load color picker from effect property (ex. Chrome Key color)
3. Try moving the color picker away from the render view

**Expected behavior:**
Color picker can be moved out of the way and any portion of the render view can be selected

**System Details:**
- OpenShot Version: 2.5.1
- Operating System / Distro: Ubuntu 18.04

**Screenshot:**
picking_behind_dialog

ferdnyc:
The "Pick a screen color" feature isn't too useful if you can't move the dialog out of the way of the rest of the interface.

Fixes #3799

### DIFF
--- a/src/windows/views/blender_listview.py
+++ b/src/windows/views/blender_listview.py
@@ -222,20 +222,41 @@ class BlenderListView(QListView):
         # Get translation object
         _ = get_app()._tr
 
-        # Show color dialog
         color_value = self.params[param["name"]]
         currentColor = QColor("#FFFFFF")
         if len(color_value) >= 3:
             currentColor.setRgbF(color_value[0], color_value[1], color_value[2])
-        newColor = QColorDialog.getColor(currentColor, self, _("Select a Color"),
-                                         QColorDialog.DontUseNativeDialog)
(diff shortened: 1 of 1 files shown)


ferdnyc:
Hmm. It does strike me now that it might make sense to do the same thing for the _non_-animated Titles dialog, as well.

jonoomph:
Nice, that makes sense! Thanks!

ferdnyc:
@jonoomph 

\*nod\* I've now concluded we should do the same thing for both the regular Title Editor, _and_ any of the color Properties values, for the exact same reason. The "eyedropper" functionality is useless if you can't reach the rest of the application's UI.

I'll pull this out into a separate module under `windows/` at some point and share the love.

PR #3806, opened 2020-10-29
New OpenCV Effect Pre-Processing UI

jonoomph:
Heavily related to https://github.com/OpenShot/libopenshot/pull/585. This PR adds in some new Effect pre-processing, which can happen prior to the effect actually being added to a Clip. This allows for custom options, and a new ClipProcessing integration with libopenshot. Primarily this is to allow for processing a video, frame by frame, and save all the tracking/motion data into a file which can be quickly read at runtime... for fast previews and crazy cool AI integration.

Screenshot from 2020-10-29 03-11-16


jonoomph:
Merging in opencv processing UI

PR #3814, opened 2020-11-01
Title templates: Make all transparent backgrounds black

ferdnyc:
As a follow-on to #3813 and the issues I brought up there, this PR modifies the source of all of our title template SVGs, and sets the background fill color to `#000` on all of the templates with a fully transparent background.

In theory, the background color doesn't _matter_ when it's fully transparent, but it does matter in terms of how the color-change button is shown and what shows in the color picker, and this is much more representative of what you'll actually see.

(This change also won't affect Inkscape, It still shows a white background by default, or a checkerboard background if checkerboard transparency is turned on for the file.)


ferdnyc:
With this change (but without #3813), the same Title Editor template I used for my screenshots there:

Screenshot from 2020-11-01 09-06-02

And, how it shows in the regular QColorPicker:

Screenshot from 2020-11-01 09-06-07

(But it's still misleading, because if you change that color it will still have no effect unless you raise the alpha level.)

ferdnyc:
Also merging this, as it's purely a perception thing and I can't even fathom someone arguing that the _white_ transparent backgrounds our templates currently have is _deliberate_ or _useful_ in some way. It's just that nobody ever got around to changing it. Until now.

PR #3815, opened 2020-11-01
Use ColorPicker for all color dialogs

### LINKED ISSUE #3798: Color picker picks the wrong value if dialog has dark background
**Describe the bug:**
When using the chroma tool I noticed it was working very poorly. After playing with the colors I noticed the ones grabbed by the picker were much darker, which was the problem. I realized the color picker is actually pulling values off the screen--not the render view--so the dark overlay the dialog puts over the window (at least on my dialog rendering implementation) will alter the picked color. See screenshot below. 

1. Loading OpenShot
2. Load color picker from effect property (ex. Chrome Key color)
3. Select color from render view
4. If dialog has darked the 

**Expected behavior:**

**System Details:**
- OpenShot Version: 2.5.1
- Operating System / Distro: Ubuntu 18.04

**Screenshots:**
color_picker_background

### LINKED ISSUE #3799: Color picker blocks the render view

ferdnyc:
OK, building on #3813 (and, in fact, incorporating it, as it's a necessary prerequisite), this PR makes use of the replacement ColorPicker instead of QColorDialog, in all places where we accept a color as user input.

It also adds `*args, **kwargs` to some of our dialog windows' `__init__()` functions, as I noticed they weren't there. As a result of that, we can't (and Qt can't) pass any standard Qt args to those functions, which means you can't do things like parent them.

(Of course, the fac that our dialogs aren't parented or parentable is the only thing preventing them from also being modal-by-default, so it could be seen as a happy accident. Then again, it might be a good idea to properly parent them, so that Qt can clean them up for itself.)

Oh, yes, this one finally...
Fixes #3799 and fixes #3798


jonoomph:
LGTM! Is this ready to merge? If so, go for it. :+1:

jonoomph:
@ferdnyc I resolved conflicts on this one. Can you take a quick looky, and make sure this all still works correctly. If so, let's merge it! Thanks! 👍

jonoomph:
Merging this one as well. I'm gonna be doing a bunch of testing over the weekend, and I really wanted this one included as well. :+1:

PR #3816, opened 2020-11-01
Titles: Rewrite XML parsing/mods, add style_tools

ferdnyc:
#### Update

Just pushed a few more commits that re-work the title editor data handling and template parsing. Now:

1. `style_to_dict()` is used for all style parsing (`find_in_list()` is now gone)
1. The `TitlesModel` wrapper class is now a `QObject`, and gets parented to `MainWindow`
1. The underlying `QStandardItemModel`-based data model now only has one column. Each "row" in the model holds a self-contained `QStandardItem` with all of the data for a given template.
    1. Former columns 0 and 1 (the icon and name) are combined
    1. The data from column 2 (the full path to the template) is instead stored in the main/only "column" under a data role which can be accessed by using `TitleRoles.PathRole` as the role
1. The editor dialog now persists user-customized filenames for generated titles across launches of the Advanced Editor, instead of always resetting to the default name. Fixes a bug reported in a comment on #3832.

#### Original description

This PR grew out of my testing for the previous three PRs.

There was a lot of redundant code in the Title Editor source, most of it to do with the manipulation of the SVG file's contents to update the `style=` attribute on various nodes.

I hit on one major efficiency that really streamlined things.

The code was already doing the work to `split()` the style attribute's string on `;`, to extract individual properties. But then it was doing a ton of laborious substring matching and whatnot to find the _specific_ CSS properites it wanted to modify.

What I realized was, just like the property list could be split apart using `;`, the _properties_ themselves could be split at the first `:` to form key-value pairs, and then naturally the whole thing could be loaded into a `dict` which makes manipulation trivial.

So, this PR adds `classes/style_tools.py`, a module containing three functions:

```python
def style_to_dict(style: str) -> dict:
    """Explode an SVG node style= attribute string into a dict representation"""

def dict_to_style(styledict: dict) -> str:
    """Turn an exploded style dictionary back into a string"""

def set_if_existing(d: dict, existing_key, new_value):
```

That last one (which I left out the docstring for, oops) has a pretty obvious name, and does what it says: It updates the value of a dictionary key **ONLY IF** that key already exists in the dictionary. If not, then it won't insert a new key. Some of the code in the Title Editor was concerned with only updating existing style attributes where they already existed.

Armed with the new style tools, I then rewrote the update-handling methods of the TitleEditor class to use them. The result is a drop of almost 100 lines of code, not to mention a ton of inefficiency and redundancy, from `title_editor.py`


src/windows/title_editor.py
```diff
@@ -586,83 +530,52 @@ def update_background_color_button(self):
...
-                ar[ff] = "font-family:" + self.font_family
-            if fp:
-                ar[fp] = "font-size:%spx" % self.font_size_pixel
-            # rejoin the modified parts
-            t = ";"
-            self.title_style_string = t.join(ar)
+            ard = style_to_dict(s)
+            set_if_existing(ard, "font-style", self.font_style)
+            set_if_existing(ard, "font-family", self.font_family)
+            set_if_existing(ard, "font-size", f"{self.font_size_pixel}px")
```
ferdnyc, line 544: Oh, yes, purely by coincidence this PR also introduces the first [f-string](https://realpython.com/python-f-strings/) info the OpenShot codebase, making it officially require Python 3.6 or later to run.

It's time. Python 3.5 has [actually been end-of-life since September](https://www.python.org/dev/peps/pep-0478/). And f-strings are just too damn convenient _not_ to use!

src/windows/title_editor.py
```diff
@@ -426,83 +442,45 @@ def btnFont_clicked(self):
...
-            self.display_svg()
+            self.save_and_reload()
 
     def find_in_list(self, l, value):
         '''when passed a partial value, function will return the list index'''
-        for item in l:
+        for i, item in enumerate(l):
             if item.startswith(value):
-                return l.index(item)
+                return i
```
ferdnyc, line 451: I also goosed the efficiency of `find_in_list()` a bit, since it's still used in a couple of parts of the code.

src/windows/title_editor.py
```diff
@@ -540,7 +540,7 @@ def set_font_style(self):
             s = text_child.attributes["style"].value
             ard = style_to_dict(s)
             set_if_existing(ard, "font-style", self.font_style)
-            set_if_existing(ard, "font-family", self.font_family)
+            set_if_existing(ard, "font-family", f"'{self.font_family}'")
```
ferdnyc: The old code wasn't doing this either, and it admittedly _seems_ to work fine without it, but technically font family names with a space are supposed to be inside single quotes, in CSS. (No real harm in doing it for all font family names, even one-word ones.)

jonoomph:
LGTM! For some reason I thought this was already merged in!? Any-who, it looks great. Nice work!

PR #3826, opened 2020-11-06
Remove pre-filled title text from issue report templates

ferdnyc:
Users too-often submit bugs titled "Give your bug report a descriptive title" (Or "Give your bug report a descriptive title" with something else written after that.)

That's because our templates pre-fill the title field with that text. But if the title starts out _blank_, the user won't be allowed to submit the report **at all** unless they write a real title of their own — Github disables the Submit button until some text has been entered into the field.

So, starting out with a blank field seems like the better option.


ferdnyc:
I'm going to merge this, as yet another Issue was opened today with the title "Give your feature request a descriptive title".

Issue #3825, opened 2020-11-06
Properties for transitions are not visible in daily build


rexdk:
Daily build from 29Oct20
Win 10

Right click on a transition on a clip; choose properties. The properties panel shows but there are no longer any transition properties there (eg no Brightness, Contrast etc). Something has changed I think - it's been working fine, eg with a daily build from July.

ferdnyc:
That's gotta be my fault. I'll take a look thanks.

ferdnyc:
@rexdk 
Hmm. No, works OK for me with the current `develop` branch code under Linux.

Can you reproduce, then [post your `openshot-qt.log` file](https://github.com/OpenShot/openshot-qt/blob/develop/CONTRIBUTING.md)? There's probably a Python traceback in there that will point to the location of the bug.

rexdk:
1.delete log files
2.. open openshot
3. import an image
4. add to timeline and extend length
5. drag transition to end of image
6. right click on transition > Properties
7. no properties shown
8. close openshot without saving



log files and screen shot attached

Sorry to burden you - again ;)

![missing transition properties](https://user-images.githubusercontent.com/44414261/98435934-f161c380-20ce-11eb-9469-f29cb1d15e32.png)

[libopenshot.log](https://github.com/OpenShot/openshot-qt/files/5504462/libopenshot.log)
[openshot-qt.log](https://github.com/OpenShot/openshot-qt/files/5504463/openshot-qt.log)

rexdk:
just checked - my daily build is 25 Oct I think, so possibly it was found and fixed after that?

ferdnyc:
@rexdk 
Very possible, that code has been changing some lately. 

Ah! Yes, you managed to grab a Daily Build generated during a (hopefully brief) interval when I was a dumbass and merged broken code:
```
07:48:53 ERROR exceptions: Unhandled Exception
Traceback (most recent call last):
  File "C:\Program Files\OpenShot Video Editor\windows\models\properties_model.py", line 105, in update_item_timeout
    t = get_app().window.timeline_sync.timeline.GetTimelineEffect(item_id)
AttributeError: 'Timeline' object has no attribute 'GetTimelineEffect'
```
That was fixed the next day, with 4d6b05d9efa42223b279849cb9d3a31bdb48e38f. Any build generated after October 26 should have the fix.

rexdk:
[...]

PR #3828, opened 2020-11-08
Stable build dates

ferdnyc:
This PR, along with OpenShot/libopenshot-audio#107 and OpenShot/libopenshot#588, implements an idea I had while discussing Daily Build versions with @rexdk in OpenShot/openshot-qt#3825:

By using the UNIX epoch date of the _openshot-qt **commit**_ that the build is generated from, instead of the value of `time.time()`, we can ensure that the set of Daily Build files generated for a given build run all have the _same_ version identifiers in their filename, and differ only by the specifics of the format/architecture targeted.

Before, due to the use of `int(time.time())` to set the first part of the build identifier, each of the builders' files was named slightly differently. 

Now, instead of e.g. (from the most recent Daily Build run):
```
OpenShot-v2.5.1-dev2-1604704263-b910d2f6-2cb4eeff-x86_64.AppImage
OpenShot-v2.5.1-dev2-1604704263-b910d2f6-2cb4eeff-x86_64.AppImage.torrent
OpenShot-v2.5.1-dev2-1604704807-b910d2f6-2cb4eeff-x86.exe
OpenShot-v2.5.1-dev2-1604704813-b910d2f6-2cb4eeff-x86_64.exe
OpenShot-v2.5.1-dev2-1604704807-b910d2f6-2cb4eeff-x86.exe.torrent
OpenShot-v2.5.1-dev2-1604704813-b910d2f6-2cb4eeff-x86_64.exe.torrent
OpenShot-v2.5.1-dev2-1604707336-b910d2f6-2cb4eeff-x86_64.dmg
OpenShot-v2.5.1-dev2-1604707336-b910d2f6-2cb4eeff-x86_64.dmg.torrent
```
We'd instead have (based on the timestamp of the triggering commit):
```
OpenShot-v2.5.1-dev2-1604699062-b910d2f6-2cb4eeff-x86_64.AppImage
OpenShot-v2.5.1-dev2-1604699062-b910d2f6-2cb4eeff-x86_64.AppImage.torrent
OpenShot-v2.5.1-dev2-1604699062-b910d2f6-2cb4eeff-x86.exe
OpenShot-v2.5.1-dev2-1604699062-b910d2f6-2cb4eeff-x86_64.exe
OpenShot-v2.5.1-dev2-1604699062-b910d2f6-2cb4eeff-x86.exe.torrent
OpenShot-v2.5.1-dev2-1604699062-b910d2f6-2cb4eeff-x86_64.exe.torrent
OpenShot-v2.5.1-dev2-1604699062-b910d2f6-2cb4eeff-x86_64.dmg
OpenShot-v2.5.1-dev2-1604699062-b910d2f6-2cb4eeff-x86_64.dmg.torrent
```
Much nicer, IMHO.

This PR adds capture of the openshot-qt commit timestamp (as a UNIX epoch time), and storage in the `share/openshot-qt` metadata file.

It also updates `build-server.py` to:
1. Expect a new field `COMMIT_DATE_UNIX`, when reading the metadata files for all three architectures
1. Prefer `int(version_info['openshot-qt']['COMMIT_DATE_UNIX'])` as the source of the timestamp component of the build filename, although it will fall back to `int(time.time())` if for some reason the commit date isn't available.

This change may conceivably cause the Daily Build timestamps to initially go _backwards_ slightly in time, if the most recent commit ends up being timestamped _before_ the wall-clock timestamps of the previous set of builds. But even if that happens, it should hopefully be a one-time thing.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -18,7 +18,7 @@ linux-builder:
     - unzip artifacts.zip
     - cp -r "$CI_PROJECT_DIR/build/install-x64/python/." "$CI_PROJECT_DIR"
     - export LD_LIBRARY_PATH=$CI_PROJECT_DIR/build/install-x64/lib:$LD_LIBRARY_PATH
-    - echo -e "CI_PROJECT_NAME:$CI_PROJECT_NAME\nCI_COMMIT_REF_NAME:$CI_COMMIT_REF_NAME\nCI_COMMIT_SHA:$CI_COMMIT_SHA\nCI_JOB_ID:$CI_JOB_ID" > "build/install-x64/share/$CI_PROJECT_NAME"
+    - echo -e "CI_PROJECT_NAME:$CI_PROJECT_NAME\nCI_COMMIT_REF_NAME:$CI_COMMIT_REF_NAME\nCI_COMMIT_SHA:$CI_COMMIT_SHA\nCI_JOB_ID:$CI_JOB_ID\nCI_PIPELINE_ID:$CI_PIPELINE_ID" > "build/install-x64/share/$CI_PROJECT_NAME"
     - git log $(git describe --tags --abbrev=0 @^)..@ --oneline --pretty=format:"%C(auto,yellow)%h%C(auto,magenta)% %C(auto,blue)%>(12,trunc)%ad %C(auto,green)%<(25,trunc)%aN%C(auto,reset)%s%C(auto,red)% gD% D" --date=short > "build/install-x64/share/$CI_PROJECT_NAME.log"
     - cd doc; make html SPHINXOPTS="-D html_theme_options.analytics_id=UA-4381101-5"; cd ..;
     - ~/auto-update-sphinx "$CI_PROJECT_DIR/build" "$CI_COMMIT_REF_NAME"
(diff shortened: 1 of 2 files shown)


ferdnyc:
Aw... crap. I used the _author_ commit timestamp (`git log --date=unix --format="%ad"`)... but I **should've** used the _committer_ timestamp (`git log --date=unix --format="%cd"`). That way, the timestamps are guaranteed to be in numerical order, and e.g. builds from two different branches can't end up with the same timestamp. (Well, except for the highly unlikely coincidence of them having been committed at the **exact same** second).

...Now I have to go edit all those files! :sob:

ferdnyc:
> Aw... crap. I used the _author_ commit timestamp (`git log --date=unix --format="%ad"`)... but I **should've** used the _committer_ timestamp (`git log --date=unix --format="%cd"`). 

Huh. Surprisingly enough, now that I've changed it everywhere, this doesn't actually seem to make a difference. At least locally on my system, `%ad` and `%cd` always seem to return the same timestamp, though I can't fathom how that's possible for things like merged branches. 

`¯\_(ツ)_/¯` Oh well. `%cd` is "more correct" even if it doesn't actually matter in practical terms.

jonoomph:
@ferdnyc Also, just FYI, I think the reason we based each daily build on the current time was to prevent collisions when uploading Artifacts to our Daily Build release on GitHub. I seem to recall lots of failures when trying to upload the same daily build again. So, fair warning. :+1:

jonoomph:
I also think those dates (i.e. commit date) is an ENV variable made available on GitLab for each build.

ferdnyc:
@jonoomph 

> @ferdnyc Also, just FYI, I think the reason we based each daily build on the current time was to prevent collisions when uploading Artifacts to our Daily Build release on GitHub. I seem to recall lots of failures when trying to upload the same daily build again. So, fair warning.

Well, in _theory_ there shouldn't really be collisions — at least, I'm hoping not. In theory no two sets of build jobs should be generated with the same "fingerprint", since any new build will be triggered by some commit that changes one of the name components. I suppose if a successful job is manually restarted for the exact same set of commits, even after the first successful one has uploaded artifacts, then there might be an issue. 

Or... oh, hm. There is one situation I can think of where that might happen, and it's one I didn't account for. If a previous job _partly_ succeeded, but had to be re-started, the re-builds on the jobs that **succeeded** the previous time might fail to upload, because the previous run already uploaded the same output file. Hmm.

I wonder if there's a way to make "file already exists" a non-fatal condition, since in those cases it's actually fine for the job to succeed without uploading; it doesn't need to upload since there was already a successful build uploaded.

ferdnyc:
@jonoomph 

I wouldn't actually mind the current timestamp, as long as all the files built from the same run used the **same** current timestamp, so that it was easier to determine which files "go together" and are built with the same contents.

Maybe the right approach is to just have a script job in the pipeline, in between the `libopenshot` build and the multiple `openshot-qt` architecture jobs, that does nothing but generate the base filename for all of the build artifacts in that run, and passes it on to all of the individual architecture builders?

ferdnyc:
@jonoomph 

The more I think about this, the more I think my **last** comment is the way to go, instead of the changes in this PR. What we need is a "basename generation" pipeline step that runs at the beginning of the openshot-qt CI pipeline. That way, all of the builds in any given pipeline will use the same filenames (with different extensions/suffixes), but that base filename can be generated using the current time so every _pipeline_ is guaranteed to generate files that are named differently from any of the others.

I'm just not 100% sure whether that can be done by just editing the `.gitlab-ci.yml` file, or if there's configuration that needs to be done administratively in the Gitlab instance to modify the pipeline steps. Any pointers?

ferdnyc:
Hmmm. Looking at [the docs](https://docs.gitlab.com/12.10/ee/ci/variables/), another option would be to use the `CI_PIPELINE_ID` instead of the current time. That should still be unique per run, but the same for all builders in the run. It's also waay shorter (four digits, currently, eventually 5), so that's nice too.

jonoomph:
@ferdnyc Yes, I like the idea of tacking the CI_PIPELINE_ID onto the filename! Like you say, much shorter, and less emphasis than a full datetime epoch.

ferdnyc:
I shall update the PR!

ferdnyc:
Done, and I closed the two in the other repos. There's really no need to capture `CI_PIPELINE_ID` in the metadata files for the libraries.

ferdnyc:
:roll_eyes: Random Travis failure is random. ("Couldn't get key" for the libopenshot-daily PPA, on some of the build jobs.)

ferdnyc:
Merging this. The project Mac builder still isn't finished because it's waiting for Notarization, but all of the builds were successful.

ferdnyc:
Me:

> Or... oh, hm.

> Hmm.

> Hmmm.

..Not _intentional_, but subconsciously my pensiveness is surprisingly linear in its progression.

I guess the only thing to say regarding that is: "Hmmmm."

PR #3839, opened 2020-11-16
Relocate libopenshot.log recovery code to exceptions.py

ferdnyc:
The `MainWindow` method `create_lock_file()`, of all things, has been repeatedly flagged as a high-complexity function. (One of four in `windows.main_window`, the remaining three being `MainWindow`'s `findAllMarkerPositions`, `keyPressEvent`, and most disturbing of all, `__init__`.)

In the case of `create_lock_file()`, the complexity came about because, right in the middle of doing its job (creating a lock file), it went off on this long, completely-unrelated tangent where it would parse the end of the last session in `libopenshot.log` looking for stack traces.

Since that has _nothing_ to do with creating a lock file, I pulled that entire tangential block of code, plus the `tail_file()` function it (exclusively) made use of, out of `windows.main_window` entirely. Instead I converted them both to functions in `classes.exceptions`. (Not a perfect spot, but a _much_ better one.)

Theoretically now, `exceptions.libopenshot_crash_recovery()` could even be called by `app.py` instead. Like it probably should be.

(There's still too much stuff in `MainWindow` that's really application-startup stuff, IMHO. I'd remind any skeptics that, again, **_`MainWindow.__init__()`_ is flagged for excessive complexity...**)

### DIFF
--- a/src/classes/exceptions.py
+++ b/src/classes/exceptions.py
@@ -25,17 +25,117 @@
  along with OpenShot Library.  If not, see <http://www.gnu.org/licenses/>.
  """
 
+import os
 import traceback
+import platform
+
+from classes import info
 from classes.logger import log
-from classes.metrics import track_exception_stacktrace
+from classes.metrics import track_exception_stacktrace, track_metric_error
 
 
 def ExceptionHandler(exeception_type, exeception_value, exeception_traceback):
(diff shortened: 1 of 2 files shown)


jonoomph:
LGTM!

PR #3850, opened 2020-11-19
Install a workflow to label PRs with merge conflicts

ferdnyc:
Let's try this again...

Label Merge Conflicts is a Github Actions tool that labels and comments on PRs with unresolved merge conflicts. Github very regrettably doesn't provide any visual indication of conflict status in a way that can be searched, filtered, or scanned for, the only way to discover merge conflicts is to open the individual PR page and scroll to the bottom.

Now, every push to the `develop` branch will trigger a workflow run that scans each PR, and if Github indicates that there are conflicts (based on the results of an async detection job that it runs, which are queryable using the Github API) then the designated label is applied. A comment is also left by the workflow. (The comment will generate an email notification for the submitter, something merely adding a label will not do).

If conflicts have been resolved, the label should eventually be removed automatically, though due to the async nature of the detection process and etc. there may be a delay. It's reasonable for submitters who have commit access to clear the label on their own if they choose to, once they've completed conflict resolution.

### DIFF
--- a/.github/workflows/label-merge-conflicts.yml
+++ b/.github/workflows/label-merge-conflicts.yml
@@ -1,0 +1,30 @@
+
+name: Label merge conflicts
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the develop branch
+on:
+  push:
+    branches: [ develop ]
+  pull_request:
+    branches: [ develop ]
+    types: [opened, synchronize, reopened]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


ferdnyc:
The workflow is hitting a permission error, which I suspect is due to the fact that it's defined on a branch in my fork. It _appears_ to run fine there, though I don't have any PRs with conflicts so I can't be sure. I don't know if the repo is configured to allow all Github Actions to run (as my fork is), or if it needs to be granted explicit permission. A question for @jonoomph .

Assuming the repo is _not_ locked down to only specific actions, then I suspect that simply merging this to `develop` will fix the permissions issue.

ferdnyc:
I'mma try merging this, I suspect it'll fix the permission errors. It works fine in my fork. If things break then I'll email @jonoomph to see if he can grant the workflow access to the repo.

PR #3870, opened 2020-11-26
Add __init__.py files to all Python source file paths (including all parent dirs), rename Blender scripts

ferdnyc:
Discovery of Python files when packaging is greatly facilitated by the presence of an `__init__.py` file in the same directory as the source, and all parent directories up to the root of the package. This marks each directory as a Python package/subpackage, and causes tools like cx_Freeze and setuptools to automatically collect all of the Python source files within. 

Without the `__init__.py` placeholder, inclusion can be piecemeal, as it's often dependent on either discovery of _classes_ within each source file (not all of our Python sources contain classes, or even functions), or on discovery of _dependencies_ on each source file.

Placing an `__init__.py` is a signal to packaging and distribution tools  that says, "all of these sources should be kept together".

To avoid the Blender Python scripts being treated as package sources, they have all been renamed from `src/blender/scripts/foo.py` to `src/blender/scripts/foo.py.in`. The preprocessing
code will rename the script to `foo.py` when it writes the modified version to the output dir.


ferdnyc:
I'm going to merge this as well — the changes are perhaps less uncontroversial, but having the `__init__.py` files in place shouldn't cause any problems, and renaming the Blender files is the right thing to do in order to keep them from being treated as Python sources when packaging OpenShot.

We're jumping through hoops to work around these and other issues, in our current packaging setup, and it would be nice to get that stuff more streamlined. This is how (or at least, it's the first step).

PR #3898, opened 2020-12-05
"Copy Keyframes > All" includes "Crop *" settings

spt3m4:
(Feel free to ignore / reject this PR or tell me i'm out of my mind)

But this may help someone who maybe expecting the Copy Keyframes > All feature to include the Crop keyframes (which work together with scaling X/Y and location X/Y to "frame" a video nicely).

Added data elements during copy in my proposal:
-`crop_x`
-`crop_y`
-`crop_height`
-`crop_width`

Tested it locally and seems to work. Not sure if there is a catch. Again, please just tell me if there is.

Hope this helps


ferdnyc:
@spt3m4 Thanks! Yeah, weird that those weren't included.

There are still a bunch of others missing, actually — really seems like "All" should include _all_ of them. We should probably change the approach entirely, instead of relying on a hardcoded list.

But until that's done, we have what we have. Now at least Crop will be covered as well!

PR #3909, opened 2020-12-09
Solve bug: Credits window not popping up

### LINKED ISSUE #3908: About/Credits not popping up
**Describe the bug:**
Hello,
When I run OpenShot (pulled from the "develop" branch), and I want to access the credits through Help > About OpenShot > Credits, the Credits window doesn't pop up and an error message is printed in the terminal.

1. Go to Help > About OpenShot
2. Click on Credits

**Expected behavior:**
A window containing the Credits pops out.

**System Details:**
- Operating System / Distro: Ubuntu 20.04.1 LTS
- Python: 3.8.5

**Log Files:**
[openshot-qt.log](https://github.com/OpenShot/openshot-qt/files/5669163/openshot-qt.log)

**Exception / Stacktrace:**
  File "/home/maesk/openshot/openshot-qt/src/windows/about.py", line 195, in load_credit
    windo = Credits()
  File "/home/maesk/openshot/openshot-qt/src/windows/about.py", line 287, in __init__
    self.developersListView = CreditsTreeView(
  File "/home/maesk/openshot/openshot-qt/src/windows/views/credits_treeview.py", line 88, in __init__
    self.refresh_view()
  File "/home/maesk/openshot/openshot-qt/src/windows/views/credits_treeview.py", line 47, in refresh_view
    self.credits_model.update_model(filter=filter)
  File "/home/maesk/openshot/openshot-qt/src/windows/models/credits_model.py", line 59, in update_model
    data = defaultdict("")
TypeError: first argument must be callable or None

**Solving the bug**
I wanted to confirm that other people (maybe on other platforms) have the same bug, to be sure that the problem doesn't come from my configuration. I would like to solve this bug, I'm going to propose a PR very soon.

KilianMaes:
The variables from the tuple returned by icon_mapping.get() were in wrong order (+ minor change). Could you verify that this PR solves the bug? I'm not totally sure about the expected window that should pop up because I never saw it before solving this issue.

Fixes #3908

### DIFF
--- a/src/windows/models/credits_model.py
+++ b/src/windows/models/credits_model.py
@@ -56,7 +56,7 @@ class CreditsModel():
 
         for person in self.credits_list:
             # Get details of person
-            data = defaultdict("")
+            data = defaultdict(list)
             data.update({
                 "name": person.get("name"),
                 "email": person.get("email"),
@@ -82,9 +82,9 @@ class CreditsModel():
             # Append type icon (PayPal, Kickstarter, Bitcoin, or Patreon)
             item = QStandardItem()
             for contrib in [n for n in self.icon_mapping if n in data["icons"]]:
-                (icon, tooltip) = self.icon_mapping.get(contrib, (None, None))
+                (tooltip, icon) = self.icon_mapping.get(contrib, (None, None))
                 item.setIcon(icon)
-                item.setToolTip(tooltip)
+                item.setToolTip(_(tooltip))
             item.setFlags(flags)
             row.append(item)
 
--- a/src/windows/models/credits_model.py
+++ b/src/windows/models/credits_model.py
@@ -56,7 +56,7 @@ class CreditsModel():
 
         for person in self.credits_list:
             # Get details of person
-            data = defaultdict(list)
+            data = defaultdict(str)
             data.update({
                 "name": person.get("name"),
                 "email": person.get("email"),
@@ -84,7 +84,7 @@ class CreditsModel():
             for contrib in [n for n in self.icon_mapping if n in data["icons"]]:
                 (tooltip, icon) = self.icon_mapping.get(contrib, (None, None))
                 item.setIcon(icon)
-                item.setToolTip(_(tooltip))
+                item.setToolTip(tooltip)
             item.setFlags(flags)
             row.append(item)
 


src/windows/models/credits_model.py
```diff
@@ -56,7 +56,7 @@ def update_model(self, filter=None, clear=True):
 
         for person in self.credits_list:
             # Get details of person
-            data = defaultdict("")
+            data = defaultdict(list)
```
ferdnyc, line 59: Hmm. This is a tricky one. I never noticed that `icons` was a _list_ of characters, in the JSON file, which would seem to support a `list` default.

<details>
<summary>Digression</summary>

If you look at the comprehension on 84:
```python
for contrib in [n for n in self.icon_mapping if n in data["icons"]]
```
I suspect `data["icons"]` was originally intended to be a string, and the icons would be single-character flags packed into that string. IOW, the idea was that `data["icons"]` would contain e.g. `"knd"` for a kickstarter contributor and Patreon supporter who's also a developer. The comprehension would match each of those characters, and it would display all of the appropriate icons.

Of course, whoever wrote this never got that to work right, because you can't set multiple icons for one cell of a data model. So then they added that `s` override for "multiple contributions". And then `"icons"` got turned into a list, because it's actually easier to export a list of characters than a packed string.
</details>

But the code also contains a lot of calls like `data["name"].lower()`, which would break on a list item. So, I suspect this should actually be `defaultdict(str)`. Which will work for `data["icons"]` as well, since an empty string is still compatible with the `in` usage in the comprehension.

```suggestion
            data = defaultdict(str)
```

Just for fun, I notice now that there are also several rows in the  JSON with a `"t"` in the icons list. What does `"t"` mean? Search me, it's not defined anywhere in the code. I suspect they might be unused translator credits. Those ended up getting pulled from the translation file itself, instead.

ferdnyc (commented):
Hey @KilianMaes, thanks for submitting this. Feedback follows, purely from reading the code as I haven't had a chance to test the branch yet (but will, after I post this).

src/windows/models/credits_model.py
ferdnyc: Heh. Oh, the `defaultdict` was me. (As was the `icon_mapping`, so the bug here is definitely my fault.)

Well, then I can say confidently that it _should_ be `defaultdict(str)`, because the previous code (which was convoluted and verbose, but basically equivalent) definitely treated the `data` values as strings. I just misunderstood how to create a `defaultdict`, apparently.

src/windows/models/credits_model.py
```diff
@@ -82,9 +82,9 @@ def update_model(self, filter=None, clear=True):
             # Append type icon (PayPal, Kickstarter, Bitcoin, or Patreon)
             item = QStandardItem()
             for contrib in [n for n in self.icon_mapping if n in data["icons"]]:
-                (icon, tooltip) = self.icon_mapping.get(contrib, (None, None))
+                (tooltip, icon) = self.icon_mapping.get(contrib, (None, None))
```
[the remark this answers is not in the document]
KilianMaes: Thanks!

src/windows/models/credits_model.py
```diff
@@ -82,9 +82,9 @@ def update_model(self, filter=None, clear=True):
...
                 item.setIcon(icon)
-                item.setToolTip(tooltip)
+                item.setToolTip(_(tooltip))
```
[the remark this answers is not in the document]
KilianMaes: I didn't notice that, you're right. I changed this

src/windows/models/credits_model.py
KilianMaes: Ok I replaced the "list" by "str" in the PR. Everythings seems to work fine. I didn't notice the "t" before you mention it so I have no idea what it stands for...

KilianMaes:
Thanks for the feedback, @ferdnyc . I made the changes in the two lines you mentioned, I hope it's ok now

ferdnyc:
@KilianMaes Sorry for the delay! I recently upgraded to Fedora 33, and since then OpenShot has been crashing on startup when I try to run it from the development tree! Haven't figured out what the issue is there, yet. I had to use a VM to test this.

I can confirm that this fixes the issue with the credits window not showing up.

The _filter_ field is COMPLETELY broken, I notice — anything you type in there causes a few lines to be displayed from the table, which have no relationship to what you typed in. So, I'll have to debug that next. But it's clearly unrelated to your fix, which is the only reason we can even see the problem. Thanks again, merging!

PR #3911, opened 2020-12-10
main_window: Move model/view init into separate method

ferdnyc:
This PR lowers the complexity of `main_window:__init__` to passable levels, by separating out the initialization of the model/view docks.


ferdnyc:
This is pretty uncontroversial, all it does is move a bunch of code to a different spot in the file. I'm going to merge it.

PR #3929, opened 2020-12-18
Credits: Fix data model

ferdnyc:
Following on from #3909, it turns out my use of a `defaultdict` in the credits model wasn't practical, because the input data file explicitly contains `null` values, which get converted into `None` on JSON import. If a dict contains _explicit_ `None` values, it will return them when those keys are dereferenced, ignoring the type set for the `defaultdict`.

So, I scrapped the `defaultdict`, wrote a loop to scrub out any `None` keys, _and_ protected all of the accesses to `person` keys by using `person.get()`.

Eliminates tracebacks when viewing credits, and fixes text filtering in the credits views.

### DIFF
--- a/src/windows/models/credits_model.py
+++ b/src/windows/models/credits_model.py
@@ -55,25 +55,22 @@ class CreditsModel():
         self.model.setHorizontalHeaderLabels(["", "", _("Name"), _("Email"), _("Website")])
 
         for person in self.credits_list:
-            # Get details of person
-            data = defaultdict(str)
-            data.update({
-                "name": person.get("name"),
-                "email": person.get("email"),
-                "website": person.get("website"),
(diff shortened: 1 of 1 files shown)


ferdnyc:
Hrrrm, as I feared we are seeing spurious crash-on-exit issues in the unit tests due to PyQtWebEngine (most likely).

The unit tests shouldn't _need_ to load a web backend **at all**, it's only the spaghetti interdependencies of `classes.app`, `windows.main_window`, `windows.views.webview`, and `classes.project_data` that forces creation of a web view. Perhaps there's a way to get that under control better.

ferdnyc:
Well, the CI failures with ubuntu-20.04 are explainable, so I'm going to push ahead and merge this despite them, as credits is pretty broken right now without this fix.

PR #3930, opened 2020-12-18
clip.rst: "it's" --> "its"

cboling2:
Removed two apostrophes that don't belong in the possessive word.


ferdnyc:
Heh, Yeah... I can almost guarantee there are plenty more of those, but two less (two fewer?) is a start! Thanks.

PR #3931, opened 2020-12-18
developers.rst: fixed "many bugs...can be added"

### LINKED ISSUE #3108: User manual content
While glancing over the [Developers section of the User Guide](https://www.openshot.org/static/files/user-guide/developers.html#the-big-picture), I happened to notice this passage, in particular the first sentence of the second paragraph:

UserGuide-Developer-ManyBugsQuote_1

_\*cough\*_ **_Ahem..._**

UserGuide-Developer-ManyBugsQuote_2

While in some ways I appreciate the brutal honesty there, we should probably fix that.

More practically, I noticed there were places where the Guide directs the user to some directory location under `/.openshot_qt/...`, which is just _wrong_. (`$HOME/.openshot_qt/...` or `~/.openshot_qt/...`, sure, that's acceptable, and even correct outside of Windows. But `/.openshot_qt/` is a very different thing.)

All of which is to say, #2989 is still open, and would clearly benefit from some attention.

cboling2:
True -- you can add many bugs -- but probably not what you want to say. :-)
This is one possible rewording.

Fixes #3108

### DIFF
--- a/doc/developers.rst
+++ b/doc/developers.rst
@@ -32,7 +32,7 @@ OpenShot Video Editor has 3 main components, a Python & PyQt user interface
 (:openshot-github:`libopenshot`). If you are not familiar with Python,
 PyQt, or C++, those would be great topics to research and learn more about at this point.
 
-However, many bugs and new features can be added with only Python knowledge, since the C++
+However, many bugs can be fixed and new features added with only Python knowledge, since the C++
 components are not involved in the user interface at all. Python is an amazing language, and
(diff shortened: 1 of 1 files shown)


ferdnyc:
At some point I had an issue filed for that. Ah, here it is. Stale got to it.

Fixes #3108

ferdnyc:
(I do like the graphical aid I made for that issue, though. My finest work.)

ferdnyc:
Rewording LGTM. It's concise, yet (and this is the kicker) _correct_. Always a plus. Thanks again!

PR #3937, opened 2020-12-26
Eliminate the Ubuntu Unity integration

ferdnyc:
I honestly don't think the Unity progress-bar integration was working even on Ubuntu anymore, but turns out now it's also causing crashes on Fedora 33 when Python tries to `import gi`.

This PR rips out the Unity integration, and moves the export-progress signals from `main_window` to `export` as they're now only used locally.

It also removes an `aboutToQuit` signal connection I had added, which is causing double unsaved-project prompts on exit if you say "No" to the first one.

### DIFF
--- a/src/windows/export.py
+++ b/src/windows/export.py
@@ -41,14 +41,15 @@ except ImportError:
 
 from xml.parsers.expat import ExpatError
 
-from PyQt5.QtCore import Qt, QCoreApplication, QTimer, QSize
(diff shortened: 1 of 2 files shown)


ferdnyc:
Going to merge this as well. There's some possibility that the Unity integration _is_ still useful to people, so its removal might impact them, but it's a non-critical piece of functionality. (Worst case, a progress bar is no longer displayed on the application icon during export — that's the only thing lost here. Assuming it even worked.)

We could certainly restore it, with a better implementation. But preventing full-on crashes (on some non-Ubuntu systems) during export trumps the convenience of a progress bar.

PR #3938, opened 2020-12-26
Create new dummy webview backend, fix unit-test launching

ferdnyc:
This PR changes the application startup substantially to make it more resilient. Specifically, it:

1. Creates a new `webview_backend.dummy`, containing a `DummyWebView` class derived from `QWidget` and providing a stub `run_js` method. This webview can be used in place of the existing WebKit and WebEngine options in situations where the webview is not expected to display anything.
2. Replaces the old `mode="unittest"` argument to `OpenShotApp` and `MainWindow` with an info variable, `info.LAUNCH_MODE`. The default value is `"gui"`, and many more things in the startup are made dependent on the value of `info.LAUNCH_MODE`.
3. Removes the `QApplication` configuration previously being done at the top of `classes.app`, and instead does it at the top of `launch.py`. This ensures that the necessary settings/imports will be completed _**before**_ instantiating the `QApplication`-derived `OpenShotApp` object.
4. Updates `query_tests.py` to set `info.LAUNCH_MODE="unittest"` and `info.WEB_BACKEND="dummy"`, as well as `info.LOG_LEVEL_CONSOLE="WARNING"`, so that startup will be constrained and logging output will be kept to a minimum during the unit test runs. This prevents crashing due to unnecessarily loaded GUI elements (especially the QML-based QWebEngineView) and keeps the unit test runs streamlined for CI use.
5. Enhances `query_tests.py` to process Qt standard command line arguments, and adds `-platform minimal` to the arguments when the unit tests are run in the Github Actions CI. This is necessary when executing the code in an environment without an X server, and is a lighter alternative to the `xvfb` virtual-server environment that was previously used.

This PR also incorporates the change from #3937, eliminating the attempts to communicate with Ubuntu Unity.

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -39,27 +39,6 @@ from PyQt5.QtCore import Qt, pyqtSlot
 from PyQt5.QtGui import QPalette, QColor, QFontDatabase, QFont
 from PyQt5.QtWidgets import QApplication, QStyleFactory, QMessageBox
 
-try:
-    # This apparently has to be done before loading QtQuick
-    # (via QtWebEgine) AND before creating the QApplication instance
-    QApplication.setAttribute(Qt.AA_ShareOpenGLContexts)
-    from OpenGL import GL  # noqa
-except (ImportError, AttributeError):
-    pass
-
-try:
-    # QtWebEngineWidgets must be loaded prior to creating a QApplication
-    # But on systems with only WebKit, this will fail (and we ignore the failure)
-    from PyQt5.QtWebEngineWidgets import QWebEngineView
-except ImportError:
-    pass
-
-try:
-    # Enable High-DPI resolutions
-    QApplication.setAttribute(Qt.AA_EnableHighDpiScaling)
-except AttributeError:
-    pass  # Quietly fail for older Qt5 versions
-
 
 def get_app():
     """ Returns the current QApplication instance of OpenShot """
@@ -67,12 +46,10 @@ def get_app():
 
 
 class OpenShotApp(QApplication):
-    """ This class is the primary QApplication for OpenShot.
-            mode=None (normal), mode=unittest (testing)"""
+    """ This class is the primary QApplication for OpenShot."""
 
-    def __init__(self, *args, mode=None):
-        QApplication.__init__(self, *args)
-
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
         try:
             # Import modules
             from classes import info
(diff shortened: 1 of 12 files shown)


src/classes/app.py
```diff
@@ -203,11 +179,77 @@ def __init__(self, *args, mode=None):
...
+
+        # Load ui theme if not set by OS
+        ui_util.load_theme(self.settings)
+
         # Set Font for any theme
         if self.settings.get("theme") != "No Theme":
             # Load embedded font
             try:
-                log.info("Setting font to %s" % os.path.join(info.IMAGES_PATH, "fonts", "Ubuntu-R.ttf"))
+                log.info("Setting font to %s", os.path.join(info.IMAGES_PATH, "fonts", "Ubuntu-R.ttf"))
```
jonoomph, line 252: Does this still need the %s placeholder?

src/windows/views/webview_backend/dummy.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def run_js(self, code, callback=None):
+        log.debug("run_js request received: %s", code)
+        if callback:
+            callback("{}")
+        else:
+            return None
```
jonoomph, line 47: nitpick, empty line at end of file 😉

jonoomph:
Thanks @ferdnyc ! This looks really nice! I had 1 question, but otherwise, I say merge away once you are ready!

jonoomph:
I'm trying to merge in things that look very critical / and not too scary. I want to create a release branch soon. 🤞

src/classes/app.py
ferdnyc: It does. logging's formatter processes template strings with `%` escapes, the remaining arguments are the inputs to the formatting.

There's one key difference between passing the format string and inputs as arguments, and doing the string interpolation up front: logging calls benefit from lazy evaluation. If logging is disabled (or the `logging.INFO` level isn't configured for output), not only will the final, formatted string not be generated, but the `os.path.join()` won't even be executed, since its output isn't needed.

`"Setting font to %s" % os.path.join(info.IMAGES_PATH, "fonts", "Ubuntu-R.ttf")` always has to do the `os.path.join()` and generate the string, in order to pass it to the `log.info()` call — whether or not it will ultimately be logged.

ferdnyc:
I _believe_ I have a better way to do this (see [ferdnyc/app-launch](https://github.com/ferdnyc/openshot-qt/tree/app-launch), so I'm going to close this PR.

PR #3954, opened 2021-01-07
STY: Use enumerate, avoid Yoda conditions

MartinThoma:
Found by flake8-simplify


jonoomph:
Thanks for the PR! This looks good to me, and def simplifies things a bit. 👍  Thanks!

PR #3995, opened 2021-01-30
Reducing size of AppImage (removing unused and duplicate files during freeze)

jonoomph:
Linux freeze.py script was creating a bloated frozen folder, with about 200+ MB of extra unused or duplicate files. This PR removes those files, and makes things lighter.


jonoomph:
Final AppImage went from 300MB to 200MB. Not too bad! I still think we can shave off more though.

PR #4007, opened 2021-02-11
keyboard shortcut "R" to toogle the razor tool

Spyna:
It would be such a handy feature to have a keyboard shortcut to toggle the razor tool. 
I choose the R keyboard shortcut but feel free to change it. 
I'd like not to add any other keyboard combination, such as "CTRL+R" or "SHIFT+R"


ferdnyc:
@Spyna 

The first thing I thought when reading this PR was, "Why _don't_ we have a shortcut for that!?" We definitely should, thanks for taking the time to submit this!

I'm fine with <kbd>R</kbd> as the default shortcut. I just pushed a commit to create a new setting variable, which makes the shortcut configurable. That way users can customize it however they like.

PR #4012, opened 2021-02-18
Refactor due to profiling improvements (in libopenshot)

jonoomph:
Related to profiling and performance improvements in libopenshot: https://github.com/OpenShot/libopenshot/pull/638

Improvements to openshot-qt:
 - I've removed some unused settings, related to OpenMP
 - Updated effect icons & cache icons
 - Updated translation templates
 - Fixed a regression causing the Caption dock to appear when upgrading OpenShot


jonoomph:
@ferdnyc Hiya! Do you know why this keeps failing? I haven't really dug into yet, just curious: https://github.com/OpenShot/openshot-qt/pull/4012/checks?check_run_id=1931336145

ferdnyc:
@jonoomph Yeah, that's what I was trying to sort out with #3938. Basically, setting up the QtWebEngine just to run the unit tests is... problematic, as it's not being shut down properly. After struggling with it for a while and getting nowhere, I eventually decided that we _shouldn't_ really **NEED** the Timeline web interface for the unit tests, _at all_, so I was trying to come up with a workable way to use a "dummy" Timeline backend with just enough functionality to let the unit tests execute, and without actually loading any browser components.

IIRC I didn't get all the way there with it, at least not yet. Haven't looked at it in like a month, though.

In _theory_ this could also be solved by making sure the web backend is properly and completely spun up and 
then shut down again before the unit test run terminates. Problem is, it doesn't seem like the WebEngine actually **starts** loading until the interface is shown, which we don't actually do, and then can't be shut down until it's fully loaded. Which just to run the unit tests would take forever. (Like potentially ~20 seconds.)

Oh, right, _and_ the `xvfb` environment doesn't support OpenGL (obviously), which is also problematic for QtWebEngine.

Issue #4047, opened 2021-03-12
OpenShot cannot render animated titles


Pythonic456:
OpenShot does not seem to be able to render any animated title frames.

**Steps to reproduce the behavior:**
1. Go to 'Title > Animated Title'.
2. Click on a title template.
3. See error.

**Expected behavior:**
I expected to be able to render some animated titles with OpenShot.

**System Details:**
- OpenShot Version 2.5.1, libopenshot 0.2.5
- Operating System / Distro: Windows 10 Pro
- Blender Version: 2.92.100.0

A windows opens and says this:
```Blender, the free open source 3D content creation suite is required for this action (http://www.blender.org).

Please check the preferences in OpenShot and be sure the Blender executable is correct.  This setting should be the path of the 'blender' executable on your computer.  Also, please be sure that it is pointing to Blender version 2.80 or greater.

Blender Path:
blender

Error Output:
No frame was found in the output from Blender
```

Help would be appreciated!

Mikrotherion:
Can half confirm. Most animated titles work for me (so blender is there and does work), but for some (Dissolving Text and Exploding Text) I get the same error message.
For those that do work I'm not sure they work correctly. The World Map, for example, is very dark and oceans are black

System Details:

    OpenShot Version 2.5.1, libopenshot 0.2.5-r1
    Operating System / Distro: Gentoo 64bit
    Blender Version: 2.91.2

ferdnyc:
@Pythonic456 

On Windows, you'll need to set the path to `blender.exe` in the Preferences. Go to Edit > Preferences, and on the first tab there's a setting "Blender command (path)". Click the Browse button at the end of that field and select the location of the Blender executable (probably `C:\Program Files\Blender Foundation\Blender 2.92\blender.exe`).

If you're using Blender installed from the Windows Store, those packages use a container format that's not compatible with OpenShot (it can't directly access the `blender.exe` file), so you'll have to download and install a full version from [blender.org](https://blender.org/).

That _should_ solve it, assuming there are no other issues using Blender 2.92 with OpenShot 2.5.1. I don't believe there are, but I haven't tested it personally. My only access to a Windows install is as a virtual machine under Linux, and Blender won't run like that because it requires GPU acceleration that isn't available in the VM.

ferdnyc:
@Mikrotherion 

[...]

PR #4053, opened 2021-03-20
Blender: Add sunlight intensity parameter to earth.xml, enhance UI support

ferdnyc:
This PR grew out of https://github.com/OpenShot/openshot-qt/issues/4047#issuecomment-801169343, where @Mikrotherion noted that the rendering of the World Globe animated title was a bit dark. I agreed, and determined that it was a lighting issue.

Rather than impose a new lighting scheme on everyone, I decided to add a `sun_intensity` parameter that allows the user to adjust the intensity of the light source from the default `1.0`. It can go down to `0.0`, or up to `20.0` which is _extremely_ bright and cartoonish.

Along the way I also enhanced the form builder a bit, so that there are some new features the XML files can take advantage of (and `earth.xml` already does, now):

* `<param>` tags are no longer required to include an empty `description=""` attribute
* The existing support for unknown-type `<param>` tags creating labels is now taken advantage of to format the parameters for `earth.xml` more hierarchically (details below)
* In addition, a new `<param type="bold_label" title="Some text/>` tag can be used to create, well, **bold** labels
* Support for the existing `<param type="spinner">` child tags `<step>` and `<digits>` is now implemented. The defaults for both are the previously-hardcoded `0.01` and `2` (which signifies two digits **_after_ the decimal point**, not two total digits), respectively.
    So, for example, the following will display values to 2 decimal places, and raise or lower in increments of `0.10`:
    ```xml
    <param name="sun_intensity" type="spinner" title="Sunlight intensity">
      <min>0.0</min>
      <max>10.0</max>
      <default>1.0</default>
      <step>0.1</step>
      <digits>2</digits>
    </param>
    ```

With the new label support, I relabeled a lot of parameters in `earth.xml` to show as a hierarchy, and shortened the labels on the values considerably because the wide labels weren't really fitting in the window. So, now, the settings will load like this:

image

I'm not 100% happy with it (I'd like a bit _more_ of a true hierarchy, at the bold-labeled levels, and perhaps even the ability to collapse those sections on their bold headings), but it's a start.

Oh, yes, I also migrated my `best_contrast()` function from the Title Editor into `classes.ui`, and it's now used by the Animated Titles interface for the same purpose: To show text labels on color buttons:

image

The color labels are enabled on all color buttons for all of the titles, that's not specific to `earth.xml`. The `<step>`, `<digits>`, `<param type="bold_label" .../>`, etc. XML features are also enabled everywhere, but `earth.xml` is currently the only file which makes use of them.

#### Sunlight Intensity

We should probably pick a new default that's higher than `1.0`, as well. Here's the current (non-adjustable) lighting level at `1.0` intensity:
Screenshot from 2021-03-24 11-51-16

Raising it to `3.0`:
Screenshot from 2021-03-24 11-51-35

And, maxed out at `20.0`:
Screenshot from 2021-03-24 11-52-06

### DIFF
--- a/src/blender/earth.xml
+++ b/src/blender/earth.xml
@@ -12,130 +12,130 @@
 	</param>
 
 	<!-- DEPART COORDINATES -->
-	<param name="depart_title" type="text" title="Depart Title" description="">
+	<param name="depart_section_label" type="bold_label" title="Departure Location"/>
+	<param name="depart_title" type="text" title="Departure Title" description="">
 		<default>New York</default>
 	</param>
-	<param name="depart_lat_dir" type="dropdown" title="Depart Latitude (Equator)" description="">
+	<param name="depart_lat_label" type="label" title="Departure Coordinates, Latitude"/>
+	<param name="depart_lat_dir" type="dropdown" title="• Direction" description="">
 		<values>
 			<value name="North" num="N"/>
 			<value name="South" num="S"/>
 		</values>
 		<default>N</default>
 	</param>
-	<param name="depart_lat_deg" type="spinner" title="Depart Latitude (degrees)" description="">
+	<param name="depart_lat_deg" type="spinner" title="• Degrees" description="">
 		<min>0</min>
 		<max>90</max>
 		<default>40</default>
 	</param>
-	<param name="depart_lat_min" type="spinner" title="Depart Latitude (minutes)" description="">
+	<param name="depart_lat_min" type="spinner" title="• Minutes" description="">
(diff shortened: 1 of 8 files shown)


ferdnyc:
Oh, yes, on the coordinate labels: I'm afraid the bullets are _included_ in the label strings right now (part of what I was saying about not being 100% happy with it), so unfortunately translators will have to translate strings like `• Minutes` and `• Degrees`. (And they'll need to include the bullets in their translated strings, for them to show up in the interface that way.)

PR #4058, opened 2021-03-23
Effect parenting

BrennoCaldato:
This PR adds in functionalities for using the new OpenCV-related effects. The main functionalities are:
- Added transform handlers for the Tracker and Object Detection effects. The user can now modify and correct the tracked objects.
- Fixed Tracker initial bounding box selection window
- Allow the user to animate Clips or Images with the tracked objects. The clip will be transformed to follow the tracked object
- Allow parenting clips. The transformations applied to a clip can be replicated to other clips.
- Effects can be parented and replicated to other clips.
- Improved usability of Tracker effect

Heavily related to https://github.com/OpenShot/libopenshot/pull/649

**Tracker and ObjectDetection transform handlers:**
image

**Clip attached to a tracked object**
image

**Effect parenting**
image

**Object Detection with transform handlers**
image


jonoomph:
@BrennoCaldato A few UI suggestions:
 - Let's rename "Parent Effect ID" to "Parent"
 - Let's rename "Parent ID" to "Parent"
 - Remove the single "Clips" menu level (since it will slow down users probably)

src/classes/effect_init.py
```diff
@@ -25,7 +25,8 @@
  You should have received a copy of the GNU General Public License
  along with OpenShot Library.  If not, see <http://www.gnu.org/licenses/>.
  """
-
+from classes.info import USER_PATH
```
jonoomph, line 28: You might add a new YOLO_PATH to info.py, so it gets auto created on launch... and thus, the user will already have an empty ~/.openshot_qt/yolo/ folder existing.

src/classes/project_data.py
```diff
@@ -407,6 +407,58 @@ def rescale_keyframes(self, scale_factor):
         scaled = scaler(copy.deepcopy(self._data))
         return scaled
 
+        # Create copy of active project data
```
jonoomph, line 410: Is this block of code from a merge? perhaps a regression? I'm thinking this is not supposed to be here. :smile:

src/windows/process_effect.py
```diff
@@ -275,13 +281,10 @@ def rect_select_clicked(self, widget, param):
...
                 # If data found, add to context
                 if topLeft and bottomRight:
-                    self.context[param["setting"]].update({"x": topLeft.x(), "y": topLeft.y(),
-                                                           "width": bottomRight.x() - topLeft.x(),
-                                                           "height": bottomRight.y() - topLeft.y(),
-                                                           "scaled_x": topLeft.x() / viewPortSize.width(), "scaled_y": topLeft.y() / viewPortSize.height(),
-                                                           "scaled_width": (bottomRight.x() - topLeft.x()) / viewPortSize.width(),
-                                                           "scaled_height": (bottomRight.y() - topLeft.y()) / viewPortSize.height(),
-                                                           "first-frame": win.current_frame
+                    self.context[param["setting"]].update({"normalized_x": x1, "normalized_y": y1,
```
jonoomph, line 284: :+1:

src/windows/process_effect.py
```diff
@@ -344,11 +347,15 @@ def accept(self):
             if(self.cancel_clip_processing):
                 processing.CancelProcessing()
 
+        fps = get_app().project.get("fps")
```
jonoomph, line 350: We have some strange blank lines / spacing issues here

src/windows/region.py
```diff
@@ -119,8 +119,13 @@ def __init__(self, file=None, clip=None):
...
         self.verticalLayout.insertWidget(0, self.videoPreview)
 
+        # Set aspect ratio to match source content
+        aspect_ratio = openshot.Fraction(self.width, self.height)
+        aspect_ratio.Reduce()
+        self.videoPreview.aspect_ratio = aspect_ratio
+
         # Set max size of video preview (for speed)
-        self.viewport_rect = self.videoPreview.centeredViewport(self.videoPreview.width(), self.videoPreview.height())
+        self.viewport_rect = self.videoPreview.centeredViewport(self.width, self.height)
```
jonoomph, line 128: Does this hurt performance? Or is this only used when processing effects, etc..?

src/windows/views/properties_tableview.py
```diff
@@ -577,21 +754,34 @@ def contextMenuEvent(self, event):
                 # Divide into smaller QMenus (since large lists cover the entire screen)
                 # For example: Transitions -> 1 -> sub items
                 SubMenu = None
-                SubMenuRoot = menu.addMenu(_(choice["name"]))
+                if choice["icon"] is not None:
+                    SubMenuRoot = menu.addMenu(choice["icon"], _(choice["name"]))
```
jonoomph, line 758: I don't think we need the _() translation function here. You have already added translated strings to the self.choices list, right?

jonoomph (commented):
LG otherwise :+1:

src/windows/models/properties_model.py
```diff
@@ -267,9 +279,14 @@ def color_update(self, item, new_color, interpolation=-1, interpolation_details=
...
+                # Create reference 
+                clip_data = c.data
+                if object_id:
+                    clip_data = c.data.get('objects').get(object_id)
+
                 # Update clip attribute
-                if property_key in c.data:
-                    log.debug("color update: %s" % c.data)
+                if property_key in clip_data:
+                    log.debug("color update: %s" % clip_data)
```
ferdnyc, line 289: (I know it was in the original code, but...)
```suggestion
                    log.debug("color update: %s", clip_data)
```
It's always better to pass log data as separate arguments, to take advantage of Python's lazy evaluation. If logging (for that level) is disabled, it doesn't even need to build the output string. When passing a single `%`-formatted string, it always has to build the string before calling the log function, even if it does nothing.

src/windows/models/properties_model.py
```diff
@@ -407,17 +427,23 @@ def value_updated(self, item, interpolation=-1, value=None, interpolation_detail
...
+            # Create reference 
+            clip_data = c.data
+            if object_id:
+                clip_data = c.data.get('objects').get(object_id)
+
             # Update clip attribute
-            if property_key in c.data:
-                log.debug("value updated: %s" % c.data)
+            if property_key in clip_data:
+                log.debug("value updated: %s" % clip_data)
```
ferdnyc, line 438:
```suggestion
                log.debug("value updated: %s", clip_data)
```

src/windows/models/properties_model.py
```diff
@@ -542,6 +571,228 @@ def value_updated(self, item, interpolation=-1, value=None, interpolation_detail
             # Clear selection
             self.parent.clearSelection()
 
+    def add_property(self, property, filter, c, item_type, object_id=None):
```
ferdnyc, line 574: Kind of a nitpick, but I'd rather name this something like `set_property` or `update_property`, since it _both_ creates and modifies property values. (`add_property()` would be the bit of the code just inside the `if self.new_item:` condition.)

src/windows/models/properties_model.py
```diff
@@ -190,15 +191,23 @@ def remove_keyframe(self, item):
...
+        if not c:
+            return
+
+        # Create reference 
+        clip_data = c.data
+        if object_id:
+            clip_data = c.data.get('objects').get(object_id)
+    
+        if property_key in clip_data:  # Update clip attribute
+            log.debug("remove keyframe: %s" % clip_data)
```
ferdnyc, line 203:
```suggestion
            log_id = f"{clip_id}/{object_id}" if object_id else clip_id
            log.debug("%s: remove %s keyframe. %s", log_id, property_key, clip_data.get(property_key))
```
With the expanded property contents, continuing to log the _entire_ clip/effect dict will make the debug spam unworkably massive. (It kind of was already, TBH.)

src/windows/models/properties_model.py
```diff
@@ -582,222 +835,14 @@ def update_model(self, filter=""):
...
-                        col.setBackground(QColor(red, green, blue))
-
-                    # Update helper dictionary
-                    row.append(col)
-
-                # Keep track of items in a dictionary (for quick look up)
-                self.items[name] = {"row": row, "property": property}
+                self.add_property(property, filter, c, item_type)
+
+            # Insert objects properties from custom effetcs
```
ferdnyc, line 840: nit: spelling
```suggestion
            # Insert objects properties from custom effects
```

src/windows/models/properties_model.py
ferdnyc: Hmm. These ended up out of order due to later editing, so make that "same objection as below..."

src/windows/models/properties_model.py
```diff
@@ -227,7 +236,9 @@ def remove_keyframe(self, item):
                     keyframe["Points"].remove(point_to_delete)
 
             # Reduce # of clip properties we are saving (performance boost)
-            c.data = {property_key: c.data[property_key]}
+            clip_data = {property_key: clip_data[property_key]}
+            if object_id:
+                clip_data = {'objects': {object_id: clip_data}}
```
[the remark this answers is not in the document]
BrennoCaldato: @ferdnyc For some reason, this new way does not update the nested properties correctly. That's why I kept the non-optimized way. I just retested and the problem persist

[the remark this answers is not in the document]
ferdnyc: @brennocaldato

Hm! That's odd. I'll check out the branch this evening and take a look. 

There's gotta be some solution. Dumping the ENTIRE property list into the history at every step is gonna murder us on both performance and project file size.

PR #4097, opened 2021-04-19
WIP: [added a replace_file button]

JacksonRG:
Test Problem 2
===
## Goal: 
    Add a new 'replace file' item to the right click file menu

1. Checked file.rst
** Verified that menu when you right click on a file is called the 'file menu'
just so I know what to look for.

1. Searched "DetailsView" as it's going to be a sibling of the button I'm adding.

1. Looked up pyqt documentation on rightclick menus
** Found that they're called context menus, and created with 
    `def contextMenuEvent(...`

1. Searched the repo for that function definition, in .py files related to the file menu.

1. Found files_listview.py and files_treeview.py

1. Each of the actions in files_listview and files_treeview appear in main_window.py and main-window.ui
** Realize this is how the functions are associated to names, Icons and Tooltips.

1. Add an action for replace file to the main-window.ui, and added it to the context
menu in listview and treeview. ReplaceFile now shows up in rightclick menu.

1. Implemented replacement (incompletely) by deleting the selected files, and calling the importFiles function.

Remaining Issues
===
1. A way of keeping the clipping the same between a file and it's replacement
2. Making the correct icon display

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -678,6 +678,19 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
             # Restore cursor
             app.restoreOverrideCursor()
 
+    def actionReplace_File_trigger(self):
+        log.debug("actionRemove_from_Project_trigger")
+        #TODO: make hide this option until a file is selected
+        #Using the technique from Remove_from_Project
+
+        f = self.files_model.current_file()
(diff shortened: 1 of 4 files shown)


jonoomph:
Closing this for the moment (this was a test PR)

PR #4105, opened 2021-04-26
AppStream metadata: update content ratings

ferdnyc:
So, I ran through the [OARS ratings tool](https://hughsie.github.io/oars/generate.html) for generating AppStream `<content_rating>` metadata blocks, and this is what it came up with. Two important changes:

1. Turns out we're _not_ supposed to include all of those tags marked `none` for content warnings that _don't_ apply; in fact, counter-intuitively, `none` is [documented as meaning](https://freedesktop.org/software/appstream/docs/chap-Metadata.html) "No rating given", whereas an _empty_ `<content_rating>` block means "the component was checked for age ratings and no age restrictions apply". So, the _absence_ of a rating category inside the `<content_rating>` block is the clearest way to indicate that the rating is not applicable to the software in question.
    
    (Also, this is what the official tool generated, so far be it from me to contradict it. Especially when its version is preferable to what we had before.)

1. Our diagnostic telemetry qualifies us for a "`moderate`" rating in the `social-info` category, though, according to the OARS scale:
    
    image
    
    (I'm basing this on our _intended_ use of the information, though I'm not 100% confident that the info provided by the metrics system **couldn't** be reversed to at least a partial identity for an individual user, if someone really wanted to.)

### DIFF
--- a/xdg/org.openshot.OpenShot.appdata.xml
+++ b/xdg/org.openshot.OpenShot.appdata.xml
@@ -47,26 +47,7 @@
     </screenshot>
   </screenshots>
   <content_rating type="oars-1.0">
-    <content_attribute id="violence-cartoon">none</content_attribute>
-    <content_attribute id="violence-fantasy">none</content_attribute>
-    <content_attribute id="violence-realistic">none</content_attribute>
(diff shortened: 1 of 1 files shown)


ferdnyc:
Hearing no objections, merging.

PR #4107, opened 2021-04-27
Make Color properties show keyframe-management, interpolation menu items

### LINKED ISSUE #3768: Bars Effect: impossible to delete Bar Colour keyframe
Setting a bar colour property inserts a keyframe as you might expect.

The only way I know of to remove a keyframe is to jump to it, and right click on the property settings box, selecting Remove Keyframe.

However, right clicking on the Bar Colour property simply brings up the colour selector.
Result: the keyframe little green tick remains, and the property is still set at that time point.

This may well happen with other effect colour select properties too - I've not checked.

ferdnyc:
This PR, which requires OpenShot/libopenshot#666 (hah!) as a prerequisite, changes the behavior of Color and Font type keyframes to no longer immediately show their selection dialog on right-click.

Instead, a menu will be shown containing the option "Select color" or "Select font" (translated, so this does add new strings), which activates the dialog. In addition, for Color keyframes, the menu will show any appropriate items for inserting/removing points, as well as adjusting interpolation type.

The keyframe value can still be double-clicked, to open the selection dialog immediately.

Fixes #3768

### DIFF
--- a/src/windows/views/properties_tableview.py
+++ b/src/windows/views/properties_tableview.py
@@ -28,10 +28,11 @@
 import os
 import functools
 from operator import itemgetter
+import pprint
 
 from PyQt5.QtCore import Qt, QRectF, QLocale, pyqtSignal, pyqtSlot
 from PyQt5.QtGui import (
-    QCursor, QIcon, QColor, QBrush, QPen, QPalette, QPixmap,
+    QIcon, QColor, QBrush, QPen, QPalette, QPixmap,
     QPainter, QPainterPath, QLinearGradient, QFont, QFontInfo,
 )
 from PyQt5.QtWidgets import (
@@ -421,8 +422,9 @@ class PropertiesTableView(QTableView):
             self.choices = cur_property[1]["choices"]
             property_key = cur_property[0]
             clip_id, item_type = selected_value.data()
-            log.info("Context menu shown for %s (%s) for clip %s on frame %s" % (property_name, property_key, clip_id, frame_number))
-            log.info("Points: %s" % points)
+            log.info(
+                "Context menu shown for %s (%s) for clip %s on frame %s",
+                property_name, property_key, clip_id, frame_number)
 
             # Clear menu if models updated
             if self.menu_reset:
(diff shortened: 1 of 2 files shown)


src/windows/views/properties_tableview.py
```diff
@@ -545,6 +534,16 @@ def contextMenuEvent(self, event):
 
             # Add menu options for keyframes
             menu = QMenu(self)
+
+            if self.property_type == "font":
+                font_action = menu.addAction(_("Select font"))
```
ferdnyc, line 539:
```suggestion
                font_action = menu.addAction(_("Change font"))
```
I noticed we already have this string in the translations.

src/windows/views/properties_tableview.py
```diff
@@ -545,6 +534,16 @@ def contextMenuEvent(self, event):
...
+                font_action.triggered.connect(functools.partial(
+                    self.font_select_callback, cur_property))
+            if self.property_type == "color":
+                color_action = menu.addAction(_("Select color"))
```
ferdnyc, line 543:
```suggestion
                color_action = menu.addAction(_("Select a color"))
```
Same deal, this string is already translated.

src/windows/views/properties_tableview.py
```diff
@@ -28,10 +28,11 @@
 import os
 import functools
 from operator import itemgetter
+import pprint
```
ferdnyc, line 31: Whoops.
```suggestion
```

src/windows/views/properties_tableview.py
```diff
@@ -617,6 +615,33 @@ def Linear_Action_Triggered(self):
...
+        _ = get_app()._tr
+        # Get font from user
+        current_font_name = cur_property[1].get("memo", "sans")
+        current_font = QFont(current_font_name)
+        font, ok = QFontDialog.getFont(current_font, caption=_("Change Font"))
+
+        # Update font
+        if ok and font:
+            fontinfo = QFontInfo(font)
+            self.clip_properties_model.value_updated(self.selected_item, value=fontinfo.family())
```
ferdnyc, line 643: Here's a billion-dollar question: (Hah! As if I'm worth even one billionth of that.)

...Why in ever-loving _HELL_ would I stick these two callback methods _**RIGHT BETWEEN**_ the trigger slots for the Linear action and the Constant action!?!? I can't _think_ of a more stupid spot, yet there they are!

I will... fix my epic stupidity.

ferdnyc:
Well, this was sort of derailed by @jonoomph adding similar code in #4490, the functionality differs but reconciling the two would be more trouble than it's worth.

PR #4113, opened 2021-04-30
New Zoom Slider Widget - Modernizing and simplifying timeline scrolling, panning, and zooming :partying_face: 

jonoomph:
New zoom slider widget replaces the old +/- buttons and simple zoom slider from previous versions of OpenShot. This new widget draws a mini timeline preview, allows the user to select any portion of the timeline, and also pan/scroll around the timeline with great accuracy. The playhead is also previewed, to give a better context of "where" you are on the timeline.

Zoom-Slider

https://user-images.githubusercontent.com/5551883/116649037-1deeed80-a944-11eb-9977-18aba7e1d669.mp4

### DIFF
--- a/src/windows/views/zoom_slider.py
+++ b/src/windows/views/zoom_slider.py
@@ -31,7 +31,7 @@ from PyQt5.QtCore import (
 from PyQt5.QtGui import (
     QPainter, QPixmap, QColor, QPen, QBrush, QCursor, QPainterPath
 )
-from PyQt5.QtWidgets import QSizePolicy, QWidget, QPushButton
+from PyQt5.QtWidgets import QSizePolicy, QWidget
 
 import openshot  # Python module for libopenshot (required video editing module installed separately)
 
--- a/src/windows/views/zoom_slider.py
+++ b/src/windows/views/zoom_slider.py
@@ -56,22 +56,20 @@ class ZoomSlider(QWidget, updates.UpdateInterface):
             layers[layer.data.get('number')] = count
 
         # Wait for timeline object and valid scrollbar positions
-        if get_app().window.timeline and self.scrollbar_position[2] != 0.0:
+        if hasattr(get_app().window, "timeline") and self.scrollbar_position[2] != 0.0:
             # Get max width of timeline
             project_duration = get_app().project.get("duration")
             pixels_per_second = self.width() / project_duration
-            project_pixel_width = max(0, project_duration * pixels_per_second)
 
             # Determine scale factor
-            horizontal_factor = self.width() / project_pixel_width
             vertical_factor = self.height() / len(layers.keys())
 
             for clip in Clip.filter():
                 # Calculate clip geometry (and cache it)
-                clip_x = (clip.data.get('position', 0.0) * pixels_per_second) * horizontal_factor
+                clip_x = (clip.data.get('position', 0.0) * pixels_per_second)
                 clip_y = layers.get(clip.data.get('layer', 0), 0) * vertical_factor
-                clip_width = ((clip.data.get('end', 0.0) - clip.data.get('start',0.0))
-                              * pixels_per_second) * horizontal_factor
+                clip_width = ((clip.data.get('end', 0.0) - clip.data.get('start', 0.0))
+                              * pixels_per_second)
                 clip_rect = QRectF(clip_x, clip_y, clip_width, 1.0 * vertical_factor)
                 if clip.id in get_app().window.selected_clips:
                     # selected clip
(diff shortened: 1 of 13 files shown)


jonoomph:
@ferdnyc Is there a known solution for getting our ubuntu-20.04 GitHub builds passing? Or should we disable them, or disable unit tests on them? I'm open to suggestions, lol. It's psychologically damaging me to keep seeing all these failures, haha. :hear_no_evil:

ferdnyc:
@jonoomph I'm HOPEFUL the solution is my `app-launch` branch, which I still haven't had a chance to test on Windows or open a PR for, yet. (I was working on it last night and have some unpushed changes, even.)

If it works out, that'll solve the issue by never even starting the WebEngine process that's crashing when the unit tests complete. (It never even gets a chance to start up properly, never mind tear itself down, in the time the tests run.)

Disabling the tests until then is probably the only other option, but TBH it's not a bad one. They don't test anything that any of the current PRs -- or current development efforts -- even touch, anyway,. So their usefulness is really kind of minimal, at least in terms of CI needs.

ferdnyc:
Please don't remove `main()` from `launch.py`, tho. That's the opposite of a solution, and just a bad idea generally. This is a standard, and GOOD, Python pattern:

```python
def main():
    # code


if __name__ == "__main__":
    main()
```

If my `app-launch` branch ends up merged it'll completely overhaul both `launch.py` and `classes/app.py`.

ferdnyc:
@jonoomph 

My `app-launch` branch is now submitted as #4114. Though I should probably push it to a project branch and re-submit, for Gitlab builder testing.

ferdnyc:
Done, make that #4115.

jonoomph:
> Please don't remove `main()` from `launch.py`, tho

This was an experiment, to test scoping of the QApplication. Reverting now. :+1:

src/windows/views/zoom_slider.py
```diff
@@ -0,0 +1,508 @@
...
+
+        # Get layer lookup
+        layers = Track.filter()
+
+        # Wait for timeline object and valid scrollbar positions
+        if get_app().window.timeline and self.scrollbar_position[2] != 0.0:
+            # Get max width of timeline
+            project_duration = get_app().project.get("duration")
+            pixels_per_second = event.rect().width() / project_duration
+            project_pixel_width = max(0, project_duration * pixels_per_second)
```
ferdnyc, line 155: You're dividing `event.rect.width()` by `project_duration`, then multiplying that value by `project_duration`? I hate to break it to you, but that's always going to come out as `event.rect().width()`. :laughing:

src/windows/views/zoom_slider.py
```diff
@@ -0,0 +1,508 @@
...
+            scroll_width = (self.scrollbar_position[1] - self.scrollbar_position[0]) * event.rect().width()
+
+            # Get FPS info
+            fps_num = get_app().project.get("fps").get("num", 24)
+            fps_den = get_app().project.get("fps").get("den", 1)
+            fps_float = float(fps_num / fps_den)
+
+            # Determine scale factor
+            horizontal_factor = event.rect().width() / project_pixel_width
```
ferdnyc, line 164: ...Which means *THIS* is always going to come out as `1`. Well, unless your `max()` was hit and `project_pixel_width` became `0`, in which case you've got a `ZeroDivisionError` exception to contend with.

ferdnyc (commented):
I haven't tried it yet, but some comments from reading the code...

src/windows/views/zoom_slider.py
```diff
@@ -0,0 +1,508 @@
...
+    def delayed_resize_callback(self):
+        """Callback for resize event timer (to delay the resize event, and prevent lots of similar resize events)"""
+        # Get max width of timeline
+        project_duration = get_app().project.get("duration")
+        normalized_scroll_width = self.scrollbar_position[1] - self.scrollbar_position[0]
+        scroll_width_seconds = normalized_scroll_width * project_duration
+        tick_pixels = 100
+        if self.scrollbar_position[3] > 0.0:
+            # Calculate the new zoom factor, based on 100 pixels per tick
+            zoom_factor = scroll_width_seconds / (self.scrollbar_position[3] / tick_pixels)
```
ferdnyc, line 369: The calculation **here** isn't based on 100 pixels per tick, it's based on `tick_pixels`. If the value of `tick_pixels` is changed, that comment becomes a lie, so it shouldn't specify a value that could change.

ferdnyc: (I will pay you **#ALL_THE_MONEY** if you can find me someone who'll claim that they'll change the comment when they change `tick_pixels`, and not be lying.)

jonoomph: I like those odds! haha

src/windows/views/zoom_slider.py
jonoomph: :thinking:

src/windows/views/zoom_slider.py
jonoomph: horizontal_factor was variable in my first iteration, and would change as you zoomed in/out, but now it's more based on the fixed size of the widget and duration of the project. I need to tweak this.

ferdnyc:
@jonoomph Eek! Didn't realize you merged this first. It contains a call to `settings.get_settings()` that my merge of #4115 just made invalid. I'll fix that...

ferdnyc:
Hm, I guess it doesn't! Well, cool.

PR #4114, opened 2021-04-30
Overhaul app launching, unit tests

ferdnyc:
This PR, tested on both Linux and Windows, completely revamps the initial creation of the `QApplication` instance and the `MainWindow` class, as OpenShot is being started by `launch.py` (also significantly rewritten), all in the service of one central goal: completely divorce the application setup (which is done in `OpenShotApp's __init__`, same as always) from the interface startup, which now lives in a new `OpenShotApp.gui()` method. Calling that is optional, and the unit tests simply... don't.

When launching the unit tests (from `tests/query_tests.py`), the interface is never even started, and therefore (by implication) neither is the web backend. Which alleviates the problems caused by `QtWebEngine` not shutting down properly because it hasn't even had a chance to start yet before the unit tests are completed.

I've also moved all of that OpenGL setup stuff and other housekeeping out of the top of classes.app and put it in `launch.py` instead, so that it's truly done BEFORE creating the `QApplication` (or, `OpenShotApp`) instance. I've always been wary of the fact that we were effectively doing it "while" the instance was being created, not technically "before".

And having it in `launch.py` means the unit tests can also skip that stuff. It's now possible, in fact, to run the unit tests in a headless environment without needing `xrdb`, just by adding a `-platform minimal` or `-platform offscreen` on the command line for `query_tests.py`. (The platform plugin has to be changed because Qt's default on Linux is xcb, which still expects an X display even if nothing makes use of it.)

There were some other changes involved, like deferring any `QMessageBox` calls that previously occurred during `OpenShotApp.__init__()`. Now they're placed on a queue, and not shown until requested by `launch.py` (or automatically, if there are any outstanding when `OpenShotApp.gui()` is called.

#### Other changes
* I've completely removed `classes.settings.get_settings()`. Settings should now be requested from the app instance, using `get_app().get_settings()`. There's also a convenience function `get_settings()` in `classes.app`, just like the old one in `classes.settings`, but you shouldn't use `classes.app.get_settings()` either. Go through the application instance.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,4 +30,5 @@ jobs:
 
     - name: Test
       shell: bash
-      run: xvfb-run --auto-servernum --server-num=1 --server-args "-screen 0 1920x1080x24" python3 ./src/tests/query_tests.py
+      run: python3 ./src/tests/query_tests.py -platform offscreen
+
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -36,8 +36,7 @@ import json
 
 from PyQt5.QtCore import PYQT_VERSION_STR
 from PyQt5.QtCore import QT_VERSION_STR
-from PyQt5.QtCore import Qt, pyqtSlot
-from PyQt5.QtGui import QPalette, QColor, QFontDatabase, QFont
+from PyQt5.QtCore import pyqtSlot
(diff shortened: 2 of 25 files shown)


ferdnyc:
Replaced by #4115

PR #4119, opened 2021-05-03
SHIFT + Right click bug preventing clip/transition movement

### LINKED ISSUE #4112: Immovable Clip
## Description
I found that while using openshot I would occasionally find that I couldn't move clips on the timeline. Through experimentation I found this way to reliably reproduce this bug. 
## Reproducing the bug
1. Put a file in the timeline
1. Shift+right click anywhere
    - Releasing right click before dismissing the context (right click) menu
1. Try to move it
1. Find that you can only move it up and down

From this I conclude that the shift "keyup" event is not being received if a context window was being shown.

## Expected behavior
- When shift is held, a clip can only be moved vertically. Keeping it's place in time the same.
- As shift isn't being held, the clips should be able to move.

**System Details:**
- OpenShot Version [e.g. 2.4.3]: {{ log_context.openshot_qt_version }}
- Operating System / Distro: [Linux / Pop!_OS 20.04]

Video Example:
https://ipfs.io/ipfs/QmSr8BEwrW8rCPUvQcP9EWYmz5R4mzXZE8wunyw952edLN

JacksonRG:
Closes #4112
# The Bug
- Holding shift while opening a context menu, and releasing before closing it caused clips to be immovable except vertically until shift was pressed again and released.
- Shift keydown events set a variable 'shift_pressed' to true. This prevents dragging clips in the x direction.
# The Search
- Traced the event structure of the code. (Good to get familiar with)
   - Hoped I could make the context menu pass on keyup events.
- Found that the event handler wasn't even getting keydown/keyup events when a context menu was active.
# The Fix
- Check the state of the shift key directly in the event when a clip is moved.
- Removed the shift_pressed variable, as it was only used for for this


jonoomph:
`openshot_qt/timeline/js/controllers.js` this is the wrong file path... oops.

JacksonRG:
> `openshot_qt/timeline/js/controllers.js` this is the wrong file path... oops. Should be `src/timeline...`

BRB. Updating my git-ignore & pushgin a change

jonoomph:
@JacksonRG You need to remove this file from your branch: `openshot_qt/timeline/js/controllers.js`, and modify this file: `src/timeline/js/controllers.js`, and then push your branch again. Otherwise, we don't have your changes to the real controllers.js file. Hope that makes sense.

src/timeline/app.js
```diff
@@ -57,25 +57,6 @@ $(document).ready(function () {
     $(".playhead-line").height(body_object.scope().playhead_height);
   });
 
-  // Bind to keydown event (to detect SHIFT)
```
jonoomph, line 60: Nice job removing all this code!

jonoomph (approved):

jonoomph:
LGTM!

PR #4122, opened 2021-05-06
Fixing a regression caused by renaming .env files

jonoomph:
Fixing a regression caused by renaming .env files with version_info. This caused the `version.json` file to not be created, and thus, all version info to vanish from all builds.

### DIFF
--- a/freeze.py
+++ b/freeze.py
@@ -175,7 +175,7 @@ for share_name in ["install-x64", "install-x86"]:
             git_log_filepath = os.path.join(share_path, git_log_filename)
             if os.path.isfile(git_log_filepath):
                 src_files.append((git_log_filepath, "settings/%s" % git_log_filename))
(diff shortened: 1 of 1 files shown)


freeze.py
```diff
@@ -175,7 +175,7 @@ def find_files(directory, patterns):
             git_log_filepath = os.path.join(share_path, git_log_filename)
             if os.path.isfile(git_log_filepath):
                 src_files.append((git_log_filepath, "settings/%s" % git_log_filename))
-                if os.path.splitext(git_log_filepath)[1] == "":
+                if os.path.splitext(git_log_filepath)[1] == ".env":
                     # No extension, parse version info
```
ferdnyc, line 179: Heh. At the risk of beating a dead horse, that change made this comment a lie. :wink:

PR #4130, opened 2021-05-10
GitHub Action to lint Python code

cclauss:


ferdnyc:
Hmm. That's an awful _lot_ of stuff to run, and seems kind of kitchen-sink-ey — like, we **don't have** a `requirements.txt` ("yet"), so there's not a lot of point in testing `pip -r requirements.txt` ffor our config.

I'm going to approve the action run, because I'm kind of curious about the results, but I think we'll have to prune this down for our use if we don't want this to just be a perennially failing check (which does nobody any good).

Still, I like the idea, and good catch on the missing Exception identifier.

cclauss:
Its a lota stuff but it all runs in just 1 minute.

Issue #4131, opened 2021-05-10
Timeline Scale Issues, Slider Not Matching Up With Playhead


Raziel0315:
I tried to find a report on this bug.  I was unsuccessful.

**Describe the bug:**

When scaling the timeline, aka zooming in, the scale has issues.  Using the scaling bar (?) doesn't immediately get you the size you want.  On top of that, IF you are able to zoom into the timeline at the size you want, the sliding bar begins to move separately from the playhead.
When you're at the beginning of the project, both of these items match up, but as you move further into the project, they start to separate, and when the sliding bar reaches the end of the display, (the one just above the timeline), the playhead gets severed.  I have added images to clarify.

**Steps to reproduce the behavior:**

Easy to reproduce: zoom in as much as possible, then move the slider all the way to the end.

**Expected behavior:**

The slider should match up with the playhead, I believe, and the playhead shouldn't get severed.

**System Details:**
- OpenShot Version:  2.5.1, May 9th, 2021 daily build
- Operating System:  Windows 10, 64 Bit

**Screenshots:** (Optional)

In the top part of the image, you can see where the slider doesn't match up with the playhead.
In the bottom half, you can see where the playhead is severed because the slider went off of the end of the timeline display or scale or whatever it is.
In both cases, you can see that the slider is well past where the clips ends.

![OpenShot Bug Report](https://user-images.githubusercontent.com/47090619/117676842-ea9a3300-b17b-11eb-9082-68525c4c5ee8.png)

JacksonRG:
I've done a rework of the ruler above the timeline.

From my testing the 'severed playhead' only occurs at the highest degree of zoom.

Looks like we're getting close to solving this.

PR #4153, opened 2021-05-21
WIP Ruler Performance Refactor

JacksonRG:
issue #4131
Drew attention to the fact that in long projects, the playhead symbol on the ruler, and the playhead line can become separated.

So far, in this branch, I've changed the ruler to only redraw the tick marks that are on the screen.

We could go ahead and merge that change, but the 'severed playhead' bug persists.

Severed

I've only been able to reproduce this at the highest level of zoom. And at that zoom level, the zoom slider seems to have a mathematical glitch. On dragging it, I find that it moves to about half it's previous time.

https://user-images.githubusercontent.com/42394129/119235015-11e6ed80-baf6-11eb-8de9-ee93f3a60e2d.mp4

This may need a separate issue. But I'll start taking a look at the zoom slider.


src/timeline/js/directives/ruler.js
```diff
@@ -144,6 +143,8 @@ App.directive("tlRuler", function ($timeout) {
           });
         });
 
+        drawTicks();
```
jonoomph, line 146: I don't have any issues with refactoring the code into a method (i.e. drawTicks()) as you have done... and invoking it "on scroll" and "on scope.watch". I think this is a relatively simple fix of a very big performance issue (especially when zoomed into the timeline, causing a HUGE timeline pixel width).

src/timeline/js/directives/ruler.js
```diff
@@ -156,60 +157,68 @@ App.directive("tlRuler", function ($timeout) {
...
 
-      //watch the scale value so it will be able to draw the ruler after changes,
-      //otherwise the canvas is just reset to blank
-      scope.$watch("project.scale + markers.length + project.duration", function (val) {
-        if (val) {
-
-          $timeout(function () {
+      function drawTicks() {
+      // drawTicks : function () {
+        $timeout(function () {
```
jonoomph, line 162: I wonder if the removal of $timeout will cause any issue. I "think" that was originally added to prevent the UI from freezing while drawing thousands upon thousands of tick marks on the Ruler. However, since we are only drawing the visible ruler now... I guess it makes sense that this is not needed anymore, lol.

jonoomph (approved):
LGTM! I haven't tested it personally, but assuming it still works with the same behavior as before, and all the ticks are drawn at the correct offsets/positions, this is great! Nice work!

src/timeline/js/directives/ruler.js
JacksonRG: Trouble is, "on scroll" Only recieves scrolling when the mouse is over the ruler. And the tlBody or tlScrollableTracks doesn't have access to drawTicks where it is. Not sure what scope to put this function in so it can be called elsewhere in the webview

src/timeline/js/directives/ruler.js
jonoomph: @JacksonRG 
We could always init a new scope variable in controllers.js:
`$scope.scrollLeft = 0;`

And extend the scope.watch() with a new custom scope variable:
`"project.scale + markers.length + project.duration + scrollLeft"`

And update it on scroll:
```
element.on("scroll", function () {
        scope.$apply(function () {
          scope.scrollLeft = element.scrollLeft();
        });
```

jonoomph: This would also remove the need to refactor drawTicks() into it's own method. I'm just not sure if this would have any serious performance implications. I'm guessing if the only part of Angular that is watching/bound to $scope.scrollLeft is this one watch... and if that watch is super fast at drawing ticks... it might be a non-issue.

jonoomph: That seems to work nicely in my tests

jonoomph: Also, one final suggestion... perhaps make the "right side" calculation go off the screen, so we don't have a gap on the right side of the ruler...
Screenshot from 2021-05-21 13-26-23

JacksonRG: Sounds good! Thanks for the tips. I'll push an update to this later tonight

JacksonRG: Fixed. Rounded up to go off the screen. And you were right. scope.scrollLeft was a simple solution to let ruler watch for changes.

src/timeline/js/directives/ruler.js
```diff
@@ -57,7 +56,7 @@ App.directive("tlScrollableTracks", function () {
         // Send scrollbar position to Qt
         if (scope.Qt) {
            // Calculate scrollbar positions (left and right edge of scrollbar)
-           var timeline_length = Math.min(32767, scope.getTimelineWidth(0));
```
jonoomph, line 60: :+1:

src/timeline/js/directives/ruler.js
```diff
@@ -95,6 +94,11 @@ App.directive("tlScrollableTracks", function () {
         element.removeClass("drag_cursor");
       });
 
+      element.on("scroll", function(){
```
jonoomph, line 97: We do already have an .on("scroll" listener above that we could use, correct?

src/timeline/js/directives/ruler.js
```diff
@@ -158,58 +160,61 @@ App.directive("tlRuler", function ($timeout) {
...
+            var rulerWidthPixels = document.getElementById("scrolling_ruler").clientWidth;
+            var num_ticks = ~~(rulerWidthPixels / each_tick);
+            var startingTick = ~~(leftScreenPixels / each_tick);
+            for (var x = 0; x < num_ticks + 2; x++) {
               ctx.beginPath();
 
               //if it's even, make the line longer
               var line_top = 0;
-              if (x % 2 === 0) {
+              if ((x + startingTick)% 2 === 0) {
```
jonoomph, line 192: Nit pick: space before `%` operator

src/timeline/js/directives/ruler.js
JacksonRG: You're right. I'll move that

src/timeline/js/directives/ruler.js
JacksonRG: Readability is good! I'll take nitpicks.

ferdnyc:
@JacksonRG @jonoomph 

> the zoom slider seems to have a mathematical glitch. On dragging it, I find that it moves to about half it's previous time.

PART of that may also be due to the feedback loop between the ZoomSlider and the timeline.

I don't know if you (or anyone) has tested the ZoomSlider when running with the WebEngine backend (macOS or Linux), but  the backend's async nature really amplifies the effects of that feedback loop — performance of the ZoomSlider is, TBH, kind of horrible. Dragging it is an exercise in frustration, as it keeps pulling back from whatever direction you drag it. The issue isn't math, but rather the circular update signaling and the way it's all connected together. And a lot of THAT is down to the backwards way the OpenShot code uses signals.

(Just look at all the hoops code like `windows/cutting.py` has to jump through in order to reuse the `windows/preview_thread.py` code, with all of the useless signals that have to be duplicated because the `PreviewParent` expects to reference them either by connecting the `PreviewWorker` to them or calling them in response to `PreviewWorker` changes.
    
Whereas, if `MainWindow` handled connecting its own signals to `PreviewWorker` slots in order to trigger functionality, and the `PreviewWorker` _emitted_ its own signals to indicate when changes occurred, none of that would be necessary because any code using the `PreviewWorker` would only have to connect to the signals it was interested in, and only connect its own signals to the slots it needed to control.)

OpenShot uses signals as a way of saying to the rest of the interface, "Do something". Which is why most signals live on the `MainWindow` object, so that they're central and can be disseminated.

That's actually wrong, though.

In Qt's own code, you'll notice, objects emit signals as a way of saying, "something happened" — mostly user input, though sometimes as a result of program changes. They're always reactive, though. If that "something" that happened is supposed to cause something else to occur in the _rest_ of the interface, that will be indicated by connecting the signal to a corresponding slot. A **SLOT** is what says to the interface, "Do something". A signal is a notification that something has _already_ occurred, which might in turn trigger other code automatically.

There are two big advantages to using only local signals:
1. Better encapsulation. If objects send local signals about local events, they should only have to rely on local data to do it. That means they don't have to query other objects for information — any information an object needs from other objects should be delivered to _its_ slots in the signals it **receives**. Each object is now a complete, working component that can be re-used elsewhere, or NOT used at all, because other objects make no assumptions about its availability, and it makes no assumptions about other objects' availability.

2. Better isolation. When objects emit OTHER objects' signals, those other objects have to exist, and they have to survive until that signal is delivered — if an owning object is destroyed while another object is trying to emit one of its signals, the code will crash. Whereas, if an object is emitting its own signals, it will obviously live to deliver them to the receiving slots: doesn't matter if there are many, one, or zero of them. If any of the other objects haven't come up yet, there will be no connections and the signal will be silently ignored. When other objects are DESTROYED, Qt automatically disconnects its slots from all of the signals it was connected to. (That guarantee is one of the reasons to always use `.deleteLater()` when manually destroying Qt objects.) Each object is now a self-contained, independent part of the codebase that can be created or destroyed as needed, and in any order, without destabilizing the codebase.

Fun fact: I actually happened across an old Qt Forum posting (this was in the Qt 4 days) where it was mentioned that Qt were considering making it _unsupported_ for an object to emit another object's signals. In the end they didn't go that route, and I assume there are occasionally valid reasons to do things that way. But that doesn't mean it's a good idea, or something that should be done cavalierly. Emitting foreign signals should be a rare event, but currently many of OpenShot's objects PRIMARILY emit other objects' signals. And that causes problems.

What does all of my rambling have to do with the ZoomSlider specifically?

The problem:
1. Currently, whenever the ZoomSlider is dragged, it emits `MainWindow.TimelineScroll` with a _relative_ position change.
2. The webview's `update_scroll` is connected to that signal, and runs `.setScroll()` in the JS.
3. Due to the async nature of WebEngine, _at some point in the future_ that `.setScroll()` runs and updates the positioning of `#scrolling_tracks` by querying its `scrollLeft()` and applying the relative movement.
4. That movement then (again, at some point in the future) triggers the `onscroll` handler, which moves the other three main `<div>`s to the "current" `scrollLeft()` of `#scrolling_tracks`. It also then calls `ScrollbarChanged()` in the webview to report the new positioning.
5. `ScrollbarChanged()` emits <code>MainWindow.TimelineScroll<b>ed</b></code> with a set of new scroll parameters, based indirectly on the position data from that old event way back in step 1.
6. The ZoomSlider's `update_scrollbars()` is connected to that signal, so it re-applies the details from that old event to its own positioning.
7. Meanwhile, if the user has been dragging the scroll rectangle, part of that drag is now undone as the ZoomSlider's positioning is rolled back to old values.

Because WebKit is synchronous, all of those events happen immediately and the feedback loop is largely invisible. But it's still there, and may be causing the detachment issues you're seeing. With WebEngine, the async updates make the feedback loop glaringly obvious, and it severely impacts the ZoomSlider's drag handling. Async updates may also cause repositionings of `#scrolling_tracks` to be handled late or out of order by the `onscroll` handler, which may be the cause of the desynchronization of the other `<div>`s.

How _should_ it work, instead?
1. When the ZoomSlider is being dragged, it continually emits `ZoomSlider.PositionChanged` signals containing the updated scroll positioning, an absolute (_not_ relative) offset to position the timeline at.
2. The webview's `update_scroll` slot has been connected to `ZoomSlider.PositionChanged` (by `MainWindow`, the only code that can make assumptions about both objects' existence), so it then triggers and issues `.setScroll()`,  which calls `$("#scrolling_tracks").scrollLeft(new_position);`. The previous `.scrollLeft()` value is irrelevant, because we're moving it to an exact location.
3. The `onscroll` handler then repositions the other three `<div>`s accordingly. (Though I halfway think `.setScroll()` should just directly reposition the ruler, as well.)
3. Because the scroll position was set _programmatically_, **no signal is issued** about the repositioning from the Qt code. (That may be decided in the onscroll handler, or once it calls `ScrollbarChanged()` in the webview. Either is workable.)
4. When the user actually scrolls the timeline themselves, the `onscroll` handler calls `ScrollbarChanged()` and causes the webview to emit `TimelineWebView.ScrollPositionChanged` with new, absolute parameters
5. ZoomSlider's `update_scrollbars()` has been connected to that signal (by `MainWindow`, the only code that can make assumptions about both objects' existence), so it updates its positioning. (Optionally, while the user is actively dragging, it may choose to ignore that signal anyway.)

No feedback loops, and no issues caused by the async nature of WebEngine updates. The timeline will respond to ZoomSlider drag motions at its leisure, but any delays in updating won't impact ZoomSlider's own drag responsiveness.

Simply making ZoomSlider ignore `MainWindow.TimelineScrolled` while dragging _may_ resolve the immediate issue, since it would disconnect the feedback loop. But that'd really be just a band-aid.

ferdnyc:
> Simply making ZoomSlider ignore `MainWindow.TimelineScrolled` while dragging _may_ resolve the immediate issue, since it would disconnect the feedback loop. But that'd really be just a band-aid.

Making the scroll updates issued by `ZoomSlider` use _absolute_ positioning rather than relative motions, OTOH, I feel is absolutely vital. The relative changes are almost certainly what's causing the playhead detachment you're seeing.

Right now each repositioning of the webview depends on its current position at that moment, so delayed or out-of-order updates will throw everything out of whack. Especially since the other `<div>`s' positioning is then dependent on sampling the position of `#scrolling_tracks` at _that_ moment, even as it may be in motion. 

Simplifying the onscroll handler may also help — like, it should probably just set a scope variable with the updated scroll parameters, instead of calling `timeline.ScrollbarChanged` itself. A `$scope.watch()` on the output variable can be responsible for tossing the new parameters back over the wall to Qt.

JacksonRG:
@ferdnyc Insightful as always. I'll take a look at the signaling.

src/timeline/js/directives/ruler.js
```diff
@@ -57,14 +56,18 @@ App.directive("tlScrollableTracks", function () {
...
            var right_scrollbar_edge = (scroll_left_pixels + element.width()) / timeline_length;
 
            // Send normalized scrollbar positions to Qt
            timeline.ScrollbarChanged([left_scrollbar_edge, right_scrollbar_edge, timeline_length, element.width()]);
         }
 
+        scope.$apply( () => {
+          scope.scrollLeft = element[0].scrollLeft;
+        })
+
```
ferdnyc, line 70: This is probably a bad idea. Angular can always  just query the value directly from `$("#scrolling_tracks")` when it needs it. If a value that's meant to be ephemeral (like an element's current scroll position) gets copied and stored, the copy is almost instantly stale. Then you get issues like ZoomScroll's  janky dragging.

There are a few cases where you specifically want the value "at the time of the event". (Pointer coordinates are stored for mouse events because what matters is where the user clicked, _not_ where the pointer is now.) But for things like scroll position, keeping a copy of an old value is just inviting update lag.

src/timeline/js/directives/ruler.js
```diff
@@ -158,58 +158,61 @@ App.directive("tlRuler", function ($timeout) {
...
-            var num_ticks = pixel_length / 50;
 
             ctx.lineWidth = 1;
             ctx.strokeStyle = "#c8c8c8";
             ctx.lineCap = "round";
 
             //loop em and draw em
-            for (var x = 0; x < num_ticks + 1; x++) {
+            var leftScreenPixels = document.getElementById("scrolling_ruler").scrollLeft;
+            var rulerWidthPixels = document.getElementById("scrolling_ruler").clientWidth;
```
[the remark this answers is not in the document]
JacksonRG: I think I left off the `#` when I tried that. Should be more concise with the dollarsign

JacksonRG:
Closing, as I started fresh on ruler-refactor branch. PR coming soon

PR #4155, opened 2021-05-21
Removed broken index (Copy/Paste keyframe issue)

### LINKED ISSUE #4138: Copy/Paste ALL Keyframes Does Not Work
I tried to find a report on this bug. I was unsuccessful.

Describe the bug:

The copy/paste function doesn't work correctly, specifically, the copy keyframes, which I use A LOT.
It seems like when copying keyframes you can choose one of the options, e.g., scale, and it pastes just fine.
However, the ALL selection doesn't seem to work.  If I try copy/paste ALL keyframes, nothing happens.
Any single copying/pasting of keyframes will work, just not ALL.

Easy to reproduce: copy a clip, choose Keyframes ALL, then paste to see that keyframes were not pasted.

Expected behavior:

ALL of the copied clips keyframes should paste onto the new clip, as before.

System Details:

OpenShot Version: 2.5.1, May 9th, 2021 daily build
Operating System: Windows 10, 64 Bit

JacksonRG:
Fixes #4138

Copy/Paste all keyframes was retrieving data from a clip property that didn't exist. This was crashing the copy, and only allowing keyframes for properties before it to be copied.

A rework of this is coming, but here's a quick fix.

### DIFF
--- a/src/windows/views/webview.py
+++ b/src/windows/views/webview.py
@@ -1456,10 +1456,6 @@ class TimelineWebView(updates.UpdateInterface, WebViewClass):
                 self.copy_clipboard[clip_id] = clip.data
             elif action == MENU_COPY_KEYFRAMES_ALL:
                 self.copy_clipboard[clip_id]['alpha'] = clip.data['alpha']
-                self.copy_clipboard[clip_id]['crop_height'] = clip.data['crop_height']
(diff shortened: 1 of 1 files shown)


jonoomph:
Nice, I forgot we removed the crop_ attributes from a Clip. LGTM!

Issue #4158, opened 2021-05-23
Truncation of generated assets folder filename clobbers versioning


Geep1:
**Describe the bug & Steps to reproduce the behavior:**
When a project is saved, it appears as _filename1_.osp and as folder _filename1_ _assets.
But, for the folder only, the filename1 is truncated to 32 characters before appending "_assets".

This causes silent overwrite of the assets folder when _filename1_ is not unique in the first 32 characters across projects. The use case where this is most likely to be encountered is in versioning, i.e.:

Save as: "My video project with a long name version 1"
Save as: "My video project with a long name version 2"

Both will generate an assets folder like:
"My video project with a long nam _assets"

The second save will cause assets folder overwrite, losing version 1 edit.

**Expected behavior:**
Possible solutions:
1) Remove truncation of _filename1_ when generating _assets name. This is preferred solution.
2) Truncate .osp name as well. Check for existing file, and if found provide Overwrite. Are You Sure? dialog THAT SHOWS THE TRUNCATED NAME.

**System Details:**
- OpenShot Version 2.5.1
- Windows 10

JacksonRG:
Thanks for reporting this! I'll get testing.

JacksonRG:
Thanks for reporting this! We've raised the truncation length to the file name limit for Linux and Windows. Tomorrow's [daily build](https://www.openshot.org/download/) should have these changes!

PR #4159, opened 2021-05-24
Prevent Assets Folder Name Collision

JacksonRG:
Issue #4158 pointed out that if project names were longer than 30 characters, they were truncated to 30 characters, and `_assets` was appended.

# Solution
Aside from MacOS, 255 characters the typical limit for file or folder names. So to avoid, neighboring projects running into each other as much as possible, we now truncate only if the addition of '_assets' would push the folder name over 255.

# Reverse Backwards Compatability.
As Frank pointed out, moving the folder to the new naming convention would cause trouble for anyone with long project files moving back to the stable build.

To solve this, we search for and **copy** the truncated folder if if a full length one doesn't exist. I believe this is the most seamless user experience, as renaming the folder, would mean the first project opened would take the folder that had been shared.

## Testing
1. Created two long named projects on develop
1. Created titles in both with different titles
2. Switched branches
3. Opened each project, finding both had their own assets

### DIFF
--- a/src/classes/assets.py
+++ b/src/classes/assets.py
@@ -40,23 +40,25 @@ def get_assets_path(file_path=None, create_paths=True):
         # Generate asset folder name filename + "_assets"
         file_path = file_path
         asset_filename = os.path.splitext(os.path.basename(file_path))[0]
-        asset_folder_name = asset_filename + "_assets"
+        asset_folder_name = asset_filename[:248] + "_assets" #Windows max name size is 255. 248 = 255 - len("_assets")
         asset_path = os.path.join(os.path.dirname(file_path), asset_folder_name)
 
         # Previous Assets File Name Convention.
         asset_folder_name_30_char = asset_filename[:30] + "_assets"
         asset_path_30_char = os.path.join(os.path.dirname(file_path), asset_folder_name_30_char)
 
+
         # Create asset folder, if necessary
         if create_paths:
 
-            if os.path.exists(asset_path_30_char):
+            if not os.path.exists(asset_path):
+                if os.path.exists(asset_folder_name_30_char):
                 #update assets folder, if it follows the previous naming convention
-                os.rename(asset_path_30_char, asset_path)
-                log.info("Updating 30 character path to full length {}".format(asset_path))
-            elif not os.path.exists(asset_path):
-                os.mkdir(asset_path)
-                log.info("Asset dir created as {}".format(asset_path))
+                    os.rename(asset_path_30_char, asset_path)
+                    log.info("Updating 30 character path to full length {}".format(asset_path))
+                else:
+                    os.mkdir(asset_path)
+                    log.info("Asset dir created as {}".format(asset_path))
             else:
                 log.info("Using existing asset folder {}".format(asset_path))
 
--- a/src/classes/assets.py
+++ b/src/classes/assets.py
@@ -51,7 +51,6 @@ def get_assets_path(file_path=None, create_paths=True):
 
         # Create asset folder, if necessary
         if create_paths:
-
             if not os.path.exists(asset_path):
                 if os.path.exists(asset_path_30_char):
                     #copy assets folder, if it follows the previous naming convention
--- a/src/classes/assets.py
+++ b/src/classes/assets.py
@@ -53,10 +53,10 @@ def get_assets_path(file_path=None, create_paths=True):
         if create_paths:
 
             if not os.path.exists(asset_path):
-                if os.path.exists(asset_folder_name_30_char):
+                if os.path.exists(asset_path_30_char):
                     #update assets folder, if it follows the previous naming convention
                     try:
-                        shutil.copytree(asset_folder_name_30_char, asset_folder_name)
+                        shutil.copytree(asset_path_30_char, asset_path)
                         log.info("Copying shortened asset folder. {}".format(asset_path))
                     except:
                         log.error("Could not make a copy of assets folder")
--- a/src/classes/assets.py
+++ b/src/classes/assets.py
@@ -26,6 +26,7 @@
  """
 
 import os
+import shutil
 from classes import info
 from classes.logger import log
 
@@ -53,9 +54,12 @@ def get_assets_path(file_path=None, create_paths=True):
 
             if not os.path.exists(asset_path):
                 if os.path.exists(asset_folder_name_30_char):
-                #update assets folder, if it follows the previous naming convention
-                    os.rename(asset_path_30_char, asset_path)
-                    log.info("Updating 30 character path to full length {}".format(asset_path))
+                    #update assets folder, if it follows the previous naming convention
+                    try:
+                        shutil.copytree(asset_folder_name_30_char, asset_folder_name)
+                        log.info("Copying shortened asset folder. {}".format(asset_path))
+                    except:
+                        log.error("Could not make a copy of assets folder")
                 else:
                     os.mkdir(asset_path)
                     log.info("Asset dir created as {}".format(asset_path))
--- a/src/classes/assets.py
+++ b/src/classes/assets.py
@@ -40,12 +40,20 @@ def get_assets_path(file_path=None, create_paths=True):
         # Generate asset folder name, max 30 chars of filename + "_assets"
         file_path = file_path
         asset_filename = os.path.splitext(os.path.basename(file_path))[0]
-        asset_folder_name = asset_filename[:30] + "_assets"
+        asset_folder_name = asset_filename + "_assets"
+        asset_folder_name_30_char = asset_filename[:30] + "_assets"
         asset_path = os.path.join(os.path.dirname(file_path), asset_folder_name)
+        asset_path_30_char = os.path.join(os.path.dirname(file_path), asset_folder_name_30_char)
+
+
 
         # Create asset folder, if necessary
         if create_paths:
-            if not os.path.exists(asset_path):
+
+            if os.path.exists(asset_path_30_char):
+                os.rename(asset_path_30_char, asset_path)
+                log.info("Updating 30 character path to full length {}".format(asset_path))
+            elif not os.path.exists(asset_path):
                 os.mkdir(asset_path)
                 log.info("Asset dir created as {}".format(asset_path))
             else:
--- a/src/classes/assets.py
+++ b/src/classes/assets.py
@@ -45,6 +45,7 @@ def get_assets_path(file_path=None, create_paths=True):
         asset_path = os.path.join(os.path.dirname(file_path), asset_folder_name)
 
         # Previous Assets File Name Convention.
+        # We can remove the 30_char variables after 05/27/2022
         asset_folder_name_30_char = asset_filename[:30] + "_assets"
         asset_path_30_char = os.path.join(os.path.dirname(file_path), asset_folder_name_30_char)
 
@@ -54,7 +55,8 @@ def get_assets_path(file_path=None, create_paths=True):
 
             if not os.path.exists(asset_path):
                 if os.path.exists(asset_path_30_char):
-                    #update assets folder, if it follows the previous naming convention
+                    #copy assets folder, if it follows the previous naming convention
+                    #must leave a copy for possible projects that shared the folder.
                     try:
                         shutil.copytree(asset_path_30_char, asset_path)
                         log.info("Copying shortened asset folder. {}".format(asset_path))
--- a/src/classes/assets.py
+++ b/src/classes/assets.py
@@ -37,20 +37,21 @@ def get_assets_path(file_path=None, create_paths=True):
         return info.USER_PATH
 
     try:
-        # Generate asset folder name, max 30 chars of filename + "_assets"
+        # Generate asset folder name filename + "_assets"
         file_path = file_path
         asset_filename = os.path.splitext(os.path.basename(file_path))[0]
         asset_folder_name = asset_filename + "_assets"
-        asset_folder_name_30_char = asset_filename[:30] + "_assets"
         asset_path = os.path.join(os.path.dirname(file_path), asset_folder_name)
+
+        # Previous Assets File Name Convention.
+        asset_folder_name_30_char = asset_filename[:30] + "_assets"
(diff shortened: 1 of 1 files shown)


jonoomph:
Based on a quick google:

Folder name length limits per OS:
 - Windows 8/10: 255 to 260
 - Mac OS: 1024
 - Linux: 255

So, perhaps the 30 char truncate is not the solution, but I do think we need to truncate it at some reasonable length... such that Project_Name + "_assets" length < 255. But as long as we are truncating the folder name, it seems possible that we can still have a conflict. But perhaps less often. :+1:

jonoomph:
Oh, and if we change this truncation (if that is a word), will this prevent different versions of OpenShot from finding the _assets folder?

src/classes/assets.py
```diff
@@ -37,15 +37,24 @@ def get_assets_path(file_path=None, create_paths=True):
...
     try:
-        # Generate asset folder name, max 30 chars of filename + "_assets"
+        # Generate asset folder name filename + "_assets"
         file_path = file_path
         asset_filename = os.path.splitext(os.path.basename(file_path))[0]
-        asset_folder_name = asset_filename[:30] + "_assets"
+        asset_folder_name = asset_filename + "_assets"
         asset_path = os.path.join(os.path.dirname(file_path), asset_folder_name)
 
+        # Previous Assets File Name Convention.
```
jonoomph, line 46: We could also tackle this with the same code:
```
for asset_folder_name in [asset_filename[:30], asset_filename[:250], ""...]:
  ...
```

JacksonRG:
> Based on a quick google:
> 
> Folder name length limits per OS:
> 
> * Windows 8/10: 255 to 260
> * Mac OS: 1024
> * Linux: 255
> 
> So, perhaps the 30 char truncate is not the solution, but I do think we need to truncate it at some reasonable length... such that Project_Name + "_assets" length < 255. But as long as we are truncating the folder name, it seems possible that we can still have a conflict. But perhaps less often.

Right. I feel 255 - 7 is enough of a name (nearly a paragraph) for all reasonable cases.

JacksonRG:
> Oh, and if we change this truncation (if that is a word), will this prevent different versions of OpenShot from finding the _assets folder?

That was my thinking, in checking for a truncated folder first, and moving it to the new naming system

src/classes/assets.py
JacksonRG: That looks very clean!

ferdnyc: > Based on a quick google:
> 
> Folder name length limits per OS:
> 
> * Windows 8/10: 255 to 260
> * Mac OS: 1024
> * Linux: 255

One sticking point: That Windows size was for the _total_ length of the **entire** path, all the way from `C:\` to the extension.

That limitation was **finally** lifted in (an _update_ to!) Windows 10, and I believe that change has been automatically pushed to all systems if they've been running Windows Update. But for a long time, Windows path strings were very restricted when you were using legacy APIs (meaning non-UNC, the `//host/share/path/` etc. names.) IIRC that was the reason for the short-ish folder naming. If the user tried to name their project _file_ something over the 260-character limit, the system would stop them, but the worry was that their name might be, say, 258 characters, so adding `_assets` would be the thing that pushed it over.

That being said, if someone is naming their project file something insane and long, and it's accepted as a legal path, these days our extra 7 characters won't often push it over the limit. (Sorry, no... actually I just remembered that it's only _three_ extra characters: the cost of adding `_assets` is offset by the fact that we've dropped the `.osp` extension.)

Except, of course, when `_assets` **is** the straw that breaks 260. There is still a slight chance of causing some headaches for any remaining Win7 users, by allowing significantly longer name lengths. (Which isn't to say that I'm opposed to it, just a factor to consider.)

> But as long as we are truncating the folder name, it seems possible that we can still have a conflict. But perhaps less often.

That's kind of the thing, I mean... if people are going to really push their luck, eventually they're going to break something.

The biggest danger is people creating long-ass project names, then _duplicating_ the files and letting Windows do its `My_long_ass_project_name (1).osp`, `My_long_ass_project_name (Copy).osp`, ... type disambiguation. Because those names only differ at the very far end of the string, any truncation potentially results in a collision.

JacksonRG: Good point! I forgot the .osp extension. I was going to go with 
`asset_folder_name = asset_filename[:248] + "_assets"` 255 - 7.

src/classes/assets.py
```diff
@@ -37,17 +38,33 @@ def get_assets_path(file_path=None, create_paths=True):
...
+        # We can remove the 30_char variables after 05/27/2022
+        asset_folder_name_30_char = asset_filename[:30] + "_assets"
+        asset_path_30_char = os.path.join(os.path.dirname(file_path), asset_folder_name_30_char)
+
+
         # Create asset folder, if necessary
         if create_paths:
+
```
jonoomph, line 55: Nitpick: remove blank line

src/classes/assets.py
```diff
@@ -37,17 +38,33 @@ def get_assets_path(file_path=None, create_paths=True):
...
         asset_filename = os.path.splitext(os.path.basename(file_path))[0]
-        asset_folder_name = asset_filename[:30] + "_assets"
+        asset_folder_name = asset_filename[:248] + "_assets" #Windows max name size is 255. 248 = 255 - len("_assets")
         asset_path = os.path.join(os.path.dirname(file_path), asset_folder_name)
 
+        # Previous Assets File Name Convention.
+        # We can remove the 30_char variables after 05/27/2022
+        asset_folder_name_30_char = asset_filename[:30] + "_assets"
+        asset_path_30_char = os.path.join(os.path.dirname(file_path), asset_folder_name_30_char)
+
```
jonoomph, line 51: Netpick, remove one of these 2 blank lines

jonoomph:
LG Otherwise... :+1:

ferdnyc:
@jonoomph, @JacksonRG :

Er, crap. Turns out, the 30-character prefix length is also hardcoded into libopenshot's support for opening `.osp` files.

@[Timeline.cpp:133-139](https://github.com/OpenShot/libopenshot/blob/d9ea98eb2de96665725071ccee352c84e1bda5de/src/Timeline.cpp#L133-L139):
```cpp
	// Determine asset path
	QString asset_name = filePath.baseName().left(30) + "_assets";
	QDir asset_folder(filePath.dir().filePath(asset_name));
	if (!asset_folder.exists()) {
		// Create directory if needed
		asset_folder.mkpath(".");
	}
```

JacksonRG:
@ferdnyc Yikes! Thanks for pointing that out. We should probably make project wide variable or function for the filePath, but for now, I'll fix the timeline and do a search for any other path code.

ferdnyc:
@JacksonRG That _should_ be it, the `.osp`  support in libopenshot is a relatively recent addition.

I was actually toying with the idea of storing the asset folder name -- NOT the path, just the name of the folder -- in the project file itself.

#### Pros
1. We know unambiguously where the assets for _this_ file were stored.
1. We keep track of them even if the user renames the project file, but not the folder (which seems more likely than not).
1. We won't have this problem again even if we change the formula for generating the name, because locating it won't be dependent on the formula.
1. If there ARE name collisions, and we come up with some way of detecting that, we could use an alternate name for the folder.

#### Cons
1. We lose the folder if the user DOES rename the project file _and_ the folder. (Though, we could still fall back to checking the standard generated names, to mitigate that. And also, honestly, this will never ever happen.)
1. We'd have to always reference all asset folder contents with a standard placeholder -- ok, well, we do that already.
1. We won't have the folder name until we've parsed the project data.

Con 3 is the big one. As I recall, right now we do string replacement on the JSON form of the project data _first_, to correct relative paths and placeholder paths. Then, we parse the corrected JSON into project data. But if we were going to get the folder name **from** the project data, we'd have to parse the JSON _first_, with the relative and placeholder paths intact. Then, we could correct the paths when parsing the other data.

There's no reason that *couldn't* work. Both (for sure) Python's `os.path()` and (I _believe_) Qt's `QFileInfo()` should accept paths like `@assets/thumbnails/blah.png` without complaint -- they shouldn't validate the path against disk until asked to. Until then, it's just a relative path with a funny directory name. (Although that funny directory name may also be an _invalid_ directory name on Windows, and we'd have to check whether that causes issues for `QFileInfo()`.)

Handling the paths this way would also be **much** safer than regexp-editing the JSON string _before_ parsing it (something that's caused us problems in the past).

But, because it would change the form of path strings as they come in via `SetJson()` in libopenshot, we'd probably have to change how those parts of the data are read.

Like, any JSON that might involve a placeholder path, we'd have to queue processing until after we've read all the keys in the root object, just in case it contains an `"assetFolderName"`. If it does, the relevant `SetJsonValue()` methods would all have to be taught how to correct placeholder-relative paths, when provided with the _path_ of the asset folder (already checked for validity and made absolute by the Timeline `SetJsonValue()`).

As far as OpenShot goes, because the JSON is read into the `get_app().project._data` Python `dict` en masse, before anything else is done with that data, we could actually just recursively examine its contents and update the path strings before they're parsed.  

Right now the path placeholders are subbed in during save right after `json.dumps()`-ing the data to string, by `classes.json`. Which makes sense, as we don't want to modify the in-memory data **before** we dump it. 

And so they're swapped out again _before_ `json.loads()`, in the same class. But there's no reason that process has to be symmetric. 

The data dict created by `json.loads()` could still contain "`@assets/`" style paths initially, just as long as it's immediately walked to correct them. Unlike in libopenshot, nothing is done with those path strings until after the entire data object has been loaded. And if we defer the path replacements until after we've parsed the JSON, then there's no reason we can't take the folder name **from** there.

ferdnyc:
@JacksonRG @jonoomph 

I may also be overcomplicating this, on the `libopenshot` front, saying we have to "defer" parsing. Like, we might be able to start off `Timeline::SetJsonValue()` like this:
```c++

QDir assetDir;
QString assetPath;
QString baseName;

if (!root["path"].isNull())
    path = root["path"].asString();

// member var 'path' can also be set in the constructor
if (path) {
    QFileInfo filePath(QString::fromStdString(path));
    baseName = filePath.baseName();
    assetDir = QDir(filePath.dir()).makeAbsolute();
}

if (!root["assetFolderName"].isNull()) {
    QString assetFolder(root["assetFolderName"].asCString());
    if (assetDir.cd(assetFolder))
        assetPath = assetDir.canonicalPath();
}

// If we didn't find the asset folder (or none was specified)
// but we have a project file, try some generated paths
if (baseName && !assetPath) {
    if (   !assetDir.cd(baseName.left(250) + "_assets")
        && !assetDir.cd(baseName.left(30) + "_assets")
      ) {
        // As a last resort, create a directory
        assetDir.setPath(baseName.left(250) + "_assets");
        assetDir.mkPath(".");
    }
    assetPath = assetDir.canonicalPath();
}
```
...And then just call each `Clip::SetJsonValue` and `Effect::SetJsonValue` with an optional argument `assetPath` (default `"."`) that they in turn pass to the Reader `SetJsonValue()`s, so they can do:
```c++
if (!root["path"].isNull()) {
    QString json_path(root["path"].asCString());
    if (json_path.startsWith("@assets/"))
         json_path = assetPath + json_path.mid(7);
    path = QFileInfo(json_path).canonicalFilePath().toStdString();
}
```
I guess we'd need a second optional arg for the `@transitions/` mapping, too.

It adds a bunch of code, but also replaces a lot of [much more expensive code](https://github.com/OpenShot/libopenshot/blob/d9ea98eb2de96665725071ccee352c84e1bda5de/src/Timeline.cpp#L133-L186):
```c++
	// Determine asset path
	QString asset_name = filePath.baseName().left(30) + "_assets";
	QDir asset_folder(filePath.dir().filePath(asset_name));
	if (!asset_folder.exists()) {
		// Create directory if needed
		asset_folder.mkpath(".");
	}


	// Load UTF-8 project file into QString
	QFile projectFile(QString::fromStdString(path));
	projectFile.open(QFile::ReadOnly);
	QString projectContents = QString::fromUtf8(projectFile.readAll());


	// Convert all relative paths into absolute paths (if requested)
	if (convert_absolute_paths) {


		// Find all "image" or "path" references in JSON (using regex). Must loop through match results
		// due to our path matching needs, which are not possible with the QString::replace() function.
		QRegularExpression allPathsRegex(QStringLiteral("\"(image|path)\":.*?\"(.*?)\""));
		std::vector<QRegularExpressionMatch> matchedPositions;
		QRegularExpressionMatchIterator i = allPathsRegex.globalMatch(projectContents);
		while (i.hasNext()) {
			QRegularExpressionMatch match = i.next();
			if (match.hasMatch()) {
				// Push all match objects into a vector (so we can reverse them later)
				matchedPositions.push_back(match);
			}
		}


		// Reverse the matches (bottom of file to top, so our replacements don't break our match positions)
		std::vector<QRegularExpressionMatch>::reverse_iterator itr;
		for (itr = matchedPositions.rbegin(); itr != matchedPositions.rend(); itr++) {
			QRegularExpressionMatch match = *itr;
			QString relativeKey = match.captured(1); // image or path
			QString relativePath = match.captured(2); // relative file path
			QString absolutePath = "";


			// Find absolute path of all path, image (including special replacements of @assets and @transitions)
			if (relativePath.startsWith("@assets")) {
				absolutePath = QFileInfo(asset_folder.absoluteFilePath(relativePath.replace("@assets", "."))).canonicalFilePath();
			} else if (relativePath.startsWith("@transitions")) {
				absolutePath = QFileInfo(openshotTransPath.absoluteFilePath(relativePath.replace("@transitions", "."))).canonicalFilePath();
			} else {
				absolutePath = QFileInfo(filePath.absoluteDir().absoluteFilePath(relativePath)).canonicalFilePath();
			}


			// Replace path in JSON content, if an absolute path was successfully found
			if (!absolutePath.isEmpty()) {
				projectContents.replace(match.capturedStart(0), match.capturedLength(0), "\"" + relativeKey + "\": \"" + absolutePath + "\"");
			}
		}
		// Clear matches
		matchedPositions.clear();
	}
```

JacksonRG:
@ferdnyc Still reading through this, but I just thought: If we store the asset folder name in the project data, would we give the option for the user to set it? Then there's an optional path to having a shared asset folder between projects.

PR #4163, opened 2021-05-28
Some query class / unit tests fixes and enhancements

ferdnyc:
My ambitious PRs tend to sit for months on end with no review, often becoming so divergent from the target branch that the conflicts are unresolvable. As a result, I'm trying to submit more, smaller PRs. This is one, containing a first round of tweaks to the project-data query classes and their associated unit tests.

* Add a unit test for `filter(intersect=)`
  Also, make the test data slightly more interesting by varying `Position` values for the clips/transitions in the test data.
* classes.query: Don't hold a reference to `app.project`
  By calling `get_app()` each time we need to access the project data, we delay the lookup until it's needed. As a result, query classes can now be imported even if an application instance has not yet been initialized. (As long as it's initialized before the first call to a query class method).
* Unit tests: 
    * Import query classes at top of file
      (There's no longer any need to delay the imports.)
   * Rename the `TestQueryClass` class to `QueryTests`, which is more in line with typical `unittest` naming
   * Replace explicit `TestQueryClass` references in the class code with `cls` or `self` as appropriate, so any future renaming will be far less of a pain

### DIFF
--- a/src/classes/query.py
+++ b/src/classes/query.py
@@ -32,11 +32,6 @@ from classes import info
 from classes.app import get_app
 
 
-# Get project data reference
-app = get_app()
-project = app.project
-
-
 class QueryObject:
     """ This class allows one or more project data objects to be queried """
 
@@ -56,7 +51,7 @@ class QueryObject:
         if not self.id and self.type == "insert":
 
             # Insert record, and Generate id
-            self.id = project.generate_id()
+            self.id = get_app().project.generate_id()
 
             # save id in data (if attribute found)
             self.data["id"] = copy.deepcopy(self.id)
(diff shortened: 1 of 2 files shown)


src/classes/query.py
```diff
@@ -242,30 +237,25 @@ def get(**kwargs):
...
-            project_folder = os.path.dirname(project.current_filepath)
-
-        # Convert relative file path into absolute (if needed)
         file_path = self.data["path"]
-        if not os.path.isabs(file_path) and project_folder:
-            file_path = os.path.abspath(os.path.join(project_folder, self.data["path"]))
+        if os.path.isabs(file_path):
+            return file_path
+
+            # Try to expand path relative to project folder
```
ferdnyc, line 244:
```suggestion
        # Try to expand path relative to project folder
```

jonoomph:
LGTM

PR #4171, opened 2021-06-02
Ruler refactor

JacksonRG:
## Super Zoom!
I've changed the ruler code so that you can zoom as much as you want with +/- keys. This does some math so that the number of pixels between ruler ticks is always a factor of the FPS (rounded). That way, as I start drawing the tick marks on a the first frame of a second, every following second on the timeline should be marked.

### The ruler math
I did the factoring by adding to `functions.js`:
- I added code to find prime numbers
    - Uses [sieve of eratosthenes](https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes) method.
- A set to store the primes in, to avoid repeated work
- Starting at 1 frame per ruler tick, I multiply by prime factors of FPS until the tick marks are at least 40px apart.

## Issues:
1) On Webkit, the timeline's shift-scroll (horizontal scroll) doesn't work.
2) On Webengine, the ctrl-scroll (zoom) doesn't work.

(These two issues are present in dev)


JacksonRG:
Shoot! Before anyone says it, I need to get it to draw times on first load.

jonoomph:
@JacksonRG This PR seems to include all the other changes made in develop. It's a bit crazy looking now. @ferdnyc Is this normal, and will this merge correctly? Just double checking...

JacksonRG:
> @JacksonRG This PR seems to include all the other changes made in develop. It's a bit crazy looking now. @ferdnyc Is this normal, and will this merge correctly? Just double checking...

Sorry, Totally missed your comment. Anyway, I just rebased onto develop (I think without breaking anything).

Also since we were talking about openshot just letting users keep zooming in, I removed the scroll limmits. (at least if you scroll with `-` and `=`.

jonoomph:
Looking much better!

jonoomph:
@JacksonRG I just resolved the conflict on `src/timeline/js/functions.js`... but please double check everything still works.

PR #4174, opened 2021-06-04
Don't kill the WebView on aboutToQuit

ferdnyc:
I screwed up. A while back, I connected PyQt5's `QApplication.aboutToQuit` signal to the `deleteLater()` slot of the WebView, intending it to clean up just before shutdown.

Turns out, PyQt5's implementation of QApplication emits that signal every time the user types anything on the input line, when running Python interactively, making the interactive shell essentially useless.

Merging this might bring back some crash-on-quit issues that the signal's destruction of the Timeline was preventing, but we should be able to find another solution. (Maybe connecting  `QApplication.quit` to the slot, instead of `aboutToQuit`.)

I also added a debug log in `OpenShotApp.cleanup()`. It's connected to `OpenShotApp.aboutToQuit` as well, and I'm concerned the same issue could be causing excessive calls to that method, even though we can't see them.

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -308,7 +308,7 @@ class OpenShotApp(QApplication):
     def show_errors(self):
         count = len(self.errors)
         if count > 0:
-            self.log.warning("Displaying {} startup messages".format(count))
+            self.log.warning("Displaying %d startup messages", count)
         while self.errors:
             error = self.errors.pop(0)
             error.show()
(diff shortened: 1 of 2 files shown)


ferdnyc:
It should be safe to use `app.lastWindowClosed()` as our trigger to `deleteLater()` the Webview, so I've added that in to replace the `aboutToQuit` connection.

ferdnyc:
That passes, so I'm going to merge this to get the code out there. Hopefully we won't see any uptick in crash-on-exit reports, but if we do we can always revert.

PR #4175, opened 2021-06-04
Logging overhaul, to solve issues caused by stdout/stderr redirection

ferdnyc:
This is a major rewrite of `classes.logging`, which I'm **hopeful** solves the long-standing issues caused by the redirection of `sys.stdout` and `sys.stderr` for capture purposes.

This PR doesn't do away with the redirection, instead it does it _cleanly_ and in a manner that doesn't break Python features that depend on the standard I/O streams.

In the new logging, the `StreamToLogger` shims that take over for `sys.stdout` and `sys.stderr` return to buffering the messages written to the streams, logging _only_ when the stream is `flush()`ed. In my testing, this means that multi-line texts like tracebacks will be captured as a _single_ log message. Or in the worst case, as a very small number of partial messages, when other output interrupts the output when it's only partially complete.

All output that's captured on `sys.stdout` or `sys.stderr` is _immediately_ passed along to the real stream outputs; the only thing the shim does is buffer a copy first. When a `flush()` comes through, the buffer is then logged to an `OpenShot.stderr` child logger of our main `OpenShot` logging context, and those log messages are **suppressed** from being re-echoed to the console, to avoid redundant duplication of output.

This means that any output send to the standard output streams, whether by OpenShot, by Python, or by some other code, will be displayed on the terminal unaltered from its original form (no more logging decoration), which should make tools that depend on the format of those messages much happier. (I'm hopeful this includes Sentry, as well.)

### DIFF
--- a/src/classes/logger.py
+++ b/src/classes/logger.py
@@ -133,3 +133,4 @@ def set_level_file(level=logging.INFO):
 def set_level_console(level=logging.INFO):
     """Adjust the minimum log level for output to the terminal"""
     sh.setLevel(level)
+
--- a/src/classes/logger.py
+++ b/src/classes/logger.py
@@ -40,65 +40,96 @@ streams = {}
 
 class StreamToLogger(object):
     """Custom class to log all stdout and stderr streams (from libopenshot / and other libraries)"""
-    def __init__(self, logger, log_level=logging.INFO):
-        self.logger = logger
+    def __init__(self, parent_stream, log_level=logging.INFO):
+        self.parent = parent_stream or sys.__stderr__
+        self.logger = logging.LoggerAdapter(
+            logging.getLogger('OpenShot.stderr'), {'source': 'stream'})
         self.log_level = log_level
-        self.linebuf = ''
+        self.logbuf = ''
 
-    def write(self, buf):
-        for line in buf.rstrip().splitlines():
-            self.logger.log(self.log_level, line.rstrip())
+    def write(self, text):
+        self.logbuf += str(text) or ""
+        self.parent.write(text)
 
     def flush(self):
-        pass
+        if self.logbuf.rstrip():
+            self.logger.log(self.log_level, self.logbuf.rstrip())
+        self.logbuf = ''
 
     def errors(self):
         pass
 
 
-# Create logger instance
-log = logging.Logger('OpenShot')
+class StreamFilter(logging.Filter):
+    """Filter out lines that originated on the output"""
+    def __init__(self, predicate):
+        self.predicate = predicate
+
+    def filter(self, record):
+        source = getattr(record, "source", "")
+        return self.predicate(source)
+
 
 # Set up log formatters
 template = '%(levelname)s %(module)s: %(message)s'
 console_formatter = logging.Formatter(template)
 file_formatter = logging.Formatter('%(asctime)s ' + template, datefmt='%H:%M:%S')
 
-# Add normal stderr stream handler
-sh = logging.StreamHandler()
-sh.setFormatter(console_formatter)
-sh.setLevel(info.LOG_LEVEL_CONSOLE)
-log.addHandler(sh)
-handlers['stream'] = sh
+# Configure root logger for minimal logging
+logging.basicConfig(level=logging.ERROR)
+root_log = logging.getLogger()
 
-# Add rotating file handler
+# Set up our top-level logging context
+log = root_log.getChild('OpenShot')
+log.setLevel(info.LOG_LEVEL_FILE)
+# Don't pass messages on to root logger
+log.propagate = False
+
+#
+# Create rotating file handler
+#
 fh = logging.handlers.RotatingFileHandler(
-    os.path.join(info.USER_PATH, 'openshot-qt.log'), encoding="utf-8", maxBytes=25*1024*1024, backupCount=3)
-fh.setFormatter(file_formatter)
+         os.path.join(info.USER_PATH, 'openshot-qt.log'),
+         encoding="utf-8",
+         maxBytes=25*1024*1024, backupCount=3)
 fh.setLevel(info.LOG_LEVEL_FILE)
+fh.setFormatter(file_formatter)
+
 log.addHandler(fh)
-handlers['file'] = fh
+
+#
+# Create typical stream handler which logs to stderr
+#
+sh = logging.StreamHandler(sys.stderr)
+sh.setLevel(info.LOG_LEVEL_CONSOLE)
+sh.setFormatter(console_formatter)
+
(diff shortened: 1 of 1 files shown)


src/classes/logger.py
```diff
@@ -40,65 +40,96 @@
...
+sh.setLevel(info.LOG_LEVEL_CONSOLE)
+sh.setFormatter(console_formatter)
+
+# Filter out redirected output on console, to avoid duplicates
+#
+# The lambda is used to test the "source" property of the
+# logging record, if present.
+# {"source": "stream"} will be attached to any records generated
+# by the redirection, thanks to LoggingAdapter)
+filt = StreamFilter(lambda x: x != "stream")
```
jonoomph, line 113: :100:

jonoomph: I like it! Nice

src/classes/logger.py
```diff
@@ -40,65 +40,96 @@
...
-            'stdout': sys.stdout,
-            })
-        # Re-route output streams
-        handlers['stdout'] = StreamToLogger(log, logging.INFO)
-        sys.stdout = handlers['stdout']
-        handlers['stderr'] = StreamToLogger(log, logging.ERROR)
-        sys.stderr = handlers['stderr']
+    if (getattr(sys, 'frozen', False)
+       or sys.stdout != sys.__stdout__):
+        return
```
jonoomph, line 123: I don't recall why this frozen block was needed. Do you recall? If so, we should add a comment here to demystify it, lol

src/classes/logger.py
```diff
@@ -40,65 +40,96 @@
...
+    sys.stdout = StreamToLogger(sys.stdout, logging.INFO)
+    sys.stderr = StreamToLogger(sys.stderr, logging.WARNING)
```
jonoomph, line 125: Should we be using `info.LOG_LEVEL_CONSOLE` and `info.LOG_LEVEL_FILE` here?

src/classes/logger.py
```diff
@@ -40,65 +40,96 @@
...
-    handlers['file'].setLevel(level)
+    """Adjust the minimum log level written to our logfile"""
+    fh.setLevel(level)
 
 
 def set_level_console(level=logging.INFO):
-    handlers['stream'].setLevel(level)
-
+    """Adjust the minimum log level for output to the terminal"""
+    sh.setLevel(level)
```
jonoomph, line 135: Nitpick: is there a blank line at the end of this file? Or is the PR just not showing it... :thinking:

jonoomph:
Thanks @ferdnyc, this looks very nice! I was also spending too many hours messing with this logging class. I would make a bunch of changes... almost solve it, and then get angry and revert it all, lol. I'm very happy you found a pretty slick solution, looking forward to testing it out. Feel free to merge this into the `integrate-sentry-tracing` branch when you are ready!

src/classes/logger.py
ferdnyc: I honestly have never had any idea. I assumed it had something to do with the AppImage, but that's pure conjecture.

src/classes/logger.py
ferdnyc: Those are the levels the messages are emitted AT, not the filtering levels for the loggers, so I'd say no. (In theory sys.stderr could be `info.ERROR`, but then we'd have a lot more errors in the logs. :laughing: )

src/classes/logger.py
ferdnyc: The file ENDS with a newline (my editor makes sure of that automatically), but I'm not sure there's a _blank_ line. I can check.

ferdnyc:
> Thanks @ferdnyc, this looks very nice! I was also spending too many hours messing with this logging class. I would make a bunch of changes... almost solve it, and then get angry and revert it all, lol. 

Ayup. The reason it took me ALL night is, this was version... three, I think. The first two went nowhere.

In the end it all came down to one line:

https://github.com/OpenShot/openshot-qt/blob/98bc546b5dfc317ea388255e1b3721ee9a0feb0e/src/classes/logger.py#L85-L86

Until I stopped propagation to the root logger, it would still get the messages and do its own thing, so I had no real control over what was done with them. I didn't figure that out until I gave every level a completely different logging format. Then I finally figured out _where_ the messages I couldn't seem to eliminate were actually coming from.

(Reading [this flowchart in the documentation](https://docs.python.org/3.9/howto/logging.html#logging-flow) was also immensely helpful, in the end.)

image

> I'm very happy you found a pretty slick solution, looking forward to testing it out. Feel free to merge this into the `integrate-sentry-tracing` branch when you are ready!

Will do, first I'm gonna fire off another CI run (which I believe should be fixed now, a couple of hours ago I merged my libopenshot changes that fixed the Python bindings — if not, it'll be fixed as soon as the PPAs are regenerated), and also see what bug is up Codacy's ass this time.

src/classes/logger.py
ferdnyc: This is actually a bit of an overcomplicated design (as the fact that it requires 5 lines of comment to explain one line of code makes pretty clear)... originally I thought I was going to need two filters, one on the stream handler and one on the file handler, and I wanted to write it in a way that I could use the same class for both filters.

In the end, the issue was the propagation to the root logger, and one I fixed that I didn't need a file handler filter. (Which makes sense, we want ALL log messages to go to the logfile, so what's there to filter?) So really the filter class could just have `bool(source != "stream")` hardcoded into it, which would probably be clearer.

src/classes/logger.py
ferdnyc: So, turns out, there was a surprising amount of nuance to that question.

Like I said, the file **ended** on a blank line, my editor is set to make sure of that. (Actually, thanks to the `.editorconfig` file at the root of the repo, so is any editor that respects those files.) That means that the file is protected against concatenation damage and other issues, so you could do this:
```sh
$ echo "=======" >> logger.py; tail -3 logger.py
    """Adjust the minimum log level for output to the terminal"""
    sh.setLevel(level)
=======
```
You'd **THINK** it would also mean that an editor would let you position your cursor on line 136, since the last line of the file (line 135) ends with a newline. But surprisingly enough, the first two editors I checked (`vim` and `gedit`) didn't; they eat the final newline, so in those editors the last line of the file indeed appeared to be line 135.

So, I added a _second_ newline to the end of the file, ensuring that editors would show a blank line at the end.

...Naturally, my own editor (Github Atom) **doesn't** eat the final newline, so in Atom it now appears as though there are _two_ blank lines at the end. :rofl: :

image

src/classes/logger.py
ferdnyc: I did indeed end up eliminating it. The cleverness just didn't justify the complexity.

PR #3604, opened 2020-07-13
Initial support for QtWebEngine and Async callbacks for timeline/Qt Integration

jonoomph:
- Requires QtWebEngine and Qt 5.6+
 - Requires OpenGL python module (to fix black screens on some systems)
 - Enables smooth scrolling on timeline
 - All callbacks from JS to Python are Async

Based on the initial work: https://github.com/OpenShot/openshot-qt/pull/2512

Related to PRs:
https://github.com/OpenShot/libopenshot-audio/pull/100
https://github.com/OpenShot/libopenshot/pull/549

PR #4178, opened 2021-06-05
Prevent feedback loop when scrolling timeline on WebEngine backend

jonoomph:
Prevent feedback loop when scrolling timeline on WebEngine backend triggered by ZoomSlider widget. Also, it appears that we might be invoking run_js too often, and it gets queued up in blocks and processed in chunks. Not very happy with the performance on web engine at this point, but this fixes the obscene jerkiness.

### DIFF
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -69,6 +69,7 @@ App.controller("TimelineCtrl", function ($scope) {
   $scope.track_label = "Track %s";
   $scope.enable_sorting = true;
   $scope.ThumbServer = "http://127.0.0.1/";
+  $scope.ignore_scroll = false;
 
   // Method to set if Qt is detected (which clears demo data
   // and updates the document_is_ready variable in openshot-qt)
@@ -285,13 +286,19 @@ App.controller("TimelineCtrl", function ($scope) {
     var scrolling_tracks = $("#scrolling_tracks");
     var horz_scroll_offset = normalizedScrollValue * timeline_length;
     scrolling_tracks.scrollLeft(horz_scroll_offset);
+
+    $scope.ignore_scroll = true;
+    $("#scrolling_ruler").scrollLeft(horz_scroll_offset);
   };
 
-  // Scroll the timeline horizontally of a certain amount (scrol_value)
+  // Scroll the timeline horizontally of a certain amount
   $scope.scrollLeft = function (scroll_value) {
     var scrolling_tracks = $("#scrolling_tracks");
     var horz_scroll_offset = scrolling_tracks.scrollLeft();
     scrolling_tracks.scrollLeft(horz_scroll_offset + scroll_value);
+
+    $scope.ignore_scroll = true;
+    $("#scrolling_ruler").scrollLeft(horz_scroll_offset + scroll_value);
   };
 
   // Center the timeline on a given time position
--- a/src/timeline/js/directives/ruler.js
+++ b/src/timeline/js/directives/ruler.js
@@ -54,6 +54,22 @@ App.directive("tlScrollableTracks", function () {
         $("#scrolling_ruler").scrollLeft(element.scrollLeft());
         $("#progress_container").scrollLeft(element.scrollLeft());
 
+        if (scope.ignore_scroll == true) {
+          // Reset flag and ignore scroll propagation
+          scope.$apply( () => {
+            scope.ignore_scroll = false;
+            scope.scrollLeft = element[0].scrollLeft;
+          })
+          // exit at this point
+          return;
+
+        } else {
+          // Only update scroll position
+          scope.$apply( () => {
+            scope.scrollLeft = element[0].scrollLeft;
+          })
+        }
+
         // Send scrollbar position to Qt
         if (scope.Qt) {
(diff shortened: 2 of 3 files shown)


jonoomph:
@ferdnyc This makes me want to do the following things:
1) Delete webkit
2) Delete webengine
3) Build a custom QWidget (or series of widgets) that just draws the timeline (visible portion).

Although it might take a few weeks of development, it would drastically decrease the size of our installers / images, and remove the most annoying and difficult part of our UI. I'll do it!  Just give me the slightest push, lol. :laughing:

ferdnyc:
@jonoomph 

The only thing I would suggest is... well, two things:

1. **We should at least consider doing it in QML instead.**
    QML is apparently Da Future™ of Qt, and combines the best of JavaScript and the Qt framework ­— the limitations of QtWidgets mean interfaces built that way will never be as dynamic and interactive as QML allows; exactly the reason the Timeline was built in AngularJS to begin with.
2. **More importantly, it should be done as part of a move to the Timeline Model [I've mentioned us needing](https://github.com/OpenShot/openshot-qt/pull/3604#issuecomment-671615627).**
    The problem isn't really that the Timeline and the ZoomSlider communicate too much, or incorrectly. It's that they communicate **_at all_**. In the Qt Way™, they'd both be totally independent, totally self-contained views on the same model, each able to exist without any knowledge of the other, indeed without the _existence_ of the other. Interacting with either one would update the relevant model data, and the changes would automatically be reflected in any other views on that data — whether there be zero, one, or many. We're breaking that model (no pun) every time we have any view class communicate with any other view class, for any reason.

ferdnyc:
> @jonoomph
> 
> 2. **More importantly, it should be done as part of a move to the Timeline Model 

Model or ModelS, in fact. Just like all Qt data models have a separate SelectionModel attached to them (which is 100% how we should be handling selection on the Timeline), we might write a separate ScrollingModel just to handle positioning of the scroll rectangle in the ZoomSlider and the scrollbars of the Timeline. No reason the data all has to be shoehorned into a single model.

jonoomph:
@ferdnyc Thanks for the suggestions. I'm not super familiar with QML (I know the basics though). At a previous company I worked at, we built a timeline QWidget, which was relatively simple, and mostly just used the paint() event to draw on top of the widget. It emitted signals when certain events happened, so other areas of the UI could respond to certain things. It did not have a dedicated Qt Model, but we had some custom classes to handle state (and model related functionality). It was simple, fast, and worked great!

I really regret using a web-view for OpenShot now, however if I recall correctly, I thought a web-view might have some huge benefits:
 - Easier to style (CSS)
 - Smoother performance / animated CSS transitions / Smooth scroll / etc...
 - Based on JS and common web-frameworks (maybe we would attract more contributors around this feature)
 - Easier to theme (just change the CSS file), etc...

However, in reality:
 - We've gotten very few PRs for additional timeline functionality
 - The JS == Python connections have been much more difficult to maintain (especially with web engine)
 - We quickly ran into limitations with HTML Canvas elements (and max length) issues
 - We required Angular and data binding, to prevent manual redrawing of the timeline on state changes (adding more complexity)
 - In general, the JS and Python integration got way too complicated
 - Performance on web engine is actually worse in many examples (i.e. scrolling, installer size, builder dependencies)

I love simple things. So far, all my interactions with Qt Models have been very complex and frustrating: QStandardItemModel => ProxyModel => QItemSelectionModel, lots of conversions for selection/model indexes, etc... OpenShot currently has a data model for the timeline (`ProjectDataStore`), it's just not in Qt's standard formats. 

However, I do agree with a few things:
 - QML should be investigated and considered (I'll take a look). I'm worried about embedding QML in a QWidget project though... as this become Web Engine 2.0, lol
 - Signals should be owned by their respective widgets (and not all on MainWindow). This makes the connections more difficult, but exposing some methods to connect signals should be simple enough.
 - Changes to the data model (i.e. ProjectDataStore) should be the preferred way to share data in the UI. So, ZoomSlider should probably be updating this data, vs emitting signals for another view to consume. But since the data was not persistent, didn't need to be saved, and only 1 other widget needed it, I went with signals vs model changes.
  
Signals and Emitting: 
I still don't have any fundamental issue with views communicating by emitting signals to each other, and having a hard dependency between them. This really boils down to the design of our program and widgets, and if it's our desire to support more flexible use cases (i.e. another app which wants to reuse our Zoom Slider, for example). The more hard dependencies we add, the less useful our widgets are for other apps... however, that is not necessarily a bad or good thing. It's just a thing. How do we want to invest our time during development?  To support our own app, or to make things more generic to support other apps. Again, it's not a right/wrong, or good/bad thing... it's just a design decision. And at the end of the day, I want to build a great video editor application, not necessarily a collection of reusable video editor QWidgets. I'm sure we'll disagree on this point, lol, but it's an interesting discussion for sure!

Prototype:
I'm going to try adding a 3rd timeline backend: `qtimeline`, and make a super simple prototype... proof of concept. Primarily to test performance of hundreds/thousands of clips... and see how it feels. I'm happy to play around with Qt Models at this time, and experiment with QML also. Let's explore these approaches, and see what makes the most sense. But regardless, let's drop the Webkit and WebEngine dependencies as quickly as possible. Life is too short, lol.

Issue #3953, opened 2021-01-07
HTTPS for anonymized metrics


ferdnyc:
It's been raised (I don't have the issue reference handy) that our metrics — however anonymized and devoid of user personal data they may be — are being sent back to the "mothership" (either of them, either Analytics' or our own) in-the-clear over HTTP.

Switching to HTTPS should be technically feasible, and @jonoomph is on board with the idea, the only potential hurdle is making sure the packaging is done correctly. I looked into that some; my results:

1. The Python `requests` module we use for HTTP communication supports HTTPS natively, with certificate verification offloaded to a separate `certifi` module.
2. `certifi` uses a certificate file (a standard `.pem`-format keystore), which it obtains one of two ways. I'll call them 'internal' and 'external'.
    1. **Internal**: The `certifi` Python package, as installed, contains a `cacert.pm` file, so everything is self-contained within the package.
       **Used by**: So far, I've found an internal certificate store in all of the following Python situations: 
       * Linux, when `certifi` is installed by `pip` from PyPi
       * Windows MSYS2, when installed with `pacman -S mingw-w64-x86_64-python-requests`
       * Windows native Python, when `certifi` is installed using `pip`
   2. **External**: In this mode, `certifi` is built to reference a standard certificate store located in some shared location within the OS, outside of `certifi`'s package directory.
      **Used by**: The only place I've seen this, and I suspect the only place I _will_ see it, is in Python installs packaged by Linux distributions. For example:
       * Fedora 33 has a certificate file at `/etc/pki/tls/certs/ca-bundle.crt`, and `certifi` uses that.
       * Ubuntu 20.04 keeps it at `/etc/ssl/certs/ca-certificates.crt`, same deal.

3. But best of all, no matter _how_ the certificates are managed, running a simple command will have `certifi` _tell us_ where they are:
    ```sh
    # Fedora system python
    $ python3 -m certifi
    /etc/pki/tls/certs/ca-bundle.crt
    
    # Windows MSYS2 shell
    $ /mingw64/bin/python -m certifi
    C:/msys64/mingw64/lib/python3.8/site-packages/certifi/cacert.pem

    # Windows MinGW64 shell
    % python3 -m certifi
    C:/msys64/mingw64/lib/python3.8/site-packages/certifi/cacert.pem
    
    # Fedora system, inside a venv
    venv$ python3 -m pip install requests certfi
    ...
    venv$ python3 -m certifi
[...]

PR #4182, opened 2021-06-10
Make sentry_sdk (and distro) optional dependencies

ferdnyc:
### Optional `sentry_sdk`

This PR moves all code that depends on the `sentry_sdk` module into a new `classes.sentry`, and locally attempts to `import sentry_sdk as sdk` if available. If not, `sdk` is set to `None`.

The `sentry_sdk` methods `set_tag`, `set_user`, and `set_context` are mirrored into `classes.sentry`. They are safe to call from other code, and will silently do nothing if `sentry.sdk` is `None`. `sentry.init_tracing()` and `sentry.disable_tracing()` offer the same protections.

Any other `sentry_sdk` methods can be used as needed, but those calls should take the form:
```python3
from classes import sentry
if sentry.sdk:
    sentry.sdk.some_method(...)
```

The Sentry init is moved just _slightly_ later in the `launch.py` initialization, because there's really no point in initializing it just to exit after showing the command-line `--help` or similar. (Also, importing multiple modules before adjusting `sys.path()` has been problematic in the past.)

### Optional `distro`
Additionally, the `distro` module is made optional in `classes.sentry` (as it's not a standard Python module), and is also (optionally) used in `classes.metrics` (in place of the long-since-deprecated `platform.linux_distribution()`).

### Secure Google Analytics metrics
Additionally-additionally, the URLs to deliver Google Analytics metrics are made `https://`, since `sentry_sdk` is using HTTPS so it seems that train has sailed.
Fixes: #3953

cc: @jonoomph 

Fixes: #4194

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -34,7 +34,6 @@ import platform
 import traceback
 import json
 
-from classes import exceptions
 from PyQt5.QtCore import PYQT_VERSION_STR
 from PyQt5.QtCore import QT_VERSION_STR
 from PyQt5.QtCore import pyqtSlot
@@ -84,7 +83,7 @@ class OpenShotApp(QApplication):
 
         try:
             # Import modules
-            from classes import info
+            from classes import info, sentry
             from classes.logger import log, reroute_output
 
             # Log the session's start
@@ -139,7 +138,7 @@ class OpenShotApp(QApplication):
 
         # Check to disable sentry
         if not self.settings.get('send_metrics'):
-            exceptions.disable_sentry_tracing()
+            sentry.disable_tracing()
 
     def show_environment(self, info, openshot):
         log = self.log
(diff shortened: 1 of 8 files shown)


ferdnyc:
Actually... looking at this code, use of Sentry doesn't appear to be properly gated on the `send-metrics` preference. Initializing it in `launch.py` will unconditionally enable transmission of data, with `sentry.disable_tracing()` only called if the checkbox is unchecked when the first tutorial screen is shown. Which means we're back in the situation where data upload is enabled prior to the user being shown the data collection opt-in, no?

ferdnyc:
(The primary impetus behind relocating `sentry_sdk` and making it optional is that the Linux distros are never gonna go for hard-requiring it. But we _might_ get away with bundling it, as long as it's still optional.)

jonoomph:
I wonder if `distro.linux_distribution` can be invoked from Windows? If you take a look at Sentry traces so far from Windows, it appears that this is not working correctly, although I haven't tracked it down yet. Otherwise, this looks great! Nice work! Merge away.

ferdnyc:
> I wonder if `distro.linux_distribution` can be invoked from Windows? If you take a look at Sentry traces so far from Windows, it appears that this is not working correctly, although I haven't tracked it down yet.

Mm, I can change that to
```python3
if distro and platform.system() == "Linux":
    sdk.set_tag("distro", " ".join(distro.linux_distribution()))
```

PR #4195, opened 2021-06-21
Remove Latin translation

ferdnyc:
While I'm sure it's amusing to have a Latin translation of the interface, there were only two strings actually translated, and *it's a dead language*.


ferdnyc:
Ugh, _never mind_. It's up to 8 strings translated now, and is being actively contributed to over @ Launchpad. For some reason!

PR #4208, opened 2021-06-26
Fixing a few sentry.io related issues:

jonoomph:
1) undefined variable
2) accessing null objects
3) matching angular-animate version to angular


src/timeline/media/js/angular-animate.min.js
```diff
@@ -1,57 +1,57 @@
 /*
- AngularJS v1.7.2
- (c) 2010-2018 Google, Inc. http://angularjs.org
+ AngularJS v1.6.0
+ (c) 2010-2016 Google, Inc. http://angularjs.org
```
ferdnyc, line 3: No chance we could move the main AngularJS version _up_ to 1.7.2, instead? (Still wildly out of date, but... 2 years less so.)

Also, if replacing `angular-animate.min.js`, the source map in `angular-animate.min.js.map` should either also be replaced, or it should be deleted, as it's no longer valid for the file it maps.

ferdnyc: My preference would be to store these in files named e.g. `angular-animate-1.6.0.min.js`, and reference them that way in the HTML. Yeah, it means we have to edit the HTML if we change versions, but... we _should_ have to edit the HTML, and possibly the JS, when we change versions! And it makes the version in use more obvious.

src/windows/main_window.py
```diff
@@ -175,8 +175,9 @@ def closeEvent(self, event):
         self.preview_parent.background.wait(5000)
 
         # Close Timeline
-        self.timeline_sync.timeline.Close()
-        self.timeline_sync.timeline = None
+        if self.timeline_sync and self.timeline_sync.timeline:
+            self.timeline_sync.timeline.Close()
+            self.timeline_sync.timeline = None
```
ferdnyc, line 180: A far better solution to this would probably be to make `timeline_sync` a `QObject` subclass, parented to the `OpenShotApp` instance (not the `MainWindow`), and deal with closing the `openshot.Timeline` object in _its_ destructor.

Same thing with the preview thread, actually. (Which at least _is_ a QThread, already.)

Then the preview thread wouldn't have to go through all that nonsense of checking whether `MainWindow` is `.initialized` yet, either.

The only barrier to doing that is that the `VideoPreview` widget won't exist yet, when it's created, so that assignment has to be done after-the-fact via a method. Or a signal.

...Actually, the only thing `preview_thread` (or, really, `preview_thread.worker` ever uses `videoPreview` for is to:
1. Pass it into a call to `self.player.SetQWidget()`
2. Call `self.videoPreview.connectSignals(self.renderer)`
    ...Where `self.renderer` is derived from a call to `self.player.GetRendererQObject()`. That call to `connectSignals` is also the only use of `self.renderer`.

So, seems to me that if the _caller_ passes the `timeline_sync.timeline` object to the `VideoWidget`, then it can do all of that its damn self and doesn't need to go through `preview_thread` to do it. The preview thread could conceivably have no relationship whatsoever to the `VideoWidget`; it's not like they actually interact directly anywhere in the code.

It **might** affect the thread membership of some of the signals, but really I don't think so because `VideoWidget` is already on the main thread, and `timeline_sync` is already on the main thread, so I'm thinking the preview thread isn't really as isolated as we think it is. It may really just be a thread that continuously interrupts the main thread in a tight loop.

I have a feeling that's why those `QCoreApplication.processEvents()` calls are necessary in the preview thread's inner loop, because it's never really escaping thread affinity with the main thread anyway.

ferdnyc: Wait, sorry, no, some of that is wrong. `self.player` isn't the `timeline_sync.timeline`, it's the `openshot.QtPlayer()`. That **does** soooorta live on the preview thread, maybe, though there's still the wrinkle that it's interacting entirely with both `timeline_sync.timeline` (which is set as its `Reader()`) and with `VideoPreview`, both of which live entirely on the main thread.

I mean, that's why the tight loop in the `preview_thread.worker` doesn't actually **do** anything with the preview frames — it just monitors `self.player.Position()` and updates when there's a change. If the `QtPlayer` (or, really, the `VideoRenderer`) emitted a signal for each frame it shows (separate from the `present` signal that _contains_ the frame image), then the preview_thread would have precious little to do.

I've often though that a `void frame_shown(int number);` signal coming from the player object would address a lot of issues, particularly with all the calls constantly being made to `get_app().window.preview_thread.player.Position()` in the GUI code.

src/timeline/media/js/angular-animate.min.js
jonoomph: We have some incompatibilities with newer versions of Angular, which is why this got downgraded in the first place. I think the next step will be to rewrite the timeline using non-web technologies... so this is a temporary solution.

src/windows/main_window.py
jonoomph: Noted, appreciate the feedback. I'm going to merge this for the moment, but will continue to dig into your comments. :+1:

PR #4210, opened 2021-06-30
added guards where we were getting exceptions

JacksonRG:
Sentry was showing `keyerror: 'Points'` exceptions.

because I'm still testing to reproduce this exception, and see if this prevents it.

### DIFF
--- a/src/windows/models/properties_model.py
+++ b/src/windows/models/properties_model.py
@@ -301,7 +301,7 @@ class PropertiesModel(updates.UpdateInterface):
                         # Keyframe
                         # Loop through points, find a matching points on this frame
                         found_point = False
-                        for point in clip_data[property_key][color].get('Points', []):
+                        for point in clip_data[property_key][color].get("Points", []):
                             log.debug("looping points: co.X = %s" % point["co"]["X"])
                             if interpolation == -1 and point["co"]["X"] == self.frame_number:
                                 # Found point, Update value
@@ -348,7 +348,7 @@ class PropertiesModel(updates.UpdateInterface):
                         if not found_point:
                             clip_updated = True
                             log.debug("Created new point at X=%d", self.frame_number)
-                            clip_data[property_key][color].setdefault('Points', []).append({
+                            clip_data[property_key][color].setdefault("Points", []).append({
                                 'co': {'X': self.frame_number, 'Y': new_value},
                                 'interpolation': 1,
                                 })
--- a/src/windows/models/properties_model.py
+++ b/src/windows/models/properties_model.py
@@ -195,11 +195,11 @@ class PropertiesModel(updates.UpdateInterface):
         if not c:
             return
 
-        # Create reference 
+        # Create reference
         clip_data = c.data
         if object_id:
             clip_data = c.data.get('objects').get(object_id)
-    
+
         if property_key in clip_data:  # Update clip attribute
             log_id = "{}/{}".format(clip_id, object_id) if object_id else clip_id
             log.debug("%s: remove %s keyframe. %s", log_id, property_key, clip_data.get(property_key))
@@ -281,7 +281,7 @@ class PropertiesModel(updates.UpdateInterface):
                 c = Effect.get(id=clip_id)
 
             if c:
-                # Create reference 
+                # Create reference
                 clip_data = c.data
                 if object_id:
                     clip_data = c.data.get('objects').get(object_id)
@@ -357,7 +357,7 @@ class PropertiesModel(updates.UpdateInterface):
                 clip_data = {property_key: clip_data[property_key]}
                 if object_id:
                     clip_data = {'objects': {object_id: clip_data}}
-                
+
                 # Save changes
                 if clip_updated:
                     # Save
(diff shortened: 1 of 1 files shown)


src/windows/models/properties_model.py
```diff
@@ -447,51 +447,52 @@ def value_updated(self, item, interpolation=-1, value=None, interpolation_detail
                     # Loop through points, find a matching points on this frame
                     found_point = False
                     point_to_delete = None
-                    for point in clip_data[property_key]["Points"]:
```
ferdnyc, line 450: I'm on my phone and can't do proper suggestions right now, but this should just be changed to:
```python
for point in clip_data[property_key].get("Points", []):
```
That will protect it without adding an additional level of nesting. If `"Points"` isn't a key, an empty list will be returned and the loop skipped.

ferdnyc: Ugh. And as the comment I just LOST was saying: neither your change or mine will be safe if there's any possibility that `clip_data[property_key]` isn't a `dict` AT ALL. If that's a possibility, we probably need more sanity-checking earlier in the code.

jonoomph: The error on Sentry is listed as `KeyError 'Points'`. So I'm thinking this is a `dict`, but just does not always contain a Points keys. At least, that is what it appears so far.

JacksonRG: @jonoomph That makes sense, so get("Points", []) would prevent that on this line.

But on line 506, we append to 'Points' and that will be a problem if 'Points' isn't an attribute of `clip_data[property_key]`

[Here](https://stackoverflow.com/questions/2052111/efficient-way-to-either-create-a-list-or-append-to-it-if-one-already-exists) I found this solution:
```python
clip_data[property_key].setdefault('Points', []).append(<point>)
```
I think it's pretty robust, because it will create, or append to an existing list.

Together, these let us deal with clips having no points.

ferdnyc: Yup, that should work nicely.

ferdnyc: If we're seeing property entries that _are_ dicts, but lack `'Points'` keys, then that probably means users are deleting the last keyframe point on their properties. IIRC that could cause this. (It certainly causes _other_ problems!)

There's code in my keyframe-processing rewrite (#3840) that refused to do that even if the user asked to. But as I suspect that PR is too diverged from `develop` to ever be mergeable, the last-keyframe protection should probably get ported to the existing code.

ferdnyc: > There's code in my keyframe-processing rewrite (#3840) that refused to do that even if the user asked to. But as I suspect that PR is too diverged from `develop` to ever be mergeable

Actually, maybe not! I just finished rebasing it on `develop`, included fixes for the issues here, and even found a solution for the data-optimization bug that crept in with the object_id changes a while back. (A fix which I can also apply to `develop`. I'll create a PR now for that change.)

ferdnyc: Done, in #4213. Any and _all_ review welcome. It's **immensely** pared down from #3840, but still a lot to review. The only thing I can suggest is to review it commit-by-commit, instead of all the changes at once. I broke it into a bunch of commits, because each one on its own is pretty self-contained and readable (relatively), even if the full diff is a nightmare. (Primarily because of a bunch of un-indenting I did, thanks to some newly-added early returns.)

JacksonRG: Taking a look. And thanks for breaking it up!

src/windows/models/properties_model.py
```diff
@@ -301,7 +301,7 @@ def color_update(self, item, new_color, interpolation=-1, interpolation_details=
                         # Keyframe
                         # Loop through points, find a matching points on this frame
                         found_point = False
-                        for point in clip_data[property_key][color].get('Points', []):
+                        for point in clip_data[property_key][color].get("Points", []):
```
ferdnyc: :confused: 

Bit of an odd commit — there's nothing _wrong_ with it, per se, but single and double quotes are 100% interchangeable in Python... unlike other languages, there are zero implicit differences between the two. Ditto triple-quoted strings, which can use either `'''` or `"""` as long as they match on both ends.

So, the code wasn't really any less "_safe_" with the single quotes. (Though, as I said, nothing wrong with changing them either, so I'm fine either way.)

ferdnyc: Or... maybe this was supposed to be `clip_data[property_key].get(color, {}).get("Points", [])`?

jonoomph:
@JacksonRG Is this PR still needed? Or has @ferdnyc addressed these in another PR?  Just double checking. Thx!

ferdnyc:
@jonoomph It's definitely worth merging, I'll do that now. My other PR may be partly redundant with it, but that's fine.

ferdnyc:
Oh, heh, this was only targeting `sentry_fixes` anyway.

...Ohhh, THAAAAT'S why Codacy kept failing to scan the PR! It doesn't have data for the HEAD of `sentry_fixes` to compare it to.

They could show a better error message, in that case, but I guarantee the issue is a missing target branch commit on their end.

PR #4211, opened 2021-07-01
Don't make paths to other drives relative

JacksonRG:
## Bug:
When saving a project that has files on different windows drives, We get an exception trying to create relative paths.

## Fix

**Needs Testing**

I don't currently have a windows machine or vm to test this on. But we're seeing exceptions being thrown when we try making a relative path to a different drive.

If anyone's willing to check this out and build it.

### DIFF
--- a/src/classes/json_data.py
+++ b/src/classes/json_data.py
@@ -253,8 +253,8 @@ class JsonDataStore:
             # Optimized regex replacement
             data = re.sub(path_regex, self.replace_string_to_absolute, data)
 
-        except Exception as ex:
-            log.error("Error while converting relative paths to absolute paths: %s" % str(ex))
+        except Exception:
+            log.error("Error while converting relative paths to absolute paths", exc_info=1)
 
         return data
 
@@ -266,14 +266,16 @@ class JsonDataStore:
 
         # Determine if thumbnail path is found
         if info.THUMBNAIL_PATH in folder_path:
-            # Convert path to relative thumbnail path
+            log.debug("Generating relative thumbnail path to %s in %s",
+                      file_path, folder_path)
             new_path = os.path.join("thumbnail", file_path).replace("\\", "/")
             new_path = json.dumps(new_path, ensure_ascii=False)
             return '"%s": %s' % (key, new_path)
 
         # Determine if @transitions path is found
         elif os.path.join(info.PATH, "transitions") in folder_path:
-            # Yes, this is an OpenShot transition
+            log.debug("Generating relative @transitions path for %s in %s",
+                      file_path, folder_path)
             folder_path, category_path = os.path.split(folder_path)
 
             # Convert path to @transitions/ path
@@ -283,15 +285,15 @@ class JsonDataStore:
 
         # Determine if @emojis path is found
         elif os.path.join(info.PATH, "emojis") in folder_path:
-            # Yes, this is an OpenShot emoji
-            # Convert path to @emojis/ path
+            log.debug("Generating relative @emojis path for %s in %s",
+                      file_path, folder_path)
             new_path = os.path.join("@emojis", file_path).replace("\\", "/")
             new_path = json.dumps(new_path, ensure_ascii=False)
             return '"%s": %s' % (key, new_path)
 
         # Determine if @assets path is found
         elif path_context["new_project_assets"] in folder_path:
-            # Yes, this is an OpenShot transitions
+            log.debug("Replacing path to %s in %s", file_path, folder_path)
             folder_path = folder_path.replace(path_context["new_project_assets"], "@assets")
 
             # Convert path to @assets/ path
@@ -305,16 +307,19 @@ class JsonDataStore:
             orig_abs_path = os.path.abspath(path)
 
             # Determine windows drives that the project and file are on
-            project_win_drive = os.path.splitdrive(get_app().project.current_filepath)[0]
+            project_win_drive = os.path.splitdrive(path_context.get("new_project_folder", ""))[0]
             file_win_drive = os.path.splitdrive(path)[0]
             if file_win_drive != project_win_drive:
+                log.debug("Drive mismatch, not making path relative: %s", orig_abs_path)
                 # If the file is on different drive. Don't abbreviate the path.
-                return orig_abs_path
+                clean_path = orig_abs_path.replace("\\", "/")
+                clean_path = json.dumps(clean_path, ensure_ascii=False)
+                return f"{key}: {clean_path}"
 
             # Remove file from abs path
             orig_abs_folder = os.path.dirname(orig_abs_path)
 
-            # Calculate new relateive path
+            log.debug("Generating new relative path for %s", orig_abs_path)
             new_rel_path_folder = os.path.relpath(orig_abs_folder, path_context.get("new_project_folder", ""))
             new_rel_path = os.path.join(new_rel_path_folder, file_path).replace("\\", "/")
             new_rel_path = json.dumps(new_rel_path, ensure_ascii=False)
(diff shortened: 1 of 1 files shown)


src/classes/json_data.py
```diff
@@ -263,6 +263,11 @@ def replace_string_to_relative(self, match):
         key = match.groups(0)[0]
         path = match.groups(0)[1]
         folder_path, file_path = os.path.split(os.path.abspath(path))
+        project_win_drive = os.path.splitdrive(get_app().project.current_filepath)[0]
+        file_win_drive = os.path.splitdrive(path)[0]
+        if file_win_drive != project_win_drive:
+            # IF the file is on different drive. Don't abbreviate the path.
+            return path
```
ferdnyc, line 270: The logic here seems ok, but it needs to come **_after_** the `else` on line 308. All of the changes before that point still need to happen. 

(Things like replacing the transitions path with `\@transitions`, which is always valid no matter what drive everything is on.)

ferdnyc (changes requested):

src/classes/json_data.py
jonoomph: In theory, this would only be true for files on different drive letters.. and thus, it shouldn't matter before or after the `else` on 308, right?  Regardless, it might be easier logic to follow with this logic after the `else`.

Also, what happens on Linux and Mac with this function call? `os.path.splitdrive`

JacksonRG: `os.splitdrive('/home/user/Videos') ` :arrow_right: `[ '', '/home/user/Videos`]`

JacksonRG: And I'll get testing on how this treats transition paths

ferdnyc: @jonoomph The code was doing an early `return` if the drive letters differed, which would've prevented the stuff before that `else` from executing. But, yeah, on Mac/Linux it's fine, the "drive letter" will always be `""` for all paths.

ferdnyc: (If the project is being saved on a drive that's _not_ the `C:\` drive, the drive letter **will** differ between the transition path (`C:\Program Files\OpenShot Video Editor\transitions\foo.png`) and the project file path (`F:\My Projects\something.osp`), but we still want `C:\Program Files\OpenShot Video Editor\transitions\` to be replaced with `@transitions`.)

JacksonRG: I moved it to inside the else. So it should still handle transitions normally (if it wasn't)

src/classes/json_data.py
```diff
@@ -309,6 +304,13 @@ def replace_string_to_relative(self, match):
             # Convert path to the correct relative path (based on the existing folder)
             orig_abs_path = os.path.abspath(path)
 
+            # Determine windows drives that the project and file are on
+            project_win_drive = os.path.splitdrive(get_app().project.current_filepath)[0]
+            file_win_drive = os.path.splitdrive(path)[0]
+            if file_win_drive != project_win_drive:
+                # If the file is on different drive. Don't abbreviate the path.
+                return orig_abs_path
```
ferdnyc, line 312: I just realized (sorry), we **could** still do the path string cleanup here, i.e.
```suggestion
                return orig_abs_path.replace("\\", "/")
```

ferdnyc: (I've never been super big on the path munging, as it's dangerous and really shouldn't be necessary... but if we're going to do it, the safest place is here, not after the project data is already embedded in a giant JSON `str` object.)

ferdnyc: Oh, duh! Also, this should still be returning a JSON string-ified version of the path. So, actually:
```suggestion
                return f"{key}: {orig_abs_path}".replace("\\", "/")
```
Sorry, I totally spaced on what the actual purpose of this function was.

ferdnyc (changes requested):
Sorry, I noticed _one_ little thing I'd forgotten to consider before. 😉 

But overall this looks good, I'll try to give it a spin on Windows. And since I only _have_ one letter drive on my Windows machine, I'll get to see how it deals with UNC network paths as well. (It should DTRT, those paths are e.g. `\\server\share\path\to\file`, and Python `os.path.splitdrive` turns that into `('\\server\share', 'path\to\file')`.)

src/classes/json_data.py
JacksonRG: > (I've never been super big on the path munging, as it's dangerous and really shouldn't be necessary... but if we're going to do it, the safest place is here, not after the project data is already embedded in a giant JSON `str` object.)

Sorry, 'munging'? Is this because windows's `\`'s are usually an escape character, and would cause problems in reading the json?

JacksonRG: I do see that all the other ends of this function return a key-value, not a plain path. My bad

ferdnyc:
Sorry _that_ took so long. I was still hitting errors, but different ones. So I threw in a whole bunch of `log.debug()` calls (can't hurt anyway), and eventually worked out that the problem was `get_app().project.current_filepath` having a value of `None`, when saving a new project for the first time. So that's fixed, and now files that live at non-relativeable paths (it is _so_ a word!) will finally save with no errors!

ferdnyc (approved):

src/classes/json_data.py
ferdnyc: @JacksonRG 

Yes and no. The backslashes never actually caused any problems for `json.dumps`/`json.loads`; it knows to double them and deals fine. Our own string-manipulating code — which at the time was running over the _entire_ string object containing the **full** project JSON! — would get confused by them, was the issue.

src/classes/json_data.py
```diff
@@ -305,16 +307,19 @@ def replace_string_to_relative(self, match):
             orig_abs_path = os.path.abspath(path)
 
             # Determine windows drives that the project and file are on
-            project_win_drive = os.path.splitdrive(get_app().project.current_filepath)[0]
+            project_win_drive = os.path.splitdrive(path_context.get("new_project_folder", ""))[0]
```
JacksonRG, line 310: I see... Again safer to use get, and a default value

ferdnyc: Well, sort of. I mean, yeah, that's kind of always true in Python, unfortunately. It's one of the language's biggest down sides... but in this case, getting an empty or default value wouldn't have been any better. It just wasn't the right place to look for the information we want, that was already supplied in the context dict.

ferdnyc: (That I'm STILL using `get()` to access that info, when I KNOW it will be set... is an abundance of paranoia.)

ferdnyc: (There's actually a subclass of `dict`, I think in `collections`, called the `DefaultDict`, that can be configured on creation to return a default value for **any** missing key that's queried. Problem is, it always returns the same default, so to create complex hierarchies you have to instantiate a lot of separate, nested `DefaultDict`s. And even then, it's not very compatible with tree-walking tools like `json.dumps` as it creates lots of phantom values -- if not infinite recursions.)

jonoomph:
LGTM! Nice work @JacksonRG . Merging this now.

jonoomph:
Not sure why Codacy is not working on a few of these PRs. I've tried to "trigger" it, but it fails instantly with no feedback.

src/classes/json_data.py
ferdnyc: Better than `DefaultDict`, for the project data — and the reason I keep harping on it, mostly to motivate myself to get back to it — would be to have a JSON Schema for the project file (#3556).

A fully realized schema would mean that we could at least _validate_, and potentially repair, project data both out to and in from the file, meaning that damaged project files wouldn't result in damaged `project.data` dicts. It wouldn't protect us from bugs _in the code_ that corrupt the data while it's live, but it would mean we could be certain that any corrupted dicts have to have resulted **from** bugs in the code, hopefully making them easier to track down and fix.

PR #4212, opened 2021-07-01
Default ignore_ids to empty object

JacksonRG:
Keep this method safe if ignore_ids has no value

### DIFF
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -1092,7 +1092,7 @@ App.controller("TimelineCtrl", function ($scope) {
   };
 
   // Search through clips and transitions to find the closest element within a given threshold
-  $scope.getNearbyPosition = function (pixel_positions, threshold, ignore_ids=[]) {
+  $scope.getNearbyPosition = function (pixel_positions, threshold, ignore_ids={}) {
     // init some vars
     var smallest_diff = 900.0;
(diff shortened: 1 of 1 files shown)


src/timeline/js/controllers.js
```diff
@@ -1092,7 +1092,7 @@ App.controller("TimelineCtrl", function ($scope) {
   };
 
   // Search through clips and transitions to find the closest element within a given threshold
-  $scope.getNearbyPosition = function (pixel_positions, threshold, ignore_ids) {
+  $scope.getNearbyPosition = function (pixel_positions, threshold, ignore_ids=[]) {
```
ferdnyc, line 1095:
```suggestion
  $scope.getNearbyPosition = function (pixel_positions, threshold, ignore_ids={}) {
```
Looking at the only call to this function in the codebase (`functions.js` line ~313, IIRC?), when it IS passed in `$ignore_ids` will be an object, not a list. (And on line 1129 here, `.hasOwnProperty()` is used on it, reinforcing that.)

An empty list might work as well (lists are objects, right?) but might as well keep consistent.

ferdnyc (changes requested):

src/timeline/js/controllers.js
jonoomph: +1 on this suggestion

ferdnyc (approved):

PR #4213, opened 2021-07-04
Restore optimized saving of property updates

ferdnyc:
This PR borrows several fixes from my PR #3840 rewrite of the properties-update processing, but in a manner that's less disruptive to the overall code, and hopefully easier to review.

Without _unduly_ destabilizing `properties_model.py`, these changes:
1. Restore the optimized saving of history entries for Clip and Effect property updates, broken since the merge of #4058 
2. Fixes a longstanding exposure in `Effect.filter()`, where live references to the `project_data` members were being returned, instead of copied objects
3. Fixes the even-longer-standing bug where changing the `interpolation_type` of a keyframe point would _also_ incorrectly change the interpolation of the **previous** point
4. Consolidates a bunch of repeated code from `PropertiesModel` into `@staticmethod`s, for re-use

In doing so, I'm pretty sure this **breaks** the handling of TrackedObject properties introduced in #4058, but that handling was already broken in ways we didn't realize, and the correct fix for it will have to happen elsewhere. I'll explain via inline review comments on the relevant bits of source code.


src/classes/query.py
```diff
@@ -356,7 +356,7 @@ def filter(**kwargs):
                             object = Effect()
                             object.id = child["id"]
                             object.key = ["clips", {"id": clip["id"]}, "effects", {"id": object.id}]
-                            object.data = child
+                            object.data = copy.deepcopy(child)
```
ferdnyc, line 359: So, this was the ur-bug: unlike all of the other query classes which inherited their `filter()` method from `QueryObject`, the ad-hoc code in `Effect.filter()` failed to perform a `deepcopy()` of the relevant parts of the project data, instead returning a live reference that allowed the project data to be _directly_ modified. (All such modifications are supposed to go through the update manager.)

src/windows/models/properties_model.py
```diff
@@ -416,163 +419,152 @@ def value_updated(self, item, interpolation=-1, value=None, interpolation_detail
...
-        elif item_type == "effect":
-            # Get effect object
-            c = Effect.get(id=clip_id)
-
-        if c:
-            
-            # Create reference 
-            clip_data = c.data
-            if object_id:
-                clip_data = c.data.get('objects').get(object_id)
```
ferdnyc, line 437: Because of the `Effect.filter()` bug, whenever an effect's properties were being modified the reference made here (whether to the `Effect` object `data` member or the `data["objects"][object_id]` value it contained) allowed direct modification of the relevant `project.data['clips'][clip_id_pos]['effects'][effect_id_pos]` dict that held the property being changed.

src/windows/models/properties_model.py
```diff
@@ -416,163 +419,152 @@ def value_updated(self, item, interpolation=-1, value=None, interpolation_detail
...
-
-            # Reduce # of clip properties we are saving (performance boost)
-            clip_data = {property_key: clip_data.get(property_key)}
-            if object_id:
-                clip_data = {'objects': {object_id: clip_data}}
-
-            # Save changes
-            if clip_updated:
-                # Save
-                c.save()
```
ferdnyc, line 567: _Fortunately_ (in a sense), the reassignment of `clip_data` here broke that reference, and created a new temporary object instead of modifying `c.data` to only contain the changed property. The `c` object therefore still contained the entire property set, without optimization. Then, `c.save()` sent that full, modified property list to the update manager.

ferdnyc (commented):
With the query bug fixed, changes to effect properties will no longer bypass the update system, which means effect property changes will be forced to go through it — and turns out, modifying deep contents of the effect properties this way:
```python
if object_id:
    clip_data = c.data['objects'][object_id]
# more code...
clip_data = {'objects': {object_id: clip_data}}
c.save()
```
Isn't the right way to go about it. To modify the properties of _objects_-inside-effects-inside-clips, we need to go about it the same way we do for effects-inside-clips: With a new `QueryObject` subclass.

*`Clip` object changes go to the update manager with a `key` of `['clips', {'id': clip_id}]`
* `Effect` object changes have a key of `['clips', {'id': clip_id}, 'effects', {'id': effect_id}]`
* `TrackedObject` changes will need a key of `['clips', {'id': clip_id}, 'effects', {'id': effect_id}, 'objects', {'id': object_id}]`

A new `QueryObject` subclass, one that has the ability to create the deeper, three-level key, is the only way the optimized property changes are going to be correctly handled by the update manager. (And they won't need any special-casing to do it!)

There may be some more nuances to this that I have to pin down. (For instance, I assume many of the property changes for object tracking will be made _automatically_ by the tracking code. Presumably those **shouldn't** go through the update manager, they should be treated like interpolated values that don't set keyframes or create undo history entries. So I have to figure out how that fits into all of this.) But this is the basic idea; the code here so far un-breaks (and re-optimizes) regular Clip/Effect updates, but doesn't tackle repairing the `object_id` case yet.

src/windows/models/properties_model.py
```diff
@@ -305,109 +305,98 @@ def color_update(self, item, new_color, interpolation=-1, interpolation_details=
...
+                    log.debug(
+                        "updating interpolation mode point: co.X = %d to %d",
+                        point["co"]["X"], interpolation)
+                    log.debug("use interpolation preset: %s", str(interpolation_details))
+
+            # Create new point (if needed)
+            if not found_point:
+                clip_updated = True
+                log.debug("Created new point at X=%d", self.frame_number)
+                d[property_key].setdefault(color, {}).setdefault("Points", []).append({
```
JacksonRG, line 354: Looks like as well as cleaning up, you addressed the safety issue my PR was aimed at :+1:

ferdnyc: Yeah, along the way I kind of did incorporate all of the excellent fixes you'd already made. I was already in there, and they were fresh in my mind! :laughing:

src/windows/models/properties_model.py
```diff
@@ -166,92 +155,113 @@ def update_frame(self, frame_number, reload_model=True):
...
-            c = Clip.get(id=clip_id)
-        elif item_type == "transition":
-            # Get transition object
-            c = Transition.get(id=clip_id)
-        elif item_type == "effect":
-            # Get effect object
-            c = Effect.get(id=clip_id)
-
-        if not c:
+        c = c = PropertiesModel.get_query_object(item_id, item_type)
```
jonoomph, line 221: `c = c =` ??

jonoomph:
LGTM Otherwise

jonoomph:
I'm good with merging this, and let's see what fallout we get from the Object Detection effect. Like you said, we can fix those in another PR.

src/windows/models/properties_model.py
ferdnyc: Ick, that's not supposed to be there!

ferdnyc: Fixed, and I couldn't resist sneaking in one more `@staticmethod` helper while I was at it.

JacksonRG:
So, I pulled this an played with it for a bit. I just tried setting the scale of a video, and couldn't change it by dragging left/right, or right clicking to see the menu of options.

I saw this on all the properties with lists of options, like gravity. Couldn't change them in the properties menu.

ferdnyc:
@JacksonRG 

Hmm, thanks, I'll take a look.

Have you tried running OpenShot from the command line with `--debug`? There are several debug statements in the code that _should_ be displayed even if things go right, and when anything goes **wrong** I'd expect either tracebacks or warning logs would be produced.

ferdnyc:
Hmm, my latest commit fixed _that_, but something still feels... a bit off. I think it's pre-existing, though.

Here's an interesting philosophical question: Say a 30FPS Clip has a Start time of 0.5, and you make an adjustment to a property at its first frame. What should happen?

1. The value at x=1 is updated
2. A new keyframe point is added at x=16

And now, for that same clip, say you make an adjustment with the playhead positioned **before** its first frame. Now  what should happen?

1. The value at x=1 is updated
2. A keyframe point at x=16 is created/updated
3. Something else

And finally, how many keyframe points is that property now considered to have?

1. One
2. Two

I can tell you from just trying it that the answer to the first question is choice 2, and the answer to the second question is also choice 2, which is at least consistent. But the answer to the _third_ question is ALSO choice 2, which ironically feels **inconsistent**.

Because, when you have a Clip with a non-zero start position, none of the keyframe values **prior** to that start position are really accessible to you at all. There's no way to modify them, because anything you do will affect X=16, not X=1.

That includes deleting points.
* If you have an X=1 clip and you create another keyframe point, you can still then delete the X=1 point, and that other point you created will govern the value for the entire clip in both directions.
* But with a clip that starts from X=16, creating a new keyframe point X=50 will set a value there, and then attempting to _delete_ at X=16 will delete X=50 instead of X=1, because you're not positioned at X=1 (you can't be) and that's the way the rules work. When you're not positioned on a keyframe point anything you do affects the next-closest point _past_ your position.

So, effectively, while you can "move" the sole keyframe location for a Start=0 (X=1) clip to any frame you choose, by adding a new point and then deleting X=1, it's impossible to do that for a Start=N clip (where N is anything greater than 0.0) because that inaccessible, unmodifiable X=1 keyframe point will always be there.

(Of course, you _could_ still modify it by temporarily extending the clip's left edge to Start=0, making changes to or even deleting X=1, and then setting the start position back to its original value. But that's really just a further inconsistency.)

...Still, I didn't create that situation, and how it works now is no different from before — with one exception, namely that the code will no longer allow you to break your properties by deleting X=1 when it's the only point left.

src/windows/models/properties_model.py
```diff
@@ -445,163 +445,151 @@ def value_updated(self, item, interpolation=-1, value=None, interpolation_detail
...
-                                point["co"]["X"], interpolation)
-                            log.debug("use interpolation preset: %s", str(interpolation_details))
-
-                    # Delete point (if needed)
-                    if point_to_delete:
-                        clip_updated = True
-                        log.debug("Found point to delete at X=%s" % point_to_delete["co"]["X"])
-                        clip_data[property_key]["Points"].remove(point_to_delete)
+        # Update clip attribute
+        log_id = "{}/{}".format(clip_id, object_id) if object_id else clip_id
```
JacksonRG, line 462: Probably just not understanding. But I can't see what this `if` is doing.

BrennoCaldato:
@ferdnyc I tested this PR and as you pointed in the first message I am facing problems when updating the keyframes of the Tracker and ObjectDetection effect

Here is an example: I dragged a clip to the timeline and applied the Tracker effect. Then I changed the delta_x and delta_y keyframes. The first displacement works but when you try to change the delta_y it raises an error.
  
    INFO properties_model: delta_y for ILBYRDMRRG changed to 0.10000000149011612 at frame 1 with interpolation: -1 at closest x: 1
    INFO project_data: _set key: ['clips', {'id': 'WPF9VXRKF2'}, 'effects', {'id': 'ILBYRDMRRG'}] values: {'objects': {'0': {'delta_y': {'Points': [{'co': {'X': 1, 'Y': 0.10000000149011612}, 'interpolation': 2}]}}}} add: False partial: False remove: False
    INFO main_window: updateStatusChanged
    INFO preview_thread: player Position(): 1
    INFO properties_model: Item ILBYRDMRRG: changed delta_y to 0.10000000149011612 at frame 1 (x: 1)
    INFO properties_model: delta_x for ILBYRDMRRG changed to 0.20000000298023224 at frame 1 with interpolation: -1 at closest x: 1
    ERROR properties_model: effect ILBYRDMRRG: can't update float property delta_x to None


@jonoomph Here is the JSON for the Tracker effect. Please notice the new tags ```objects ``` and ``` 0 ``` that contain the keyframes of the first tracked object. For the ObjectDetection effect, there will be more indexes inside ``` objects ```
[json_out.txt](https://github.com/OpenShot/openshot-qt/files/6835186/json_out.txt)

ferdnyc:
@BrennoCaldato 

Yeah, as I mentioned I already know (or, I _think_ I know) how to fix this, it just requires a new class in `classes.query`. I just need a bit of guidance on how it's _supposed_ to work.

So, if I'm understanding correctly, the effect will have a property `objects`, which will be... what, exactly? A list of JSON objects (or Python `dict`s, when it's loaded into memory)? And will each of those contain a set of Clip objects, including static properties like ID and reader info? Or will it just be a set of "trackable" property values?

Once I know the structure I'm targeting, I can whip up a `TrackedObject` subclass of `QueryObject` (hmm, unfortunate name clash there) in just a few minutes, really.

ferdnyc:
Hmm... OK, that JSON file gives me _most_ of what I need... I think.

I guess my only question, would all of these parameters be user-adjustable? (I mean, not the non-keyframable ones obviously,) If everything here is meant to have a user interface control associated with it... well, actually, come to think of it I guess that doesn't matter. Whatever you provide a UI for will be controllable through the class, anything you don't, won't. I really don't have to worry about those details.

ferdnyc:
OK, so what I'm going to do, pretty much as I originally proposed, is create a `TrackedObject` subclass of `QueryObject`. And whereas the `Clip` class has an update key of `['clips', '$ID']` and effects have a key of `['clips', '$CLIP_ID', 'effects', '$EFFECT_ID']`, this new class will have a key of `['clips', '$CLIP_ID', 'effects', '$EFFECT_ID', 'objects', '$OBJECT_ID']` (where `$OBJECT_ID` isn't a unique ID like the others, it's just a numeral that's used as its dictionary key).

Then, when parameters are adjusted, the values sent to the update manager can be trimmed to just what's under that key, in the project file, same as we do for clip, effect, transition, etc. parameters.

ferdnyc:
@BrennoCaldato 

Hmmm. One snag... the `QueryObject` class automatically uses the `QueryObject.id` as part of the object key, and sort of implicitly expects those `id` members to represent unique identifiers usable to look up the values in question.

Would it be a huge disruption to your design, if instead of using simple numeral identifiers, you were to assign a `generated_id()` to each member of the `objects` dict? It would allow the object's data to be retrieved with code like this:
```python3
my_obj = TrackedObject.get(id=object_unique_id)
log.debug("Updating properties for object %s", my_obj.id)
```
...which is sort of our norm for such things. Right now, because the IDs collide, I'd have to index the TrackedObjects by the _effect_ ID they're attached to, which could be a bit messy. (Those still aren't unique across all objects, when there are multiple objects on a single effect. Though, the combination of `effect.id` and `tracked_obj.number` _is_ unique, so I could make it work if I had to.)

ferdnyc:
@BrennoCaldato

(You can still number the objects _as well_, of course. In fact if you prefer you can still key them by numerals, and just stick an `id` member somewhere in the **inner** dict. I know you already have `child_clip_id`, but that carries the same uniqueness problem.)

ferdnyc:
Eh, you know what? I'll just make the unique IDs something like `$EFFECTID_$num` (so for your example JSON it would be `ILBYRDMRRG_0`) — the lookup methods can parse that out into components easily enough, and `_` isn't a valid ID character anyway so there's no risk of misinterpretation.

ferdnyc:
Ooh, in fact, this is good. I just realized that if I do it that way, because I'm parsing out the ID I can "overload" the lookups.
```python3
# If you do:
effect = Effect.filter(class_name="Tracker")[0]
my_obj = TrackedObject.get(id=f"{effect.id}_0")
# you'll get that object's subdict.

# But if you do:
effect = Effect.filter(class_name="Tracker")[0]
my_obj = TrackedObject.filter(id=effect.id)
# you'll get _all_ of the tracked objects for that effect.
```

BrennoCaldato:
I think I got it. It wouldn't be necessary to change the tracked object Key in the ``` objects ``` dictionary, right? And the modifications will be just on the openshot-qt side. Please let me know when you commit these changes so I can give it a try

Issue #2440, opened 2018-12-10
Openshot_qt does not match the FreeDesktop directory specs


ThePreviousOne:
it seems that Openshot places its files in$HOME/.openshot_qt
directory which doesn't match FreeDesktop directories specs :

The default for $XDG_CONFIG_HOME is $HOME/.config, the default for
$XDG_DATA_HOME is $HOME/.local/share. So all applications should look for
those environment variables and use those default values if the variables
are not set.

This causes a lot of clutter when applications do this a makes backups harder
when files are not in the right place

See [here](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html)

TuxVinyards:
Appimage 2.5.0

This link creator is a great feature >> kudos.

$HOME/.local/share is fine but this categories line is needed or the desktop link ends up classified as 'Other' and it doesn't get placed in 'sound and video' section:

Categories=AudioVideo;AudioVideoEditing;  

Also, a request:  I downloaded the image, not the .deb, deliberately so I didn't install.  Try: "Would you like to add a desktop menu link for this appimage y/n/later " on the popup.    Not   "Do you want to install bla bla bla "  I found myself auto-clicking 'no' the first time round ....

ferdnyc:
@TuxVinyards I've migrated your comment to a new issue #3222, because it has nothing to do with the issue @ThePreviousOne opened here. Unless something is directly related to an existing issue, it should be filed as its own separate issue. That way, we can keep track of different things as we discuss/investigate/fix them.

ThePreviousOne:
I was filing this issue against another Application and realized that link was wrong. Fixed it

anestv:
The issue still exists in the latest versions. In the file `src/classes/info.py` , the specific lines are:

> HOME_PATH = os.path.join(os.path.expanduser("~"))
> USER_PATH = os.path.join(HOME_PATH, ".openshot_qt")

We can change them, to check if a directory "~/.openshot_qt" already exists, then to use that one as USER_PATH for backwards compatibility; and otherwise to create and use "$XDG_DATA_HOME/openshot_qt" according to the FreeDesktop directories specs (if $XDG_DATA_HOME is not set it should be $HOME/.local/share ).

By the way, the CACHE_PATH should also be placed inside $XDG_CACHE_HOME (default $HOME/.cache)

I can make a PR to demonstrate the needed changes

ferdnyc:
[...]

PR #4216, opened 2021-07-06
Fix for Issue #2440

ThePreviousOne:
This attempts to fix the Issue raised in #2440.
I couldn't properly test it against the develop branch as it seems to be broken currently but I was able to test it successfully against master and everything seemed fine.

This was I a bit of a lazy attempt I will admit, I didn't split data and cache and config into the seperate directories, I just moved everthing out of the home directory into QStandardPaths.DataLocation for now. Still I figured this was a start and deciding where do put specific files was the real hard part (assuming we even real want to do that).

I renamed USER_PATH to APP_PATH because it was a little confusing and because now QStandardPaths should only need to be used in info.py leaving everything else relatively untouched, save for the variable rename.


JacksonRG:
This code looks good. I'm just gonna pull the code, test that it moves my `.openshot-qt` correctly.

src/classes/info.py
```diff
@@ -47,30 +50,38 @@
...
-USER_PRESETS_PATH = os.path.join(USER_PATH, "presets")
-USER_TITLES_PATH = os.path.join(USER_PATH, "title_templates")
-PROTOBUF_DATA_PATH = os.path.join(USER_PATH, "protobuf_data")
-YOLO_PATH = os.path.join(USER_PATH, "yolo")
+OLD_HOME_PATH = os.path.join(os.path.expanduser("~"))
+OLD_USER_PATH = os.path.join(OLD_HOME_PATH, ".openshot_qt")
+
+HOME_PATH = QStandardPaths.writableLocation(QStandardPaths.HomeLocation)
+USER_PATH = QStandardPaths.writableLocation(QStandardPaths.DataLocation)
+APP_PATH  = os.path.join(USER_PATH, NAME)
```
ferdnyc, line 58: At the risk of bikeshedding, rather than `APP_PATH` I'd prefer to call this something like `CONFIG_PATH` or `SETTINGS_PATH`. "`APP_PATH`" sounds too much like the _install_ location — it doesn't clearly indicate "**user** data".

src/classes/info.py
```diff
@@ -47,30 +50,38 @@
...
+
+if (os.path.exists(OLD_USER_PATH)) :
+    shutil.move(OLD_USER_PATH, APP_PATH)
```
ferdnyc, line 61: We should copy, not move, to preserve backwards compatibility. I'd rather leave extra files lying around than break people's installs if they switch back to an older version.

Oh! And also because existing user projects may contain pathnames that point into the old directory location. (For situations like Windows projects that were saved to a different drive, etc... We can't rely on the assumption that all paths follow our imperfect conventions, because they are imperfect.)

ferdnyc: (So, I'd prefer something more like...)
```suggestion
sentry_file = os.path.join(OLD_USER_PATH, ".settingsMigrated")
if (os.path.exists(OLD_USER_PATH) and not os.path.exists(sentry_file)):
    try:
        # Make sure we have write permission, first
        os.makedirs(APP_PATH, exist_ok=True)
        shutil.copytree(
            OLD_USER_PATH,
            APP_PATH,
            dirs_exist_ok=True,
            ignore_dangling_symlinks=True)
        os.mknod(sentry_file)
    except (PermissionError, FileNotFoundError) as ex:
        ERROR_MOVING_CONFIGS = ex
```
...And we'll deal with showing a message or something, when the gui launches.

src/classes/info.py
ferdnyc: Also, _currently_ — for reasons I'm not entirely clear on — the application name is being set to `openshot` (not `openshot-qt` _or_ `OpenShot Video Editor`) in `launch.py`. Which means things like `QStandardPaths.AppDataLocation` and `QStandardPaths.AppConfigLocation` will be `<DataLocation>/openshot/`, and `<ConfigLocation>/openshot/` once the QApplication instance is set up. To avoid possibly shooting ourselves in the foot for later uses of QStandardPaths, we should probably use compatible pathnames. (Whether that means changing `launch.py` or `classes.info`, I could go either way.)

ferdnyc (commented):
This looks like a good start, thanks @ThePreviousOne ! I started down this road once before, myself, but never followed through.

If we're going to do this, I'd really like to go **farther**, and do away with this dumping everything into one directory thing that we've held on to for far too long.

`QStandardPaths.DataDir` on Linux is `$HOME/.local/share/`, but that's not really a suitable path for everything:
* Config files, like the settings and the default project file, should really be in `QStandardPaths.AppConfigDir`, which resolves to `$HOME/.config/<app name>` on Linux. (I think it might be the same as `QStandardPaths.AppDataDir`, on Windows.)
* The preview cache, at least, should really be in `QStandardPaths.AppCacheDir` (`$HOME/.cache/<app name>/` on Linux)
* The lock file belongs in `QStandardPaths.RuntimeLocation` (that's `/run/user/<uid>/` on Linux), and it should be named something like `.openshot.lock`
* The log files, and possibly backups, should ideally go into `QStandardPaths.AppLocalDataDir`, which on Linux is the same as `AppDataDir`, but on Windows will point to `AppDir/Local/<appname>` instead of `AppDir/Roaming/<appname>` for roaming profiles.

src/classes/info.py
ferdnyc: <s>In fact... currently, the only reasons `classes.info` is loaded before the application instance is set up are:

1. To test loading of the application modules, or adjust `sys.path` if it fails. The `sys.path` manipulation is terrible and we need to get rid of it _anyway_, but in the mean time we can certainly use something other than `classes.info` as our module-loading canary.
1. To get the program version from `info.SETUP['version']`. But that's the wrong way to handle that anyway. The version should be stored in the `__version__` attribute of the package. We can move that to `__init__.py` where it belongs.
1. To hold some configuration stuff used by the command line args. But I can move that stuff somewhere else, like a `classes.config` module. It doesn't need to be in `classes.info`.
1. To display `info.PATH` on the console output. That's silly and can be moved.
1. It'll also get loaded when `classes.sentry` is imported, but only so _it_ can load `info.VERSION` and `info.CURRENT_LANGUAGE`. Again, those can be migrated elsewhere and we can remove that dependency.

If we fix those few exposures, we can hold off loading `classes.info` until after we've set the application name, and then we could use paths like `QStandardPaths.AppConfigDir`, `.AppCacheDir`, `.App[Local]DataDir` directly in `classes.info`, instead of having to ad-hoc them. I kind of like that possibility, myself.</s>

...Actually, never mind. I just tried it, and it's perfectly OK to call these _before_ the application instance is created:
```python3
from PyQt5.QtCore import QCoreApplication
QCoreApplication.setApplicationName('something')
QCoreApplication.setApplicationVersion('1.2.3-four5')
```

...So, we can (and probably should) just do that at the top of `classes.info`, instead of where it's located now in `launch.py`, and then use the `App`-specific versions of the `QStandardPaths`. (Or,  we could move the app name setting to the very top of `launch.py`, before it imports `classes.info`. Same effect, and it's already hardcoded so it's not like it needs `classes.info`.)

ferdnyc: (Also confirmed...)
```python3
from PyQt5.QtCore import QCoreApplication, QStandardPaths
QCoreApplication.setApplicationName('openshot')
print(QCoreApplication.writableLocation(QStandardPaths.AppConfigLocation))
# ==> '/home/ferd/.config/openshot/`
```

src/classes/info.py
ferdnyc: (Hmm. Should the sentry file be located in the _old_ config dir, or the _new_ config dir? I'm kind of conflicted on that one.

I think probably the old dir. Then we can tell users, "Delete the file `.openshot_qt/.settingsMigrated` and OpenShot will re-import that configuration on its next launch (overwriting any settings changes in the new location)."

ferdnyc:
Hah! The documentation build workflow failed, because it imports `classes.info` which now imports from `PyQt5`. But that build image never installed PyQt5, it never needed it before.

Just goes to show, you never know what a change might break. :laughing:  I'll fix that.

ferdnyc:
@ThePreviousOne 

I "had" to (not really, but I did) force-push your branch — Sorry! — please make sure to pull before committing any more changes locally, so you won't be out in sync.

ThePreviousOne:
I agree `APP_PATH` isn't really the best name either, and splitting thing up is for the better. As I said before this was a rather lazy attempt. I am happy to finish this properly, this was more or less to get things started.

ferdnyc:
@ThePreviousOne 

Eep! Something seems to have gone awry, it looks like your branch is now _only_ my commit — hopefully that _wasn't_ what you intended to happen! (Did you try a form of `git rebase`, maybe `git rebase --onto`? I've blown away more than my share of commits accidentally, attempting those.)

I should still have your commits checked out in my clone, so I can probably restore them from there if you need.

ThePreviousOne:
@ferdnyc no worries I decided to revert my previous commit and just rewrite it with the changes you suggested. but then I realized that most of that commit was completely unnecessary and just rebased

ThePreviousOne:
this gets the branch back to its previous state, It still feels lazy though.
The problem now is QApplication sets the application name after OStandardPath calls AppDataLocation so AppDataLocation returns "" and everything is dumped into the ~/.local/share/ instead of ~/.local/share/<appname>  , So I still had to use an os.path.join to make it work. Ideally this shouldn't need to be done, thats my next step.

ferdnyc:
@ThePreviousOne 
> this gets the branch back to its previous state, It still feels lazy though.
> The problem now is `QApplication` sets the application name after `OStandardPaths` calls `AppDataLocation` so AppDataLocation returns `""` and everything is dumped into the `~/.local/share/` instead of `~/.local/share/<appname>` , So I still had to use an `os.path.join` to make it work. Ideally this shouldn't need to be done, that's my next step.

I mentioned in one of my review comments, don't know if you saw it — turns  out, it's perfectly OK to do this at the top of `info.py`, even with no QApplication instance yet:
```python3
from PyQt5.QtCore import QCoreApplication, QStandardPaths
QCoreApplication.setApplicationName('openshot')
```
...And then the `QStandardPaths` will all use the application-directory-postfixed versions of their paths (where applicable).

ThePreviousOne:
We can do that but, should I? Wouldn't that mean a second instance of `QApplication` ?

ferdnyc:
For the record, here's the complete list of (writable) standard paths on Windows 10 (where my username is "DP" for reasons that will remain an enduring mystery...), with the application name set:
```python3
>>> from PyQt5.QtCore import QCoreApplication, QStandardPaths
>>> from pprint import pp
>>> import string
>>> pp([(x, QStandardPaths.writableLocation(getattr(QStandardPaths, x)))
...  for x in dir(QStandardPaths) if x.startswith(tuple(list(string.ascii_uppercase)))
...  and isinstance(getattr(QStandardPaths, x), QStandardPaths.StandardLocation)])
[('AppConfigLocation', 'C:/Users/DP/AppData/Local/openshot'),
 ('AppDataLocation', 'C:/Users/DP/AppData/Roaming/openshot'),
 ('AppLocalDataLocation', 'C:/Users/DP/AppData/Local/openshot'),
 ('ApplicationsLocation',
  'C:/Users/DP/AppData/Roaming/Microsoft/Windows/Start Menu/Programs'),
 ('CacheLocation', 'C:/Users/DP/AppData/Local/openshot/cache'),
 ('ConfigLocation', 'C:/Users/DP/AppData/Local/openshot'),
 ('DataLocation', 'C:/Users/DP/AppData/Local/openshot'),
 ('DesktopLocation', 'C:/Users/DP/Desktop'),
 ('DocumentsLocation', 'C:/Users/DP/Documents'),
 ('DownloadLocation', 'C:/Users/DP/Downloads'),
 ('FontsLocation', 'C:/Windows/Fonts'),
 ('GenericCacheLocation', 'C:/Users/DP/AppData/Local/cache'),
 ('GenericConfigLocation', 'C:/Users/DP/AppData/Local'),
 ('GenericDataLocation', 'C:/Users/DP/AppData/Local'),
 ('HomeLocation', 'C:/Users/DP'),
 ('MoviesLocation', 'C:/Users/DP/Videos'),
 ('MusicLocation', 'C:/Users/DP/Music'),
 ('PicturesLocation', 'C:/Users/DP/Pictures'),
 ('RuntimeLocation', 'C:/Users/DP'),
 ('TempLocation', 'C:/Users/DP/AppData/Local/Temp')]
```
And, same list on Linux (Fedora 33):
```python3
>>> pp([(x, QStandardPaths.writableLocation(getattr(QStandardPaths, x)))
... for x in dir(QStandardPaths) if x.startswith(tuple(list(string.ascii_uppercase)))
... and isinstance(getattr(QStandardPaths, x), QStandardPaths.StandardLocation)])
[('AppConfigLocation', '/home/ferd/.config/openshot'),
 ('AppDataLocation', '/home/ferd/.local/share/openshot'),
 ('AppLocalDataLocation', '/home/ferd/.local/share/openshot'),
 ('ApplicationsLocation', '/home/ferd/.local/share/applications'),
 ('CacheLocation', '/home/ferd/.cache/openshot'),
 ('ConfigLocation', '/home/ferd/.config'),
 ('DataLocation', '/home/ferd/.local/share/openshot'),
 ('DesktopLocation', '/home/ferd/Desktop'),
 ('DocumentsLocation', '/home/ferd/Documents'),
 ('DownloadLocation', '/home/ferd/Downloads'),
 ('FontsLocation', '/home/ferd/.local/share/fonts'),
 ('GenericCacheLocation', '/home/ferd/.cache'),
 ('GenericConfigLocation', '/home/ferd/.config'),
 ('GenericDataLocation', '/home/ferd/.local/share'),
 ('HomeLocation', '/home/ferd'),
 ('MoviesLocation', '/home/ferd/Videos'),
 ('MusicLocation', '/home/ferd/Music'),
 ('PicturesLocation', '/home/ferd/Pictures'),
 ('RuntimeLocation', '/run/user/1000'),
 ('TempLocation', '/tmp')]
```
I think the Windows `RuntimeLocation` surprises me the most, and is making me regretfully reconsider it as the location of the lockfile. :cry:

ferdnyc:
@ThePreviousOne 
> We can do that but, should I? Wouldn't that mean a second instance of `QApplication` ?

Nope — doesn't create an instance! Somehow! It just squirrels the information away somewhere. It's wacky.
```python3
>>> from PyQt5.QtCore import QCoreApplication, QStandardPaths
>>> QCoreApplication.setApplicationName('openshot')
>>> QCoreApplication.instance()
>>> QStandardPaths.writableLocation(QStandardPaths.AppConfigLocation)
'C:/Users/DP/AppData/Local/openshot'
>>> QCoreApplication.instance()
>>>
```

ferdnyc:
(It's _possible_ the application name won't be persistent outside of that module, without a Q*Application instance to persist it... but that's fine, we can set it again after we instantiate `OpenShotApp`.)

ThePreviousOne:
Ok cool, well with that taken care of, all that left is to decide were to put everything.

Also I can wager a guess as to why your Username is DP on windows, if you feel the same way about windows as me at least.

ferdnyc:
> Also I can wager a guess as to why your Username is DP on windows, if you feel the same way about windows as me at least.

Heh. :laughing: Close, it actually stands for "Dirty Pirate", as I feel warmly enough about Windows to never, ever pay for it. (Wow, I should not post that publicly. ...Meh, it's a VM, purely for QA purposes. They can sue me for my no money!)

ThePreviousOne:
> (It's _possible_ the application name won't be persistent outside of that module, without a Q*Application instance to persist it... but that's fine, we can set it again after we instantiate `OpenShotApp`.)

I think I already took care of that, fixed the hard coded name in launch.py

ThePreviousOne:
lol, I use win the free way too,

ThePreviousOne:
> Ok cool, well with that taken care of, all that left is to decide were to put everything.

that is assuming you still want to move away from the monolithic directory structure

ThePreviousOne:
I don't actually like force pushing like this but the change was hardly worth a fresh commit and I don't know how `USER_PATH = os.path.join(QStandardPaths.writableLocation(QStandardPaths.AppDataLocation, NAME)` even managed to work with the closing parentheses on `writableLocation`

ThePreviousOne:
What does each directory actually do? I realize some if not most locations move per project that's fine we just need to set a baseline.

I moved `CACHE_PATH` to `CacheLocation` and put `PREVIEW_CACHE_PATH` inside that. `THUMBNAIL_PATH` could go there too. File Managers store them there why shouldn't we.
For `TRANSLATIONS_PATH`, `EMOJI_PATH`, `PROTOBUF_DATA_PATH`, `USER_TITLES_PATH`, `USER_PRESETS_PATH`, `BACKUP_PATH`, and  `RECOVERY_PATH` I think its fine to leave them in`AppDataLocation`.
Judging by the name `YOLO_PATH` can probably move to `TempLocation`.

As for `BLENDER_PATH` I googled around to see if I could figure out what it was for, but all I got was that its not very straight forward setup, still don't know what its for, animations? then `AppDataLocation` makes sense.
All that leaves is `TITLE_PATH` and `USER_PROFILES_PATH` what are they for?

What do you think?

ThePreviousOne:
Ok so this is broke but I can't seem to figure out how, its something in info I expect but can't seem to consistently reproduce the error, Maybe I'm just tired and missing something simple, but I thought I'd push it and anyway maybe you can see what I can't

ThePreviousOne:
ha oops that commit message was supposed to be 'Break up the Monolith' but thats fitting I guess since its kind broke

ThePreviousOne:
I accidentally forgot some "debug" lines in there, I'll get them next commit

ThePreviousOne:
That was a pain, and an embarrassment, but it works now.

ThePreviousOne:
> @ThePreviousOne
> 
> Eep! Something seems to have gone awry, it looks like your branch is now _only_ my commit — hopefully that _wasn't_ what you intended to happen! (Did you try a form of `git rebase`, maybe `git rebase --onto`? I've blown away more than my share of commits accidentally, attempting those.)
> 
> I should still have your commits checked out in my clone, so I can probably restore them from there if you need.

do you not know about `git reflog` ?

ThePreviousOne:
@JacksonRG That was actually a really lazy implementation. There were a bunch of mistakes following it too, but it properly works now if your interested

jonoomph:
I would love to see a table, with all the current PATHs compared to what this PR is changing them to... on each OS. I'm a bit concerned we are changing a bunch of PATHs, and I don't quite understand the reason behind the change. Thx

ThePreviousOne:
@jonoomph 
By separating PATHs we are following directory conventions that all applications should follow from the start.
Separating Cache and data specifically makes backups (and configs) easier as cache is not something that needs to be preserved.
As for a table its rather simple previously everything was in `$USER_HOME` where-ever that may be on your system
Windows > C:\Users\username\.openshot_qt
Linux       > /home/username/.openshot_qt
on MacOS I actually don't know, but probably /Users/username/.openshot_qt

now caches go to :
Windows  > C:\Users\username\AppData\Local\openshot_qt\cache\<PATH>
Linux        > /home/username/.cache/
MacOS     > /Users/username/Library/Caches/openshot_qt/<PATH>
with the exception of `CACHE_PATH` which is their root folder

and anything considered data (by me) goes to :
Windows > C:\Users\username\AppData\Local\openshot_qt\<PATH>
Linux       > /home/username/.local/share/openshot_qt/<PATH>
MacOS    > /Users/username/Library/Application Support/openshot_qt/<PATH>
with the exception of `DATA_PATH` which is their root folder

and for now as I believe `YOLO_PATH` is Temp I have moved it as such
On my system specifically this means a RAM Drive to avoid wasted writes
https://www.thewindowsclub.com/temporary-files-folder-location-windows I run Linux but this article is pretty universal

Windows > C:/Users/username/AppData/Local/Temp/openshot_qt
Linux       > /tmp/openshot_qt
MacOS    > randomly generated by the OS (so I assume RAM otherwise that's a lot of clutter)

this info all came from https://doc.qt.io/qt-5/qstandardpaths.html and @ferdnyc has confirmed this https://github.com/OpenShot/openshot-qt/pull/4216#issuecomment-876641968

Still though I expect some things to move again as I don't actually know what everything is and there are some things that haven't had there location decided.

ThePreviousOne:
This is "Work in Progress" but I can't make anymore progress without feedback as I don't even know what some of the folders I'm moving are actually for and therefore can't make an appropriate decicision.

ThePreviousOne:
@jonoomph I just noticed I didn't do a great job addressing your query

the locations mentioned above are correct, though I didn't say specifically which went to cache and which to data.
So Caches are going to be THUMBNAIL_PATH, PREVIEW_CACHE_PATH, CACHE_PATH with CACHE_PATH as the Parent Directory
And DATA_PATH, BACKUP_PATH, RECOVERY_PATH, BLENDER_PATH, TITLE_PATH, TRANSITIONS_PATH, _USER_PROFILES_PATH, USER_PRESETS_PATH, USER_TITLES_PATH, EMOJIS_PATH, PROTOBUF_DATA_PATH with DATA_PATH as the Parent Directory

jonoomph:
Thanks for the work on this, but at this time, I am not interesting in changing things in this direction. There is a huge amount of documentation and support for the current directory structure, and current I fear this would lead to massive confusion.

PR #4220, opened 2021-07-08
Fixes and Improvements to ZoomSlider (fixing feedback and jitters)

JacksonRG:
To address the sliding of the zoom slider out from under the mouse, I'm saving the starting position of the zoom slider, and adding the horizontal movement of the mouse (from the position it was pressed)


jonoomph:
LGTM!

PR #4223, opened 2021-07-10
Audio effects

BrennoCaldato:
Adding new audio effects to Openshot

- Compressor: Reduce the volume of loud sounds or amplify quiet sounds.
- Delay: Adjust the synchronism between the audio and video track.
- Distortion: Alter the audio by clipping the signal.
- Echo: Reflection of sound with a delay after the direct sound.
- Expander: Louder parts of audio becomes relatively louder and quieter parts becomes quieter.
- Noise: Random signal having equal intensity at different frequencies.
- ParametricEQ:  Filter that allows you to adjust the volume level of a frequency in the audio track.
- Robotization: Transform the voice present in an audio track into a robotic voice effect.
- Whisperization: Transform the voice present in an audio track into a whispering voice effect.

Icons:
image


jonoomph:
I've updated translation templates with new audio effects, and added updated icons.

jonoomph:
LGTM!

PR #4229, opened 2021-07-12
renamed whirpool whirlpool

### LINKED ISSUE #4227: whirpool vs. whirlpool
Should "whirpool" be spelled "whirlpool"?
In https://github.com/OpenShot/openshot-qt/blob/develop/src/language/OpenShot/OpenShot_transitions.pot#L363-L364 and the four files in https://github.com/OpenShot/openshot-qt/tree/develop/src/transitions/extra

    whirpool_1.jpg
    whirpool_2.jpg
    whirpool_3.jpg
    whirpool_4.jpg

### LINKED ISSUE #4228: barr vs. bar
I am translating OpenShot but i don't understand the word "barr"
Example: https://github.com/OpenShot/openshot-qt/blob/develop/src/language/OpenShot/OpenShot_transitions.pot#L63-L64

Is that a spelling mistake?
Should it be "bar"?

JacksonRG:
# Don't merge until:
- [ ] Openshot handles missing a transition file more gracefully (doesn't crash)

Just a typo fix.

Replacing:
1) 'whirpool' with 'whirlpool' Fixes #4227
1) 'barr' with 'bar' Fixes #4228


src/language/OpenShot/OpenShot_transitions.pot
```diff
@@ -360,8 +360,8 @@ msgstr ""
 msgid "Ray light right %s"
 msgstr ""
 
-#: /home/jonathan/apps/openshot-qt-git/src/transitions/extra/whirpool_3.jpg
-msgid "Whirpool %s"
+#: /home/jonathan/apps/openshot-qt-git/src/transitions/extra/whirlpool_3.jpg
+msgid "whirlpool %s"
```
ferdnyc, line 364: The name should be kept uppercase, in the translation master. (It'll actually be automatically capitalized when the file is regenerated by `src/language/generate_translations.py`, so it won't matter much anyway, but might as well keep it consistent so there's less churn in the translations.)

ferdnyc (commented):
My only concern here is that renaming the file will break existing projects that use that transition, but now won't be able to find it when they load. I wonder if it's worth special-casing a fix?

 I wish we had a really robust system in place for making repairs to project data, on load. There's some ad-hoc code in `classes.project_data`, for the few cases we have had, but it's very one-off and manual. It'd be nice to have a well-defined, modular system: "OK, here's a list of 'plugins' for the loader, and here are the version ranges or other criteria for when each should be applied to incoming project files". That would ensure all of the necessary fixes get applied before the project data is "released" to the application, and let us add and age out fixes as needed.

That would lower the friction, so we could add little fixes like, "change 'whirpool' to 'whirlpool' in any transition paths encountered" just as a matter of course when changes are made that impact user data. (Not that this is _that_ challenging a fix to add anyway.)

ferdnyc:
@JacksonRG 
Ohhhhh, this is... not great. You've led me to discover a _new_ bug that I was previously unaware of. (Not your fault, of course.)

I had added a quick change to `upgrade_data_structures` in `classes.project_data`, to try and make corrections for this. (Because, as I said, it was easy enough.) But in testing, I was getting crashes — _with_ **OR WITHOUT** my fixup code.

Turns out, now that the project file path is passed to libopenshot and parsed _there_ as well, if any of the transition paths can't be found, it throws an `openshot::InvalidFile` exception and crashes the entire application!

So, it's all the more imperative that we do this carefully, because doing it _wrong_ will (it turns out) make older projects completely non-readable. Ugh.

("Doing it carefully", I _think_, means that we have to not only do fixes in `upgrade_data_structures`, but also write the updated data back out to disk so that it can be loaded by libopenshot. That means I should generalize my backup logic from the 2.5.0 Unicode fixes, because I'd really like to make a backup of the user's file before we silently rewrite it on them.)

src/language/OpenShot/OpenShot_transitions.pot
JacksonRG: Ohhhh. I think I had case sensitivity off on my editor. I'll be more careful with that

JacksonRG:
Thanks for pointing that out! I'll make a big not not to merge until we handle missing files better

JacksonRG:
Closing until I verify this is still an issue

PR #4230, opened 2021-07-13
Add default case form move_clips

JacksonRG:
We have a few thousand sentry events from dragging when a given clip isn't in the array move_clips.

This PR should ensure the clip is in move_clips before accessing the array.


JacksonRG:
Just realized I would need to add the same guards to the transitions.

Unless I made a 'timeline item' class to consolidate the duplicated code like the `draggable` attribute.

jessej37:
Looks like when you're updating your newX value you forgot to change + y_offset to + x_offset and this.style.top to this.style.left

jessej37:
Also, this line very well may be correct, but just making sure you meant to place 2 assignments in this line instead of 1 assignment that adds all the following values:
newY = move_clips[transition_name]['top'] = this.style.top + y_offset;

jonoomph:
@JacksonRG This broke our ability to drag multiple clips/transitions at the same time. :thinking:

PR #4238, opened 2021-07-20
Fix race condition on Export dialog, which would create random glitch at beginning of export

jonoomph:
Fix race condition on Export dialog, which would create random glitch at beginning of export, by caling ApplyMapperToClips() after the export had begun... usually introducing a glitch into the 1st 10 frames written to the video file (when exporting to a different framerate). Nasty bug... tricky to find.

### DIFF
--- a/src/windows/export.py
+++ b/src/windows/export.py
@@ -94,14 +94,6 @@ class Export(QDialog):
         self.close_button.setVisible(False)
         self.exporting = False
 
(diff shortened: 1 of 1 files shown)


jonoomph:
@JacksonRG This probably closes some old GitHub issues as well. I can recall many people directly emailing about this issue, now that I understand it better. Usually something about a stutter in the very beginning of a video export.

PR #4256, opened 2021-07-25
Change "View Toolbar" to "Toolbar"

scootergrisen:
Since the menu item is inside the View menu.

### DIFF
--- a/src/windows/ui/main-window.ui
+++ b/src/windows/ui/main-window.ui
@@ -1178,7 +1178,7 @@
     <bool>true</bool>
    </property>
    <property name="text">
-    <string>View Toolbar</string>
+    <string>Toolbar</string>
    </property>
    <property name="toolTip">
     <string>View Toolbar</string>


JacksonRG (approved):
LGTM

jonoomph:
Where is this change visible? In other words, does this text actually show up on the UI? Can we see a before/after screenshot?

ferdnyc:
@jonoomph It's at the top of the view menu. Here's the before:

image

The after would be, it just reads "Toolbar" with a checkmark next to it. Seems correct to me.

(To say  nothing of the redundancy found in e.g. "View > Views > Advanced View", but one thing at a time.)

ferdnyc:
And in fact, the visibility context menu (right-click on any dead area of the main window) does  just label it "Toolbar", as that's its name:

image

We could maybe rename it to "Main Toolbar" (since there _are_ others, though they're created by hand in the code rather than defined in the UI file, so they aren't included in the menus), but then the action would similarly be "Main Toolbar".

ferdnyc:
One alternative approach we _could_ take would be to do away with the explicitly-created `QAction` altogether.

`QToolBar`s, like `QDockWidget`s, have a default `toggleViewAction()` method — it's what I used (on the docks) to construct the `View > Docks` submenu. None of those actions are defined by hand, they're auto-generated by Qt. I just call `docks_menu.addAction(dock.toggleViewAction())` in a loop over all the `dock`s. We could do the same thing for the toolbar, just add `toolbar.toggleViewAction()` to the menu instead of creating a new, redundant one.

...Hmm. I wonder if there's a way to do that directly from the UI file?

ferdnyc:
> ...Hmm. I wonder if there's a way to do that directly from the UI file?

Doesn't appear to be. Not that it's hard to do, but it means moving something from the UI file to the code, which always feels like the wrong direction for things to be going. At least to me.

But, hey, while I was in there poking around at the UI file, I:
1. Removed all of the "Clear" actions for the text filtering fields that we haven't used since we turned on their built-in clear buttons
2. Finally moved all of our custom action icons _out_ of the Humanity/ directory (since they're not part of the Humanity icon set)
3. Updated my Makefile for the resource file and generated a new one, but only after I...
4. Got fed up and made a whole new icon for the Title Editor, because I was sick of that invisible black-on-black generic "font" icon.

So, now Title > Title (oy) will look like this
new-title-editor-icon

PR incoming.

jonoomph:
LGTM, seems simple enough. Thx!

jonoomph:
[PR: Change "View Toolbar" to "Toolbar"](https://trello.com/c/JqyPjpHH/28-pr-change-view-toolbar-to-toolbar)

Issue #4265, opened 2021-07-25
Menu items like "Rotate 90 (Right)" is missing degree "°" sign


scootergrisen:
OpenShot 2.5.1-dev3
Windows 7

Right click a clip and select rotation to see:

    Rotate 90 (Right)
    Rotate 90 (Left)
    Rotate 180 (Flip)

I guess it is not correct to just write the number. 90 what? Degree i presume.
So maybe:

    Rotate 90° (Right)
    Rotate 90° (Left)
    Rotate 180° (Flip)

scootergrisen:
https://github.com/OpenShot/openshot-qt/pull/4266

JacksonRG:
> #4266

Just so you know: If you write something like "Closes: #4265" in your pull request, this issue will automatically close when that PR is merged.

PR #4266, opened 2021-07-25
Add ° sign to Rotate menu items

scootergrisen:


JacksonRG (changes requested):
I really like this idea!

My only concern is if anyone happens to open these files in an encoding other than UTF-8, or if they don't remember the unicode `00b0`.

Something like this:
```python
deg = "\N{DEGREE SIGN}"
string = "Rotate 90" + deg + " (RIGHT)"
```
Should be clearly readable, and repeatable.

scootergrisen:
Closes: https://github.com/OpenShot/openshot-qt/issues/4265

jonoomph:
I'm worried about this one. Not all languages are left to right, and I'm not even sure if all languages use this same symbol to represent a degree sign. This might take further research, to see how to best localize and translate this one. But just concatenating the unicode symbol does not look correct to me.

ferdnyc:
Honestly the first version was safer, as the translatable string would always be extracted as "Rotate 90° (Right)" — it's up to the translator to include the degree sign or not, in their translation, as appropriate for their locale. But the encoding on our translation files is set to UTF-8, so they can handle a degree sign.

(It's not like we're composing these translations from enumerable values, each string is individually translated. But with the concatenated version, `gettext` will _definitely_ extract the string incorrectly when building the `.pot` file.)

ferdnyc:
While we're at it, though, there's also the issue that the Rotate menu doesn't really _rotate_ your clip, it _imposes_ a rotation. IOW, if you've already rotated it to 90° (right/clockwise) and you choose "Rotate > Rotate 90 (Right)", nothing will happen. If you've manually rotated it 45° and then choose "Rotate > Rotate 90 (Right)", it will be set to 90° rotation. So maybe the labels should be "Rotate **to** 90° (Right)", etc.

Oh, and regarding the Unicode — if we're really timid there is, of course, always the option to just spell out the word "degrees". "Rotate to 90 degrees (Right)" is a very translatable string.

jonoomph:
Thanks for the work on this one, but due to issues with translations, I am not interested in this solution at the moment. Sorry and thanks again for the attempt!

PR #4270, opened 2021-07-26
Avoid space after <a> tags

scootergrisen:
In the About OpenShot dialog there is trailing space in <a> links (Learn more and OpenShot on GitHub) so there is a blue line after the link because of the way it is written with space between <a> and <span>.

### DIFF
--- a/src/windows/about.py
+++ b/src/windows/about.py
@@ -141,14 +141,12 @@ class About(QDialog):
         about_html = '''
             <html><head/><body><hr/>
             <p align="center">
-            <span style=" font-size:10pt; font-weight:600;">%s</span>
+            <span style="font-size:10pt; font-weight:600;">%s</span>
             </p>
             <p align="center">
-            <span style=" font-size:10pt;">%s </span>
-            <a href="https://www.openshot.org/%s?r=about-us">
-                <span style=" font-size:10pt; text-decoration: none; color:#55aaff;">%s</span>
-            </a>
-            <span style=" font-size:10pt;">.</span>
+            <span style="font-size:10pt;">%s </span>
+            <a href="https://www.openshot.org/%s?r=about-us"><span style="font-size:10pt; text-decoration: none; color:#55aaff;">%s</span></a>
+            <span style="font-size:10pt;">.</span>
             </p>
             </body></html>
             ''' % (
@@ -160,12 +158,9 @@ class About(QDialog):
             <html><head/>
             <body style="font-size:11pt; font-weight:400; font-style:normal;">
             <hr />
-            <p align="center"
-               style="margin:12px 12px 0 0; -qt-block-indent:0; text-indent:0;">
+            <p align="center" style="margin:12px 12px 0 0; -qt-block-indent:0; text-indent:0;">
                <span style="font-size:10pt; font-weight:600;">%s </span>
-               <a href="http://www.openshotstudios.com?r=about-us">
-               <span style="font-size:10pt; font-weight:600; text-decoration: none; color:#55aaff;">
-               OpenShot Studios, LLC<br /></span></a>
+               <a href="http://www.openshotstudios.com?r=about-us"><span style="font-size:10pt; font-weight:600; text-decoration: none; color:#55aaff;">OpenShot Studios, LLC<br /></span></a>
             </p>
             </body></html>
             ''' % (copyright_text)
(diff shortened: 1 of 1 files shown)


src/windows/about.py
```diff
@@ -160,12 +158,9 @@ def __init__(self):
...
             <body style="font-size:11pt; font-weight:400; font-style:normal;">
             <hr />
-            <p align="center"
-               style="margin:12px 12px 0 0; -qt-block-indent:0; text-indent:0;">
+            <p align="center" style="margin:12px 12px 0 0; -qt-block-indent:0; text-indent:0;">
                <span style="font-size:10pt; font-weight:600;">%s </span>
-               <a href="http://www.openshotstudios.com?r=about-us">
-               <span style="font-size:10pt; font-weight:600; text-decoration: none; color:#55aaff;">
-               OpenShot Studios, LLC<br /></span></a>
+               <a href="http://www.openshotstudios.com?r=about-us"><span style="font-size:10pt; font-weight:600; text-decoration: none; color:#55aaff;">OpenShot Studios, LLC<br /></span></a>
```
JacksonRG, line 163: 176 characters (not including tabs) is a pretty long line. I sure wish I had a monitor that big :laughing:.

but I don't see the need to save lines if it might be less readable for edditing..

src/windows/about.py
```diff
@@ -396,9 +389,7 @@ def __init__(self):
         github_html = '''
             <html><head/><body>
             <p align="center">
-                <a href="https://github.com/OpenShot/">
-                <span style=" text-decoration: underline; color:#55aaff;">%s</span>
-                </a>
+                <a href="https://github.com/OpenShot/"><span style="text-decoration: underline; color:#55aaff;">%s</span></a>
```
JacksonRG, line 392: Again. I don't think we need all this in one line.

But if you'd like to save some space. You could definitely make a css class in one of the `.css` filese for all this repeated style info.

src/windows/about.py
```diff
@@ -141,14 +141,12 @@ def __init__(self):
         about_html = '''
             <html><head/><body><hr/>
             <p align="center">
-            <span style=" font-size:10pt; font-weight:600;">%s</span>
+            <span style="font-size:10pt; font-weight:600;">%s</span>
```
JacksonRG, line 144: :+1: good eye. consistency is nice

JacksonRG (changes requested):
Thanks for the code cleanup. Just a couple little changes, and I'll approve

src/windows/about.py
scootergrisen: Well we can consider this the first stage of improving the code.
There don't really need to be an span and an br tag inside the a tag and the URL redirects to a shorter URL and i suggested using "Website" as link test so i guess in the end it will be:

               <a href="https://www.openshot.org/">Website</a>

src/windows/about.py
scootergrisen: Same af the other link might end up being:

                 <a href="https://github.com/OpenShot/">%s</a>

src/windows/about.py
JacksonRG: the `<br />` is there so that the about window is evenly spaced. But you're right about the web link. Feel free to change that to `https://www.openshot.org/company` as that's the about current about page.

scootergrisen: Better use CSS margin/padding to control space between elements.

jonoomph:
LGTM

src/windows/about.py
ferdnyc: Regardless, the point is, it's HTML — split your lines! There's no reason to write 180 characters wide when this has the _exact_ same meaning:
```suggestion
               <a href="http://www.openshotstudios.com?r=about-us">
                 <span style="font-size:10pt; font-weight:600; text-decoration: none; color:#55aaff;">
                 OpenShot Studios, LLC<br />
                 </span>
               </a>
```

... I am 100% there with you about the `<br />`, though. That is just silliness.

ferdnyc: (And  whyTF is the `<br />` **INSIDE** the LINK!!?!! :angry: )

ferdnyc: (The `<br />`, btw, is the reason for the blue line after the link. Not anything to do with the source formatting. That's exactly the problem with having it _inside_ the `<a>` tags.)

ferdnyc: In fact, even if we want the break the tag nesting should really be
```html
<span><a>Text</a></span><br />
````

ferdnyc:
@scootergrisen 

I'm going to close this, with our thanks, since I ended up solving it a slightly different way in #4497. That PR updates our theme to set the link color across the entire application, and re-works the HTML in the dialog a bit so it's more hierarchical.

But thanks again for pointing it out, I wouldn't have made the time to look at the theme color palette and the dialog HTML if you hadn't!

PR #4308, opened 2021-08-10
Fixing a regression on Export dialog (FrameMapper was sometimes missing)

jonoomph:
Fixing a regression on Export dialog, where we were not correctly applying FrameMapper's to all the clips.. causing audio desync, timing issues with Clips, audio pitch issues. Whoops. :1st_place_medal:


JacksonRG:
Just tested with a video at 48kHz audio.

Without the change, the audio was almost inaudibly low.
With, it was fuzzy, but definitely synchronized. LGTM! :rocket:

Raziel0315:
The new build looks good.  Great job!

PR #4317, opened 2021-08-17
Allow mousewheel zoom on webengine

JacksonRG:
Handle mousewheel zooming in the webview, so that it works the same in either web backend.


jonoomph:
This looks great, other than some Codacy issues with syntax/formatting

PR #4318, opened 2021-08-17
Fix middle click drag, and ruler jitter

JacksonRG:
Issues fixed here:
---
1) Middle click drag was sending the timeline to 00:00:00,0. I made some changes so that it works relative to the timeline's original position.

2) The ruler draws a time on every other mark. As you scroll left, it was always starting with the second mark on screen, breaking up the continuous motion of the ruler. I made sure that the ruler always draws times on the *same* marks, so the scrolling feels continuous.

3) If you resized the timeline in some way, the ruler wasn't redrawing until you scrolled or zoomed. I added `element.width()` to the watched variables that calls `drawRuler`, so that this now causes the ruler to redraw.

### DIFF
--- a/src/timeline/js/directives/ruler.js
+++ b/src/timeline/js/directives/ruler.js
@@ -222,7 +222,7 @@ App.directive("tlRuler", function ($timeout) {
           let tickSpan = $('<span style="left:'+pos+'px;"></span>');
           tickSpan.addClass("tick_mark");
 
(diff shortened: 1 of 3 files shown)


jonoomph:
LGTM, other than Codacy issues. Also, let's update the description on this one, and make sure we give enough context into what has been fixed. :+1:

jonoomph:
Merge away! :pray:  Thx!

PR #4321, opened 2021-08-18
Webkit feature parity

JacksonRG:
1) When the window is resized, parts of the ruler that weren't visible had no ruler marks. Now, changes in `width` attribute the ruler's parent container should trigger redrawing. (Thought that was fixed in my last PR)
2) Horizontal scrolling hasn't worked for me when running with the webkit backend. Now both web backends should respond to `shift+mousewheel` as horizontal scrolling

### DIFF
--- a/src/timeline/js/directives/ruler.js
+++ b/src/timeline/js/directives/ruler.js
@@ -60,6 +60,10 @@ App.directive("tlScrollableTracks", function () {
             timeline.zoomIn();
           }
         }
+        if (e.shiftKey) {
+          let current_scroll = $("#scrolling_tracks").scrollLeft()
+          $("#scrolling_tracks").scrollLeft(current_scroll + e.originalEvent.deltaY);
+        }
       });
 
       // Sync ruler to track scrolling
@@ -146,7 +150,6 @@ App.directive("tlBody", function () {
     }
   };
 });
-
 
 // The HTML5 canvas ruler
 App.directive("tlRuler", function ($timeout) {
--- a/src/timeline/js/directives/ruler.js
+++ b/src/timeline/js/directives/ruler.js
@@ -62,7 +62,8 @@ App.directive("tlScrollableTracks", function () {
           }
         }
         else if (e.shiftKey) {
-          let current_scroll = $("#scrolling_tracks").scrollLeft()
+          e.preventDefault();
+          let current_scroll = $("#scrolling_tracks").scrollLeft();
           $("#scrolling_tracks").scrollLeft(current_scroll + e.originalEvent.deltaY);
         }
       });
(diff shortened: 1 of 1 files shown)


src/timeline/js/directives/ruler.js
```diff
@@ -60,6 +60,10 @@ App.directive("tlScrollableTracks", function () {
             timeline.zoomIn();
           }
         }
+        if (e.shiftKey) {
+          let current_scroll = $("#scrolling_tracks").scrollLeft()
+          $("#scrolling_tracks").scrollLeft(current_scroll + e.originalEvent.deltaY);
+        }
```
ferdnyc:
```suggestion
        if (e.shiftKey) {
          e.preventDefault();
          let current_scroll = $("#scrolling_tracks").scrollLeft()
          $("#scrolling_tracks").scrollLeft(current_scroll + e.originalEvent.deltaY);
        }
```

ferdnyc: Confirmed, this fixes all jerkiness.

ferdnyc:
Oh, nice, and tilt-wheel horizontal scrolling still works as well.

src/timeline/js/directives/ruler.js
ferdnyc: HAHAHA! However, <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+wheel is a bit craaaazy. I think the Ctrl-key response and the Shift-key response may need to be exclusive-or.

ferdnyc: Opened a PR on your PR, re: <kbd>Ctrl</kbd>+<kbd>Shift</kbd> zooming, since the lines were too far away from the changes to use Suggest. See #4492.

src/timeline/js/directives/ruler.js
```diff
@@ -52,14 +52,19 @@ App.directive("tlScrollableTracks", function () {
...
+          if (delta > 0) { // Scroll down: Zoom out
             /*global timeline*/
             timeline.zoomOut();
           } else { // Scroll Up: Zoom in
             /*global timeline*/
             timeline.zoomIn();
           }
         }
+        else if (e.shiftKey) {
+          let current_scroll = $("#scrolling_tracks").scrollLeft()
```
ferdnyc: This still _also_ needs the `preventDefault()`:
```suggestion
          e.preventDefault();
          let current_scroll = $("#scrolling_tracks").scrollLeft();
```

src/timeline/js/directives/ruler.js
ferdnyc: I merged #4492 into this PR, and re-posted my suggested change for the updated code.

src/timeline/js/directives/ruler.js
ferdnyc: I think I'll apply this suggestion and merge, this works perfectly in my testing. It's a nice, focused PR so there's no sense in letting it languish here forever.

PR #4322, opened 2021-08-19
Sentry & Metric Improvements

jonoomph:
Improving Sentry.io error reporting to dynamically set the error rate, based on rates obtained from openshot.org when retrieving the current version info. This helps us adjust this rate as a release matures, and no longer needs to send in as many error reports... but initially, right after a release, we can turn up the dial, and receive more reports. Also, I've optimized the metric tracking to always report the initial launch (only on 1st install), and then opt the user out (and prompt the user to opt back in for all subsequent metric tracking). This is essentially the same as before, except we will get accurate anonymous metrics on how many downloads/installs of OpenShot are happening.

Fixed some issues with the Tutorial dock, which does not always position itself correctly. Also, changed the default simple view of OpenShot, to place the Emojis dock in tabbed mode: Files, Transitions, Effects, **Emojis**. Lastly, fixed the exit of OpenShot to not hang on our thumbnail server and ZMQ polling.


jonoomph:
Merging

src/classes/sentry.py
```diff
@@ -47,12 +48,16 @@ def init_tracing():
...
-        # Dev mode, trace all exceptions
-        traces_sample_rate = 1.0
-        environment = "development"
+    traces_sample_rate = 0.0
+    if info.VERSION == info.ERROR_REPORT_STABLE_VERSION:
+        traces_sample_rate = info.ERROR_REPORT_RATE_STABLE
+        environment = "production"
+    else:
+        traces_sample_rate = info.ERROR_REPORT_RATE_UNSTABLE
+        environment = "unstable"
```
JacksonRG, line 57: If I'm understanding... We will a dial for stable, and unstable? That'll be very useful when we come up to another release

src/settings/_default.settings
```diff
@@ -78,14 +78,14 @@
...
     "title": "",
     "type": "hidden",
     "category": "Qt",
     "setting": "window_state_v2"
   },
   {
-    "value": "AdnQywACAAAAAAGbAAAAcQAABeAAAAOoAAABmwAAAI0AAAXgAAADqAAAAAAAAAAAB4A=",
+    "value": "AdnQywADAAAAAAGbAAAAaAAABeAAAAOoAAABmwAAAI0AAAXgAAADqAAAAAAAAAAACJAAAAGbAAAAjQAABeAAAAOo",
```
JacksonRG, line 88: I'm sure there's a reason for this. I'd just like to know it.

JacksonRG (commented):
Looks good! Just a couple questions for understanding.

src/classes/sentry.py
jonoomph: Exactly, the website will return `{"error_rate_unstable": 0.15, "openshot_version": "2.5.1", "error_rate_stable": 1.0}` when reporting the current stable version.

src/settings/_default.settings
jonoomph: I think this is base64, but perhaps I'm wrong about that. It's what we get from the MainWindow::saveState() and MainWindow.saveGeometry() functions... so it's provided by Qt, and understood by Qt.

PR #4324, opened 2021-08-20
Change "Openshot logo" to "OpenShot logo"

scootergrisen:


jonoomph:
This will break lots of translations, and seems pretty minor. I suppose it should be fixed though. :thinking:

jonoomph:
This is auto generated from a file name in transitions... and is not fixable in the POT directly. I suggest we just delete this transition, which will remove this from our POT file. Closing this PR.

PR #4325, opened 2021-08-20
Change "Edit Decision Lists" to "Edit Decision List"

scootergrisen:
I'm guessing this should be without the "s".

There are two other string without "s":

    Import Edit Decision List
    Export Edit Decision List

https://github.com/OpenShot/openshot-qt/blob/develop/src/windows/ui/main-window.ui#L1652
https://github.com/OpenShot/openshot-qt/blob/develop/src/windows/ui/main-window.ui#L1676

### DIFF
--- a/src/classes/exporters/edl.py
+++ b/src/classes/exporters/edl.py
@@ -57,7 +57,7 @@ def export_edl():
     else:
         recommended_path = recommended_path.replace(".osp", ".edl")
     file_path = QFileDialog.getSaveFileName(app.window, _("Export EDL..."), recommended_path,
-                                            _("Edit Decision Lists (*.edl)"))[0]
+                                            _("Edit Decision List (*.edl)"))[0]
     if not file_path:
         return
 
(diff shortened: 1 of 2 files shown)


ferdnyc:
@scootergrisen 
> I'm guessing this should be without the "s".

Turns out, that's an unexpectedly nuanced question!

The other strings represent a menu action, and you can only import one EDL at a time. So, singular is clearly correct for those.

The strings you brought up, though, are file-dialog _filter_ strings, and there doesn't really appear to be a norm there.

image

Windows Wordpad's "Open File" dialog shows a muddled mix of singular and plural both — "Office Open XML Document", but "Unicode Text Documents"

image

The import file dialog for Inkscape on Linux is even worse: "CorelDRAW ... files", "Enhanced Metafiles" and "WordPerfect Graphics", but "Microsoft Visio XML Diagram" and "HP Graphics Language file".

Because the selection tied to those strings is filtering a list, in theory plural is correct — the **dialog** will show _all_ of the EDL files in the directory, from which you can pick one.

However, because the acronym EDL is a singular one, I'm inclined to agree with you that "Edit Decision Lists" is incorrect. The plural of "EDL" is either "EDLs" or (better) "EDL files". And since the filter should describe the _type_ of files being filtered, "Edit Decision List" (singular) or "Edit Decision List files" (plural) would be correct.

ferdnyc:
(This is bolstered by the fact that the other import/export filter string we use is "Final Cut Pro (*.xml)", singular. Not "Final Cut Pros" or "Final Cut Pro XMLs".)

PR #4333, opened 2021-08-21
Fixing libopenshot version warning message + Sentry.io fixes

jonoomph:
Fixing some issues in the release branch, found in my testing and in Sentry.io (with some of our 2.6.0 testers)

- Fixed libopenshot version mis-match error was crashing OpenShot
- Fixed timeline bounding box errors reported on sentry (needed protection)
- Added many guards to accessing variables (reported by sentry)
- Added "Select a Color" context menu for color keyframes, instead of popping up a dialog and blocking our context menu options on the color keyframes
- Fixed default QSpinBox values (which were using float instead of int)
- Fixed windows drive letter relative path detection (to prevent trying to make relative paths across drives)
- Replaced log.error with log.warning on Blender output (since this is a valid outcome of using the wrong version of Blender)

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -185,8 +185,6 @@ class OpenShotApp(QApplication):
                 },
             level="error",
         ))
-        raise RuntimeError(
-            "libopenshot version {} found, minimum is {}".format(ver, min_ver))
 
     def gui(self):
         from classes import language, sentry, ui_util, logger_libopenshot
--- a/src/launch.py
+++ b/src/launch.py
@@ -183,10 +183,7 @@ def main():
     argv = [sys.argv[0]]
     argv.extend(extra_args)
     argv.extend(args.remain)
-    try:
-        app = OpenShotApp(argv)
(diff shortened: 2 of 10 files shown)


src/classes/app.py
```diff
@@ -185,8 +185,6 @@ def check_libopenshot_version(self, info, openshot):
                 },
             level="error",
         ))
-        raise RuntimeError(
```
jonoomph, line 188: This was preventing the OpenShotApp from initializing

src/launch.py
```diff
@@ -183,10 +183,7 @@ def main():
     argv = [sys.argv[0]]
     argv.extend(extra_args)
     argv.extend(args.remain)
-    try:
-        app = OpenShotApp(argv)
-    except Exception:
-        app.show_errors()
```
jonoomph, line 189: app was never initialized here... and below a few lines, we call `app.gui()` which displays the errors... if a gui is created.

PR #4366, opened 2021-08-27
repair project files with invalid json for paths

JacksonRG:
#4211 caused a specific case where a user's osp would be written without valid json. (No quotes on the key for a path attribute)

This is meant to find any such errors in a project file, and fix them before attempting to open.

### DIFF
--- a/src/classes/json_data.py
+++ b/src/classes/json_data.py
@@ -54,11 +54,15 @@ class JsonDataStore:
         self._data = {}  # Private data store, accessible through the get and set methods
         self.data_type = "json data"
 
-        # Regular expression for project files with possible corruption
-        self.version_re = re.compile(r'"openshot-qt".*"2.5.0')
+        # Regular expressions for file version detection of possible corruption
+        self.version_re_250 = re.compile(r'"openshot-qt".*"2.5.0')
+        self.version_re_260 = re.compile(r'"openshot-qt".*"2.6.0')
 
         # Regular expression matching likely corruption in project files
         self.damage_re = re.compile(r'/u([0-9a-fA-F]{4})')
+
+        # Regular expression matching likely windows drive letter corruption in project files
+        self.damage_re_windows_drives = re.compile(r'(\n\s*)(\w*):')
 
         # Regular expression used to detect lost slashes, when repairing data
         self.slash_repair_re = re.compile(r'(["/][.]+)(/u[0-9a-fA-F]{4})')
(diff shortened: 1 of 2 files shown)


src/classes/json_data.py
```diff
@@ -153,6 +153,10 @@ def read_from_file(self, file_path, path_mode="ignore"):
                 contents = f.read()
             if not contents:
                 raise RuntimeError("Couldn't load {} file, no data.".format(self.data_type))
+            
+            if os.path.splitext(file_path) and os.path.splitext(file_path)[1] == '.osp':
```
jonoomph, line 157: Do we need to qualify this replacement? In other words, the only "content" that gets passed to this function is our *.osp project file.

jonoomph: Or should we protect it by version and if this regex finds any matches... Something like this
```
if self.damage_re.search(contents) and self.version_re.search(contents):
```

jonoomph: This approach makes more sense probably, because it limits the impact of this change to only project files from 2.6.0 which have at least 1 missing ", right?  However, I would rename `version_re` to something more specific, and create another `version_re_260` for 2.6.0 detection, etc...

JacksonRG: Sounds good. When debugging. I did see that `.project` files get passed through here. But checking version sounds good.

src/classes/json_data.py
```diff
@@ -153,13 +157,28 @@ def read_from_file(self, file_path, path_mode="ignore"):
                 contents = f.read()
             if not contents:
                 raise RuntimeError("Couldn't load {} file, no data.".format(self.data_type))
-            
-            if os.path.splitext(file_path) and os.path.splitext(file_path)[1] == '.osp':
-                # Repair lost quotes on json keys
-                contents = re.sub('(\n\s*)(\w*):', r'\1"\2":', contents)
+
+            # Scan for and correct possible OpenShot 2.6.0 corruption
+            if self.damage_re_windows_drives.search(contents) and self.version_re_260.search(contents):
```
JacksonRG, line 162: Fixing just the projects from when the bug was present! Looks great.

Earlier you mentioned `verison <= 2.6.0` (but that might mean the file's already been parsed if we know the version.)

There may be a handful of users windows users who had the daily build in the last month, and had files on different drives. (A swiss cheese bug to be sure) But 
`r'"openshot-qt".*"2.(5..|6.0)')`

jonoomph: I merged the windows drive letter fix right before creating the release branch. So I think we are probably good... minimal people this might impact. If so, they can manually update the "2.5.1-dev3" in the *.osp file to "2.6.0". Pretty easy work around, if anyone finds themselves in this situation

jonoomph:
LGTM! Merge away

PR #4367, opened 2021-08-28
Avoid dividing by zero when resizing app/video widget

jonoomph:
This is another sentry error, which happens a few hundred times a day. Not sure how or why the video widget is resized to a height of 0, but I guess it happens and that is an invalid size... so no reason to calculate a ratio or emit any signals for an invalid height.

### DIFF
--- a/src/windows/video_widget.py
+++ b/src/windows/video_widget.py
@@ -345,9 +345,9 @@ class VideoWidget(QWidget, updates.UpdateInterface):
                     tmp2 = tmp.keys()
                     obj_id = list(tmp.keys())[0]
                     raw_properties_effect = raw_properties_effect.get('objects').get(obj_id)
-                    
+
                     # Check if the tracked object is visible in this frame
-                    if raw_properties_effect.get('visible'): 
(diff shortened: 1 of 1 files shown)


src/windows/video_widget.py
```diff
@@ -1143,13 +1143,14 @@ def delayed_resize_callback(self):
         project_size = QSize(get_app().project.get("width"), get_app().project.get("height"))
         project_size.scale(self.delayed_size, Qt.KeepAspectRatio)
 
-        # Calculate height/width divisible by 2
-        ratio = float(project_size.width()) / float(project_size.height())
-        width = round(project_size.width() / 2.0) * 2
-        height = (round(width / ratio) / 2.0) * 2
+        # Calculate height/width divisible by 2 (if valid height)
+        if project_size.height() > 0:
```
jonoomph, line 1147: Actually, I just realized we don't even use this data

jonoomph: Fixing

JacksonRG:
:+1:

PR #4374, opened 2021-08-30
WIP: don't translate unless translation has same number of placeholders

JacksonRG:
If a string that we're translating has any placeholders to substitute data into, only return the translation if it has the same number of placeholders. This will prevent OpenShot from crashing if a bad translation gets through.

__Two froms of python placeholders__

- The C way
```python
print("Here's the value: %s" % variable)
```
- and the Rust way
```python
print("Here's the value: {}".format(variable))
```
__Exception__
The one type of error this won't catch is if a `%s` is replaced with a `%d` or some substitution that can't accept the data given. To ensure against that, I would have to count the different characters after `%` for order number and order, and I believe that's more computation than we want here  

Not merging
---
I'm now planning to incorporate these regular expressions into our CI translation checking. They shouldn't be needed at runtime if we're testing our translation files properly

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -84,11 +84,13 @@ class OpenShotApp(QApplication):
         # For _tr() to ensure translated string can have
         # data substituted inside of it.
         # C Style: "%s" % "insert"
+        #   type '%%' to display a '%'
         # Rust Style: "{}".format("insert")
-        self.c_style_sub = re.compile(r"(?<!%)%(?!%)") # '%' on it's own
-        self.rust_style_sub = re.compile(r"(?<!{){(?!{).*(?<!})}(?!})") # Style: "{}".format("insert")
-        self.c_sub_names = re.compile(r"(?<!%)%\(.*?\).") # Style: "%s" % "insert"
-        self.rust_sub_names = re.compile(r"(?<!{){.*?}(?!})") # Style: "{}".format("insert")
+        #   type '{{' or '}}' to display '{' or '}'
+
+        self.c_style_sub = re.compile(r"%{1}") # '%' on it's own
+        self.rust_style_sub = re.compile(r"{{1}.*?}{1}") # curly brackets with 0 or more characters between
+        self.c_style_sub_named = re.compile(r"%{1}\(.*?\).") # match from '%(' to the first ')'
 
         try:
             # Import modules
@@ -334,8 +336,8 @@ class OpenShotApp(QApplication):
         # If substitution names in the translation aren't present in the message string,
         # Fall back to english string
         # (This way the translated string will work as long as the english one does)
-        for match in list(re.finditer(self.rust_sub_names, translation)) + \
-        list(re.finditer(self.c_sub_names, translation)):
+        for match in list(re.finditer(self.rust_style_sub, translation)) + \
+        list(re.finditer(self.c_style_sub_named, translation)):
             if message.find(match.group()) == -1:
                 return message
         return translation
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -83,10 +83,12 @@ class OpenShotApp(QApplication):
 
         # For _tr() to ensure translated string can have
         # data substituted inside of it.
-        self.c_style_sub = re.compile(r"(?<!%)%(?!%)") # Style: "%s" % "insert"
+        # C Style: "%s" % "insert"
+        # Rust Style: "{}".format("insert")
+        self.c_style_sub = re.compile(r"(?<!%)%(?!%)") # '%' on it's own
         self.rust_style_sub = re.compile(r"(?<!{){(?!{).*(?<!})}(?!})") # Style: "{}".format("insert")
-        self.c_sub_names = re.compile(r"(?<!%)%(?!%)") # Style: "%s" % "insert"
-        self.rust_sub_names = re.compile(r"(?<!{){(?!{).*(?<!})}(?!})") # Style: "{}".format("insert")
+        self.c_sub_names = re.compile(r"(?<!%)%\(.*?\).") # Style: "%s" % "insert"
+        self.rust_sub_names = re.compile(r"(?<!{){.*?}(?!})") # Style: "{}".format("insert")
 
         try:
             # Import modules
@@ -333,8 +335,8 @@ class OpenShotApp(QApplication):
         # Fall back to english string
         # (This way the translated string will work as long as the english one does)
         for match in list(re.finditer(self.rust_sub_names, translation)) + \
-        list(re.finditer(self.c_sub_names, message)):
-            if translation.find(match.group()) == -1:
+        list(re.finditer(self.c_sub_names, translation)):
+            if message.find(match.group()) == -1:
                 return message
         return translation
 
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -82,8 +82,10 @@ class OpenShotApp(QApplication):
         self.errors = []
 
         # for _tr() to ensure translated string 
-        self.placeholder_old_re = re.compile(r"(?<!%)%(?!%)") # Style: "%s" % "insert"
-        self.placeholder_new_re = re.compile(r"(?<!{){(?!{).*(?<!})}(?!})") # Style: "{}".format("insert")
+        self.c_style_sub = re.compile(r"(?<!%)%(?!%)") # Style: "%s" % "insert"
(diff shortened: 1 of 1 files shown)


jonoomph:
Just a quick comment on this one. This PR makes me a bit nervous. I think I would rather see this implementing on our `test_translations.py` code, to help verify and raise an error if translations don't match our English strings.

For example...
```
 self.app._tr(trans_name) % suffix_number
```
This is really not an issue with the `self.app_tr` method, but an issue with string replacements after the translation has successfully resolved into a string. It feels wrong trying to complicate the ` self.app._tr(trans_name)` side of things, when that is not where the error is being thrown. Just thinking out loud. :thinking:

JacksonRG:
@jonoomph I see what you mean, but I think this is probably the best way to avoid having to wrap every translation formatting in a try/catch

ferdnyc:
@JacksonRG @jonoomph : I think I have a way. It'll still mean changing every use of Python string interpolation on translated strings, but it should better protect the code from Python's finickeyness.

This is a minimal subclass of `str` that adds one new method, `.args()`:
```python3
class trstr(str):
    _rawstr = None
    def __init__(self, *args):
        super().__init__()
        self._rawstr = str(*args) if args else ""
    def args(self, *args, **kwargs):
        try:
            if len(args) >= 1 and isinstance(args[0], dict):
                return trstr(self._rawstr % args[0])
            elif kwargs:
                return trstr(self._rawstr % kwargs)
            else:
                return trstr(self._rawstr % args)
        except TypeError:
            return self
    def __str__(self):
        return str(self._rawstr)
```

When `.args()` is called on a `trstr`, it attempts to perform string interpolation using `%`. An exception just causes it to give up:
```python3
>>> trstr("A normal string")
'A normal string'
>>> trstr("A string with %s").args("substitutions")
'A string with substitutions'
>>> trstr("A string with %s %s").args("many", "substitutions")
'A string with many substitutions'
>>> trstr("A string with %s %s").args("too", "many", "substitutions")
'A string with %s %s'
>>> trstr("A string with %d %s %s").args("too", "many", "substitutions")
'A string with %d %s %s'
>>> trstr("A string with %d %s %s").args(2, "many", "substitutions")
'A string with 2 many substitutions'
#
# The check for a `dict` and the `kwargs` support make tagged
# interpolation work:
>>> trstr("%(first)s %(second)d").args(first="red", second=2)
'red 2'
>>> trstr("%(first)s %(second)d").args({"first": "blue", "second": 12})
'blue 12'
# Order is still arbitrary with dict/kwargs...
>>> trstr("%(first)s %(second)d").args(second=2, first="red")
'red 2'
>>> trstr("%(first)s %(second)d").args({"second": 99, "first": "bottles"})
'bottles 99'
```

Because it's a subclass of `str`, it should hopefully be usable where a string is usable. So, our Qt translation function could return a `trstr()` of its output, instead of just passing back a Python `str`. And then in the code, we call...
```python3
_("Some translatable string").args(any, arguments, needed)
```

It'll preclude any use of .format()-style interpolation, but those were a mistake and I never should've complicated the translations with them, since `%` strings are familiar to translators.

Only thing we'd still have to test is whether `trstr` can _really_ be used anywhere a `str` is, like as arguments to PyQt5 classes. But if not, we could always wrap all of the translations in a final `str()` to make the output always a real `str`:
```python3
QMessageBox(
    parent,
    str(_("Translatable %s title").args("window")),
    str(_("Some translatable string").args(any, arguments, needed)),
)
```

ferdnyc:
One cool thing is that, because the return value of `trstr.args()` is a `trstr`, you can re-do a failed interpolation:
```python3
>>> x = trstr("A string with %d %s %s").args("too", "many", "substitutions")
>>> x
'A string with %d %s %s'
>>> x.args(2, "many", "substitutions")
'A string with 2 many substitutions'
```

Admittedly, that's not very useful _in practice_ since we only have one translation to attempt. I was hoping I might be able to wire up chained translations, the way they work in C++ with `QString`, so that you could do:
```python3
trstr("A string with %d %s %s").arg(1).arg("many").arg("subs")
```
...buuuuuut, Python raises an exception if the RHS argument to `%` is too long _or_ too short, so partial/incremental substitutions seem to be out.

ferdnyc:
@jonoomph 

> This is really not an issue with the `self.app_tr` method, but an issue with string replacements after the translation has successfully resolved into a string. It feels wrong trying to complicate the ` self.app._tr(trans_name)` side of things, when that is not where the error is being thrown.

Can't  say I totally agree with that. I mean, on _principle_ I agree that `translate()` isn't at fault, but it would be good to have protection against bad interpolations instead of requiring the data to be pristine or it'll lead to crashes. (Which makes it harder for users to edit their local translations, for one thing, without suddenly having to worry about causing crashes.)

Python's `%` interpolation is fragile and clearly designed to be used on hardcoded strings. It was never meant for things like translation, or really any operations on strings determined at runtime. 

(One of the big advantages to f-strings is that they _can't_ be used on anything other than fixed strings, which makes it impossible to create problems like this. And even `.format()` did _improve_ on the situation, because using it means that this no longer raises:
```python3
"My {} {} {}".format("Very", "Excellent", "Mother", "Just", "Served", "Us", "Nachos")
```
It'll just discard the remaining args and set the string to `"My Very Excellent Mother"`. As it should.)

I've always wondered why Riverbank didn't implement _some_ sort of argument-handling for `tr()` and `translate()`, since returning `str` instead of `QString` loses you the ability to do standard `QString::arg()` substitution. And `QString::arg()` is much more robust:

If you do this:

```c++
auto qs = QString("This %1 of brothers").arg("band").arg("rabbits");
```

`qs` is `QString("This band of brothers")` just like you'd expect, and nothing crashes.

ferdnyc:
Oh. Never mind, we don't even need that.

Turns out Python has a whole class [designed just for this purpose](https://docs.python.org/3/library/string.html#template-strings) (and it's been there since [PEP 292](https://www.python.org/dev/peps/pep-0292/) in 2002): `string.Template`.

It handles `$identifier` and `${identifier}` substitution in strings (chosen for familiarity to anyone who's ever used shell variables), and critically, it has a `safe_substitute()` method that, well, does exactly what it says on the tin.

```python3
>>> from string import Template
>>> Template("This is a $type string").safe_substitute(
...     type="template")
'This is a template string'
>>> Template("This is a $different string").safe_substitute(
...     type="template")
'This is a $different string'
>>> Template("This is a $different string").safe_substitute(
...     type="template", different="separate")
'This is a separate string'
```

So, yeah, really our `translate()` wrapper should be returning `Template(app.translate("", arg))`. Or, if people really like that awful syntax, a `string.Template` subclass that overloads `__mod__` in  the familiar way:

```python3
>>> class ModTemplate(Template):
...     def __mod__(self, other):
...         return self.safe_substitute(other)
... 
>>> ModTemplate("a $type string") % {"type": "template"}
'a template string'
>>> ModTemplate("a $type string") % {
...     "type": "template", "color": "red"}
'a template string'
>>> ModTemplate("a $type string") % {"color": "red"}
'a $type string'
>>> 
```

There's no way to make it support `%`-escapes (good!), so the translation strings would definitely have to change. There's also no way to support positionally-defined arguments, because `$0`, `$1`, etc. aren't valid identifiers. It has to be an alphanumeric string starting with `[_a-zA-Z]`. Though I suppose we could set a rule that positional arguments are named `$a`, `$b`, `$c`, and so on, then do something like this:

```python3
>>> class ModTemplate(Template):
...     def __mod__(self, other):
...         from typing import Sequence
...         if isinstance(other, Sequence):
...             other = {chr(ord('a')+i): v
...                      for i, v in enumerate(other)}
...         return self.safe_substitute(other)
... 
>>> ModTemplate("a $b $a") % ("template", "smart")
'a smart template'
```
That's probably as close as we can ever hope to get to matching the current behavior.

Wait, no, that has a bug. Strings are sequences (bluh.), so this happens:
```python3
>>> ModTemplate("a $a") % "template"
'a t'
>>> ModTemplate("a $a") % ("template")
'a t'
```

OK, this time for sure:
```python3
>>> class ModTemplate(Template):
...     def __mod__(self, other):
...         from typing import Sequence
...         if isinstance(other, str):
...             other = {"a": other}
...         elif isinstance(other, Sequence):
...             other = {chr(ord('a')+i): v
...                      for i, v in enumerate(other)}
...         return self.safe_substitute(other)
... 
>>> ModTemplate("a $a") % "template"
'a template'
>>> ModTemplate("a $a") % ("template")
'a template'
>>> ModTemplate("a $a $b") % ("template")
'a template $b'
```

JacksonRG:
This issue was taken care of by Frank's translation test PR

PR #4434, opened 2021-09-20
Remove exec bit from presets

ferdnyc:
Three of our presets files inexplicably have their executable bits set, even though presets are decidedly not executable. This confuses some packaging systems.


huntekah (approved):
LGTM

PR #4440, opened 2021-09-21
Fixing mask "source" menu in properties

jonoomph:
Fixing mask "source" menu in properties, which should be showing a list of project files but was breaking due to a missing "icon" property.


JacksonRG (approved):
LGTM

PR #4445, opened 2021-09-23
Remove vscode config, and add a gitignore item

JacksonRG:
Undoing a silly mistake I made in the 2.6.0 update.


ferdnyc:
@JacksonRG 
You should add `.vscode/` to your local `.git/info/exclude/` instead. As the `gitignore(5)` man page notes (emphasis mine):

> A project normally includes such `.gitignore` files in its repository, **containing patterns for files generated as part of the project build.**

The `.git/info/exclude` file is the local counterpart to that, containing files that are likely to be generated locally but will never be generated by the project itself.

(The Git developers have done a terrible job disseminating information about how this works.)

**ETA:** The first four lines of the existing `.gitignore` are similarly wrong, and they shouldn't be there either. Fully conceded. All I can say is that I wasn't consulted on their addition. `¯\_(ツ)_/¯`

JacksonRG:
@ferdnyc Thanks for the info! I'll amend the `.gitignore`. (and my exclude file)

It shouldn't happen again, as I've also gotten into the habit of using `git add -p` so I don't accidentally have debug statements in my pull requests.

.gitignore
```diff
@@ -1,13 +1,15 @@
+# Exclude common editor config
+.vscode/
 *.DS_Store
 .idea
 .settings
+# Exclude results of the build process
 idea/*
```
ferdnyc:
```suggestion
idea/*
# Exclude results of the build process
```

`idea/` is also a non-build exclusion.

JacksonRG: Sounds good. And as you can see, I decided all engineering is compromise. It's good to know the intended use of `.gitignore`. But, I think there's merit to preventing a common error if someone does `git add --a`.

ferdnyc:
Huh. Turns out there's also a `core.excludesFile`. ("Defaults [on Linux] to `$XDG_CONFIG_HOME/git/ignore`. If `$XDG_CONFIG_HOME` is either not set or empty, `$HOME/.config/git/ignore` is used instead.") Adding a path there excludes it from all repos everywhere (when managed by a `git` that can see that config / that file).

.gitignore
ferdnyc: I'll apply that small change and merge this now. Thanks!

ferdnyc: > I think there's merit to preventing a common error if someone does `git add --a`.

Yeah... I'm of two minds, really. I _do_ think that, to an extent, but the problem is that as soon as you start down that road, you end up with a file full of literally dozens of different software packages' particular ephemera. It's impossible to cover all the bases.

Plus, PRs with random spurious files in them _should_ be caught and fixed in review. If they're not, then that makes me think that either
1. We just aren't bothering to review code before merging
2. PRs being submitted are too big/sprawling for others to review them

I know there's _defintely_ some of the second, I've been guilty of it myself. The first, I have no idea... I'm personally reluctant to merge my own code without a second pair of eyes unless it's a REALLY simple, unambiguous change... and that probably explains why I have like 25 open PRs across the three repos. :confounded: 

All that being said, though, I bet there's a bot we could install to patrol for newly-added files, and call them out in the PR comments. It _should_ be rare for a PR to introduce completely new files, so that would probably do a lot to prevent accidental adds. Hmm... to the Marketplace!

PR #4452, opened 2021-09-25
New Title Editor menu icon & resources consolidation

ferdnyc:
This PR is at cross purposes, to some extent, and I may split it up, but... two things are happening here

#### New Title Editor icon

I finally got sick of the black-on-black generic "font" icon that was set for the Title Editor launch action (as seen in the Title > Title menu entry), and  just drew a new one. It's loosely based on the filmstrip icon we use for "Select Profile" in the toolbar, but plenty different.

new-title-editor-icon

And because that got me into the Qt resource file and general housekeeping...

#### Consolidation of non-Python files into `/src/resources/`

From an installation / packaging perspective, having all of our "other" files (the ones that aren't Python modules, be they images, Blender data, XML files, etc) scattered across so many directories is problematic. Especially since the code currently doesn't treat `openshot_qt` as a single, cohesive Python package.

(That's the effect of all those `from classes import blah...`, `from windows import blah...` statements: In terms of the code, we have a Python package called "classes", and a Python package called "windows", and they aren't actually part of the _same_ package at all!)

So ANYWAY, it would be far easier for packaging tools to automatically deal with OpenShot, and we would have to maintain less code that brute-forces things manually (almost fighting AGAINST the packaging tools in the process), if there were just a single path that held all of the "non-Python ancillary files". I am starting the work of making `/src/resources/` that path.

This will also mean fewer paths to manage in `classes.info`: The path for most any file will be simply:
```python3
os.path.join(info.RESOURCES_PATH, subdir or "", subsubdir or "", "filename.ext")
```

1. `AudioThumbnail.svg` (used in many places) is now in there
2. So is the Humanity font we load (because `/src/images/fonts/` was a completely silly place for it)
3. A copy of `openshot.png` in `/src/resources/` is now the test file the unit tests will use, when creating data (replacing the previous arbitrary `/src/images/AboutLogo.png`, which was only there _for_ the unit tests. (The actual code loads it from the resource file.) That copy is now removed.

(Eventually I'm hoping to make the `resources` dir a Python module via its `__init__.py`, so that you can import it and do things like this:)
```python3
from openshot_qt.resources import resource_path
path_to_transition = resource_path("transitions", "common", "fade.svg")
path_to_icon = resource_path("cache", "fade.png")

# Or perhaps even
from openshot_qt.resources import resource_type as rtype, resource_path as rpath
path_to_file = rpath(rtype.TRANSITION, "common", "fade.svg")
```

Transitions are a  bad example, though, as they'll be hard to move now that the path hierarchy is hardcoded into libopenshot. That's something that definitely has to change.

Also, other housekeeping:
1. Removed all of those leftover "Clear" actions from the UI file, for the text filtering fields that we haven't used since we turned on their built-in clear buttons.
1. Finally moved all of our custom action icons out of the `images/Humanity/actions/custom/` directory (since they're not part of the Humanity icon set), now they live in `/images/actions/`.
1. Updated my Makefile for the resource file and generated a new one.


images/Makefile
```diff
@@ -1,11 +1,12 @@
...
-ICONS+=$(shell /usr/bin/find . -name \*-icon.png -newer ../src/classes/openshot_rc.py)
-ICONS+=$(shell /usr/bin/find . -name cursor\*.png -newer ../src/classes/openshot_rc.py)
-ICONS+=$(shell /usr/bin/find . -name keyframe\*.png -newer ../src/classes/openshot_rc.py)
-ICONS+=AboutLogo.png openshot.svg
-ICONS+=$(shell /usr/bin/find . -type f -name hw-accel\*.svg -newer ../src/classes/openshot_rc.py)
+
+DUMMY: list-outdated
+
+ICONS:=$(shell /usr/bin/find . -type f -name \*.svg -newer ../src/classes/openshot_rc.py)
+ICONS+=$(shell /usr/bin/find . -type f -name \*.png -newer ../src/classes/openshot_rc.py)
```
JacksonRG, line 6: Tell me if I'm way off. This is is finding any svgs or pngs created _after_ `openshot_rc.py`?

images/actions/title-editor.svg
```diff
@@ -0,0 +1,55 @@
...
+   <rect x="2" y="11" width="1" height="1" opacity=".81"/>
+  </g>
+ </g>
+ <g fill="#fff">
+  <path d="m 6.0267285,5 h -0.027038 3.998981 L 10,6 11,6.0440387 V 7 H 5 L 5.0355818,5.9913773 6,6 Z" stroke-linecap="round" stroke-linejoin="round" sodipodi:nodetypes="cccccccccc"/>
+  <rect x="5" y="9" width="2" height="1" style="paint-order:normal"/>
+  <rect x="8" y="8.0042" width="2" height=".99581" style="paint-order:normal"/>
+  <path d="m10 9h1v1h-1zm-1 0h1v1h-1z" stroke-linecap="round" stroke-linejoin="round"/>
+ </g>
+</svg>
```
JacksonRG, line 55: I was already pulling up inkscape just to double check that this looks right. Then found that when you view the raw of an svg, it renders. :sweat_smile: 

Looks good

images/razor_line_with_razor.png
```diff
@@ -0,0 +1 @@
+../src/timeline/media/images/razor_line_with_razor.png
```
JacksonRG, line 1: Should this be in `resources/timeline/...`? If the goal is moving non `.py` files into `resources/`

JacksonRG: Possibly answering my question.

Is the "moving non py files into `resources/`" only within `src/`? Like eventually the emojis, and blender files would go there?

JacksonRG (approved):
Looks good!
Mostly just asking questions for my benefit

PR #4455, opened 2021-09-26
project_data: Also sanity-check 2.5.1-dev*

ferdnyc:
Hey @jonoomph @JacksonRG 

Shouldn't the check for bad crop data be `openshot_version.startswith("2.5.1")`? Surely some of the development builds had the same issues as the release, no?

### DIFF
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -719,7 +719,7 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
                                             point.get("handle_right")["X"] = 0.5
                                             point.get("handle_right")["Y"] = 0.0
 
-        elif openshot_version == "2.5.1":
+        elif openshot_version.startswith("2.5.1"):
             # Replace missing crop properties from 2.5.x
             for clip in self._data["clips"]:
                 # Loop through keyframes for alpha
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -719,9 +719,10 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
                                             point.get("handle_right")["X"] = 0.5
                                             point.get("handle_right")["Y"] = 0.0
 
-        elif openshot_version.startswith("2.5.1"):
+        elif openshot_version.startswith("2.5."):
             # Replace missing crop properties from 2.5.x
-            for clip in self._data["clips"]:
+            log.debug("Scanning OpenShot 2.5 project for legacy cropping")
+            for clip in self._data.get("clips", []):
                 # Loop through keyframes for alpha
                 crop_x = clip.pop("crop_x", {})
                 crop_y = clip.pop("crop_y", {})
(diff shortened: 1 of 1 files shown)


ferdnyc:
Hey @JacksonRG any thoughts on this?

ferdnyc:
(Fortunately we know for sure we'll never have a version `2.5.10`, so the string comparison should be safe.)

JacksonRG (approved):
Good idea.

This might actually help with several errors opening projects that are being reported.

JacksonRG:
Makes me want to grep the whole project for any other version numbers... brb

---
looks like the only other `2.5*`  is in `json_data.py:58` looks like that's been there for a while and is specifically addressing a 2.5.0 bug. So we're good

ferdnyc:
@JacksonRG Yeah, the bug with JSON encoding was pretty specific to `2.5.0` dash nothing. It was changed right before release (and may have received slightly less testing than was warranted, in hindsight). `2.5.1` was rushed out to fix it right away, so the only significant numbers of users who had files impacted by that issue were all running the `2.5.0` release.

src/classes/project_data.py
```diff
@@ -728,10 +728,11 @@ def upgrade_project_data_structures(self):
...
 
-                if self.is_keyframe_valid(crop_x, 0.0) or \
-                        self.is_keyframe_valid(crop_y, 0.0) or \
-                        self.is_keyframe_valid(crop_width, 1.0) or \
-                        self.is_keyframe_valid(crop_height, 1.0):
+                if any([self.is_keyframe_valid(crop_x, 0.0),
+                        self.is_keyframe_valid(crop_y, 0.0),
+                        self.is_keyframe_valid(crop_width, 1.0),
+                        self.is_keyframe_valid(crop_height, 1.0),
+                       ]):
```
ferdnyc, line 736: @JacksonRG I couldn't help myself.

Technically this is less efficient, as it'll _always_ perform all four calls instead of short-circuiting the conditions after the first `True` result. But I think that's actually a good thing, since it means all four (potential) values will be run through `is_keyframe_valid`, and any issues will be more likely to surface.

src/classes/project_data.py
```diff
@@ -769,10 +770,13 @@ def upgrade_project_data_structures(self):
...
-        if len(keyframe.get("Points", [])) == 0:
+        points = keyframe.get("Points", [])
+        if not points or not isinstance(points, list):
             return False
-        return len(keyframe.get("Points", [])) > 1 or \
-               keyframe.get("Points", [])[0].get("co", {}).get("Y", 0.0) != default_value
+        return any([
+            len(points) > 1,
+            points[0].get("co", {}).get("Y", default_value) != default_value,
+        ])
```
ferdnyc, line 775: @JacksonRG 
The main point of this was really to fix a minor bug I just noticed in the final comparison. Since `0.0` _is_ the default value for some of these tests, but _not_ for others, using it as the missing-value filler will lead to inconsistent results: If the `"Y"` value is missing and we substitute a default of `0.0`, `is_valid_keyframe` will sometimes return _True_ (when `default_value` is not `0.0`).

Using `default_value` as the... er... default value means the comparison will always evaluate `False` on missing `"Y"` values, which is what we'd want.

ferdnyc (commented):
Meddled a bit, and fixed one small bug in the existing code that I happened to notice. I'm going to test this locally with some hand-crafted project files before I think about merging. I don't trust a successful CI build to tell me whether there are any issues here, since none of our (paltry) unit tests even remotely touch it. Heck, they don't invoke any of `project_data.py` other than the class `__init__()`s.

src/classes/project_data.py
```diff
@@ -719,41 +719,38 @@ def upgrade_project_data_structures(self):
...
+                        self.is_keyframe_valid(crop_y, 0.0),
+                        self.is_keyframe_valid(crop_width, 1.0),
+                        self.is_keyframe_valid(crop_height, 1.0),
+                       ]):
                     # Apply crop effect to clip (which wasn't available in 2.5.x)
+                    log.info(
+                        "Migrating OpenShot 2.5 crop properties for clip %s",
+                        clip.get("id", "<unknown>")
+                    )
+                    from json import loads as jl
```
ferdnyc, line 742: This is evil from a readability standpoint, I admit, but in my defense it's very local.

src/classes/project_data.py
```diff
@@ -719,41 +719,38 @@ def upgrade_project_data_structures(self):
...
-                    else:
-                        effect_json["bottom"] = json.loads(openshot.Keyframe(0.0).Json())
+                    effect_json = jl(effect.Json())
+
+                    # Attach previous clip crop values
+                    effect_json.update({
+                        "x": crop_x or jl(openshot.Keyframe(0.0).Json()),
+                        "y": crop_y or jl(openshot.Keyframe(0.0).Json()),
+                        "right": crop_width or jl(openshot.Keyframe(1.0).Json()),
+                        "bottom": crop_height or jl(openshot.Keyframe(1.0).Json()),
```
ferdnyc, line 752: This change was, again, not merely gratuitous but a minor bug fix. Because the loop that immediately follows this is going to set all of the `"bottom"` and `"right"` keyframe values to `1.0 - point["co"]["Y"]`, they need to be _created_ with initial values of `1.0` if they don't exist. Otherwise, they'll end up as `1.0` when they should be `0.0`.

src/classes/project_data.py
```diff
@@ -719,41 +719,38 @@ def upgrade_project_data_structures(self):
...
+                if any([self.is_keyframe_valid(crop_x, 0.0),
+                        self.is_keyframe_valid(crop_y, 0.0),
+                        self.is_keyframe_valid(crop_width, 1.0),
+                        self.is_keyframe_valid(crop_height, 1.0),
+                       ]):
                     # Apply crop effect to clip (which wasn't available in 2.5.x)
+                    log.info(
+                        "Migrating OpenShot 2.5 crop properties for clip %s",
+                        clip.get("id", "<unknown>")
+                    )
```
ferdnyc, line 741: Added some logging, which is always good on a one-time operation that touches user data.

ferdnyc (commented):
@JacksonRG

It's waaaaaaaaay too late to matter at this point, but I made a couple of further tweaks to this, and tested it on a bunch of project files from different versions to make sure it did the right thing.

PR #4460, opened 2021-10-01
Set window icon for child windows

### LINKED ISSUE #4285: Generic icon in Choose Profile and Preferences windows
OpenShot 2.5.1-dev3
Windows 7

The icon for the Choose dialog window is generic.
Same for Preferences window.

billede

The icon for e.g. Export Video dialog is the OpenShot icon.
billede

ferdnyc:
This PR corrects our child window icons on Windows, as some were failing to show the OpenShot icon. It also cleans up our class argument handling, and eliminates all remaining Python2-style `super(Parent, Child)` calls.

Fixes #4285

Details: 

- I've made sure any UI files that were missing a `windowIcon` property now have it set.

- All of our `QWidget`-derived classes (`QDialog` subclasses, etc.) now
  AT LEAST take `**kwargs` in their `__init__()` methods, to allow for
  arbitrary arguments.

  The ones that didn't already take arguments
  beyond `self`, I added support for both `*args` and `**kwargs`, so that
  they can be called naturally as `SomeDialog(parent, etc...)` in order
  to assign '`parent`' as the parent.

  The ones that already took positional args, I only added `**kwargs`
  at the end, so they'd have to be parented via kwarg:
  ```python3  
  win = SomeDialog(arg, arg2, parent=parent)
  ```
- Though that previous commit added support for parenting to our
  dialog classes, most windows opened from `MainWindow` SHOULD NOT
  be parented to it. If they are, Linux desktops will freeze their
  position in the middle of a dimmed parent window, which is
  inconvenient. There's an old bug somewhere asking us to fix that,
  which we did.

  Because that's not an obvious thing, I've added a comment before
  each window instantiation that should _not_ be parented, explaining
  why. For the ones that take positional args, I've also explicitly
  set '`None`' as the first argument.

### DIFF
--- a/src/windows/about.py
+++ b/src/windows/about.py
@@ -104,9 +104,9 @@ class About(QDialog):
 
     ui_path = os.path.join(info.PATH, 'windows', 'ui', 'about.ui')
 
-    def __init__(self):
+    def __init__(self, *args, **kwargs):
         # Create dialog class
-        QDialog.__init__(self)
+        super().__init__(*args, **kwargs)
 
         # Load UI from designer & init
         ui_util.load_ui(self, self.ui_path)
--- a/src/windows/add_to_timeline.py
+++ b/src/windows/add_to_timeline.py
@@ -377,7 +377,7 @@ class AddToTimeline(QDialog):
             position += (end_time - start_time)
 
         # Accept dialog
-        super(AddToTimeline, self).accept()
+        super().accept()
 
     def ImageLengthChanged(self, value):
         """Handle callback for image length being changed"""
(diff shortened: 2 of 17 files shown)


jonoomph:
@ferdnyc What dimming of the main window are you referring to? Can we get a screenshot? Is this a common thing I just don't know about?  Just curious, thanks!

ferdnyc:
@jonoomph Sure — it's the default mode for at least the GNOME desktop, not sure if/what happens under KDE. But with GNOME Shell, if I give the Animated Title dialog a parent, because it _is_ a dialog by default it's "attached" to its parent window. When the dialog opens, the parent is dimmed. (Subtly, but enough to throw the ColorPicker off if someone wanted to grab a color from the video preview, which is where this first came up.) Also, their positions are locked together — any attempt to drag the dialog drags the main window along with it, so if something is hidden under the dialog there's no way to reveal it:

Screenshot from 2021-11-10 23-47-27

But if the dialog is unparented, it's separate from the main window, can be dragged freely around the screen, and doesn't affect the colors of the interface:

Screenshot from 2021-11-10 23-48-12

JacksonRG:
LGTM!

(also finally figured out what all the `super().__init__()` I'm seeing mean. And what arbitrary `*args` are while reading it)

JacksonRG (approved):
LGTM!

(also finally figured out what all the `super().__init__()` I'm seeing mean. And what arbitrary `*args` are while reading it)

PR #4466, opened 2021-10-05
SVG-based images for marker, add_marker, lock, and transition

ferdnyc:
As th branch name gives away, I started this as an effort to update the `add_marker.png` icon. But along the way, I ended up redrawing not only the _actual_ marker icon in SVG (along with a new add_marker design based on it), but also the `lock.png` used for the track lock indicator, and the `transition.svg` background that's placed inside transitions. Code is updated where necessary to utilize the new icons.

The marker changes bear additional explanation. From my commit message:

> - A new property, 'vector', is added to marker objects in the project
>   data. It contains a (file)name _without_ extension. If the Timeline
>   sees marker.vector, it will set `{{marker.vector}}.svg` as the
>   icon for the marker.
> - If marker.vector is not present, it will look at marker.image. If
>   marker.image === blue.png, it will be replaced with blue.svg.
> - Any other marker.image filename will be used directly.
> 
> - The main_window.py code that creates new markers knows to add
>   `"vector": "blue"` to their dictionary. `"image": "blue.png"` is
>   also still added to preserve backwards-compatibility in the
>   project data.

Because the SVG files I check in to the `src/` tree are optimized, where I also had unoptimized original Inkscape files, those are checked in to `/images/`. They won't be part of the resource file or the `src/` packaging, they're just for us.

<s>Aaand, I forgot to rebase this branch on `develop`. I'd been working on it based off of my #4157 because it's just so handy to have console logging when testing timeline JS changes in the browser. I'll have to split these new commits out into their own branch first.</s> Fixed.

But first, some before/afters. The main benefit to SVG, obviously, would be crisper HiDpi rendering — at < 16px per side, there isn't that much difference to see.

### marker icon before/after

old: image    new: image

### marker at 400% zoom

old: image  new: image

### add_marker before/after

old: image  new: image

(<s>**UGH. QtSvg strikes again, that's _not_ how the icon is supposed to look. So, I'll have to sort out what the issue is there.**</s> Managed to fix add_marker for QtSvg.)

### lock icon before/after

old: image  new:  image

The transition graphic isn't even worth showing, the differences are undetectable.


ferdnyc:
Updates on this and all of my other 1000 pending projects may also be slow this week, as I've been summoned for jury duty starting in... 90 minutes from now, eep!

ferdnyc:
Heh. Codacy hates the formatting of our CSS file. You and me both, buddy. :neutral_face:

ferdnyc:
I got sick of Codacy bitching about the format of the CSS file (especially since I _agreed_ with it), so I completely reformatted the whole thing. `csstidy` as a base, then plenty of hand adjustments because it likes to squish certain things (`values,with,commas,look,like,this`, and `properties:squish,too;`), and excessively spread out other things (both curly braces each on its own separate line).

JacksonRG (commented):
Those icons look great! Very shiny.

And I appreciate the cleanup work!

JacksonRG (approved):
LGTM

jonoomph:
LGTM!

PR #4470, opened 2021-10-07
Change user 'default.project' filename to 'default.osp', migrate old files if found

ferdnyc:
In retrospect, `default.project` was a terrible name for the user default project file. It's so much easier for users to save their own default project right from OpenShot, by using a default filename with an `.osp` extension. This PR corrects the poor choice of expected filename.

### DIFF
--- a/src/classes/info.py
+++ b/src/classes/info.py
@@ -77,6 +77,15 @@ for folder in [
     if not os.path.exists(os.fsencode(folder)):
         os.makedirs(folder, exist_ok=True)
 
+# Migrate USER_DEFAULT_PROJECT from former name
+LEGACY_DEFAULT_PROJECT = USER_DEFAULT_PROJECT.replace(".osp", ".project")
+if all([
+    os.path.exists(LEGACY_DEFAULT_PROJECT),
+    not os.path.exists(USER_DEFAULT_PROJECT),
+]):
+    print("Migrating default project file to new name")
+    os.rename(LEGACY_DEFAULT_PROJECT, USER_DEFAULT_PROJECT)
+
 # Maintainer details, for packaging
 JT = {"name": "Jonathan Thomas",
       "email": "jonathan@openshot.org",
--- a/src/classes/info.py
+++ b/src/classes/info.py
@@ -65,7 +65,7 @@ PROTOBUF_DATA_PATH = os.path.join(USER_PATH, "protobuf_data")
 YOLO_PATH = os.path.join(USER_PATH, "yolo")
 # User files
 BACKUP_FILE = os.path.join(BACKUP_PATH, "backup.osp")
-USER_DEFAULT_PROJECT = os.path.join(USER_PATH, "default.project")
+USER_DEFAULT_PROJECT = os.path.join(USER_PATH, "default.osp")
 
 # Create user paths if they do not exist
 # (this is where temp files are stored... such as cached thumbnails)


ferdnyc:
The only thing holding me back from merging this is a vague uncertainty whether there are enough (or any) people using the feature as-is, that effectively disabling their existing `default.project` file might negatively impact them. In theory _both_ paths could be supported for a time, to let people transition.

Then again, I may be agonizing over something that affects a grand total of three people, none of whom will even care. :confounded:

src/classes/info.py
```diff
@@ -65,7 +65,7 @@
 YOLO_PATH = os.path.join(USER_PATH, "yolo")
 # User files
 BACKUP_FILE = os.path.join(BACKUP_PATH, "backup.osp")
-USER_DEFAULT_PROJECT = os.path.join(USER_PATH, "default.project")
+USER_DEFAULT_PROJECT = os.path.join(USER_PATH, "default.osp")
```
JacksonRG, line 68: You mentioned being worried about anyone currently using this. I'd say rename the current line `USER_DEFAULT_PROJECT_LEGACY`. And in `project_data.py:270`, add an else if to read in the legacy project.

It will save the new project along side the legacy one. But the legacy will only be read once (assuming no catastrophe before saving), because OpenShot will save under `USER_DEFAULT_PROJECT`

> 

>

ferdnyc: @JacksonRG Well, we don't _resave_ the default project, so normally it wouldn't ever be migrated. But I could add code to explicitly move `default.project` to `default.osp` the first time it's encountered, yeah. And then it'll be at the new name.

(It'd be right there in the same directory, so I'm not really worried about users not being able to find the file because I renamed it on them. They'll figure it out.)

So, yeah, that's probably the best solution. :+1:

ferdnyc: (The **ADVANTAGE** of the new name is that now, if someone wants to create/update a default project file, they can just edit an OpenShot project — either a new one, or their existing `default.osp` — and save it [back again] as `default.osp.` In the past, a project file had to be manually copied into the user's configuration directory and renamed `default.project`. Which is why that was, in hindsight, so very dumb.)

ferdnyc: In fact, I stuck the migration code right in `info.py`. No sense in cluttering up `project_data.py` with temporary code for a one-time operation.

ferdnyc: I'm _almost_ tempted to add special-case code to the _save_ logic that strips out the `"history"` object and any other undesirable elements, if the save is being directed to `os.path.join(info.USER_PATH, "default.osp")`. 

I just know someone somewhere is going to end up saving a default project file with like 30 levels of undo embedded in it. And then we'll end up with a bug report saying that OpenShot always takes forever to load and all of their project files come out massive no matter how simple the project is.

But... nah. Gotta let people do stupid things sometimes, or they'll never learn nothin'.

ferdnyc: (...I'm probably also  just projecting. I have never been a fan, and am still not a fan, of saving undo history in the project. IMHO it's a massive amount of unnecessary complexity for something that hardly any users ever find useful (or even realize is there), and the only practical effect of is to bloat project files with excessive amounts of history data that will never be used for anything.)

ferdnyc: Hm. This reminds me, if default projects are going to be just regular project files potentially saved right from OpenShot, I should make sure `project_data.py` resets the **ID** on the file, when it creates a "new" project from the defaults. I had [originally envisioned the user defaults](#2863) as a hand-crafted file (probably created from a copy of `settings/_default.project` from the installed OpenShot, which is how I originally made mine), something so-called "power users" would install  manually if they wanted to customize OpenShot's defaults.

But that was both unrealistic and unnecessarily tedious, and just saving a project configured the way you want right from the UI makes much more sense.

ferdnyc: Oh, good. It does.

https://github.com/OpenShot/openshot-qt/blob/7bcebec75b398ab226cf890443fc808ceb4a61fc/src/classes/project_data.py#L337-L338

Well, then we're fine.

ferdnyc: With that addressed, going to merge this. Thanks!

PR #4473, opened 2021-10-07
Minor refactor of export dialog, to prevent many different crashes

jonoomph:
Minor refactor of export dialog, to prevent many different crashes and issues (with accessing variables before init). I still want to completely refactor this dialog, to make it much simpler and easier to test... but this is a start.

### DIFF
--- a/src/windows/export.py
+++ b/src/windows/export.py
@@ -819,6 +819,12 @@ class Export(QDialog):
         # Apply mappers to timeline readers
         self.timeline.ApplyMapperToClips()
 
+        # Initialize
+        max_frame = 0
+
+        # Precision of the progress bar
+        format_of_progress_string = "%4.1f%% "
+
         # Create FFmpegWriter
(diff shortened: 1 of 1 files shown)


src/windows/export.py
```diff
@@ -924,12 +925,12 @@ def titlestring(sec, fps, mess):
                         format_of_progress_string
                     )
 
-                    # track largest frame processed
-                    max_frame = frame
-
                     # Process events (to show the progress bar moving)
                     QCoreApplication.processEvents()
 
+                # track largest frame processed
```
JacksonRG, line 931: Is this supposed to be largest frame, or "last" frame?

JacksonRG (approved):

src/windows/export.py
jonoomph: Technically, it's both (since we are using a range, and always incrementing the value)

PR #4475, opened 2021-10-07
Anamorphic fixes to video widget

jonoomph:
A few more fixes for anamorphic video profile support.

### DIFF
--- a/src/windows/video_widget.py
+++ b/src/windows/video_widget.py
@@ -77,27 +77,19 @@ class VideoWidget(QWidget, updates.UpdateInterface):
             if action.type == "load" and action.values.get("pixel_ratio"):
                 pixel_ratio_changed = True
                 self.pixel_ratio = openshot.Fraction(
-                    action.values.get("pixel_ratio").get("num", 16),
-                    action.values.get("pixel_ratio").get("den", 9))
+                    action.values.get("pixel_ratio").get("num", 1),
+                    action.values.get("pixel_ratio").get("den", 1))
                 log.info(
                     "Set video widget pixel aspect ratio to: %s",
                     self.pixel_ratio.ToFloat())
             elif action.key and action.key[0] == "pixel_ratio":
                 pixel_ratio_changed = True
                 self.pixel_ratio = openshot.Fraction(
-                    action.values.get("num", 16),
-                    action.values.get("den", 9))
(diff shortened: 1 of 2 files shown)


src/windows/video_widget.py
```diff
@@ -77,28 +77,20 @@ def changed(self, action):
...
+                    action.values.get("num", 1),
+                    action.values.get("den", 1))
                 log.info(
                     "Update: Set video widget pixel aspect ratio to: %s",
                     self.pixel_ratio.ToFloat())
 
-            # Update max size (to size of video preview viewport)
-            if display_ratio_changed or pixel_ratio_changed:
-                timeline = get_app().window.timeline_sync.timeline
-                timeline.SetMaxSize(
```
jonoomph, line 97: This call does not seem needed anymore

src/windows/video_widget.py
```diff
@@ -77,28 +77,20 @@ def changed(self, action):
             if action.type == "load" and action.values.get("pixel_ratio"):
                 pixel_ratio_changed = True
                 self.pixel_ratio = openshot.Fraction(
-                    action.values.get("pixel_ratio").get("num", 16),
```
jonoomph, line 80: pixel_ratios should default to 1/1, if not found

src/windows/video_widget.py
```diff
@@ -77,28 +77,20 @@ def changed(self, action):
...
-                    action.values.get("pixel_ratio").get("den", 9))
+                    action.values.get("pixel_ratio").get("num", 1),
+                    action.values.get("pixel_ratio").get("den", 1))
                 log.info(
                     "Set video widget pixel aspect ratio to: %s",
                     self.pixel_ratio.ToFloat())
             elif action.key and action.key[0] == "pixel_ratio":
                 pixel_ratio_changed = True
```
JacksonRG: I searched the project, and the only use of `pixel_ratio_changed` and `display_ratio_changed` is in the if statement that got deleted. So this and the initialization of those two bools should probably be taken out.

JacksonRG (approved):
2 variables that aren't being used anymore. Best get rid of those for future clarity.

Other than that, LG

src/windows/video_widget.py
jonoomph: Good catch! Removing them now

PR #4489, opened 2021-10-14
Fix Sentry crash on old SDK

edumucelli:
Openshot fails to start due to missing Sentry methods.

```
➜  openshot-qt    
Loaded modules from: /usr/lib/python3/dist-packages/openshot_qt
  File "/usr/bin/openshot-qt", line 11, in <module>
    load_entry_point('openshot-qt==2.6.0.dev0', 'gui_scripts', 'openshot-qt')()
  File "/usr/lib/python3/dist-packages/openshot_qt/launch.py", line 178, in main
    sentry.init_tracing()
  File "/usr/lib/python3/dist-packages/openshot_qt/classes/sentry.py", line 69, in init_tracing
    sdk.set_tag("system", platform.system())
AttributeError: module 'sentry_sdk' has no attribute 'set_tag'
```

This happens on my machine after updating Openshot from 2.4.3 to 2.6.0, Ubuntu 20.04 using the stable PPA. The problem is that the `set_tag`, `set_user`, etc methods do not exist on sentry-sdk older than `0.13.1`.

For the moment I have commented those methods from `/usr/lib/python3/dist-packages/` on my local installation to make Openshot run. I have then applied this PR locally.

Adding `packaging` is the safest way to handle semver parsing, I have added it to `freeze.py`, but I do not know if that is needed to be added anywhere else.

As a first PR to the project I imagine there might be other aspects I have not covered, just wanted to give it a try to see if you guys were interested.

### DIFF
--- a/freeze.py
+++ b/freeze.py
@@ -93,7 +93,6 @@ python_packages = ["os",
                    "zmq",
                    "webbrowser",
                    "json",
-                   "packaging"
                    ]
 
 # Modules to include
--- a/freeze.py
+++ b/freeze.py
@@ -93,6 +93,7 @@ python_packages = ["os",
                    "zmq",
                    "webbrowser",
                    "json",
+                   "packaging"
                    ]
 
 # Modules to include
--- a/src/classes/sentry.py
+++ b/src/classes/sentry.py
@@ -27,6 +27,7 @@
  """
 
 import platform
+from packaging import version
 
 from classes.logger import log
 from classes import info
@@ -66,13 +67,14 @@ def init_tracing():
         release=f"openshot@{info.VERSION}",
         environment=environment
     )
-    sdk.set_tag("system", platform.system())
-    sdk.set_tag("machine", platform.machine())
-    sdk.set_tag("processor", platform.processor())
-    sdk.set_tag("platform", platform.platform())
-    if distro and platform.system() == "linux":
-        sdk.set_tag("distro", " ".join(distro.linux_distribution()))
-    sdk.set_tag("locale", info.CURRENT_LANGUAGE)
+    if _supports_tagging():
+        sdk.set_tag("system", platform.system())
+        sdk.set_tag("machine", platform.machine())
+        sdk.set_tag("processor", platform.processor())
+        sdk.set_tag("platform", platform.platform())
+        if distro and platform.system() == "linux":
+            sdk.set_tag("distro", " ".join(distro.linux_distribution()))
+        sdk.set_tag("locale", info.CURRENT_LANGUAGE)
 
 
 def disable_tracing():
@@ -82,16 +84,26 @@ def disable_tracing():
 
 
 def set_tag(*args):
-    if sdk:
+    if sdk and _supports_tagging():
         sdk.set_tag(*args)
 
 
 def set_user(*args):
-    if sdk:
+    if sdk and _supports_tagging():
         sdk.set_user(*args)
 
 
 def set_context(*args):
-    if sdk:
+    if sdk and _supports_tagging():
         sdk.set_context(*args)
 
+
+def _supports_tagging():
+    """Returns whether the imported sentry-sdk has tag-related
+    methods such as set_tag, set_user, set_context.
+
+    Those methods were introduce on 0.13.1 version. Checking this
+    before calling those methods on the sentry-sdk avoids crashing
+    Openshot in case an old sdk is installed.
+    """
+    return version.parse(sdk.consts.VERSION) >= version.parse("0.13.1")
--- a/src/classes/sentry.py
+++ b/src/classes/sentry.py
@@ -27,7 +27,6 @@
  """
 
 import platform
-from packaging import version
 
 from classes.logger import log
 from classes import info
@@ -68,13 +67,23 @@ def init_tracing():
         environment=environment
     )
     if _supports_tagging():
-        sdk.set_tag("system", platform.system())
-        sdk.set_tag("machine", platform.machine())
-        sdk.set_tag("processor", platform.processor())
-        sdk.set_tag("platform", platform.platform())
-        if distro and platform.system() == "linux":
-            sdk.set_tag("distro", " ".join(distro.linux_distribution()))
-        sdk.set_tag("locale", info.CURRENT_LANGUAGE)
+        configure_platform_tags(sdk)
+    else:
+        sdk.configure_scope(platform_scope)
+
+
+def platform_scope(scope):
+    configure_platform_tags(scope)
+
+
+def configure_platform_tags(sdk_or_scope):
+    sdk_or_scope.set_tag("system", platform.system())
+    sdk_or_scope.set_tag("machine", platform.machine())
+    sdk_or_scope.set_tag("processor", platform.processor())
+    sdk_or_scope.set_tag("platform", platform.platform())
+    if distro and platform.system() == "linux":
+        sdk_or_scope.set_tag("distro", " ".join(distro.linux_distribution()))
+    sdk_or_scope.set_tag("locale", info.CURRENT_LANGUAGE)
 
 
 def disable_tracing():
@@ -106,4 +115,4 @@ def _supports_tagging():
     before calling those methods on the sentry-sdk avoids crashing
     Openshot in case an old sdk is installed.
     """
-    return version.parse(sdk.consts.VERSION) >= version.parse("0.13.1")
+    return all([hasattr(sdk, method) for method in ["set_tag", "set_user", "set_context"]])


src/classes/sentry.py
```diff
@@ -82,16 +84,26 @@ def disable_tracing():
...
+
+def _supports_tagging():
+    """Returns whether the imported sentry-sdk has tag-related
+    methods such as set_tag, set_user, set_context.
+
+    Those methods were introduce on 0.13.1 version. Checking this
+    before calling those methods on the sentry-sdk avoids crashing
+    Openshot in case an old sdk is installed.
+    """
+    return version.parse(sdk.consts.VERSION) >= version.parse("0.13.1")
```
ferdnyc: Only thing I'll note here is that, because we don't require sentry to be installed at all, `sdk` can be `None`.

Every (current) call to `_supports_tagging()` comes after an `if sdk`, so that's not necessarily a problem, though it might be safer to check it here, too? (I don't necessarily trust future developers to _not_ add a call without checking `sdk` first.)

What actually might be a problem is, because `packaging` isn't part of the standard library, _it_ probably has to be optional too. I just know there will be _some_ system OpenShot runs on where it's not installed.

What do you think of something like this (up top)?
```python3
try:
    from packaging.version import parse as parseversion
except ModuleNotFoundError:
    parseversion = str
```
That way, if `packaging` is somehow missing, `parseversion(...)` will just fall back to string comparisons.

(Hell, it may even be missing on our own build servers, though I can't imagine how it could've avoided getting pulled in by something. But we've never required it because we can't _use_ it on our version strings, which are nonstandard. :disappointed: )

It's required by `sip`, which ups the odds, but not necessarily by `PyQt5.sip` in the runtime.

ferdnyc: (I'm also fine with a shorter name like `parsever`/`parse_ver`/`parse_v`/whatever. I actually hate `parseversion`, but it's the most explicit option.)

freeze.py
```diff
@@ -93,6 +93,7 @@
                    "zmq",
                    "webbrowser",
                    "json",
+                   "packaging"
```
ferdnyc: Nit: trailing comma, please.
```suggestion
                   "packaging",
```

ferdnyc (commented):
Thanks @edumucelli , I hadn't realized that part of the API was so new. I wonder if there was some alternative way of applying the same contextual detail to Sentry telemetry, or if it was simply not supported, period?

ferdnyc:
@edumucelli I pushed a commit to install the packaging module in our CI builds, so you'll probably want to pull the branch before making any changes locally.

src/classes/sentry.py
edumucelli: Hey @ferdnyc thanks for the throughout explanation of the trade-offs present on Openshot current setup, that is very valuable for me. One thing I have just considered is using packaging as you have described, but in case it is not present (`except ModuleNotFoundError`), have a custom implementation of version comparison.

I have take all the tags on `sentry-sdk` and tested against this implementation below, which is strongly inspired by [this](https://stackoverflow.com/a/1714190/914874) approach -- minus the fact that tags may contain strings such as `preview`.

```
import re


def versioncmp(version1, version2):
    def cmp(a, b):
        return (a > b) - (a < b)

    def normalize(v):
        parsed = []
        for x in re.sub(r'(\.0+)*$', '', v).split("."):
            try:
                parsed.append(int(x))
            except ValueError:
                continue
        return parsed

    return cmp(normalize(version1), normalize(version2))


if __name__ == "__main__":
    for version in [
        "0.1.0", "0.1.0-preview1", "0.1.0-preview10", "0.1.0-preview11", "0.1.0-preview12", "0.1.0-preview13",
        "0.1.0-preview14", "0.1.0-preview15", "0.1.0-preview16", "0.1.0-preview2", "0.1.0-preview3",
        "0.1.0-preview4", "0.1.0-preview5", "0.1.0-preview6", "0.1.0-preview7", "0.1.0-preview8",
        "0.1.1", "0.1.2", "0.1.3", "0.10.0", "0.10.1", "0.10.2", "0.11.0", "0.11.1", "0.11.2", "0.12.0",
        "0.12.1", "0.12.2", "0.12.3", "0.13.0", "0.13.1", "0.13.2", "0.13.3", "0.13.4", "0.13.5",
        "0.14.0", "0.14.1", "0.14.2", "0.14.3", "0.14.4", "0.15.0", "0.15.1", "0.16.0", "0.16.1",
        "0.16.2", "0.16.3", "0.16.4", "0.16.5", "0.17.0", "0.17.1", "0.17.2", "0.17.3", "0.17.4",
        "0.17.5", "0.17.6", "0.17.7", "0.17.8", "0.18.0", "0.19.0", "0.19.1", "0.19.2", "0.19.3",
        "0.19.4", "0.19.5", "0.2.0", "0.2.1", "0.2.2", "0.20.0", "0.20.1", "0.20.2", "0.20.3",
        "0.21.0-beta.0", "0.3.0", "0.3.1", "0.3.10", "0.3.11", "0.3.2", "0.3.3", "0.3.4", "0.3.5",
        "0.3.6", "0.3.7", "0.3.8", "0.3.9", "0.4.0", "0.4.1", "0.4.2", "0.4.3", "0.5.0", "0.5.1",
        "0.5.2", "0.5.3", "0.5.4", "0.5.5", "0.6.0", "0.6.1", "0.6.2", "0.6.3", "0.6.4", "0.6.5",
        "0.6.6", "0.6.7", "0.6.8", "0.6.9", "0.7.0", "0.7.1", "0.7.10", "0.7.11", "0.7.12", "0.7.13",
        "0.7.14", "0.7.2", "0.7.3", "0.7.4", "0.7.5", "0.7.6", "0.7.7", "0.7.8", "0.7.9", "0.8.0",
        "0.8.1", "0.9.0", "0.9.1", "0.9.2", "0.9.3", "0.9.4", "0.9.5", "1.0.0", "1.1.0", "1.1.0-beta.0",
        "1.2.0", "1.3.0", "1.3.1", "1.4.0", "1.4.1", "1.4.2", "1.4.3"
    ]:
        target_version = "0.13.1"
        comparison = versioncmp(target_version, version)
        if comparison == 0:
            print(f"{target_version} == {version}")
        elif comparison < 0:
            print(f"{target_version} < {version}")
        else:
            print(f"{target_version} > {version}")
```

The result is correct for all existing sentry sdk tags:

```
0.13.1 > 0.1.0
0.13.1 > 0.1.0-preview1
0.13.1 > 0.1.0-preview10
0.13.1 > 0.1.0-preview11
0.13.1 > 0.1.0-preview12
0.13.1 > 0.1.0-preview13
0.13.1 > 0.1.0-preview14
0.13.1 > 0.1.0-preview15
0.13.1 > 0.1.0-preview16
0.13.1 > 0.1.0-preview2
0.13.1 > 0.1.0-preview3
0.13.1 > 0.1.0-preview4
0.13.1 > 0.1.0-preview5
0.13.1 > 0.1.0-preview6
0.13.1 > 0.1.0-preview7
0.13.1 > 0.1.0-preview8
0.13.1 > 0.1.1
0.13.1 > 0.1.2
0.13.1 > 0.1.3
0.13.1 > 0.10.0
0.13.1 > 0.10.1
0.13.1 > 0.10.2
0.13.1 > 0.11.0
0.13.1 > 0.11.1
0.13.1 > 0.11.2
0.13.1 > 0.12.0
0.13.1 > 0.12.1
0.13.1 > 0.12.2
0.13.1 > 0.12.3
0.13.1 > 0.13.0
0.13.1 == 0.13.1
0.13.1 < 0.13.2
0.13.1 < 0.13.3
0.13.1 < 0.13.4
0.13.1 < 0.13.5
0.13.1 < 0.14.0
0.13.1 < 0.14.1
0.13.1 < 0.14.2
0.13.1 < 0.14.3
0.13.1 < 0.14.4
0.13.1 < 0.15.0
0.13.1 < 0.15.1
0.13.1 < 0.16.0
0.13.1 < 0.16.1
0.13.1 < 0.16.2
0.13.1 < 0.16.3
0.13.1 < 0.16.4
0.13.1 < 0.16.5
0.13.1 < 0.17.0
0.13.1 < 0.17.1
0.13.1 < 0.17.2
0.13.1 < 0.17.3
0.13.1 < 0.17.4
0.13.1 < 0.17.5
0.13.1 < 0.17.6
0.13.1 < 0.17.7
0.13.1 < 0.17.8
0.13.1 < 0.18.0
0.13.1 < 0.19.0
0.13.1 < 0.19.1
0.13.1 < 0.19.2
0.13.1 < 0.19.3
0.13.1 < 0.19.4
0.13.1 < 0.19.5
0.13.1 > 0.2.0
0.13.1 > 0.2.1
0.13.1 > 0.2.2
0.13.1 < 0.20.0
0.13.1 < 0.20.1
0.13.1 < 0.20.2
0.13.1 < 0.20.3
0.13.1 < 0.21.0-beta.0
0.13.1 > 0.3.0
0.13.1 > 0.3.1
0.13.1 > 0.3.10
0.13.1 > 0.3.11
0.13.1 > 0.3.2
0.13.1 > 0.3.3
0.13.1 > 0.3.4
0.13.1 > 0.3.5
0.13.1 > 0.3.6
0.13.1 > 0.3.7
0.13.1 > 0.3.8
0.13.1 > 0.3.9
0.13.1 > 0.4.0
0.13.1 > 0.4.1
0.13.1 > 0.4.2
0.13.1 > 0.4.3
0.13.1 > 0.5.0
0.13.1 > 0.5.1
0.13.1 > 0.5.2
0.13.1 > 0.5.3
0.13.1 > 0.5.4
0.13.1 > 0.5.5
0.13.1 > 0.6.0
0.13.1 > 0.6.1
0.13.1 > 0.6.2
0.13.1 > 0.6.3
0.13.1 > 0.6.4
0.13.1 > 0.6.5
0.13.1 > 0.6.6
0.13.1 > 0.6.7
0.13.1 > 0.6.8
0.13.1 > 0.6.9
0.13.1 > 0.7.0
0.13.1 > 0.7.1
0.13.1 > 0.7.10
0.13.1 > 0.7.11
0.13.1 > 0.7.12
0.13.1 > 0.7.13
0.13.1 > 0.7.14
0.13.1 > 0.7.2
0.13.1 > 0.7.3
0.13.1 > 0.7.4
0.13.1 > 0.7.5
0.13.1 > 0.7.6
0.13.1 > 0.7.7
0.13.1 > 0.7.8
0.13.1 > 0.7.9
0.13.1 > 0.8.0
0.13.1 > 0.8.1
0.13.1 > 0.9.0
0.13.1 > 0.9.1
0.13.1 > 0.9.2
0.13.1 > 0.9.3
0.13.1 > 0.9.4
0.13.1 > 0.9.5
0.13.1 < 1.0.0
0.13.1 < 1.1.0
0.13.1 < 1.1.0-beta.0
0.13.1 < 1.2.0
0.13.1 < 1.3.0
0.13.1 < 1.3.1
0.13.1 < 1.4.0
0.13.1 < 1.4.1
0.13.1 < 1.4.2
0.13.1 < 1.4.3
```

Thus we could either skip `packaging` altogether and use this or we could try the `packaging` import and fallback on this. The trade-off here is that by removing packaging it is one module less to take into consideration on the build process, etc. On the other side using the custom implementation may have some edge case is not present on any of the existing tags. Using both seems to be the strongest option, but brings more code into the codebase :-) Maybe we can start just by using the custom implementation. Let me know how you feel about those options.

ferdnyc: @edumucelli I think that could work, but actually won't be necessary.

I did some digging into my earlier question about how this worked _before_, and I discovered that all these methods were present in the earlier SDK versions as well.

Before they streamlined the API, `set_tag()`, `set_context()`, `set_extra()`, etc. were methods on `sentry_sdk.scope.Scope` objects. To use them, after calling `sentry_sdk.init()` you'd call `sentry_sdk.configure_scope(fn)` — for some reason, it took a callback as an argument, and that callback was called with the current scope to act on it. so, for example:
```python3
def scopeit(scope):
    scope.set_tag(...)
    scope.set_context(...)

sentry_sdk.init()
sentry_sdk.configure_scope(scopeit)
```

...A bit more roundabout, so I can see why they streamed it. But the Scope-based API was there... well, I went as far back as `0.1.0` and it already had it.

So I think that's the route to go. And rather than dealing with version checks, we should  just use `if hasattr(sdk, "set_tag")` and  the like to determine whether we can call those methods directly, or if we have to use `configure_scope()`.

I suppose for completeness we could also check `if hasattr(sdk, "configure_scope")` though it does seem to have always been there. (And is still there, in the most current release, as an alternative to the convenience methods.)

edumucelli: That is great! Thanks for the investigation. I will work on it then considering this alternative, which is way less intrusive.

freeze.py
edumucelli: Thanks for the suggestion. Resolved.

src/classes/sentry.py
edumucelli: Changed accordingly @ferdnyc, please take a look. The `if hasattr(sdk, "configure_scope")` I have not added. I have double checked the library until earliest as you did and indeed it does exist, so I think it should be safe.

edumucelli:
@ferdnyc would you guys be interested on adding `mypy` typing into the codebase? I could slowly work on it, that would give multiple benefits (safer refactoring, easier to navigate on the code, etc), please refer to [this](https://sethmlarson.dev/blog/2021-10-18/tests-arent-enough-case-study-after-adding-types-to-urllib3) that has been done on `urllib3`.

ferdnyc (approved):
@edumucelli This looks great! I force-downgraded my system to `sentry_sdk==0.1.0` and everything started up fine, so all seems good.

Thanks for sorting this out, crashers are never fun. I've already rebased the branch onto `develop` and squashed a few commits, merging now.

ferdnyc:
@edumucelli Regarding type hints... Well, my thoughts on that are complicated, so I'll just lay things out and then you'll have a picture of the situation. (From my highly-biased perspective, of course.)

I'm of two minds about the value of type hinting in the OpenShot code, for a couple of reasons (both of which can be related to that blog post):
1. **`openshot-qt` isn't library code.**
    A big motivator for the urllib3 team in adding type hints was that users of their library wanted to type-check their own code which imported urllib3. That's a clear motivator, and supporting your users is always a worthwhile goal. OpenShot, as an enduser application, can stand to benefit its users in a lot of ways, but adding type hints is none of them. At least not _directly_.
    To the extent that type hints might help unearth bugs and make for a more robust codebase, obviously the users indirectly benefit from that. But the direct need of, "Oh, our users would really like to have this, we'd be improving the library for them" isn't really there.

1. **Lack of type hints is far from OpenShot's biggest issue.** It's not even the biggest issue with the _code_.
    The urllib3 post discusses all the benefits they saw from adding type hints to their codebase, _despite_ already having a unit test suite with an insane number — 1800 I think it was? — of test cases and 100% test coverage. In the context of a mature, well-designed library with carefully curated APIs like that, I'm sure type checking is a nice tool to add to the arsenal.
    OpenShot has 11 anemic test cases that all cover _one_ module (`src/classes/query.py`), and that is the sum total of OpenShot's entire test suite. Other than that, testing consists entirely of "run the program and manually try to do stuff, so we can see if anything breaks" ad-hoc QA. I doubt we even _know_ what types most of our APIs are **expected** to be passing, because without unit tests discovering those expectations would come down to grepping the codebase for all calls to each function, and determining how they use it. Which... sounds slightly _less_ obviously useful than the urrllib3 team's experience, let's just put it that way.

I'd never discourage positive contributions to the codebase, of any sort. In fact, I've already introduced some type annotations to a couple of modules — `src/classes/style_tools.py`, `src/classes/keyframe_scaler.py` — just to sort of "test the waters". Nobody _objected_, but that's because nobody really cared or noticed. (Nor, it seems, did I **myself** care all that much.  Adding hints is nice and all, but if I'd actually run that code through `mypy` — as I now have — I'd have discovered that there are several type violations. So, the only classes we have with any type hints still don't have _correct_ typing. Ugh.)
 
The introduction of type hints into two new modules also led to a subsequent 0.00% increase in the use of hinting across the rest of the codebase. So the "our whole team pulled together to make this happen" aspect of the urllib3 effort is also unlikely to translate. The value of one (or two, or three) coders working to add type hints while the rest of the developers ignore or even (inadvertently) _remove_ them in their own commits strikes me as... uncertain, at best.

Like I said: My intent is not to discourage contributions in any form (unless they negatively impact the code), and if the question is whether type hint additions would be accepted: Yes, absolutely I would happily merge any PRs that add type hints to our code. But in terms of systematically upgrading the entirety of OpenShot's codebase, and the value of doing that... I have to be honest and say that if I made a list of all the improvements the code needs, type hints _might_ make the top **hundred** but they'd be real close to the bottom.

edumucelli:
> @edumucelli Regarding type hints... Well, my thoughts on that are complicated, so I'll just lay things out and then you'll have a picture of the situation. (From my highly-biased perspective, of course.)
> 
> I'm of two minds about the value of type hinting in the OpenShot code, for a couple of reasons (both of which can be related to that blog post):
> 
>     1. **`openshot-qt` isn't library code.**
>        A big motivator for the urllib3 team in adding type hints was that users of their library wanted to type-check their own code which imported urllib3. That's a clear motivator, and supporting your users is always a worthwhile goal. OpenShot, as an enduser application, can stand to benefit its users in a lot of ways, but adding type hints is none of them. At least not _directly_.
>        To the extent that type hints might help unearth bugs and make for a more robust codebase, obviously the users indirectly benefit from that. But the direct need of, "Oh, our users would really like to have this, we'd be improving the library for them" isn't really there.

I see, that is understandable, my example of urrlib3 was just because it was a public example. At the company I work we have an important project (that is also not a library) and has been completed type checked with `mypy --strict` check on the CI. The gains are important when it comes to peace of mind once you need to refactor things. 

> 
>     2. **Lack of type hints is far from OpenShot's biggest issue.** It's not even the biggest issue with the _code_.
>        The urllib3 post discusses all the benefits they saw from adding type hints to their codebase, _despite_ already having a unit test suite with an insane number — 1800 I think it was? — of test cases and 100% test coverage. In the context of a mature, well-designed library with carefully curated APIs like that, I'm sure type checking is a nice tool to add to the arsenal.
>        OpenShot has 11 anemic test cases that all cover _one_ module (`src/classes/query.py`), and that is the sum total of OpenShot's entire test suite. Other than that, testing consists entirely of "run the program and manually try to do stuff, so we can see if anything breaks" ad-hoc QA. I doubt we even _know_ what types most of our APIs are **expected** to be passing, because without unit tests discovering those expectations would come down to grepping the codebase for all calls to each function, and determining how they use it. Which... sounds slightly _less_ obviously useful than the urrllib3 team's experience, let's just put it that way.
> 
> 
> I'd never discourage positive contributions to the codebase, of any sort. In fact, I've already introduced some type annotations to a couple of modules — `src/classes/style_tools.py`, `src/classes/keyframe_scaler.py` — just to sort of "test the waters". Nobody _objected_, but that's because nobody really cared or noticed. (Nor, it seems, did I **myself** care all that much. Adding hints is nice and all, but if I'd actually run that code through `mypy` — as I now have — I'd have discovered that there are several type violations. So, the only classes we have with any type hints still don't have _correct_ typing. Ugh.)
> 
> The introduction of type hints into two new modules also led to a subsequent 0.00% increase in the use of hinting across the rest of the codebase. So the "our whole team pulled together to make this happen" aspect of the urllib3 effort is also unlikely to translate. The value of one (or two, or three) coders working to add type hints while the rest of the developers ignore or even (inadvertently) _remove_ them in their own commits strikes me as... uncertain, at best.

I imagine the task is overwhelming, when adding typing to one, two files and seeing hundreds with no typing at all. I'd slowly add them, but also instrument the CI to break if the set of correctly typed files have any typing regression or mismatch. So, even if you have 10 files with the correct typing, following developers won't remove them or break them by mistake. But I understand that openshot codebase has more important tasks in priority.

> 
> Like I said: My intent is not to discourage contributions in any form (unless they negatively impact the code), and if the question is whether type hint additions would be accepted: Yes, absolutely I would happily merge any PRs that add type hints to our code. But in terms of systematically upgrading the entirety of OpenShot's codebase, and the value of doing that... I have to be honest and say that if I made a list of all the improvements the code needs, type hints _might_ make the top **hundred** but they'd be real close to the bottom.

Thanks for very complete view of the current situation. :+1:

PR #4490, opened 2021-10-15
Adding new keyframe SVG icons to timeline (circle for bezier, diamond…

jonoomph:
- Adding new keyframe SVG icons to timeline (circle for bezier, diamond for linear, square for constant)
- Adding click on keyframe point (to seek to that spot)
- Improving performance on some Angular keframe data getters
- Fixing keyframe HTML to be better organized, and better styled.

Screenshot from 2021-10-14 20-14-03

### DIFF
--- a/src/timeline/index.html
+++ b/src/timeline/index.html
@@ -69,7 +69,7 @@
 			<div ng-repeat="layer in project.layers.slice().reverse()" id="track_static_{{layer.number}}" ng-right-click="showTrackMenu(layer.id)" class="track_name">
 				<div class="track_top">
 					<div tl-clip-menu class="track_menu" ng-mousedown="showTrackMenu(layer.id)"></div>
-					<span class="track_label">{{getTrackName(layer.label, project.layers.length - $index)}}</span><span ng-show="layer.lock" class="track_lock"></span>
+					<span class="track_label">{{getTrackName(layer.label, project.layers.length - $index)}}</span><span ng-if="layer.lock" class="track_lock"></span>
 				</div>
 			</div>
 			<br>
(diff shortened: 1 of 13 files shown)


src/timeline/index.html
```diff
@@ -84,42 +84,48 @@
...
+						<img class="thumb thumb-start" ng-if="getThumbPath(clip)" ng-src="{{ getThumbPath(clip) }}"/>
 					</div>
-					<div ng-show="clip.show_audio" class="audio-container">
+					<div ng-if="clip.show_audio" class="audio-container">
 						<canvas tl-audio height="46px" width="{{canvasMaxWidth((clip.end - clip.start) * pixelsPerSecond)}}px" class="audio"></canvas>
 					</div>
+
+                    <!-- CLIP KEYFRAME POINTS -->
+                    <div class="point_region" ng-if="clip.selected">
+                        <div ng-repeat="(point, value) in getKeyframes(clip)" ng-click="selectPoint(clip, point)" id="point_{{clip.id}}_{{point}}_{{$index}}" class="point point_{{ value }}" style="left: {{(((point - 1) / (project.fps.num / project.fps.den) - clip.start) * pixelsPerSecond)}}px;"></div>
```
ferdnyc: Just a style suggestion, to start: _please_ wrap insanely long lines. It's almost impossible to read the code on GitHub when lines are 300+ characters long.

```suggestion
                        <div ng-repeat="(point, value) in getKeyframes(clip)"
                          ng-click="selectPoint(clip, point)"
                          id="point_{{clip.id}}_{{point}}_{{$index}}"
                          class="point point_{{ value }}"
                          style="left: {{(((point - 1) / (project.fps.num / project.fps.den) - clip.start) * pixelsPerSecond)}}px;"></div>
```

ferdnyc: <s>(Because everything _is_ so wide, I shortened the variables to `p` and `val`. The code's right there, it's not like anyone will forget what they mean. They aren't ever visible outside of this single HTML tag's scope. Still, YMMV on that.)</s>
Well, I _thought_ I had. :confused:

src/timeline/media/css/main.css
```diff
@@ -134,16 +134,47 @@ img {
...
-    width: 2px;
+.point_bezier {
+    position: absolute;
+    width: 10px;
+    height: 10px;
+    background-image: url(../images/keyframe-bezier.svg);
+    background-repeat: no-repeat;
+    margin: 0;
+    cursor: pointer;
+}
```
ferdnyc, line 150: Another suggestion: We can drop the SVG icons entirely, they're overkill for such simple shapes.
```suggestion
.point_bezier {
    position: absolute;
    width: 8px;
    height: 8px;
    border-radius: 8px;
    background-color: #adadad;
    border: 1px solid #adadad;
    margin: 1px;
    cursor: pointer;
}
```
Tadahhh! It's a circle. But one we can recolor with just styling changes, apply effects to in realtime, etc.

src/timeline/media/css/main.css
```diff
@@ -134,16 +134,47 @@ img {
...
+    position: absolute;
+    width: 10px;
     height: 10px;
-    background: linear-gradient(180deg, #0f08, #0f0);
-    margin: 0 3px;
+    background-image: url(../images/keyframe-linear.svg);
+    background-repeat: no-repeat;
+    margin: 0;
+    cursor: pointer;
+}
```
ferdnyc, line 160: Ditto this one (I can't suggest because there are deleted lines in the middle...)

```css
.point_linear {
    position: absolute;
    width: 8px;
    height: 8px;
    background-color: #adadad;
    transform: rotate(45deg);
    margin: 1px;
    cursor: pointer;
}

File: src/timeline/media/css/main.css [hunk 4]
```diff
@@ -134,16 +134,47 @@ img {
...
+
+.point_constant {
+    position: absolute;
+    width: 10px;
+    height: 10px;
+    background-image: url(../images/keyframe-constant.svg);
+    background-repeat: no-repeat;
+    margin: 0;
+    cursor: pointer;
+}
```
ferdnyc, line 170: And this is just `_linear` without the rotation:
```suggestion
.point_constant {
    position: absolute;
    width: 8px;
    height: 8px;
    background-color: #adadad;
    margin: 1px;
    cursor: pointer;
}
```

src/timeline/media/css/main.css
ferdnyc: > one we can recolor with just styling changes, apply effects to in realtime, etc.

(Yes, technically modern CSS allows you to do all that with SVG images as well, but it's still a lot of unnecessary overhead reading a file from disk and rendering it, when the div that contains it can just be shaped and colored in place.)

ferdnyc (commented):
Nice, I like it!

We don't really need to bother with the SVG files, though...

src/timeline/index.html
ferdnyc: That double division is hard on the brain, too.

What if we inverted the FPS fraction and multiplied, instead? (Might be more efficient, though I bet the math library optimizes it that way anyway so it probably doesn't matter performance-wise.)

There's also a superfluous pair of outer parens in there:
```html
<div ... 
     style="left: {{((point - 1) * (project.fps.den / project.fps.num) - clip.start)
                    * pixelsPerSecond}}px;">
</div>
```

src/timeline/media/css/main.css
jonoomph: Good idea, but in practice, this is actually why I started working on this one. Our AppImage has issues with some of the more advanced CSS, such as transforms and rotation. In fact, the simple green lines we had before don't appear on many distros running our AppImage (using web engine). I have no idea why not. I spent an entire day going down the rabbit hole on that one. Replacing them with simpler, SVG icons worked great though... but again, no disagreement... just a wild, confusing issue with some CSS not rendering correctly on certain distros.

src/timeline/index.html
jonoomph: Good suggestion, I'll wrap things up nicer

src/timeline/media/css/main.css
ferdnyc: @jonoomph 

Hrm. I don't know why that would be happening, offhand, but it seems worth tracking down instead of working around. I know the Timeline has no issues with any CSS or JS I throw at it, running on Qt 5.15.2 from Fedora's packages.

Are the distros reporting this all ancient-history ones, like Ubuntu 18.04 and CentOS? I know the AppImage is packaged with Qt 5.9, when QtWebEngine was still in its relative infancy. It's possible just upgrading to Qt 5.15 would solve these issues, and probably others.

...In fact, if the Qt being bundled is the one from Ubuntu, another thing to try would be to package from a downloaded Qt binary distribution, instead. Not to knock Ubuntu's packagers (OK, maybe a little...), it's more about age than quality, and Qt's 5.15 builds can still be run on older systems.

The Qt binary builds also have `RPATH`s set in the libraries, and use those to link between the various pieces, which is a much more robust solution than the AppImage's `LD_LIBRARY_PATH` trickery. Not that we wouldn't still need that for other libraries, but at least the Qt build we'd know we can safely and completely relocate anywhere we want with just a few lines in `qt.conf`.

(The distros strip out `RPATH`s from their Qt builds and render the libraries un-relocatable. You can adjust where they load plugins and things from with `qt.conf`, but the libraries will only load their pieces if the system linker knows where to find them.)

Qt's binary distributions, all the libraries are set up like this:
```sh
$ cd ~/Qt/5.15.2/lib
$ chrpath -l libQt5Widgets.so
libQt5Widgets.so: RUNPATH=$ORIGIN
```
So they'll always load the other pieces from the same directory, no matter where it is. Which is why, even though the Qt5 libs are also installed in my `/usr/lib64/`, this happens:
```sh
# No problems finding libraries at the original location
$ ldd libQt5WebEngineCore.so|grep Qt
	libQt5Quick.so.5 => /home/ferd/Qt/5.15.2/gcc_64/lib/./libQt5Quick.so.5 (0x00007f81d49c4000)
	libQt5Gui.so.5 => /home/ferd/Qt/5.15.2/gcc_64/lib/./libQt5Gui.so.5 (0x00007f81d4093000)
	libQt5QmlModels.so.5 => /home/ferd/Qt/5.15.2/gcc_64/lib/./libQt5QmlModels.so.5 (0x00007f81d3e13000)
	libQt5WebChannel.so.5 => /home/ferd/Qt/5.15.2/gcc_64/lib/./libQt5WebChannel.so.5 (0x00007f81d3bef000)
	libQt5Qml.so.5 => /home/ferd/Qt/5.15.2/gcc_64/lib/./libQt5Qml.so.5 (0x00007f81d3573000)
	libQt5Network.so.5 => /home/ferd/Qt/5.15.2/gcc_64/lib/./libQt5Network.so.5 (0x00007f81d31c3000)
	libQt5Positioning.so.5 => /home/ferd/Qt/5.15.2/gcc_64/lib/./libQt5Positioning.so.5 (0x00007f81d2f20000)
	libQt5Core.so.5 => /home/ferd/Qt/5.15.2/gcc_64/lib/./libQt5Core.so.5 (0x00007f81d272a000)
	libicui18n.so.56 => /home/ferd/Qt/5.15.2/gcc_64/lib/./libicui18n.so.56 (0x00007f81d14f0000)
	libicuuc.so.56 => /home/ferd/Qt/5.15.2/gcc_64/lib/./libicuuc.so.56 (0x00007f81d1138000)
	libicudata.so.56 => /home/ferd/Qt/5.15.2/gcc_64/lib/./libicudata.so.56 (0x00007f81cf753000)
#
# Relocate them somewhere else, and...
$ mkdir /var/tmp/qt_test; cp * /var/tmp/qt_test/
$ cd /var/tmp/qt_test
#
# Still good
$ ldd libQt5WebEngineCore.so.5 |grep -i Qt
	libQt5Quick.so.5 => /var/tmp/qt_test/./libQt5Quick.so.5 (0x00007faed2a3a000)
	libQt5Gui.so.5 => /var/tmp/qt_test/./libQt5Gui.so.5 (0x00007faed2109000)
	libQt5QmlModels.so.5 => /var/tmp/qt_test/./libQt5QmlModels.so.5 (0x00007faed1e89000)
	libQt5WebChannel.so.5 => /var/tmp/qt_test/./libQt5WebChannel.so.5 (0x00007faed1c65000)
	libQt5Qml.so.5 => /var/tmp/qt_test/./libQt5Qml.so.5 (0x00007faed15e9000)
	libQt5Network.so.5 => /var/tmp/qt_test/./libQt5Network.so.5 (0x00007faed123b000)
	libQt5Positioning.so.5 => /var/tmp/qt_test/./libQt5Positioning.so.5 (0x00007faed0f96000)
	libQt5Core.so.5 => /var/tmp/qt_test/./libQt5Core.so.5 (0x00007faed07a0000)
	libicui18n.so.56 => /var/tmp/qt_test/./libicui18n.so.56 (0x00007ff7203d4000)
	libicuuc.so.56 => /var/tmp/qt_test/./libicuuc.so.56 (0x00007ff72001c000)
	libicudata.so.56 => /var/tmp/qt_test/./libicudata.so.56 (0x00007ff71e639000)
```
That trick will never work with Ubuntu's (or any distro's) packaged library builds, which makes it a lot harder to use them in things like AppImages.

We _would_ have to replace the Ubuntu PyQt5 as well, either with Riverbank's wheels or ones we built ourselves against the downloaded Qt. I _think_ the former might work as long as libopenshot was linked with the same Qt release (I **feel** like I've been able to run OpenShot off PyQt5 wheels before, by doing it carefully). But if I'm misremembering that, I now have a fair amount of experience building PyQt5, enough to be able to make our own build work. (And it's certainly the **safer** option, to be using libopenshot and PyQt5 linked with the exact same Qt libs. Even if they still can't easily communicate with each other across the sip/SWIG divide.)

ferdnyc:
@jonoomph Hrm, the above was originally attached to a bit of the code, which GitHub didn't see fit to include here. So, for the record I'm talking about this code in `value_updated`:

https://github.com/OpenShot/openshot-qt/blob/869ffb856c7c18099fd4baad22348a976554fd31/src/windows/models/properties_model.py#L465-L478

That `if point.get("interpolation", 2) == 0` means that whenever the `previous_point_x` interpolation is NOT Bezier, its `handle_right` will be removed. Even if  the current `interpolation` for `closest_point_x` **is** Bezier.

ferdnyc:
@jonoomph 

Also, while you're focused on the timeline stuff, think you might be able to look at #4466? It's very much in the same vein as this one.

Also, because it'd be nice to get _some_ kind of feedback on **one** of my PRs, considering how many I have gathering dust. Unless you'd just prefer I stop submitting them? If that's the intent here... well, I can't say "message received" because clearly not. But I'm at least beginning to consider the possibility that I've just been failing to read between the lines.

I don't want to step on anyone's toes with submissions that aren't useful, nor do I want to waste my time working on things that'll never make it into the code. (I have already done a _lot_ of that simply because it's my norm to work on something only to ultimately second-guess myself and decide it's not worth submitting. But even the ones I _do_ feel are worth submitting kind of feel like a waste of my time and everyone else's, these days.)

PR #4493, opened 2021-10-18
Fix restoring of window geometry on startup

ferdnyc:
For a while  now, OpenShot's saved window geometry has experienced "creep" — the dimensions that were restored for the docks scaled slightly wrong each time, so that the timeline would drift shorter and shorter, the video preview would get narrower and narrower. Just a bit each time, but over time it built up.

Turns out, the issue was that we were restoring the geometry _before_  the window's `show()` , which means the window size isn't finalized yet so Qt makes something up and scales to that.

For the same reason, it's also important to restore the geometry _before_ the state, as applying the window state first will also cause the geometry to scale wrong in the other direction (docks grow with each launch).

This PR changes the timing and ordering of the saved values being applied, so that OpenShot loads with the same window geometry each launch (unless resized by the user).

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -2928,7 +2928,8 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
         self.clearSelections()
 
         # Show Property timer
-        # Timer to use a delay before showing properties (to prevent a mass selection from trying
+        # Timer to use a delay before showing properties
+        # (to prevent a mass selection from trying
         # to update the property model hundreds of times)
         self.show_property_id = None
         self.show_property_type = None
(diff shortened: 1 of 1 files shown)


ferdnyc:
> For the same reason, it's also important to restore the geometry _before_ the state, as applying the window state first will also cause the geometry to scale wrong in the other direction (docks grow with each launch).

AIUI, the reason for this, specifically, is that Qt now (since mid-`5.x`) stores/restores geometry based on the _screen_ size — as if the application was fullscreened or the window was maximized. That way, even if you run the application on a different size display, you'll get the same layout, just scaled to the current screen.

It's also why, if you only _use_ OpenShot in fullscreen/maximized mode, you've probably never noticed this issue, as it's effectively hidden when running the application at screen size.

To restore everything at the correct size for all windowing modes, we need to:
1. Get the window fully laid out (implicitly at screen size)
2. _Then_ apply the saved geometry
3. _Then_ restore any adjustments to those assumptions, like the fact that the interface is un-maximized

If we restore state too soon, then we apply fullscreen-sized geometry to a non-fullscreen window and everything is too big.

ferdnyc:
I've tested this on Windows and Linux, and it seems to do the right thing pretty much all the time. Even when I've given it a hard time, like (say):
1. maximized my window
2. exited the program
3. made my screen _larger_ (yay virtual machines)
4. re-launched

The window comes up maximized to fill the new screen area, with the layout correctly adjusted. 2.6.1-release doesn't even restore the maximized state in that scenario.

I'm going to merge this. Hopefully it doesn't shit the bed on macOS.

PR #4497, opened 2021-10-20
Fix hyperlinks in About dialog, elsewhere?

ferdnyc:
@scootergrisen pointed out in #4270 that our links in the About dialog were looking sort of poorly. That's because we had a lot of HTML tags mis-nested (`<br>` inside `<a>`, etc.), and also because we were doing furious amounts of styling directly on tags.

This PR sets the hyperlink color for the dark palette when `ui_utils` first sets it up, eliminating the need for any color styling of link text. It also replaces a lot of HTML `<span>`s with styled `<p> `, styled `<div>`, and styled `<a>`, where appropriate. (Putting `<span>`s and `<br/>`s inside `<a>`s to do formatting is wonky HTML that causes endless problems.)


ferdnyc:
"After" of the main window, on Windows 10:

image

JacksonRG (approved):
LGTM!

PR #4499, opened 2021-10-21
WIP: Remove unused code

edumucelli:
There is lots of unused code (variables, parameters, declarations). I have removed most of them, they may sign that something is wrong, in case that they actually should be used but it was not, or just because someone forgot it.

There might be also de case in which for completeness (or framework-related expectations due to PyQt) some that are not used should be there such as the never used `*args` or `checked`.


src/windows/preferences.py
```diff
@@ -560,7 +559,7 @@ def dropdown_index_changed(self, widget, param, index):
         # Check for restart
         self.check_for_restart(param)
 
-    def testHardwareDecode(self, all_decoders, decoder, decoder_card="0"):
+    def testHardwareDecode(self, all_decoders, decoder, decoder_card=0):
```
edumucelli, line 562: `decoder_card` default value was string, when it should be int. That is the only "functional" change.

src/windows/add_to_timeline.py
```diff
@@ -49,7 +49,7 @@ class AddToTimeline(QDialog):
 
     ui_path = os.path.join(info.PATH, 'windows', 'ui', 'add-to-timeline.ui')
 
-    def btnMoveUpClicked(self, checked):
+    def btnMoveUpClicked(self):
```
ferdnyc: Yeah, these should all stay, `slot(bool)` is the signature for `QAction.triggered` and `QAction.clicked` signals, which is what these slots are connected to.

src/windows/add_to_timeline.py
```diff
@@ -379,7 +379,7 @@ def accept(self):
         # Accept dialog
         super(AddToTimeline, self).accept()
 
-    def ImageLengthChanged(self, value):
+    def ImageLengthChanged(self):
```
ferdnyc: I believe that's also a slot, and if it's connected to a signal with a payload then it'll need to take the arg whether it uses it or not.

src/windows/main_window.py
```diff
@@ -1008,7 +1008,7 @@ def actionSaveFrame_trigger(self, checked=True):
         self.setStatusBar(self.statusBar)
 
         # Determine path for saved frame - Default export path
-        recommended_path = recommended_path = os.path.join(info.HOME_PATH)
```
ferdnyc, line 1011: I have never understood how or why that was like it was. I suspect just a double copy-paste error. But I know I fixed it in at least three PRs, which I guess nobody ever looked at so they were never merged.

src/windows/main_window.py
```diff
@@ -1069,8 +1069,6 @@ def actionSaveFrame_trigger(self, checked=True):
         self.cache_object.Clear()
         self.timeline_sync.timeline.SetCache(old_cache_object)
         self.cache_object = old_cache_object
-        old_cache_object = None
-        new_cache_object = None
```
ferdnyc, line 1073: Can you tell we have some C programmers obsessed with nulling variables contributing?

src/windows/main_window.py
```diff
@@ -2185,7 +2182,7 @@ def actionTutorial_trigger(self):
             self.tutorial_manager.exit_manager()
             self.tutorial_manager = TutorialManager(self)
 
-    def actionInsertTimestamp_trigger(self, event):
+    def actionInsertTimestamp_trigger(self):
```
ferdnyc: ARRRGH. Someone is obsessed with the idea that slots are event handlers, and keeps creating functions that claim to take `event` arguments — they're not event handlers, and they don't receive `event` args.

However, this _**is**_ also the slot for a `QAction` signal, and as such will be called with a `checked` boolean. So this should actually be _changed_ to `checked`.

src/windows/models/properties_model.py
```diff
@@ -237,11 +235,6 @@ def remove_keyframe(self, item):
                     log.debug("Found point to delete at X=%s" % point_to_delete["co"]["X"])
                     keyframe["Points"].remove(point_to_delete)
 
-            # Reduce # of clip properties we are saving (performance boost)
-            clip_data = {property_key: clip_data[property_key]}
-            if object_id:
-                clip_data = {'objects': {object_id: clip_data}}
-
```
ferdnyc: This should stay, only because the fact that it's Doing It Wrong™ is an outstanding bug that needs to be fixed. Removing the code will only make it harder to demonstrate the problem (which is that the code here does nothing).

src/windows/models/properties_model.py
```diff
@@ -355,11 +348,6 @@ def color_update(self, item, new_color, interpolation=-1, interpolation_details=
                                 'interpolation': 1,
                                 })
 
-                # Reduce # of clip properties we are saving (performance boost)
-                clip_data = {property_key: clip_data[property_key]}
-                if object_id:
-                    clip_data = {'objects': {object_id: clip_data}}
-
```
ferdnyc, line 362: Same.

src/windows/models/properties_model.py
```diff
@@ -556,15 +544,9 @@ def value_updated(self, item, interpolation=-1, value=None, interpolation_detail
...
                         clip_object.Close()
-                        clip_object = None
                     except Exception as ex:
                         log.warn('Invalid Reader value passed to property: %s (%s)' % (value, ex))
 
-            # Reduce # of clip properties we are saving (performance boost)
-            clip_data = {property_key: clip_data.get(property_key)}
-            if object_id:
-                clip_data = {'objects': {object_id: clip_data}}
-
```
ferdnyc, line 567: And again.

src/windows/preferences.py
ferdnyc: I suspect the original plan (and possibly future plan) was that `decoder_card` would be parsed from our settings data, in which case it would likely be a string. That's why it's explicitly wrapped in an `int()` when it's used later. You're right, though, as it's currently called it's only ever passed `0` or `1`.

The other possibility was that someone was under the impression that `dict` keys have to be strings, which of course isn't true but a lot of people believe it anyway for some reason.

src/windows/views/blender_listview.py
```diff
@@ -219,7 +219,7 @@ def dropdown_index_changed(self, widget, param, index):
         if param["name"] == "length_multiplier":
             self.init_slider_values()
 
-    def color_button_clicked(self, widget, param, index):
+    def color_button_clicked(self, widget, param):
```
ferdnyc, line 222: Speaking of unused, the next two lines totally are as well, in this function.

src/windows/video_widget.py
```diff
@@ -411,7 +409,7 @@ def centeredViewport(self, width, height):
         else:
             return QRect(((width - widthFromHeight) / 2.0) + left_padding, top_padding, widthFromHeight, height)
 
-    def present(self, image, *args):
+    def present(self, image):
```
ferdnyc: Hmm, this is a slot called from C++ code, I'm a bit nervous about changing its signature.

src/windows/add_to_timeline.py
ferdnyc: (Not so much that it's _required_ to, as PyQt is forgiving, but I think it's worthwhile to shame coders when they ignore data being handed to them gift-wrapped with a little bow on top, then proceed to look up that same data later on, only using less efficient means. Which is _exactly_ what's happening here. If these slots were written sanely then they _would_ make use of the `value` argument. I mean... this particular piece of code is in the running for dumbest function in the entire codebase. It exists **solely** to _throw away_ its argument before calling the real update function.)

I just feel like dumb code should be allowed to be blatantly, obviously dumb.

src/windows/views/properties_tableview.py
```diff
@@ -752,7 +739,6 @@ def contextMenuEvent(self, event):
                 # Add sub-choice items (for nested choice lists)
                 # Divide into smaller QMenus (since large lists cover the entire screen)
                 # For example: Transitions -> 1 -> sub items
-                SubMenu = None
```
ferdnyc, line 755: *sigh* I know I attempted fixed a bunch of these. Can't remember if that's in one of my still-open PRs, or one of the ones I had to close because conflicting changes were committed without looking at my code, and they left the PR with merge conflicts too big and messy to untangle.

Heck, I've rewritten basically the entirety of `properties_model.py` a couple of times over now, in un-merged PRs.

Aaaanyway... that's beside the point, really. As well as the reason I'm looking this over. No sense repeating history.

src/windows/title_editor.py
```diff
@@ -132,7 +132,7 @@ def __init__(self, *args, edit_file_path=None, duplicate=False, **kwargs):
             # Display image (slight delay to allow screen to be shown first)
             QTimer.singleShot(50, self.display_svg)
 
-    def txtLine_changed(self, txtWidget):
+    def txtLine_changed(self):
```
ferdnyc, line 135: Ugh. The _reasons_ this doesn't need the widget are terrible ones, and a sign that this entire module is in need to an overhaul. But as things stand, it's definitely not used. :+1:

ferdnyc (changes requested):
Thanks @edumucelli ! This looks promising. What else can I say, but... "Yep."

We have a lot of cut-and-paste code here, as if that wasn't obvious. (There are some modules that contain entire identical _functions_.)

In my more cynical moments I suspect that copy-paste code duplication makes up at least 60% of the repo. (It can be over 50% because some code is duplicated 3, 4, 5 times.) So, yeah, there's plenty of opportunity for cleanup. Which is tedious work, so thanks again for pitching in.

I'm not too keen on changing the signatures of functions that are called as Qt slots, however. (Like everything that takes a `checked` argument in `main_window.py`) PyQt5 doesn't require explicitly _decorating_ functions as slots, in order to use them that way. (So, any function can be a "stealth" slot.)

If it's connected to a signal that carries a payload, then personally I'd prefer to explicitly receive that payload and ignore it, rather than silently discarding it.

(Especially since there are a lot of those where we _should_ be using the value passed with the signal, but currently don't for Reasons™.)

The ones that take unused `*args` are probably fine to drop (except for one particular one I flagged in a comment), and of course any unused/clobbered variables can all go for sure.

ferdnyc:
@edumucelli 

Actually, you want to talk about unused code? `transfer_children` at the bottom of `classes.ui_util` is never, ever called, anywhere. Worse, if it _were_ called it wouldn't even **do anything** except log a message _lying_ about how it's transferring widgets from one class to another. I have no idea why it's there but I'd love to see it go away.

src/windows/add_to_timeline.py
edumucelli: Alright, thanks for the insight, in reality I know nothing about PyQt (more broadly Qt).

src/windows/add_to_timeline.py
edumucelli: Indeed, it worked fine (I have built and run) things after those changes, the screens opened with no problem, so you are right: it is forgiving. Thanks for comment.

src/windows/main_window.py
edumucelli: Thanks for the comment. Indeed, someone may have copied it and it appeared twice. This time it is going away :+1:

src/windows/main_window.py
edumucelli: Thanks for the comment, will address it.

src/windows/models/properties_model.py
edumucelli: I understand your point, but I wonder if that will make any positive effect by letting it here. This code is there since 2015 and it seems that it did not served to improve things. It will only make it hard to following people reading the code. WDYT?

src/windows/video_widget.py
edumucelli: Thanks for the comment. Ok, will keep it.

src/windows/views/blender_listview.py
edumucelli: Thanks for the comment, will change accordingly.

edumucelli:
> `transfer_children`

Thanks, I have also removed `conversion.py`, it had two methods that were never used.

edumucelli:
> @edumucelli
> 
> Actually, you want to talk about unused code? `transfer_children` at the bottom of `classes.ui_util` is never, ever called, anywhere. Worse, if it _were_ called it wouldn't even **do anything** except log a message _lying_ about how it's transferring widgets from one class to another. I have no idea why it's there but I'd love to see it go away.

@ferdnyc can you please re-take a look and let me know if that is good enough after the changes. Thanks!

src/windows/models/properties_model.py
ferdnyc: @edumucelli That's definitely not right, `object_id` and the code around it was only added with the OpenCV changes earlier this year (or maybe the very end of 2020). Which is also when this code was broken.

(The _previous_, very-similar code did in fact do the right thing, reducing the `clip.data` `dict`'s contents directly. Whereas this code accidentally _replaces_ `clip_data` (a variable that initially contains a reference to `clip.data`) with the new, reduced `dict`. `clip_data` is then garbage-collected and the optimized `dict` with it, while `clip.data` remains unoptimized. ...But that part of this code is a relatively recent situation.)

ferdnyc: My mistake, `clip_data` is initially a reference to **`c.data`**, which the old code updated directly when optimizing. Everything else I said holds.

src/classes/ui_util.py
```diff
@@ -288,10 +288,3 @@ def center(window):
     centerPoint = get_app().window.frameGeometry().center()
     frameGm.moveCenter(centerPoint)
     window.move(frameGm.topLeft())
-
-
-def transfer_children(from_widget, to_widget):
-    log.info(
-        "Transferring children from '%s' to '%s'",
-        from_widget.objectName(), to_widget.objectName())
```
ferdnyc, line 296: \o/ Die, lying function, die!

src/windows/models/properties_model.py
edumucelli: I am refraining to change behavior on this PR, wanted only to remove what was unused, but it seems that unused code is actually made my mistake or is untested and broken, so it is difficult to know what is unused and what is unused because it is broken. Should this once again revert the previous behavior?

```
if object_id:
  c.data = {'objects': {object_id: clip_data}}
else:
  c.data = clip_data
```

ferdnyc: @edumucelli Possibly, in some cases. I suspect it should really be something like...
```python3
# Reduce # of clip properties we are saving (performance boost)
if object_id:
    clip_data = {property_key: clip_data[property_key]}
    c.data = {'objects': {object_id: clip_data}}
else:
    c.data = {property_key: clip_data[property_key]}
```
Only thing is, I'm not 100% confident that the `object_id` case will do the right thing, when optimized that way.

As you say, though, changing behavior in this PR would muddy its focus. Best to just not touch the code in question, I'd say. (In _this_ PR, I mean. This can certainly be fixed in a different PR.)

jonoomph:
@edumucelli Thanks for working on this! Have you tested each of these modifications?  Are we assuming that nothing will break?  This one scares me, without running it locally, and testing all the various changes.

edumucelli:
I have not tested any of the modifications because what I am doing here is
removing unused code, so if it's not used at run time, I do not know how to
write tests for code that has been removed, if that's what you mean.

I am not used to the codebase, sorry, but the only way to test that I know
of is writing actual tests, please let me know what you mean by testing if
that is not the case.

But honestly, I do not see how one should test removal of code. I see the
codebase has no tests so I am curious what is the current approach.

I can close the pull request if that change is scary, but I think it is a
pity let all this dead code here, let me know what you think.

Eduardo -- http://eduar.do

edumucelli:
As there is no interest on such cleanup changes I am closing it. Good luck!

PR #4501, opened 2021-10-21
A rewrite of test_translations, adding new features & flexibility

ferdnyc:
I figure if we're gonna do this, let's really do this. The new `test_translations.py` is modular Python containing functions which can be imported, or it can be run as a standalone script. When run as a script, it will exit with a nonzero exit code when errors are encountered. If imported, the `scan_all()` function will raise an exception on error.

Filenames for one or many `.qm` files can be given on the command line in script mode, to check a certain specific file or group of files. If no arguments are given, it falls back to the same method of scanning the directory, then checks all files.

### Errors detected

The error checking is greatly expanded, and now checks all of the following conditions when comparing source strings to translations:
  - The count of `%s`, `%d`, and `%f` escapes is the same for each
  - If any <code>%(<var>var</var>)<var>x</var></code> named placeholders are used:
    - The translation must use all of the same variable names (<code><var>var</var></code>)
    - The type code (<code><var>x</var></code>) for each named variable must be the same

Raw count of `%` characters is NOT used. It's far too inaccurate, both creating false errors _and_ missing real ones. (Like translations containing `% s` instead of `%s` — we have several of those, it turns out.)

- Counts failure to load any translation file as an error, which will result in non-zero exit status

### Example output

When all is well, it just reports success in green, after a summary of the source string and translation file counts (that's always output, for both success and failure):

image

If there are problems, they're now shown grouped by `.qm` file, with the name of the `.pot` source shown next to each source string, followed by the bad translation. Here are results from a translation file I deliberately inserted errors into:

image

### `color` utility class
The color-formatting escapes are now provided via a `color` class, which has classmethods (`color.yellow()` and `color.green()` are added to `color.red()`) that will decorate the string representation of its arguments and return an escape-wrapped string. So, for example, you can do this:

image

### DIFF
--- a/src/language/test_translations.py
+++ b/src/language/test_translations.py
@@ -30,52 +30,153 @@ import os
 import re
 import fnmatch
 import sys
-from PyQt5.QtCore import QLocale, QLibraryInfo, QTranslator, QCoreApplication
+from PyQt5.QtCore import QTranslator, QCoreApplication
 
 
-# Get the absolute path of this project
-language_path = os.path.dirname(os.path.abspath(__file__))
+# Absolute path of the translations directory
+LANG_PATH = os.path.dirname(os.path.abspath(__file__))
 
-# Color for errors
-red='\033[31m'
-endc='\033[0m'
+# Match '%(name)x' format placeholders
+TAG_RE = re.compile(r'%\(([^\)]*)\)(.)')
 
-found_errors = False
+
+class color:
+    """Color for message output"""
+    _red = '\u001b[31m'
+    _yellow = '\u001b[33m'
+    _green = '\u001b[32m'
+    _reset = '\u001b[0m'
+
+    @classmethod
+    def red(cls, *args):
+        return cls._red + str(*args) + cls._reset
+
+    @classmethod
+    def yellow(cls, *args):
+        return cls._yellow + str(*args) + cls._reset
+
+    @classmethod
+    def green(cls, *args):
+        return cls._green + str(*args) + cls._reset
+
 
 # Get app instance
 app = QCoreApplication(sys.argv)
 
-# Load POT template (all English strings)
-all_templates = ['OpenShot.pot', 'OpenShot_transitions.pot', 'OpenShot_blender.pot']
-for template_name in all_templates:
-    POT_source = open(os.path.join(language_path, 'OpenShot', template_name)).read()
-    all_strings = re.findall('^msgid \"(.*)\"', POT_source, re.MULTILINE)
 
-    print("Testing {} strings in {}...".format(len(all_strings), template_name))
+def build_stringlists() -> dict:
+    """ Create a dict containing lists of strings, keyed on source filename"""
+    all_strings = {}
 
-    # Loop through folders/languages
-    for filename in fnmatch.filter(os.listdir(language_path), 'OpenShot*.qm'):
-        lang_code = filename[:-3]
-        # Install language
-        translator = QTranslator(app)
-        app.installTranslator(translator)
+    for pot in [
+        'OpenShot.pot',
+        'OpenShot_transitions.pot',
+        'OpenShot_blender.pot',
+        'OpenShot_emojis.pot',
+    ]:
 
-        # Load translation
-        success = translator.load(lang_code, language_path)
-        if not success:
-            print(red, '%s-%s' % (success, lang_code), endc)
+        with open(os.path.join(LANG_PATH, 'OpenShot', pot)) as f:
+            data = f.read()
+        all_strings.update({
+            pot: re.findall('^msgid \"(.*)\"', data, re.MULTILINE)
+        })
+    return all_strings
 
-        # Loop through all test strings
-        for source_string in all_strings:
-            if "%s" in source_string or "%s(" in source_string or "%d" in source_string:
-                translated_string = app.translate("", source_string)
-                if source_string.count('%') != translated_string.count('%'):
-                    found_errors = True
-                    print(red, '\tInvalid string replacement found: "%s" vs "%s" [%s]' %
-                          (translated_string, source_string, lang_code), endc)
 
-        # Remove translator
-        app.removeTranslator(translator)
+def check_trans(strings: list) -> list:
+    """Check all strings in a list against a given .qm file"""
+    # Test translation of all strings
+    translations = {
+        source: app.translate("", source)
+        for source in strings
+    }
+    # Check for replacements with mismatched number of % escapes
+    errors = {
+        s: t for s, t in translations.items()
+        if any([
+            s.count('%s') != t.count('%s'),
+            s.count('%d') != t.count('%d'),
+            s.count('%f') != t.count('%f')])
+    }
+    # Check for missing/added variable names
+    # e.g.: "%(clip_id)s %(value)d" changed to "%(clip)s %(value)d"
+    # or mismatched types
+    # e.g.: "%(seconds)s" changed to "%(seconds)d"
+    named_variables = {
+        s: (TAG_RE.findall(s), TAG_RE.findall(t))
+        for s, t in translations.items()
+        if s.count('%(') > 0
+    }
+    errors.update({
+        s: translations[s]
+        for s, (s_vars, t_vars) in named_variables.items()
+        if sorted(s_vars) != sorted(t_vars)
+    })
+    return list(errors.items())
 
-if found_errors:
-    raise(Exception("Errors detected during translation testing! See above."))
+
+def process_qm(file, stringlists) -> int:
+    """Scan a translation file against all provided strings"""
+    # Attempt to load translation file
+    basename = os.path.splitext(file)[0]
+    translator = QTranslator(app)
+    if not translator.load(basename, LANG_PATH):
+        print(color.red('QTranslator failed to load') + f' {file}')
+        return 1
+
+    app.installTranslator(translator)
+
+    # Build a dict mapping source POTfiles to lists of error pairs
+    error_sets = {
+        sourcefile: check_trans(strings)
+        for sourcefile, strings in stringlists.items()
+    }
+
+    app.removeTranslator(translator)
+
+    # Display any errors found, grouped by source POT file
+    error_count = sum([len(v) for v in error_sets.values()])
+    if error_count:
+        print(f'{file}: ' + color.red(f'{error_count} total errors'))
+    for pot, errset in error_sets.items():
+        if not errset:
+            continue
+        width = len(pot)
+        msg = "Invalid"
+        for source, trans in errset:
+            print(color.yellow(f'{pot}:') + f' {source}')
+            print(color.red(f'{msg:>{width}}:') + f' {trans}\n')
+    return error_count
+
+
+def scan_all(filenames: list = None) -> int:
+    all_strings = build_stringlists()
+    if not filenames:
+        filenames = fnmatch.filter(os.listdir(LANG_PATH), 'OpenShot*.qm')
+    # Loop through language files and count errors
+    total_errors = sum([
+        process_qm(filename, all_strings)
+        for filename in filenames
+    ])
+
+    string_count = sum([len(s) for s in all_strings.values()])
+    lang_count = len(filenames)
+
+    print(f"Tested {color.yellow(string_count)} strings on "
+          + f"{color.yellow(lang_count)} translation files.")
+    if total_errors > 0:
+        msg = f"Found {total_errors} translation errors! See above."
+        raise Exception(msg)
+    return sum([])
+
+
+# Autorun if used as script
+if __name__ == '__main__':
+    try:
+        string_count = scan_all(sys.argv[1:])
+    except Exception as ex:
+        print(color.red(ex))
+        exit(1)
+    else:
+        print(color.green("No errors found!"))
+        exit(0)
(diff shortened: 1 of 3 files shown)


ferdnyc:
And here are real results, run over all of the language files on the `develop` branch:

image

So, a few problems we didn't know about, including whatever translator is inserting spaces into the `%` escapes. And one or more who seem to occasionally _add_ a `%` escape that isn't in the source string?

ferdnyc:
(The CI failures are a sign the code is _working_, and not passing the translations due to the problems I just pointed out. Which is actually why that job should really be in a separate workflow, come to think of it, rather than part of the main build. (It's also not necessary to run it in each of the matrix builds, the results won't be any different in ubuntu 18.04 vs 20.04.)

ferdnyc:
My added commit moves the translation-file scanning into a separate job within the CI workflow, it runs independently and will pass/fail separately from the build jobs (which can in turn pass/fail independent of the translation checks).

src/language/test_translations.py
```diff
@@ -30,52 +30,153 @@
...
+LANG_PATH = os.path.dirname(os.path.abspath(__file__))
 
-# Color for errors
-red='\033[31m'
-endc='\033[0m'
+# Match '%(name)x' format placeholders
+TAG_RE = re.compile(r'%\(([^\)]*)\)(.)')
+
+
+class color:
```
edumucelli: This really looks line an Enum to me, a set of name constants, I'd suggest using the std Enum for it.

edumucelli:
```suggestion
class Color:
```

src/language/test_translations.py
```diff
@@ -30,52 +30,153 @@
...
-                    print(red, '\tInvalid string replacement found: "%s" vs "%s" [%s]' %
-                          (translated_string, source_string, lang_code), endc)
-
-        # Remove translator
-        app.removeTranslator(translator)
-
-if found_errors:
-    raise(Exception("Errors detected during translation testing! See above."))
+
+def build_stringlists() -> dict:
```
edumucelli: I'd suggest using the `typing` package. Although we have discussed and you thoroughly explained you think that is not a priority, but since you are making an effort here, I'd do it using `typing.Dict`, we may profit from your effort here in the future. Actually a `Dict[str, List]` if I got what the method is doing.

src/language/test_translations.py
```diff
@@ -30,52 +30,153 @@
...
+
+    # Display any errors found, grouped by source POT file
+    error_count = sum([len(v) for v in error_sets.values()])
+    if error_count:
+        print(f'{file}: ' + color.red(f'{error_count} total errors'))
+    for pot, errset in error_sets.items():
+        if not errset:
+            continue
+        width = len(pot)
+        msg = "Invalid"
```
edumucelli: I think this `msg` does not need to be reset on every iteration of the loop, I would suggest moving it to the outside.

src/language/test_translations.py
```diff
@@ -30,52 +30,153 @@
...
+    ])
+
+    string_count = sum([len(s) for s in all_strings.values()])
+    lang_count = len(filenames)
+
+    print(f"Tested {color.yellow(string_count)} strings on "
+          + f"{color.yellow(lang_count)} translation files.")
+    if total_errors > 0:
+        msg = f"Found {total_errors} translation errors! See above."
+        raise Exception(msg)
```
edumucelli: Maybe creating an specific Exception to the semantic more clear, e.g., `TranslationError` and raising it here. I am not sure if the codebase has already some policy on defining exceptions.

edumucelli (commented):
I won't approve because I do not feel legitimate to approve anything on the project, but I think this looks neat and improves a lot from the current state.

src/language/test_translations.py
```diff
@@ -30,52 +30,153 @@
...
+    # e.g.: "%(seconds)s" changed to "%(seconds)d"
+    named_variables = {
+        s: (TAG_RE.findall(s), TAG_RE.findall(t))
+        for s, t in translations.items()
+        if s.count('%(') > 0
+    }
+    errors.update({
+        s: translations[s]
+        for s, (s_vars, t_vars) in named_variables.items()
+        if sorted(s_vars) != sorted(t_vars)
```
JacksonRG: (Admittedly minor nitpick): My only question on this is if a syntactical difference in languages requires a repeat of a variable (probably repeating a name to remove ambiguity, not repeating a number). Would it make sense to compare the sets of variables rather than the lists?

src/language/test_translations.py
```diff
@@ -30,52 +30,153 @@
...
+        for source in strings
+    }
+    # Check for replacements with mismatched number of % escapes
+    errors = {
+        s: t for s, t in translations.items()
+        if any([
+            s.count('%s') != t.count('%s'),
+            s.count('%d') != t.count('%d'),
+            s.count('%f') != t.count('%f')])
+    }
```
JacksonRG, line 104: I like how simple this is to read. Didn't know the { `statement` `for ...` `if ...` } could be used to create dicts and lists! Though I've used it as a combined `map` and `filter` function before.

JacksonRG (approved):
Just 1 question about a very edge case. Approving anyways because I don't think we currently allow that case either.

src/language/test_translations.py
ferdnyc: @JacksonRG Hmm... no, that's a really good point. I'd considered the need to reorder variables (hence the use of `sorted()` when comparing the lists), but not that a variable might be repeated in the source language but not the target (or vice versa, I suppose).

Now that you've got me thinking about it, there's also the possibility that a variable might not make sense in the target language and need to be dropped. It is legal for a `%`-interpolation to _not use_ all of the variables it's passed, so really `t_vars` only needs to be a _strict subset_ of `s_vars` to be valid. 

If I compare using _sets_ I can (a) skip the `sorted()` and (b) make the condition `not set(t_vars) <= set(s_vars)` instead of the `!=` comparison. Ooh, I like that!

src/language/test_translations.py
ferdnyc: Fair point. I suspect Python may optimize such things away when byte-compiling, but there's no sense in making it have to.

(I was originally going to just hardcode it in the f-string, but I don't think there's a way to do that and still match the width of the `pot` string. So then I was going to generate a new `len(pot)`-width version each loop, but I realized that was stupid because <code>{<var>variable</var>:><var>length</variable>}</code> exists. But I apparently stopped one optimization too soon.)

ferdnyc: (Now that I'm rereading it, I recall I also wasn't worried about optimization because this loop only fires once for each _error_ found. Ideally, it'll never run! :grin: )

src/language/test_translations.py
ferdnyc: A _**policy**_, no, definitely not. But I completely agree re: semantic clarity. So, what the hey? Let's define an exception. There's a first time for everything!

(Sadly true. Looking around, I have a couple of other worktrees with exception definitions in them, but so far nothing submitted. Though I really _should_ submit the one that replaces an abuse of `ModuleNotFoundError`.)

edumucelli: Yes, because at some point (hopefully) the business exceptions are common in multiple parts. They will show similar validation mechanisms that can grow into common code. In this case here I do not now how reusable they will be, so that is just a comment, feel free to dismiss it.

edumucelli (approved):

src/language/test_translations.py
ferdnyc: @JacksonRG Yeah, [dict comprehensions (PEP 274)](https://www.python.org/dev/peps/pep-0274/) were added relatively "late" to the Python syntax.

(They were added in Python 3.0 & 2.7, released December 2008 & July 2010 respectively. As opposed to list comprehensions, which were defined in [PEP 202](https://www.python.org/dev/peps/pep-0202/) and released as part of Python 2.0 in July 2000. ...I would _love_ it if someone dug up a reference to how they originally got the name "comprehensions", because f--- if I know. Even in PEP 202 the name is unexplained and just taken as a given.)

Since Python 2.2, `dict()` has accepted a list of `(k, v)` tuples, so it's been possible to fake dict comprehensions anyway with `dict([(k, v) for k, v in some_expression_that_yields_k_and_v])`. The dict comprehension syntax just streamlines that for clarity.

Weirdly, while list comprehensions are covered nicely [in the Python Tutorial](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions), dict and set comprehensions get the shaft there.

The thing I like best about them is, because they're totally self-contained, you can use temporary variable names like `s` and `t` without feeling too guilty, since their only uses _have_ to be **very** local. (The temporaries are evaluated in a local scope to prevent their definitions escaping the comprehension, unlike in normal unrolled loops where any variables you define are hoisted to the function level.) Being able to be temporarily concise is nice when your _non_-temporary variable names are long enough to be unambiguous and self-documenting.

Even in its unrolled, brief loop form, this feels like ugly code relative to the comprehension:
```python
errors = {}
for s, t in translations.items():
    if any([s.count('%s') != t.count('%s'),
            s.count('%d') != t.count('%d'),
            s.count('%f') != t.count('%f')
           ]):
       errors[s] = t
```
Not to mention the pointlessness of having to create the empty dict first.

If it's a longer loop that's making more complex decisions about what goes into the dict, then that's probably a sign the contents should be split out into a separate function (that can then be used in a comprehension).

src/language/test_translations.py
ferdnyc: I have typed the module fully, and remembered to run it through `mypy` this time so it's actually _right_ :laughing: It even passes `mypy --disallow-untyped-defs`. Go, me.

src/language/test_translations.py
ferdnyc: Done, and all of the calls to the classmethods updated accordingly.

ferdnyc: As far as the Enum... I mean, the `Color` class definitely isn't an enum, it's a string-processing class. 

`Color.red(*args)` returns a string that _starts with_ `Color._red`, ends with `Color._reset`, and has the string representation of `*args` in between.

I could make the color escape values a separate Enum of their own, but they're never meant to be accessed outside of the classmethods so is it worth it? It'd just be swapping out a reference to an internal class string member with a reference to a different class's internal string member, arguably at the expense of clarity.

If I'm following your suggestion, something like,
```python3
ColorEscapes = Enum(
    'ColorEscapes', {
        "RED": "\u001b[31m",
        "YELLOW": "\u001b[33m",
        "GREEN": "\u001b[32m",
        "RESET": "\u001b[0m",
    })
```
And then in the `Color` class, calls like...
```python3
@staticmethod
def red(*args):
    return ColorEscapes.RED + str(*args) + ColorEscapes.RESET
```
I'm debating whether that seems more clear or _less_ clear, TBH. And it would make the escapes visible outside the `Color` class (unless I define the enum _inside_ the `Color` class), which I'm not sure is desirable.

ferdnyc:
Rather than having the CI checking (and failing) every PR in the repo whenever there are translation issues, I moved things around again so that the "Validate translations" job only runs when paths under `src/language/` are changed. (So, _this_ PR still triggers it.)

It can also be manually run from the Actions tab, by committers. (At least, _hopefully_ that'll be the case after merge to the default branch, now that I've nailed down the syntax with [some help from the GitHub Community](https://github.community/t/problem-using-workflow-dispatch-trigger-with-other-workflow-triggers/183242).)

src/language/test_translations.py
edumucelli: Great! I can work on adding the `mypy` to the ci, we can already start adding this module to the list of `opt-in` files, so that we do not accidentally remove/revert the typing effort done here.

src/language/test_translations.py
JacksonRG: Now you're telling me set overloads the comparison operators? Oh how my list of python tricks grows. Figured it'd have to be `setA.contains(setB)` or similar. That's awesome :sunglasses:

jonoomph:
@ferdnyc This is awesome! Thanks for digging into this and improving our error detection. I'm working on fixing all these translations now, and then I'll rerun the validate check after merging the fixes.

jonoomph:
I've fixed these translations issues in develop, and if we merge develop back into this branch, that test should now pass.

jonoomph:
I'll see if I can push to your branch... not sure if that is allowed.

jonoomph:
I've decided to just merge this and test locally (since it was taking too long for me to get a local version cloned... not sure why it was going to take hours and hours to clone it... weird)

src/language/test_translations.py
ferdnyc: @JacksonRG 

Oh, sets overload, like, _all_ the operators:
```python3
>>> setA = {"red", "green", "blue"}
>>> setB = {"one", "two", "red", "blue"}
>>> setA | setB  # Union
{'blue', 'one', 'two', 'green', 'red'}
>>> setA - setB  # Difference
{'green'}
>>> setB & setA  # Intersection
{'blue', 'red'}
>>> setB ^ setA  # XOR
{'one', 'two', 'green'}
>>> (setB ^ setA) < setB
False
>>> (setB & setA) < setB
True
# (There are also modifying forms: |=, ^=, -=, &=, etc.)
```
(With weird exceptions. Like, `setA | setB` is valid, but `setA + setB` is a `TypeError`. **#ForSomeReason!**)

PR #4504, opened 2021-10-25
WIP: Point symlink at correct file

### LINKED ISSUE #4502: Broken symbolic link
**Describe the bug:**

Since commit df1e4845ff7789c2735cc4a9a6691ff782b34052, the file `images/Humanity/actions/custom/razor_line_with_razor.png` has pointed to `../../../../timeline/media/images/razor_line_with_razor.png`, which does not exist.

This relative path should be adjusted to `../../../../src/timeline/media/images/razor_line_with_razor.png`

JacksonRG:
Closes #4502,

@lFam noticed and reported a broken symlink.

### DIFF
--- a/images/Humanity/actions/custom/razor_line_with_razor.png
+++ b/images/Humanity/actions/custom/razor_line_with_razor.png
@@ -1,1 +1,1 @@
-../../../../timeline/media/images/razor_line_with_razor.png
+../../../../src/timeline/media/images/razor_line_with_razor.png


images/Humanity/actions/custom/razor_line_with_razor.png
```diff
@@ -1 +1 @@
-../../../../timeline/media/images/razor_line_with_razor.png
\ No newline at end of file
+../../../../src/timeline/media/images/razor_line_with_razor.png
```
ferdnyc, line 1: I've been meaning to fix these... a different/separate issue is that none of these custom-made images should be in `images/Humanity/actions/custom/` at all, since they're not part of the Humanity icon set. We've been stuffing things in there for no good reason, when it would make more sense to keep our custom icons completely separate from Humanity. (That would allow us to update it, or let the distros drop it since they use theme icons, or take only the pieces we actually _need_ from it instead of pulling in the entire set.)

But that's a bigger thing, and the symlink fix is a good one for its current location. :+1:

ferdnyc: (Worse, there also seem to be some of our custom/borrowed icons in `images/Humanity/actions/16/`, detectable as the only `.png` files in a directory full of `.svg` icons.)

ferdnyc: Oh, that's right. I _did_ fix that, in #4452. (And this as well, as it happens.)

JacksonRG: @ferdnyc Cool I'll take a look at #4452. maybe we can get it rebased and merged

ferdnyc: @JacksonRG I sent @jonoomph a blackmail-email earlier today with that one on my list of demands :smiling_imp: , so anything's possible. That's a bigger thing, though, so it makes sense to merge this quick and have it done. I can easily conflict-resolve the one line in #4452 it'll affect, afterwards.

ferdnyc: Plus, #4452 doesn't yet address the `.png` files in `images/Humanity/actions/16/` (I'd forgotten about those) so it should probably deal with those too, first.

jonoomph:
@JacksonRG Is this still an issue on `develop`? The High DPI branch that was recently merged addressed a bunch of icon related issues.

JacksonRG:
Closing, the file is no longer a link

Issue #4326, opened 2021-08-20
Include translation PO/TS files and remove compiled QM files


scootergrisen:
The POT files with the source strings for translations is in https://github.com/OpenShot/openshot-qt/tree/develop/src/language/OpenShot

But https://github.com/OpenShot/openshot-qt/tree/develop/src/language contains compiled QM files.
Which are not plain text files so i can't see the translated strings here in the source it seems. Just on Launchpad.

I would think it is better to include the translated PO files or perhaps the TS files and remove the compiled QM files.
I would think you only need the QM files for the release. Like compile them everytime there is a new version of OpenShot released. Not include them in the source.

ferdnyc:
@scootergrisen 

I don't disagree, and that might indeed be a better way to do it. Right now the translation-management process is a very manual one, performed by @jonoomph, which makes it a bit opaque. There's room for improvement, no doubt.

But in the short term, `.qm` files are actually totally convertible back to text format, all you need is the `lconvert` tool from any Qt distribution. Using the Spanish translation file as an example...
```sh
# To convert .qm back to its .ts source
lconvert -i OpenShot_es.qm -o OpenShot_es.ts

# To convert into a .po file, instead
lconvert -i OpenShot_es.qm -o OpenShot_es.po
```
The full list of compatible formats is displayed in the `lconvert --help` output, and currently would be:
```sh
$ lconvert --help

Usage:
    lconvert [options] <infile> [<infile>...]

lconvert is part of Qt's Linguist tool chain. It can be used as a
stand-alone tool to convert and filter translation data files.
The following file formats are supported:

    pot   - GNU Gettext localization template files
    ts    - Qt translation sources
    qph   - Qt Linguist 'Phrase Book'
    xlf   - XLIFF localization files
    po    - GNU Gettext localization files
    qm    - Compiled Qt translations
```

For those who prefer a GUI, the contents of `.qm` files can also be browsed with Qt Linguist, the graphical translation tool that comes  with Qt.

scootergrisen:
I prefer pot/po files over ts/qm files when i translate but i guess if the software i using Qt that the translation files should use ts files.
Why are there pot/qm files on github but po/mo files on launchpad?
Does launchpad not support ts files since when i try to download from there i can only choose po/mo.

Seems the qm file does not contain context where the string is found in the source code.
Which makes sence when it is a compiled file meant to be small and not for translators to use.

The pot/po files contains the context where the string is found in the source code.

[...]

Issue #3589, opened 2020-06-30
i18n support for plural forms


gunchleoc:
The zoom slider above the timeline has the following text:

`{} seconds`

The gd translation for "seconds" can change depending on the number, e.g.

```
1 diog
2 dhiog
3 diogan
20 diog
```

This also affects Slavic languages.

I am using a workaround `{} diog(an)` for now, which makes the UI ugly.

This seems to be the only string that needs plural forms, so no need to give it high priority.

gunchleoc:
Translation catalog hasn't changed on Launchpad, so I'm assuming this hasn't been fixed.

JacksonRG:
@gunchleoc the zoom slider has been reworked. I don't see text on it. So this may have been fixed.

gunchleoc:
Here's the lines from the .pot file. I have redacted the file paths since it's configured to show the complete path including the local username.

```
#: src/windows/views/webview.py:802
#: src/windows/views/webview.py:2770
#: src/windows/main_window.py:2574
msgid "{} seconds"
msgstr ""
```
One of these 3 occurrences is still in the current source code:

```
        # Add Freeze menu options
        Time_Menu.addSeparator()
        for freeze_type, trigger_type in [
            (_("Freeze"), MENU_TIME_FREEZE),
            (_("Freeze && Zoom"), MENU_TIME_FREEZE_ZOOM)
        ]:
            Freeze_Menu = QMenu(freeze_type, self)

            for freeze_seconds in [2, 4, 6, 8, 10, 20, 30]:
                # Add menu option
                Time_Option = Freeze_Menu.addAction(_('{} seconds').format(freeze_seconds))
                Time_Option.triggered.connect(
                    partial(self.Time_Triggered, trigger_type, clip_ids, freeze_seconds, playhead_position))

            # Add menu to parent
            Time_Menu.addMenu(Freeze_Menu)
```
https://github.com/OpenShot/openshot-qt/blob/develop/src/windows/views/webview.py#L796

ferdnyc:
Yeah, the zoom_slider value won't appear in the UI at all in future releases, but the freeze times are still countable strings subject to pluralization. Amazingly enough, it seems (from a quick scan of the translation master) those may be the **only** remaining translatable strings that require pluralization rules. I have to think that's due to conscious avoidance, for reasons I'll get back to.

[...]

PR #4513, opened 2021-10-29
Updating translations: Oct 29, 2021

jonoomph:
Fixing broken translations mentioned in https://github.com/OpenShot/openshot-qt/pull/4501


ferdnyc:
@jonoomph 

There was a _fake_ **language** in the translations?!? Well, that's bizarre!

jonoomph:
@ferdnyc Yes, it appears that a user manually uploaded a `gaa` language in Launchpad.net, and internally it identifies as the `ga` language. But I can't seem to find a way to delete the `gaa` language now. :shrug:

ferdnyc:
@jonoomph Oy. We really should switch to Crowdin or some other service. Launchpad was... "fine", at the time, but there are way better options these days, from services dedicated to translation as their sole focus. 

Including plenty that support .ts files directly, so we wouldn't have to round-trip everything from .ts to .pot to .po and back to .ts again.

ferdnyc:
@jonoomph 

A user also made the very good point, a couple of months ago, that we should really be checking the .ts files into the repo, instead of (or in addition to) the compiled .qm files. While it's possible to decompile .qm using Linguist, useful information like the source file references is lost when they're compiled. So users who want to contribute translations have no option but to use the very disconnected approach that is Launchpad.

(Disconnected compared to using Linguist. Which, thanks to Qt for Python, can now use source file references to display translation strings in their original context, in Python same as in C++. It's actually kind of cool.)

`pylupdate` has the same ability to extract message strings from Python code as `gettext` does, and it's possible to configure it to understand `_` as an alias for `translate()` / `tr()`.

ferdnyc:
@jonoomph 

I worry that our current approach may not be providing translators with all of the information they need (or would ideally like to have, at any rate) in order to translate the interface. The whole "only one translation per unique string" restriction is a severe imposition, and ignores the reality that the same word(s) can have VERY different meanings in different contexts. The lack of any support for plural translations has also been brought up as an ongoing problem (by the same user who suggested providing access to the .ts files, IIRC).

ferdnyc:
@jonoomph 

See #3589 #4326

PR #4514, opened 2021-10-29
WIP: High DPI Support for OpenShot (thumbnails, icons, cursors, preview images, scaling support)

jonoomph:
Adding support for High DPI monitors. This involved many different issues to be resolved. Primarily, using the QIcon class, to generate the correct size pixmap (with the correct DPI for different scenarios). QImage and QPixmap do not support the auto-magic of selecting the correct sized asset (i.e. MyIcon.png vs MyIcon@2x.png).

A few places required special handling of devicePixelRatio(), passing that along to libopenshot (when setting MaxSize of our timeline), and generating thumbnails / preview images in the correct size.

Lastly, our AppImage required some special bash handling, to scrape the current scaling factor from Linux, and setting the DPI for each detected screen, prior to launching openshot-qt. Windows and Mac don't seem to need this logic, and I'm assuming this is an issue specific to AppImage ignoring the current scaling factor of the OS.

I also build a special script to generate our UI icon/thumbnail cache in 2 different sizes (1x and 2x), since transitions, blender, effects, and emojis all require cache images for the UI to load quickly. Without the image cache, it takes a very long time on initial load of OpenShot... especially on lower powered machines. This is a diskspace vs cpu optimization.

Due to the # of files in this commit... it's best reviewed by looking at the individual commits, and ignoring the ones with thousands of cached image files.

**Remaining issues:**
 - ~~Custom toolbar icons: only show low res version (slice, marker, lock)~~
 - ~~Icons in About dialog (python, star, paypal, bitcoin, etc...) only show low res version~~
 - ~~All Icons in Windows (and probably Mac) only load the /Humanity/ resource icons packaged with OpenShot, and only low res versions~~
 - ~~Add to Timeline dialog crops HD file thumbnails~~

**Before High DPI Support:**
Before-high-dpi-support

**After High DPI Support:**
After-high-dpi-support


images/generate_cache.py
```diff
@@ -0,0 +1,187 @@
...
+    from xml.dom import minidom as xml
+
+
+# Create Qt application (required for SVG font rendering)
+svg_app = QApplication([])
+icon_size = info.LIST_ICON_SIZE
+emoji_size = info.EMOJI_ICON_SIZE
+
+
+def create_cache_thumbnail(path, width, height, thumb_path):
```
edumucelli, line 49: Sugestion: I'd model the concept of image and thumbnail on classes, read them as a list and then use `create_cache_thumbnail(image: Image)` instead of passing all the parameters that compose the image. Doing so as (if) the the number of parameters around image/thumnail grows you do not need to have the number of parameters in the function signature to grow:

The classes signature would be the following:

```
@dataclasses.dataclass(frozen=True)
class Image:
    path: pathlib.Path
    width: int
    height: int
    thumbnail: Thumbnail

@dataclasses.dataclass(frozen=True)
class Thumbnail:
    path: pathlib.Path
```

Also that makes a clear interface for (if) other modules want to use the exposed method.

images/generate_cache.py
```diff
@@ -0,0 +1,187 @@
...
+    thumb_path = os.path.join(info.IMAGES_PATH, "cache", "{}.png".format(fileBaseName))
+    create_cache_thumbnail(icon_path, icon_size.width(), icon_size.height(), thumb_path)
+
+
+#################################################
+# EMOJIS UI CACHE - Create small images for UI
+# Get emoji metadata
+emoji_metadata_path = os.path.join(info.PATH, "emojis", "data", "openmoji-optimized.json")
+with open(emoji_metadata_path, 'r', encoding="utf-8") as f:
+    emoji_lookup = json.load(f)
```
edumucelli: `emoji_lookup` is not being used.

images/generate_cache.py
```diff
@@ -0,0 +1,187 @@
...
+                                         mask_path, "", "#000", True, "png", 85)
+            reader.Close()
+            clip.Close()
+
+        except Exception:
+            # Handle exception
+            log.debug('Invalid cache image file %s', path, exc_info=1)
+
+
+#################################################
```
edumucelli, line 87: Together with the `Image` class, those methods could be just a series of `read_images() -> List[Image]`, that would make way less procedural and easy to understand an test as they would be name after their semantic and not based on `#####` separating each of the parts. I have converted this one here to give an idea of what it could be:

```
def read_ui_cache() -> List[Image]:
    """TITLES UI CACHE - Create small images for UI
    get a list of files in the OpenShot package directory
    """
    titles_dir = os.path.join(info.PATH, "titles")
    titles_list = [
        os.path.join(titles_dir, filename)
        for filename in sorted(os.listdir(titles_dir))
        ]

    images = []
    for icon_path in sorted(titles_list):
        filename = os.path.basename(icon_path)
        fileBaseName = os.path.splitext(filename)[0]

        # Check for thumbnail path (in build-in cache)
        thumb_path = base_image_path / "cache" / "{}.png".format(fileBaseName)
        images.append(Image(
            path=pathlib.Path(icon_path),
            width=icon_size.width(),
            height=icon_size.height(),
            thumbnail=Thumbnail(path=thumb_path)
        ))
    return images
```

Doing that way you would at the end of the file just have a:

```
for method in [read_ui_cache, read_emojis_path, read_effects_ui_cache, read_blender_ui_cache, read_transitions_ui_cache]:
  for image in method():
     create_cache_thumbnail(image)
```
 That would make it way more fluent and clearly express its semantic and all the methods in here would share a clear interface.

JacksonRG (approved):
Can't test on a high DPI monitor. But I can verify everything still looks right when I run this branch. And the code looks good to me!

images/generate_cache.py
jonoomph: Thanks for the suggestions. I agree, and I'll add this onto my list for a future improvement. I'm a bit worried about changing anything at this point, since I've done a bunch of testing and validation on this PR.

jonoomph:
Thanks for all the suggestions and feedback! Merging this for now, since it's a big one and I don't want to create to many headaches down the road for other PRs.

images/generate_cache.py
edumucelli: No problems @jonoomph, I can also do myself this refactoring and add a unit test.

installer/launch-linux.sh
```diff
@@ -1,4 +1,13 @@
-#!/bin/sh
+#!/bin/bash
+
+# Query the current display DPI
+#   This is to fix the AppImage, and scale the UI correctly
+#   for high DPI displays.
+regex="^.*resolution:\s*([0-9]*)x"
+DPI_OUTPUT=$(xdpyinfo | grep -B 2 resolution)
+[[ $DPI_OUTPUT =~ $regex ]]
+DPI=$(bc <<<"scale=2;${BASH_REMATCH[1]}/92")
```
ferdnyc: @jonoomph 

Where did the number `92` come from, in this equation? 

Standard desktop DPI is 96, and exactly what my (and most) unscaled displays will report. So now the Linux AppImage is setting the scaling factor to `1.04` on my machine.

PR #4522, opened 2021-11-04
Protect local event arg in JS

jonoomph:
Protect local event arg in JS, since it can be undefined or null in certain cases. Related to sentry issue: https://sentry.io/share/issue/90d9593ebfba41449671552abb1c953f/

This bug represents tens of thousands of errors on Sentry


jonoomph:
going ahead and merging this, but @JacksonRG please take a look when you have a second

PR #4528, opened 2021-11-07
Tweak OpenShot icon SVG framing

ferdnyc:
This PR re-positions the (unchanged) logo in a total area enlarged from `290px × 290px` to `300px × 300px`, and trims away "invisible" drawing area outside of the frame. This helps to prevent some clipping of the filmstrip shadow effect, which was occurring on the left and right edges of the image.


jonoomph:
Not sure I understand this one, but if we are just tweaking the SVG, it should be good to merge, assuming it solves some problem, lol.

PR #4532, opened 2021-11-09
Prevent razor on locked tracks

JacksonRG:
I happened to find that clips on locked tracks can be sliced,  so here's a PR to prevent that.

Note: I noticed that layers are stored as multiples of 1 million, so I floor divide the `layer` attribute before comparing, because I suspect that's intended to be used in ordering overlapping clips.

### DIFF
--- a/src/windows/views/webview.py
+++ b/src/windows/views/webview.py
@@ -1902,15 +1902,17 @@ class TimelineWebView(updates.UpdateInterface, WebViewClass):
 
         # Get CLIPS array from IDs
         clips = get_app().project.get("clips")
-        clips = filter( lambda x: x.get("id") in clip_ids, clips)
-        # Get tracks from project
-        locked_tracks = filter( lambda x: x.get("lock") == True, get_app().project.get("layers"))
-        # For t in locked_tracks
-        for t in locked_tracks:
-            # Ignore all clips on that track
-            for c in clips:
-                if int(c.get("layer") / 1000000):
-                    clip_ids.remove(c.get("id"))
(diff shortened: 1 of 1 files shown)


src/windows/views/webview.py
```diff
@@ -1890,6 +1890,20 @@ def Slice_Triggered(self, action, clip_ids, trans_ids, playhead_position=0):
         fps_float = fps_num / fps_den
         frame_duration = fps_den / fps_num
 
+        # Get CLIPS array from IDs
+        clips = get_app().project.get("clips")
+        clips = list(filter(lambda x: x.get("id") in clip_ids, clips))
+        # Get locked tracks from project
+        locked_layers = [
+            t.get("number") // 1000000
```
jonoomph: Why are we dividing the layer # by 1000000? Let's remove this

jonoomph:
Otherwise looks fine (simple enough solution to remove the clip_ids that are on locked tracks). You could also add this protection in lower down in the `for clip` loop... preventing/skipping any locked tracks. But technically it accomplishes the same thing, so I'll leave it up to you.

src/windows/views/webview.py
JacksonRG: I figured since they're multiples of a million, the layer might be used to settle which clip or transition to draw over another.

... Speaking of which I'm gonna test this with transitions to make sure those are locked too

JacksonRG: Tried without the division having 3 clips on top of each other. They all had the same `layer`: 3000000.

So that's gone, and I made use of the loops that already exist.

jonoomph:
Much better! Nice one. Merging.

PR #4534, opened 2021-11-11
WIP: Fix babl freezing/packaging

ferdnyc:
So, it turns out Babl isn't quite so simple to package.

In addition to its library, it has a set of extension functions typically installed as shared objects (`.so`, `.dynlib`, `.dll`) in a directory `babl-0.1` wherever its `LIBDIR` would reside. These objects need to come along with the library itself.

Normally the location of this directory is computed automatically, in a way that doesn't necessarily help us (it expects the files to either be in `$PREFIX/lib/babl-0.1/`, or in the same directory as the babl library itself — _not_ a subdir, annoyingly). However, the environment variable `BABL_PATH` can be set to point at a location where it looks for those files.

This PR attempts to:
1. Package the contents of the `babl-0.1` directory from within `freeze.py`
2. Install the file `installer/babl.py` as `classes/babl.py` IFF the `babl-0.1` directory is packaged
3. In `launch.py`, attempt to load `classes.babl` and, if found, use it to set the environment variable `BABL_PATH` to (effectively) `./babl-0.1` relative to the launch script.

That way, `BABL_PATH` will only be set when launching a frozen OpenShot that includes the Babl extensions.

There are a few changes, also, to make `freeze.py` slightly more generic, so that (for instance) it doesn't require that `resvg.so` and `libbabl-0.1.so` live only in `/usr/local/lib/` on Linux, it'll also check for them in the `ARCHLIB` directory.

I'm not _sure_ any of this will work exactly as intended, which is why this is being opened as a WIP PR from a branch in the project repo — we'll have to see what the build servers do with it.


ferdnyc:
Brought down by the world's most insidious typo.

Once you know it's there, it's obvious that this:
```python3
    babl_extension_paths = [
        ARCHLIB + "babl-0.1"
        "/usr/local/lib/babl-0.1",
    ]
```
will do **very** different things from this:
```python3
    babl_extension_paths = [
        ARCHLIB + "babl-0.1",
        "/usr/local/lib/babl-0.1",
    ]
```
But wow, is the difference between the two hard to track down.

ferdnyc:
So, this is mostly working, seems like, except the macOS notarization isn't liking the `lib/babl-0.1/` directory and/or contents. Hrm.

ferdnyc:
I wonder if that directory should be in `OpenShot Video Editor.app/Contents/Frameworks/babl-0.1` instead of `OpenShot Video Editor.app/Contents/MacOS/lib/babl-0.1`? I'm getting the impression,  from trolling around the web, that `.../Frameworks/_____.dylib` is A-OK (even when the `.dylib` isn't a Framework per se`), but I'm not sure if _subdirectories_ under `Frameworks/` is also accepted.

Of course, to test that, I'll have to tease apart the existing packaging code's setting of `install_name`s and etc. in order to redirect it.

(Just thinking out loud, now...) This is why I'm starting to feel like CMake has the right idea, with their `@rpath`-install-name-by-default standard configuration. In theory, if all of the libs had an `install_name` of `@rpath/libraryname.dylib`, and the executable(s) had their `rpath` set to `@executable_path/.` (for `.dylib`s in the same directory as the executable), then everything would work like it does now, _buuut_ **moving** them would _only_ require changing that rpath (or adding an additional rpath) on the executable to point at `@executable_path/../Frameworks/`.

Though I suppose for the babl extensions, specifically, they'd need an `install_name` of `@rpath/babl-0.1/extensionname.dylib`. (So that the executable doesn't need an additional rpath of `@executable_path/../Frameworks/babl-0.1`).

ferdnyc:
Aha!! I suspect I might be getting burned by [this](https://developer.apple.com/library/archive/technotes/tn2206/_index.html#//apple_ref/doc/uid/DTS40007919-CH1-TNTAG201):

> ### Nested Code
> From Mavericks onwards, signatures record nested code by its code signature and embed that information in the (outer) signature's resource envelope, recursively. This means that when a code signature is created, all nested code must already be signed correctly or the signing attempt will fail.
> 
> This is not a problem if you follow the standard Xcode build flow, because it will build and sign targets inside out: build the innermost code, copy it into the next-outer bundle, then sign that, etc., signing the outermost bundle last.
> 
> Nested code is expected in a number of standard locations within a bundle.
> 
> [...]
> 
> These places are expected to contain only code. Putting arbitrary data files there will cause them to be rejected (since they're unsigned).
> 
> Conversely, putting code into other places will cause it to be sealed as data (resource) files. This causes this code to be sealed twice; once in the nested code, and once in the outer signature. This wastes both signing and verification time and storage space. Also, this can break the outer signature of apps that use their own update mechanisms to replace nested code. If this nested code is being treated as resources, the outer signature doesn't know that this nested content is actually code.
> 
> Always put code and data into their proper places. This applies to all signed code and is enforced by the code signing machinery regardless of how the code is distributed.
> 
> **Note that a location where code is expected to reside cannot generally contain directories full of nested code, because those directories tend to be interpreted as bundles. So this occasional practice is not recommended and not officially supported. If you do do this, do not use periods in the directory names. The code signing machinery interprets directories with periods in their names as code bundles and will reject them if they don't conform to the expected code bundle layout.**

I've renamed the extensions directory from `babl-0.1` to just `babl`, as an experiment. We'll see if that's better.

ferdnyc:
\o/

Changing the directory name from `lib/babl-0.1` to just `babl` fixed the macOS notarization issues. And my latest commit hopefully fixed the Windows path problems that were causing the babl extensions to not be discovered properly.

installer/init_babl.py
```diff
@@ -0,0 +1,34 @@
...
+ You should have received a copy of the GNU General Public License
+ along with OpenShot Library.  If not, see <http://www.gnu.org/licenses/>.
+ """
+
+import os
+import sys
+
+def init():
+    if os.path.exists("babl-0.1"):
+        sys.putenv("BABL_PATH", os.path.realpath("babl-0.1"))
```
JacksonRG: I may be wrong, but I don't see a `putenv` in `sys`

freeze.py
```diff
@@ -232,23 +233,44 @@ def find_files(directory, patterns):
...
+    ]
+    for p in libopenshot_dirs:
+        if not os.path.exists(p):
+            continue
+        # Find all related SO files
+        matching_libs = find_files(p, ["*openshot*.so*"])
+        if not matching_libs:
+            continue
+        for filename in matching_libs:
+            if '_' in filename or filename.count(".") == 2:
```
JacksonRG, line 274: Could I get a comment, (or just be told) what about our shared object naming convention this making use of?

Edit: Is it because we only need `libopenshot.so.#` and not `libopenshot.so.#.#.#`? (My understanding is that the latter is identical, and there for anyone using that numbering convention)

JacksonRG (approved):
Looks great so far. I'll come back to review those last 3 commits tomorrow

PR #4539, opened 2021-11-13
detect duration too long

JacksonRG:
Goal
---
[This comment](https://github.com/OpenShot/openshot-qt/issues/652#issuecomment-967706719) pointed out that when a very long project is shortened, the duration doesn't decrease. This can cause the zoom slider's tightest zoom to have a resolution of a minute, even if the projects is only a minute long.

They also provided this wonderful demonstration video. :1st_place_medal: 

https://user-images.githubusercontent.com/77164855/141560890-1d428a2f-f9cb-407c-b18f-e95d0d9ebdcb.mp4

Known issues (Fixed :heavy_check_mark:)
---
1. [x] If a project is empty and you try to zoom, it will crash.
2. [x] Resize currently gets added to undo history

# Additional Bug
I noticed when loading a project, the cache would start where the playhead was,
and the playhead would move to 0:00. So I added a line to move the cache to 0:00 as well

### DIFF
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -703,8 +703,10 @@ App.controller("TimelineCtrl", function ($scope) {
     return Math.min(32767, desired_width);
   };
 
-// Find the furthest right edge on the timeline (and resize it if too small)
+// Find the furthest right edge on the timeline (and resize it if too small or too large)
   $scope.resizeTimeline = function () {
+    var max_timeline_padding = 20;
+    var min_timeline_padding = 10;
     // Find latest end of a clip
     var furthest_right_edge = 0;
     for (var clip_index = 0; clip_index < $scope.project.clips.length; clip_index++) {
@@ -715,15 +717,10 @@ App.controller("TimelineCtrl", function ($scope) {
       }
     }
     // Resize timeline
-    if (furthest_right_edge > $scope.project.duration || $scope.project.duration > furthest_right_edge + 20) {
+    if (furthest_right_edge > $scope.project.duration || $scope.project.duration - furthest_right_edge > max_timeline_padding) {
       if ($scope.Qt) {
-        timeline.resizeTimeline(furthest_right_edge + 10);
-        $scope.project.duration = furthest_right_edge + 10;
-      }
-    } else if ( $scope.project.duration - furthest_right_edge > 20) {
-      if ($scope.Qt) {
-        $scope.project.duration = furthest_right_edge;
-        timeline.resizeTimeline(furthest_right_edge);
+        timeline.resizeTimeline(furthest_right_edge + min_timeline_padding);
+        $scope.project.duration = furthest_right_edge + min_timeline_padding;
       }
     }
   };
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -720,6 +720,11 @@ App.controller("TimelineCtrl", function ($scope) {
         timeline.resizeTimeline(furthest_right_edge + 10);
         $scope.project.duration = furthest_right_edge + 10;
       }
+    } else if ( $scope.project.duration - furthest_right_edge > 20) {
+      if ($scope.Qt) {
+        $scope.project.duration = furthest_right_edge;
+        timeline.resizeTimeline(furthest_right_edge);
+      }
     }
   };
 
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -705,8 +705,9 @@ App.controller("TimelineCtrl", function ($scope) {
 
 // Find the furthest right edge on the timeline (and resize it if too small or too large)
   $scope.resizeTimeline = function () {
-    var max_timeline_padding = 20;
-    var min_timeline_padding = 10;
+    let max_timeline_padding = 20;
+    let min_timeline_padding = 10;
+    let min_timeline_length = 300; // Length of the default OpenShot project
     // Find latest end of a clip
     var furthest_right_edge = 0;
     for (var clip_index = 0; clip_index < $scope.project.clips.length; clip_index++) {
@@ -717,10 +718,11 @@ App.controller("TimelineCtrl", function ($scope) {
       }
     }
     // Resize timeline
-    if (furthest_right_edge > $scope.project.duration || $scope.project.duration - furthest_right_edge > max_timeline_padding) {
+    if (furthest_right_edge > $scope.project.duration - min_timeline_padding || $scope.project.duration - furthest_right_edge > max_timeline_padding) {
       if ($scope.Qt) {
-        timeline.resizeTimeline(furthest_right_edge + min_timeline_padding);
-        $scope.project.duration = furthest_right_edge + min_timeline_padding;
+        let new_timeline_length = Math.max(min_timeline_length, furthest_right_edge + min_timeline_padding);
+        timeline.resizeTimeline(new_timeline_length);
+        $scope.project.duration = new_timeline_length;
       }
     }
   };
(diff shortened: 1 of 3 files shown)


ferdnyc:
Suggestion I can't make in the code because it's too far outside the diff:

* While you're touching this, it'd be great to delete the comment on line 708 that (falsely) claims it's changing the selection. (The loop was clearly cut-and-pasted from elsewhere in the code.)
https://github.com/OpenShot/openshot-qt/blob/4163f74aa7bfe99de15d92beb2390c6c5973cf2b/src/timeline/js/controllers.js#L708

JacksonRG:
Closing, as I'm still working on this.

ferdnyc:
I had additional thoughts that I never submitted, I was  hoping Chrome would have them in a comment buffer here but I don't see them. (Probably means that comment is still open in another tab somewhere. I'll try to find it, I only have like 70 or 80 to look through. :expressionless: )

ferdnyc:
Aha! Found it. This is incomplete, so just expect that it'll cut off abruptly at some point, possibly with half-finished ideas/points.

@JacksonRG 

I've been thinking about timeline-shrinking for a while — it's really annoying that the timeline _never_ shrinks itself. If you add and then delete a 2-hour clip, you will always have a 2+-hour project at a minimum. This would be great to have!

Additional issues that I think should be addressed at the same time as this change:

1. Currently `timeline.resizeTimeline()` uses the `update()` function which creates a separate undo entry for the resize:
    https://github.com/OpenShot/openshot-qt/blob/4163f74aa7bfe99de15d92beb2390c6c5973cf2b/src/windows/views/webview.py#L2917-L2921

    Not only is that unnecessary and annoying (the user never wants to _undo_ just the automatic resize of the timeline), it can actually get you into a bad state. Normally it's not an issue because the `resizeTimeline()` in, say, the JS Clip directive's drag/resize callbacks will be followed by an `update_clip_data()` call that adds an undo for the action _after_ the timeline resize. But if you, say, resize a clip that's sitting at the end of the timeline duration, then undo-ing that can trigger a call to `resizeTimeline()` that automatically **repeats** the resize, adding an additional undo entry for the second resize. Result being, undo is now broken completely, because you get stuck in an infinite loop trying to undo an event that's constantly re-created the moment you do.

    That situation will only grow **far** worse if timeline-shrinking is added into the mix. `resizeTimeline()` needs to use `update_untracked()` instead, so that it changes the duration _without_ creating an undo event.

ferdnyc:
@JacksonRG One other thing I thought about ­— if live-resizing presents too many issues (and it very well may), then another approach would be to shrink extra length off the timeline _only_ when saving the project. That way, it would only have to be computed in the Python code, and the JS wouldn't have to worry about it. And at least that way, when you **next** loaded your saved project, the excessive, unused Timeline area would be gone. That's probably good enough, or at the very least it's a hell of a lot better than the current situation.

JacksonRG:
> I had additional thoughts that I never submitted, I was hoping Chrome would have them in a comment buffer here but I don't see them. (Probably means that comment is still open in another tab somewhere. I'll try to find it, I only have like 70 or 80 to look through. expressionless )

(I think `Ctrl shift a` searches through open tabs?) at least when I used brave. Not sure if its a feature across chromium.

JacksonRG:
@ferdnyc Thanks for pointing that out! Had no idea resize goes in action history.

ferdnyc:
> @ferdnyc Thanks for pointing that out! Had no idea resize goes in action history. 

Yeah, I hadn't realized it either until I started looking into this. Just one of several ways our Undo/Redo functionality could stand to be improved, though in this case the improvement is simple and obvious, and exactly what `update_untracked` is meant for. There's also no down side — if the timeline is auto-sizing, then its size can just update after each action (including undo/redo actions), and there's no reason to track the size itself in the history.

Other history issues are harder to solve, like the fact that a single "Split: Keep Both Sides" timeline action results in **three** separate individual undo entries. (One for the resizing of the left clip, one for insertion of the right-hand copy, and... I can't even remember what the third is. Seems like it should only need two. Maybe it actually deletes the original clip and inserts two new ones, for some reason.) 

...That kind of thing can only be solved with the addition of history transactions, which would allow us to combine multiple steps into a single undo/redo entry. 

(Well... either that, or we could store the entire action's changes directly into a single history entry, by making them so unreasonably complex that they're effectively all transactions. But that risks adding unnecessary complexity to the vast majority of _other_ history actions that only require one step. And it puts unnecessary burden on the project/property code, if it has to construct a complex event in its entirety before registering it with the HistoryManager.)

JacksonRG:
Just came back to this and incorporated some suggestions.

src/timeline/js/controllers.js
```diff
@@ -705,7 +705,7 @@ App.controller("TimelineCtrl", function ($scope) {
 
 // Find the furthest right edge on the timeline (and resize it if too small)
```
jonoomph: "too small or too large"

src/timeline/js/controllers.js
```diff
@@ -715,11 +715,16 @@ App.controller("TimelineCtrl", function ($scope) {
       }
     }
     // Resize timeline
-    if (furthest_right_edge > $scope.project.duration) {
+    if (furthest_right_edge > $scope.project.duration || $scope.project.duration > furthest_right_edge + 20) {
```
jonoomph: Let's move these constants (10, 20) into a variable for clarity

jonoomph: Also, we might need to incorporate a "minimum" duration into this logic. Not sure though, but just a guess. For example, let's say we add a 60 min clip, and then delete it. Does the timeline get resized to 0? Does that have any effect on the size of the timeline as shown to the user?

JacksonRG: It would reset to 10, so that's a good point. I'll make sure the minimum is what the timeline starts off at for a new project.

jonoomph:
LGTM (assuming it works, lol)

jonoomph:
This looks ready for merging. Merge away.

jonoomph:
Looks like this introduced an infinite loop. Investigating

Colorjet3:
Has this been all hashed out and finalized?  What are all the final rules and decisions?  It would be very beneficial if this can be summarized (User Story) and then provide the step needed to test it with the expected results.  I can then add this to our current test plans to verify.  Additionally, it will need to be part of the new feature/function and added to the overall test plan/documentation/etc.
Update:  Need detailed information regarding the Undo/Redo as well.

jonoomph:
@Colorjet3 We are still reviewing the potential fix: https://github.com/OpenShot/openshot-qt/pull/4703. Feel free to join the discussion on the new PR. Thx

robertobriccola:
@Colorjet3 
I have the same problem with the latest version and upgrades 3.4.0 | 0.5.0., in WIN11.
The slider does not compact.

Slider problem

Colorjet3:
Thank you @robertobriccola for your confirmation.  This is in the queue now for the lead developer to work on.

Colorjet3:
By the way, there is a way to shrink the timeline.  To do so requires you to be either in Retro or Humanity:Dark theme mode.  The Cosmic Dusk theme has a bug and is not working currently (I will be submitting a bug to get this fixed).

1.  Open a project that has many clips on a track.  Let's assume that the timeline is approximately 10 min. long.
2.  You decide to delete all the clips from the 6 min. marker and beyond.
3.  Now this leaves you with approximately 4 min. of empty space on the timeline.
4.  Save your project, to be on the safe side.
5.  Switch to Retor or Humanity:Dark theme.  Does not require a restart of OpenShot.
6.  Move the Zoom slider all the way to the right.  Make sure it is flush to the right edge of the zoom slider area.
7.  Now, hover your mouse over any track get as close as possible to the end of the track.  Make sure you do not go beyond the end of the track and get close to the edge of the OpenShot main window border.
8.  As you get close to the end of the track, at some point the mouse arrow should turn to left/right arrows.  When this happens, click and hold and drag to the left to shrink the timeline.

Not the most efficient way of doing it but at least there is a way.

PR #4549, opened 2021-11-23
WIP: Enable export controls after exception

### LINKED ISSUE #4537: Video export UI stays disabled on errors
**Describe the bug**
All controls in the export window stay disabled if an encoding error happens

1. Open any video or project
2. Click on the 'Export video' button
3. Cause encoding to fail (for example by entering an invalid codec into the advanced video settings)
4. Start the export
5. Error message about invalid codec appears

**Expected behavior**

Controls should revert to enabled state after either a failed or successful encoding without having to close and reopen the dialog, because this action resets all values that were set previously.

**System Details**
- OpenShot Version: 2.6.1-dev
- libopenshot Version:  0.2.7-dev
- Platform: Windows-10-10.0.19041
- Processor: AMD64 Family 25 Model 33 Stepping 0, AuthenticAMD
- Machine: AMD64
- Python version: 3.8.9
- Qt5 version: 5.15.2
- PyQt5 version: 5.15.4
- Qt Detected Languages: ['en-GB', 'en-US']
- LANG Environment Variable: 
- Daily Build: Verified issue still exists in daily build: http://github.com/OpenShot/openshot-qt/releases/download/OpenShot-v2.6.1-dev-daily-8688-c9669180-0fb4ab09-x86_64.exe

**Log Files**
- [openshot-qt.log](https://s3.amazonaws.com/logs.openshot.org/f2zkjxuik6ic7hf6eb5cuscqo6rct20t/openshot-qt.zip) (8 KB)
- [libopenshot.log](https://s3.amazonaws.com/logs.openshot.org/f2zkjxuik6ic7hf6eb5cuscqo6rct20t/libopenshot.zip) (0 KB)

**Exception / Stacktrace**
None. This problem doesn't causes an exception.

JacksonRG:
Closes #4537

Previously, a user had to close and re-open the export window after encountering an error.

Todos:
---
- [ ] As per Ferdnyc's comment, check that the second export doesn't have some data from the original attempt, before the exception

### DIFF
--- a/src/windows/export.py
+++ b/src/windows/export.py
@@ -987,6 +987,7 @@ class Export(QDialog):
             msg.setWindowTitle(_("Export Error"))
             msg.setText(_("Sorry, there was an error exporting your video: \n%s") % friendly_error)
             msg.exec_()
+            self.enableControls()
 
         # Notify window of export started
         self.ExportEnded.emit(export_file_path)


jonoomph:
Did you test this to ensure nothing else needs to be reset after an error? Otherwise, LGTM!

ferdnyc:
@JacksonRG @jonoomph 

I fear this may not, in fact, work.

The problem is that lots of things get set up for the export process during the `accept()` method, so for the user to be able to subsequently change any of those we **probably** need to throw out the `FFmpegWriter()` instance and start over with a new one. I don't trust its parameters to be multiply-modifiable without having some sort of weird side-effects. It's very, very stateful, to a fault, unfortunately.

(Heck, in fact, just the fact that we've already potentially done an `Open()` and a `WriteHeaders()` before the Exception was thrown, all but guarantees that the writer has to be tossed out. Once it's already started writing, it can't easily be backed out of that state. Possibly can't _at all_ be backed out of that state.)

jonoomph:
Closing this one for now, as this might not be worth the effort right now

PR #4553, opened 2021-11-25
Don't autocreate user dirs immediately when importing classes.info

ferdnyc:
Our PPA builds for 2.6.1 are currently broken in Launchpad, because I turned on documentation building which immediately crashed. The problem is that the Sphinx config does an import of `classes.info` to get the OpenShot version, and as soon as it's imported `classes.info` attempts to create all of the preferences directories in `$HOME/.openshot_qt`, a no-no during package builds.

I assume this is also why the unit tests are disabled, in the launchpad configs.

All of this makes the point, though, that we **SHOULDN'T BE** creating a ton of directories simply because a class was _imported_ — especially not one we import _everywhere_, from `setup.py` to `doc/conf.py` to `freeze.py` to `src/tests/query_tests.py`. That's an unacceptable side-effect for a class that's intended as merely a metadata dumping ground.

So, these commits move the creation of directories into a function within `classes.info`, `setup_userdirs()`, which `launch.py` calls immediately prior to setting up the `OpenShotApp` instance. All of the other scripts that import the class _won't_ call it, so they don't trigger the creation of a bunch of unnecessary settings dirs as a side-effect.

To make this work, I had to rearrange a bunch of things in `classes.settings`, `classes.logger`, `classes.project`, and `classes.app` so that they could handle the user preferences dir being entirely absent, since the unit tests need to be able to import `classes.app` and create an `OpenShotApp` instance without ever calling `classes.info.setup_userdirs()`.

I also removed the load of `classes.logger` from `classes.sentry` — if that's going to be loaded at the _very start_ of the launch script, it can't trigger the initialization of the logging class. It's too early. So it just does a `print()` of the info instead, same as how `launch.py` handles its own output.

And I removed all references to the `mode` from the GUI code, as the unit tests haven't started the GUI for some time now. `window.mode` will never be `"unittest"` under any circumstances, anymore. (`app.mode` is still used by the unit tests to adjust certain startup functionality, so it's still there. Also it's now set as an instance variable, so if any of the code ever _does_ need to check the mode, it can test `if get_app().mode == "unittest"`.)

With these changes, it's now possible to do all of these things:
* Build, install, or package the application
* Build the user manual
* Run the unit tests

...WITHOUT triggering the creation of `$HOME/.openshot_qt/` and its many, many subdirs.

Unfortunately I wasn't able to arrange things so that OpenShot itself could run **entirely** without a user dir as a 0-footprint application. That's something I see value in (temporary files would just be written to system default tempdirs, is the idea) since it would allow use of OpenShot in all sorts of scenarios where writing to `$HOME/.openshot_qt` is not feasible, permitted, or desired. But, it's a bigger undertaking that shouldn't hold up these changes.

#### Update

I added a few more commits:

1. `classes.info` now captures all of the initial defaults for user paths. (Any variable with a name that ends with `_PATH`, and a value that matches `info.USER_PATH` at the start is placed into the `info._path_defaults` dict.)
    * That captured dict is now used in `info.setup_userdirs()` (replacing the hardcoded list of variables representing paths to be created)
    * The defaults can be _reapplied_ to reset any modified `info` paths, by calling `info.reset_paths()`.
    * The default value for a path can be retrieved even if the current value has been modified, by calling `info.get_default_path("NAME_OF_VAR")`.
    * Most (hopefully all, but I can't promise that) of the rest of the code that was responsible for resetting `info` paths now does so by simply calling `info.reset_paths()`. 
    * Code that had to ad-hoc things like `os.path.join(info.USER_PATH, "thumbnails")` to work around a modified `info.THUMBNAIL_PATH` now instead just calls `info.get_default_path("THUMBNAIL_PATH")`.

1. I re-worked the autosave code a bit. The `project.save()` method, which used to have two parameters `make_paths_relative` and `move_temp_files`, now has only one: `backup_only`. Setting `backup_only=True` implies both of the previous parameters _`False`_, and also disables some additional logic that's undesirable when making a backup. (Like the storing of `current_filepath` and the resetting of `has_unsaved_changes`, both of which were previously being done in `project.save()`, then immediately **un**-done by the autosave code that called it. Now it's never done in the first place, because `backup_only=True` is passed in with the call.)

1. I renamed `MainWindow.clear_all_thumbnails()` to `MainWindow.clear_temporary_files()`, because I just couldn't abide the previous name for a function that not only cleared _three_ different types of working file, but also deleted the (previous) backup file! A name that bad was just asking for someone to make a wrong assumption, possibly with consequences.

### DIFF
--- a/src/classes/info.py
+++ b/src/classes/info.py
@@ -124,7 +124,7 @@ try:
 except ImportError:
     language_path = os.path.join(PATH, 'language')
     print("Compiled translation resources missing!")
-    print("Loading translations from: {}".format(language_path))
+    print(f"Loading translations from: {language_path}")
 
 # Compile language list from :/locale resource
 try:
@@ -192,13 +192,7 @@ SETUP = {
 def setup_userdirs():
     """Create user paths if they do not exist (this is where
     temp files are stored... such as cached thumbnails)"""
-    for folder in [
-            USER_PATH, BACKUP_PATH, RECOVERY_PATH, THUMBNAIL_PATH,
-            CACHE_PATH, BLENDER_PATH, TITLE_PATH, TRANSITIONS_PATH,
-            PREVIEW_CACHE_PATH, USER_PROFILES_PATH, USER_PRESETS_PATH,
-            USER_TITLES_PATH, EMOJIS_PATH, PROTOBUF_DATA_PATH,
-            YOLO_PATH,
-            ]:
+    for folder in _path_defaults.values():
         if not os.path.exists(os.fsencode(folder)):
             os.makedirs(folder, exist_ok=True)
 
@@ -209,6 +203,20 @@ def setup_userdirs():
     ]):
         print("Migrating default project file to new name")
         os.rename(LEGACY_DEFAULT_PROJECT, USER_DEFAULT_PROJECT)
+
+
+def reset_userdirs():
+    """Reset all info.FOO_PATH attributes back to their initial values,
+    as they may have been modified by the runtime code (retargeting
+    info.THUMBNAIL_PATH to a project assets directory, for example)"""
+    for k, v in _path_defaults.items():
+        globals()[k] = v
+
+
+def get_default_path(varname):
+    """Return the default value of the named info.FOO_PATH attribute,
+    even if it's been modified"""
+    return _path_defaults.get(varname, None)
 
 
 def website_language():
--- a/src/classes/info.py
+++ b/src/classes/info.py
@@ -68,6 +68,13 @@ BACKUP_FILE = os.path.join(BACKUP_PATH, "backup.osp")
 USER_DEFAULT_PROJECT = os.path.join(USER_PATH, "default.osp")
 LEGACY_DEFAULT_PROJECT = USER_DEFAULT_PROJECT.replace(".osp", ".project")
 
+# Back up "default" values for user paths
+_path_defaults = {
+    k: v for k, v in locals().items()
+    if k.endswith("_PATH")
+    and v.startswith(USER_PATH)
+}
+
 try:
     from PyQt5.QtCore import QSize
 
--- a/src/classes/info.py
+++ b/src/classes/info.py
@@ -66,27 +66,7 @@ YOLO_PATH = os.path.join(USER_PATH, "yolo")
 # User files
 BACKUP_FILE = os.path.join(BACKUP_PATH, "backup.osp")
 USER_DEFAULT_PROJECT = os.path.join(USER_PATH, "default.osp")
-
-# Create user paths if they do not exist
-# (this is where temp files are stored... such as cached thumbnails)
-for folder in [
-    USER_PATH, BACKUP_PATH, RECOVERY_PATH, THUMBNAIL_PATH, CACHE_PATH,
-        BLENDER_PATH, TITLE_PATH, TRANSITIONS_PATH, PREVIEW_CACHE_PATH,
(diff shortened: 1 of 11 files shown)


src/classes/info.py
```diff
@@ -207,10 +189,33 @@
...
+            USER_TITLES_PATH, EMOJIS_PATH, PROTOBUF_DATA_PATH,
+            YOLO_PATH,
+            ]:
+        if not os.path.exists(os.fsencode(folder)):
+            os.makedirs(folder, exist_ok=True)
+
+    # Migrate USER_DEFAULT_PROJECT from former name
+    if all([
+        os.path.exists(LEGACY_DEFAULT_PROJECT),
+        not os.path.exists(USER_DEFAULT_PROJECT),
```
JacksonRG, line 209: neat way to make a very readable "and not"

src/classes/info.py
```diff
@@ -192,13 +199,7 @@
...
     """Create user paths if they do not exist (this is where
     temp files are stored... such as cached thumbnails)"""
-    for folder in [
-            USER_PATH, BACKUP_PATH, RECOVERY_PATH, THUMBNAIL_PATH,
-            CACHE_PATH, BLENDER_PATH, TITLE_PATH, TRANSITIONS_PATH,
-            PREVIEW_CACHE_PATH, USER_PROFILES_PATH, USER_PRESETS_PATH,
-            USER_TITLES_PATH, EMOJIS_PATH, PROTOBUF_DATA_PATH,
-            YOLO_PATH,
-            ]:
+    for folder in [v for v in _path_defaults.values()]:
```
JacksonRG: Is this different from `for folder in _path_defaults.values():`? If so, I'm missing it.

JacksonRG (approved):
LGTM

src/classes/info.py
ferdnyc: That... is a good question. I suppose not.

I think I was originally expecting a comprehension (something like `[v for k, v in _path_defaults.items() if k.endswith('_PATH')]` or `... if v.startswith(USER_PATH)`), but in the end I didn't need it. So you're right, that can be simplified. Thanks!

ferdnyc:
```suggestion
    for folder in _path_defaults.values():
```

src/classes/info.py
ferdnyc: Standard disclaimer about `any()` and `all()` not supporting short-circuit evaluation applies (all of the conditions in the list will always be evaluated, unlike in a multi-clause `if .. and/or ...` condition), but for simple, one-time tests like this one the efficiency loss is undetectable.

src/classes/info.py
ferdnyc: (Well, not so much that I didn't need the comprehension. I just ended up doing it when I was _creating_ `_path_defaults`, instead.)

ferdnyc: Fixed in a subsequent commit.

ferdnyc:
Merging this, as it should restore our Launchpad builds to working order.

ferdnyc:
(Just waitin' on the CI, to merge.)

ferdnyc:
[Success!](https://code.launchpad.net/~openshot.code/+recipe/openshot-qt-daily) And now there's an `openshot-qt-doc` package containing the user manual, as well.

ferdnyc:
<s>Oops. But I have to fix Bionic, which doesn't have `python3-sphinx-copybutton` available.</s> Fixed.

PR #4554, opened 2021-11-25
remove_repeated_variable

ksm0517:
variable 'recommended_path' is repeated in main window's method actionSaveFrame_trigger


ferdnyc:
@ksm0517 Yeah, I've never understood why that was there. I assume a cut-and-paste flub or something. Thanks, I'll merge once the CI passes it.

PR #4577, opened 2021-12-07
Export clips v2

JacksonRG:
# Summary
Create a popup window to export clips as their own video files on right click in the files window. If a file, not a clip is selected, copy it. All new files are placed in a folder of the user's choosing.

Here's an example of the ui.
ExportClipsLabelOnly
Please note that the green outline, and lack of title bar are due to my DE.

# Use Case
A very common use-case for OpenShot is splitting many small videos from a larger one. For example, when a home video tape collection is digitized, and many different moments are combined into a single video file. OpenShot has a super easy way to split videos into clips (`Project Files->Split Clip...`). However, these are only "clips" within OpenShot, and many users want an easy way to then export these "clips" as actual individual video files.  This PR is aimed at addressing this work-flow.

# Goals
- [x] If the user selects files, copy them to the destination directory
- [x] Export using crf (22) to prevent very large file exports.
    - Possibly add a setting for this.
- [x] Add a progress bar
- [x] Add documentation for this UI
- [x] If file name already exists, don't export (incredibly unlikely that a file is naturally named `file_name [15 to 20].mp4`
- [x] Add copyright info


src/windows/main_window.py
```diff
@@ -1955,6 +1955,20 @@ def actionFile_Properties_trigger(self):
...
 
+    def actionExportClips_trigger(self):
+        from windows.exportClips import clipExportWindow
+        f = self.selected_files()
+        exp = clipExportWindow(export_clips_arg=f)
+        # exp.setPath(get_app().project.current_filepath)
+        exp.setPath(get_app().project.get("export_path"))
+        # f = self.selected_files()
+        # exp = clipExportWindow()
+        # exp.setPath(get_app().project.current_filepath)
```
JacksonRG: Just noticed this  commented code. Removing them in a later commit.

src/classes/filePicker.py
```diff
@@ -0,0 +1,86 @@
+from PyQt5.QtWidgets import QPushButton, QFileDialog, QCompleter, QLineEdit, QLabel
```
jonoomph: Copyright / License info is missing

jonoomph:
@JacksonRG Can you add a screenshot of this UI to the description of this PR? Thx

src/windows/exportClips.py
```diff
@@ -0,0 +1,142 @@
+from PyQt5.QtWidgets import QVBoxLayout, QPushButton, QDialog, QLabel
```
jonoomph: Ditto... missing copyright/license

src/windows/main_window.py
```diff
@@ -1955,6 +1955,16 @@ def actionFile_Properties_trigger(self):
...
 
+    def actionExportClips_trigger(self):
+        from windows.exportClips import clipExportWindow
+        f = self.selected_files()
+        exp = clipExportWindow(export_clips_arg=f)
+        exp.setPath(get_app().project.get("export_path"))
+        try:
+            exp.exec_()
+        except:
+            print("Error occured")
```
jonoomph: Replace print with `log.error()`

jonoomph:
Here is the format of our normal Export dialog (the footer). I would suggest we copy the same style (alignment, button placement, etc...):
Export-Dialog-Footer

jonoomph:
Also, no need for a "Close" button, since we have a "Cancel" button

src/classes/filePicker.py
```diff
@@ -0,0 +1,115 @@
+"""
+ @file filePicker.py
```
jonoomph: I'm confused by this filePicker.py approach? Is there a reason we can't use a general folder picker: `QFileDialog.getOpenFileUrls`, or something similar? Do we need to create our very own `QFileDialog`?

jonoomph: I'm worried that instead of a 1 or 2 line solution, we've introduced a bunch of unneeded complexity to maintain

src/windows/exportClips.py
```diff
@@ -0,0 +1,236 @@
+"""
```
jonoomph: This filename should be all lower-case, with _ between words (like the other window classes)

src/windows/exportClips.py
```diff
@@ -0,0 +1,236 @@
...
+from classes.filePicker import filePicker
+from classes import ui_util
+from classes import info
+from classes.app import get_app
+from classes.logger import log
+import openshot, os, re, shutil
+
+_ = get_app()._tr
+
+def makeLegalFileName(s: str):
```
jonoomph: 2 blank lines between imports and the 1st function declaration

src/windows/exportClips.py
```diff
@@ -0,0 +1,236 @@
...
+                               clip.data.get("acodec", "aac"),
+                               clip.data.get("sample_rate"),
+                               clip.data.get("channels"),
+                               clip.data.get("channel_layout"),
+                               clip.data.get("audio_bit_rate"))
+
+class clipExportWindow(QDialog):
+    """A popup to export clips as mp4 files
+    in a folder of the user's choosing"""
+    fp: filePicker
```
jonoomph: Again, something like:

```
from PyQt5.QtWidgets import QFileDialog
dialog = QFileDialog()
foo_dir = dialog.getExistingDirectory(self, 'Select an awesome directory')
```

src/windows/exportClips.py
```diff
@@ -0,0 +1,236 @@
...
+        self.buttonBox.addButton(self.cancel_button, QDialogButtonBox.ActionRole)
+        self.buttonBox.addButton(QDialogButtonBox.Close).clicked.connect(lambda: self.done(0))
+        self.FilePickerArea.addWidget(self.fp)
+        self.progressExportVideo.setValue(0)
+
+    def _cancelButtonClicked(self):
+        self.exporting = False
+
+    def _updateProgressBar(self, count: int, total: int):
+        d = count - total
```
jonoomph: Perhaps a comment here to describe this logic? Its a bit confusing to read

src/windows/exportClips.py
```diff
@@ -0,0 +1,236 @@
...
+        total_frames, frames_written = 0, 0
+        for c in clips:
+            total_frames += framesInClip(c)
+        for f in files:
+            total_frames += int(f.data.get("video_length",0))
+        for f in files:
+            copyFileToFolder(f, export_destination)
+            frames_written+=int(f.data.get("video_length",0))
+        for c in clips:
+            extension = "mp4"
```
jonoomph: This variable could be removed... and "mp4" used directly in the next line?

src/windows/exportClips.py
```diff
@@ -0,0 +1,236 @@
...
+                continue
+            w = openshot.FFmpegWriter(export_path)
+            setupWriter(c, w)
+            w.PrepareStreams()
+            w.Open()
+
+            start_time = c.data.get("start")
+            end_time = c.data.get("end")
+            fps = c.data.get("fps").get("num") / c.data.get("fps").get("den")
+            start_frame, end_frame = int(start_time * fps), int(end_time*fps)
```
jonoomph: Should this use round() instead of int()?

jonoomph: In fact, you might move the "time to frame" calculation into a function...  The logic should be:

```
(time * fps) + 1   (to ensure we always start on frame 1)
```

src/windows/exportClips.py
```diff
@@ -0,0 +1,236 @@
...
+                    break
+            clip_reader.Close()
+            w.Close()
+            log.info("Finished Exporting Clip: %s" % export_path)
+        log.info("Finished exporting")
+        self._updateProgressBar(frames_written, total_frames)
+        self._updateDialogExportFinished()
+
+    def _updateProgressBar(self, count: int, total: int):
+        if total==0:
```
jonoomph: Is this function duplicated? Let's try and not duplicate it.

jonoomph:
@JacksonRG Looks like a great start! Some nitpicks and small fixes. The biggest issue I have is with the custom QFileDialog we are creating... and I don't think we need to (for simplicity sake). If you have any questions, please let me know.

src/windows/exportClips.py
JacksonRG: Oops, I meant to move it to be next to the starting and ending export updates. The functions that disable some buttons and change the title.

src/classes/filePicker.py
JacksonRG: It started with me thinking there was an existing widget for typing in file paths, and suggesting sub-directories.

When I didn't find one, I figured we could make one to plug into other widgets, like properties view, or exporter. A quick, consistent way to find files without popping up a dialog.

JacksonRG: I see that it _is_ scope creep for this PR. So if you like, I can:

1) Move this widget to a new branch for sometime down the line.
2) Move this code (minus the auto-complete) into exportClips.py

jonoomph:
@JacksonRG A few other comments:

- We should default to the home folder (or the current project folder, if any)
- Clicking "Browse" and not selecting a folder, should not clear the text box
- The progress-bar seemed incorrect to me, as it jumped to 99% immediately. Maybe the small clips were just really quickly exported... not sure, but it seemed too quick.
- Remove the "Close" button (user can always click cancel or the close icon on the window titlebar)
- Reduce the default height in the *.ui file... so there is less vertical padding between widgets (by default)
- Icon in the "Export Clips" context menu should be the same as "Export" icon
- Move "Export Clips" context option under "Split Clips...", to keep them grouped (ideally between horizontal lines), and since splitting always happens prior to exporting.

jonoomph:
Otherwise, this worked really nicely! Nice work!

JacksonRG:
@jonoomph The The progress bar jumps because it count's frames. So if you export a source file, as well as clips equaling the whole length, It will jump to 50% when the file is copied (half the frames) then start working on encoding the rest.

If you'd rather, the progress bar can ignore the file-copying progress.

jonoomph:
@JacksonRG Did we resolve the issues with audio-only files?  For now, it's probably fine to just default to our mp4/h.264 for any video or audio file. In the future, it might be nice to use *.mp3 files for any audio-only files.

JacksonRG:
Having issues testing audio files. They keep returning True `has_video`... but causing `ffmpeg` issues when I setup the writer. So I'm adding an error popup, and catching the error.

jonoomph:
@JacksonRG Some files, such as MP3, do contain both video and audio. However, it should work fine to write those files as a MP4.

src/windows/export_clips.py
```diff
@@ -101,28 +101,28 @@ def startAndEndFrames(clip):
...
-    if clip.data.get('has_audio', False):
-        writer.SetAudioOptions(True,
-                               clip.data.get("acodec", "aac"),
-                               clip.data.get("sample_rate"),
-                               clip.data.get("channels"),
-                               clip.data.get("channel_layout"),
-                               clip.data.get("audio_bit_rate")
-                               )
+    writer.SetAudioOptions(True,
+                           "aac",
```
JacksonRG, line 121: Here was the problem! I was by default using whatever audio encoding the file had. (even when they were mp4 files this caused a crash)

jonoomph:
@JacksonRG LGTM! Merge away

PR #4583, opened 2021-12-08
Fixing 2 regressions with video_widget (Region support, Zoom support)

jonoomph:
Fixing regressions with https://github.com/OpenShot/openshot-qt/pull/4527.

 - Region drawing was broken (used by Tracker effect)
 - Zooming with Scroll-Wheel on video_widget was broken

I am reverting just the sections of code required to fix these 2 issues.

### DIFF
--- a/src/windows/video_widget.py
+++ b/src/windows/video_widget.py
@@ -509,15 +509,23 @@ class VideoWidget(QWidget, updates.UpdateInterface):
     def centeredViewport(self, width, height):
         """ Calculate size of viewport to maintain aspect ratio """
 
-        window_size = QSizeF(width, height)
-        window_rect = QRectF(QPointF(0, 0), window_size)
+        # Calculate padding
+        top_padding = (height - (height * self.zoom)) / 2.0
+        left_padding = (width - (width * self.zoom)) / 2.0
 
-        aspectRatio = self.aspect_ratio.ToFloat() * self.pixel_ratio.ToFloat()
-        viewport_size = QSizeF(aspectRatio, 1).scaled(window_size, Qt.KeepAspectRatio)
-        viewport_rect = QRectF(QPointF(0, 0), viewport_size)
-        viewport_rect.moveCenter(window_rect.center())
+        # Adjust parameters to zoom
+        width = width * self.zoom
+        height = height * self.zoom
 
-        return viewport_rect.toRect()
+        # Calculate which direction to scale (for perfect centering)
+        aspectRatio = self.aspect_ratio.ToFloat()
+        heightFromWidth = width / aspectRatio
+        widthFromHeight = height * aspectRatio
+
+        if heightFromWidth <= height:
+            return QRect(left_padding, ((height - heightFromWidth) / 2) + top_padding, width, heightFromWidth)
+        else:
+            return QRect(((width - widthFromHeight) / 2.0) + left_padding, top_padding, widthFromHeight, height)
 
     def present(self, image, *args):
         """ Present the current frame """
@@ -1265,8 +1273,12 @@ class VideoWidget(QWidget, updates.UpdateInterface):
 
     def regionTriggered(self, clip_id):
         """Handle the 'select region' signal when it's emitted"""
-        # Clear transform
-        self.region_enabled = bool(not clip_id)
+        if self and not clip_id:
+            # Clear transform
+            self.region_enabled = False
+        else:
+            self.region_enabled = True
+
         get_app().window.refreshFrameSignal.emit()
 
     def resizeEvent(self, event):


JacksonRG (approved):
Rectangle math looks right! Merge away

src/windows/video_widget.py
```diff
@@ -509,15 +509,23 @@ def paintEvent(self, event, *args):
     def centeredViewport(self, width, height):
         """ Calculate size of viewport to maintain aspect ratio """
 
-        window_size = QSizeF(width, height)
-        window_rect = QRectF(QPointF(0, 0), window_size)
```
JacksonRG, line 513: If I'm reading this right, here's the error. Window is always at the top left corner.

jonoomph: You might be right.... but my brain struggles with this type of logic, lol

ferdnyc: That's not the error (I'm not sure what the error is, I didn't see what problem this was meant to address) — `window_rect` is the _window_ (the widget area being displayed in), of course its origin is in its own upper-left corner, and it  has the size indicated. You can't create a rect without an origin point, which is almost always `QPoint(0,0)` or `QPointF(0.0, 0.0)`.

This later call centers the viewport within the window:
```python
viewport_rect.moveCenter(window_rect.center())
```

src/windows/video_widget.py
```diff
@@ -509,15 +509,23 @@ def paintEvent(self, event, *args):
...
-        return viewport_rect.toRect()
+        # Calculate which direction to scale (for perfect centering)
+        aspectRatio = self.aspect_ratio.ToFloat()
+        heightFromWidth = width / aspectRatio
+        widthFromHeight = height * aspectRatio
+
+        if heightFromWidth <= height:
+            return QRect(left_padding, ((height - heightFromWidth) / 2) + top_padding, width, heightFromWidth)
+        else:
+            return QRect(((width - widthFromHeight) / 2.0) + left_padding, top_padding, widthFromHeight, height)
```
ferdnyc, line 528: This change re-broke Python 3.10. you can't pass `QRect()` float values, it only takes `int`s. In Python 3.10 that's an instant crash.

If there's an issue here, it's that my code doesn't account for zoom — I _thought_ that was intentional, I thought I'd handled that somewhere else. I'd certainly _meant_ to, because the zoom math should be independent of the viewport. The viewport math is based on the original, un-zoomed frame image.

If it gets zoomed in or out, that just scales it larger or smaller around its own center point. Zooming changes the _apparent_ size, not the _actual_ size. If the viewport is going to account for zoom, then shouldn't it have to  account for rotation as well? Rotation also changes the frame's apparent size (and aspect).

ferdnyc: (The idea here was that `centeredViewport()` would be generic — you give it dimensions, it tells you how to center the image in those dimensions based on its aspect ratio. Scaling, rotation, etc, are all separate operations that occur after that point.)

src/windows/video_widget.py
```diff
@@ -1265,8 +1273,12 @@ def keyFrameTransformTriggered(self, effect_id, clip_id):
...
     def regionTriggered(self, clip_id):
         """Handle the 'select region' signal when it's emitted"""
-        # Clear transform
-        self.region_enabled = bool(not clip_id)
+        if self and not clip_id:
+            # Clear transform
+            self.region_enabled = False
+        else:
+            self.region_enabled = True
+
```
ferdnyc, line 1281: OK, the only issue here was that I got the test backwards. It should be `self.region_enabled = bool(clip_id)`. But it's very odd to be testing `self` in a Python method. `self` is the current class instance. It's implicit. I've always wondered what those tests were meant to accomplish.

src/windows/video_widget.py
ferdnyc: Also, at the risk of ranting a bit... the code needs to stop abusing the term "heightForWidth" for simple scaling. That term has a meaning in Qt, and that's not what it means.

"HeightForWidth" means that a widget's height is dependent on its width, **INDEPENDENT of scaling**.

Picture a text widget containing some wrapped text. If the widget is wide enough that none of the lines wrap, N lines fit in a widget N lines tall:

image

But if the widget is made narrower and lines wrap, its height changes:

image

**THAT** widget would return `True` for `.hasHeightForWidth()`. Because that's what it means. Scaling a rectangle is not a height-for-width operation, it's just scaling. And it doesn't need to be done in the complicated way it's done here.

src/windows/video_widget.py
ferdnyc: (The reason I got the test backwards was that it sets `region_enabled` to `False` if `not clip_id` is `True`. Just like in speech, double negatives in code are confusing.)

src/windows/video_widget.py
ferdnyc: Well, anyway, like I said the lack of accounting for zoom was the only difference. Here's the entire change required to add zooming to my `centeredViewport()`:

```diff
diff --git a/src/windows/video_widget.py b/src/windows/video_widget.py
index a4d42969..5666a9a9 100644
--- a/src/windows/video_widget.py
+++ b/src/windows/video_widget.py
@@ -513,7 +513,9 @@ class VideoWidget(QWidget, updates.UpdateInterface):
         window_rect = QRectF(QPointF(0, 0), window_size)
 
         aspectRatio = self.aspect_ratio.ToFloat() * self.pixel_ratio.ToFloat()
-        viewport_size = QSizeF(aspectRatio, 1).scaled(window_size, Qt.KeepAspectRatio)
+        viewport_size = QSizeF(aspectRatio, 1).scaled(
+                            window_size, Qt.KeepAspectRatio
+                        ) * self.zoom
         viewport_rect = QRectF(QPointF(0, 0), viewport_size)
         viewport_rect.moveCenter(window_rect.center())
```

After that patch, it produces the same results as this code.

(Within a pixel one way or the other, attributable to the difference between passing floats to `QRect()` — which I _think_ in Python < 3.10 effectively called `int()` on all of the arguments, meaning it truncated floats — vs. using `QRectF.toRect()`, which will round to the nearest integer)

But also:
1. It doesn't break Python 3.10
2. It respects pixel aspect ratio. (_Shouldn't_ the viewport be doing that? I feel like it should. Unless it's already applied somewhere else.)

PR #4584, opened 2021-12-08
Fix Max Preview Size (New & Open Project)

jonoomph:
When opening or creating a project, we lose our timeline max-size optimization. This causes slow downs in preview and some strange visual artifacts, such as text-size changing when displaying frame #'s on a clip.

### DIFF
--- a/src/windows/video_widget.py
+++ b/src/windows/video_widget.py
@@ -509,15 +509,23 @@ class VideoWidget(QWidget, updates.UpdateInterface):
     def centeredViewport(self, width, height):
         """ Calculate size of viewport to maintain aspect ratio """
 
-        window_size = QSizeF(width, height)
-        window_rect = QRectF(QPointF(0, 0), window_size)
+        # Calculate padding
+        top_padding = (height - (height * self.zoom)) / 2.0
+        left_padding = (width - (width * self.zoom)) / 2.0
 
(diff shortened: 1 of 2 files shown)


src/windows/video_widget.py
```diff
@@ -509,15 +509,23 @@ def paintEvent(self, event, *args):
     def centeredViewport(self, width, height):
         """ Calculate size of viewport to maintain aspect ratio """
 
-        window_size = QSizeF(width, height)
-        window_rect = QRectF(QPointF(0, 0), window_size)
+        # Calculate padding
```
jonoomph, line 512: Not sure why these changes are showing different than develop (since I already merged them)

src/windows/video_widget.py
```diff
@@ -1265,8 +1273,12 @@ def keyFrameTransformTriggered(self, effect_id, clip_id):
 
     def regionTriggered(self, clip_id):
         """Handle the 'select region' signal when it's emitted"""
-        # Clear transform
-        self.region_enabled = bool(not clip_id)
+        if self and not clip_id:
```
jonoomph, line 1276: Ditto

jonoomph:
Merging to get the builders working on this one

src/windows/video_widget.py
JacksonRG: :thinking:

JacksonRG (approved):
LGTM

PR #4586, opened 2021-12-09
Adding execute permission for files inside Mac *.App bundle

jonoomph:
Adding execute permission for files inside Mac *.app bundle, to allow different users to run the *.app, and not just the user who installed OpenShot.

Related to https://github.com/OpenShot/openshot-qt/issues/4571


JacksonRG (approved):
merge away!

PR #4587, opened 2021-12-09
Ignore permission error when creating folders (info.py)

jonoomph:
Ignore permission error when creating folders in (info.py), for example, when Launchpad build servers are running Sphinx, and the home folder has no permissions. This is mostly a build server issue, so I added some print() to stderr with details on what failed, but we don't raise an actual exception anymore.


JacksonRG (approved):
Looks Good!

PR #4602, opened 2021-12-17
Update inkscape files to the new higher DPI

JacksonRG:
- Opened each inkscape file.
- Allowed inkscape to re-scale each file while preserving
  their appearance, and saved the file

Warning message from Inkscape:
Inkscape-warning

### DIFF
--- a/src/titles/Bar_1.svg
+++ b/src/titles/Bar_1.svg
@@ -385,10 +385,10 @@
      inkscape:document-units="px"
      inkscape:current-layer="g2439"
      showgrid="false"
-     inkscape:window-width="1600"
-     inkscape:window-height="850"
-     inkscape:window-x="0"
-     inkscape:window-y="26"
+     inkscape:window-width="1295"
+     inkscape:window-height="1028"
+     inkscape:window-x="625"
+     inkscape:window-y="24"
      inkscape:window-maximized="0" />
   <metadata
      id="metadata2388">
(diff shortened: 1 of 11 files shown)


jonoomph:
Did this actually update the DPI from 90 to 96? I've attached the warning image to the description of this ticket.

JacksonRG:
I believe it did. But I'll look to verify. It made the files open without that warning afterwards.

jonoomph:
Sounds good. LGTM

ferdnyc:
@JacksonRG 

> - Opened each inkscape file.
> - Allowed inkscape to re-scale each file while preserving
>   their appearance, and saved the file

Wow, I'm really sorry you went through all of that! It wasn't necessary. Pure-vector Inkscape files (like ours) don't have an intrinsic DPI (some other files do, if they're designed to specific dimensions for real-world output, like pre-press designs) but ours don't). If you look at the diff, the only thing that changed with reopening and resaving them is:
1. A few size values at literally the EIGHTH decimal point (`0.39215687` -> `0.39215686`, for example)
2. The session metadata (meaning, the dimensions of your inkscape application window were stored, replacing the previously-saved values from whoever last edited the files)
3. **The Inkscape version number**

The last one is the critical one, because Inkscape won't offer to update files if they were saved with a version newer than 0.92. So, this change will prevent that message coming up in the future. It could've been fixed with a batch script, but what's done is done.

JacksonRG:
> Wow, I'm really sorry you went through all of that! It wasn't necessary.

Oh it was no issue. I just selected all, opened, and waited. :laughing:

And good to know. While a script is a good idea, I like letting inkscape edit svgs, as that's its job.

I appreciate the info! I was dragging my feet to merge this, as I wasn't positive how I could prove I didn't mess up the scaling on any of our titles.

JacksonRG:
@ferdnyc Just rolled back everything but the version number! Planning to merge once the tests pass

Issue #4477, opened 2021-10-08
Support the XDG Base Directory Specification


itaranto:
Please support the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) on Linux/*nix systems.

Right now OpenShot dumps everything user related to `~/.openshot_qt` which is **not** nice for several reasons.

If you think that splitting this data between cache, config and user data is not feasible to do right now, **at the bare minimum** put everything under `$XDG_DATA_HOME/openshot_qt` (with a fallback to `~/.local/share/openshot_qt`)

Thanks.

ChristianTacke:
@itaranto Can you take a look at #4624?

Especially of interest: Do you think, more directories should be moved from the default XDG_DATA_HOME to other places?

itaranto:
> @itaranto Can you take a look at #4624?
> 
> Especially of interest: Do you think, more directories should be moved from the default XDG_DATA_HOME to other places?

That's looks really nice, I'm not familiar with the internals of this project, but splitting files between cache, config and user data seems to be the way to go!

I'm not a contributor here, but I have one question regarding `pyxdg`: So, you are basically detecting if `pyxdg` is installed and then use it right? Wouldn't be better to add `pyxdg` as a dependency for OpenShot?

ChristianTacke:
> That's looks really nice, I'm not familiar with the internals of this project, but splitting files between cache, config and user data seems to be the way to go!

To further order things, we need someone with more insight into the internals. For example `thumbnail` really also sounds like "cache", but I am not sure at all.

> […] I have one question regarding `pyxdg`: So, you are basically detecting if `pyxdg` is installed and then use it right? Wouldn't be better to add `pyxdg` as a dependency for OpenShot?

It might!

Two points:

* I am new to the openshot codebase (just took a look today). And I haven't found the point for dependency management. So I have no idea where to add it!
* XDG should be used on Linux/Unix. But probably not on windows or macos. So one idea was: If a user actually installs `pyxdg` on windows/macos, they hopefully know, why they're doing so. On the other hand, I hope for it to be "just" installed on a typical Linux workstation.
* Actual packaging by distros hopefully will add it later on.

itaranto:
[...]

PR #4624, opened 2021-12-31
Basic Support for XDG Base Directory Specification

ChristianTacke:
If the classic `~/.openshot_qt` is there, then continue using it, and ignore XDG.
If [pyxdg](https://www.freedesktop.org/wiki/Software/pyxdg/) is available, then use it.

Some basic XDG style directory setup:

* `XDG_CONFIG_HOME` (`~/.config/openshot_qt/`):
  * `openshot.settings`
* `XDG_CACHE_HOME` (`~/.cache/openshot_qt/`):
  * `preview-cache` as `preview`
  * `cache` as `main`
* `XDG_DATA_HOME` (`~/.local/share/openshot_qt/`):
  * Everything else

NOTE: I wonder whether more files should be moved from the "default" location. For example the `thumbnail` folder sounds like a candidate for cache?

See-Also: https://github.com/OpenShot/openshot-qt/issues/4477

### DIFF
--- a/src/classes/info.py
+++ b/src/classes/info.py
@@ -28,6 +28,12 @@
 import os
 from time import strftime
 
+try:
+    from xdg import BaseDirectory
+except ImportError:
+    BaseDirectory = None
+
+
 VERSION = "2.6.1-dev"
 MINIMUM_LIBOPENSHOT_VERSION = "0.2.7"
 DATE = "20210904000000"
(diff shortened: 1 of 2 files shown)


JacksonRG:
Hi, thanks for volunteering your code! 🏆 And cleaning up our file organization could certainly be helpful. (Though I might suggest waiting for a full release to avoid some confusion in telling users the wrong location to go find their log files.)

We're working on being more rigorous with our PRocess of reviewing PRs and merging them. The most important component of this is: tests.

Could you include some steps to make sure your code is working as it should, so we can try that on multiple computers and operating systems?

Things to include are: "What issue/improvement you're addressing", "Roughly how you fixed it", "Steps to prove it's working"

Again thanks for your help!

jonoomph:
Thanks for suggesting these changes! We really appreciate the thought, but at this time, I am not seeing the benefit for OpenShot and our users, to drastically change the folder structure only for Linux. But I am curious what others think about this change, so we'll continue to monitor feedback on this topic. Thanks again!

PR #4630, opened 2022-01-02
Fix viewport zooming and region selection

ferdnyc:
See #4583


ferdnyc:
Merging to re-fix Python 3.10.

PR #4636, opened 2022-01-07
CI: New automated docs build process, with PDF!

ferdnyc:
Inspired by this tweet I saw go by last month:

<blockquote class="twitter-tweet"><p lang="en" dir="ltr">Dear <a href="https://twitter.com/openshot?ref_src=twsrc%5Etfw">@openshot</a>,<br>on your site there&#39;s a user guide. Can I please use this offline via a download (pdf or something), please?<br>Any help much appreciated.</p>&mdash; Tommy Tom Thomas (@TomCahi99376902) <a href="https://twitter.com/TomCahi99376902/status/1473331659948924934?ref_src=twsrc%5Etfw">December 21, 2021</a></blockquote>

I spent more of the past _week_ than I care to admit or remember, completely overhauling our entire documentation build system within GitHub Actions.

The revised workflow uses a local action I wrote (in `.github/actions/sphinx/`) to generate a Docker container image (largely a clone of `latexpdf/Dockerfile` from sphinx-doc/docker@1c2c9e4c25087a9ab224865a5785b9f1cde64c78) with everything required to generate both HTML and PDF versions of our manual. Then it does two quick container runs, grabs the HTML directory and the PDF file, and publishes them as separate build artifacts.

Along the way, I:
1. **Renamed the `doc/` folder to `docs/`.** Just... trust me, it was the path of least resistance. Lots of tools default to that name, and being non-standard wasn't doing us any favors.
    I tried to update the path everywhere it appears _in_-repo, but some paths may need to be modified on external resources like the project GitLab builders. (Though I did update the `.gitlab-ci.yml` so perhaps that covers it.)

1. **Made the docs build self-contain-_able_.** Again, this was about the path of least resistance on the container build. Because it is containerized, the docs can't be referencing files from all over the rest of the filesystem, since those are areas the container doesn't have access to. So, the workflow now copies all of the external dependencies into `docs/_docker/` as part of its setup, and `docs/conf.py` knows to try there **first** when importing files. If that dir isn't found, it reverts to loading the external images and importing `classes.info` from its parent directories, just like before.

1. **Added `:alt:` texts to all of our image embeds.** Because accessibility is important, and really not that hard.

1. Added some latex code to the document preamble so that **table heading rows are now bolded in the PDF doc.**

1. **Added `sphinxcontrib-svg2pdfconverter`** to the docs requirements, and to the configs. You can read [the extensive StackOverflow answer I wrote yesterday, detailing my encounter with it](https://stackoverflow.com/questions/49898798/using-sphinx-extension-to-convert-svg-to-pdf/70589771#70589771), if you want the full scoop of its awesomeness. Long & short, **it lets us generate PDF docs incorporating SVG images, _without_ rasterizing them**. The embedded images are still vectors, and they look sharp and clean at any magnification.

    (Kind of a shame the only SVG images currently used are two toolbar icons. We should probably try to flesh that out more.)

1. Reorganized the section on data Import/Export formats to **move the XML and EDL example listings out into their own appendix section**, way at the end of the document.

    - The XML listing was just too overpowering plopped right in the middle of the doc, it's like 7 pages long. Now, with the listings gone, the remaining section is a comfortable one-page length, and there are references linking directly to the listings in the appendix.

    - Since I created `appendix_a.md` as a _MarkDown_ file (easier formatting), I **activated Sphinx's ability to combine MarkDown and ReStructuredText** when generating documents. It's just a matter of adding an extension, and installing a Python package called `myst_parser`, which is a MarkDown processor.

   - In the process of moving the XML listing, I also hand-reformatted the crazy long lines of XML attributes that a few of the output tags had, because they were causing the listing to overflow its bounding box (and sometimes the PAGE bounding box) in the latex docs. The re-wrapped version is identical except for the insertion of newlines and indentation spacing, but it all formats comfortably within the confines of a printed page (well... seven printed pages).

1. **Switched to using `sphinx.ext.autosectionlabel`** in place of the manual reference targets. Mostly because I had to in order to make links into the MarkDown document, but also because it makes for more comfortable referencing, especially when reading the document source.

    Because we have duplicate section names, an option to prefix targets with the document name is enabled. So, referencing looks like this:
    ```rst
    .. To reference a particular section in a particular document...
    
    This sentence will link to the :ref:`clips:Cutting & Slicing` section.
    
    .. To reference the title section (the entire document, IOW)...
    
    This sentence will link to the :ref:`animation:Animation` chapter.
    ```
    Pretty straightforward. The output is, as you would expect...
    > This sentence will link to the [Cutting & Slicing](#) section.
    > This sentence will link to the [Animation](#) chapter.

Aaaand... well, shit, I certainly _HOPE_ that's everything!


ferdnyc:
When the '**Sphinx build / build (pull_request)**' CI job finishes, you'll be able to check out the generated HTML or PDF manual by clicking on its Details link. There will be two Artifacts links underneath the workflow "map" on the right, one for the PDF and one for the HTML.

(Unfortunately because of the way Artifacts work, the PDF will be inside a zip archive, even though it's just one file. No way to change that, docs say.)

ferdnyc:
> * The XML listing was just too overpowering plopped right in the middle of the doc, it's like 7 pages long.

> * but it all formats comfortably within the confines of a printed page (well... seven printed pages).

A lie twice told! The XML isn't seven printed pages. It's _**sixteen**_. (Well, 15 ¾.)

ferdnyc:
> Aaaand... well, shit, I certainly _HOPE_ that's everything!

Whoops, not quite! One more small change with big impact, though I suspect it won't be very noticeable really. (It would be, if I _hadn't_ done it!)

**I removed the "Table of Contents" title from `index.rst`.**

With a title on that page, _all_ of the other pages were being placed one level _below_ it, in the latex/PDF output. So instead of a document with 14 chapters, you had a document with a **single** chapter and fourteen subsections! It looked quite silly. So, I removed the title, gave the TOC a `:caption:` instead, and everything ended up where it's supposed to be.

.github/actions/sphinx/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+      xindy \
+      tex-gyre \
+ && apt-get autoremove \
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/*
+#      graphviz \
+#      texlive-latex-extra \
+#      texlive-fonts-extra \
+
+COPY ./requirements.txt requirements.txt
```
ferdnyc, line 28: _Note: I'm still working on this a bit, but it's not a blocker to merging_

I'm not entirely happy with the constraint that the `requirements.txt` file for the docs has to live in `.github/actions/sphinx/` now. (The one in `docs/` only contains a reference to that file, and a bunch of comment lines explaining that requirements have to be added to that other file instead of the obvious one. Lame.)

Problem is, the container image fails to build if it tries to reference any files outside the 'build context'. It seems that context is established before the start of the run. Even if I copy the file in, it isn't used. (That's probably a safety measure to avoid script injection, which isn't really the worst idea TBH.)

Trying to find a way I can keep the file in `docs/` while still using it to configure the Docker image. But if I find a solution it can always be a follow-up PR. This PR works as-is for now.

.github/actions/sphinx/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+#COPY ../../../src/classes/info.py docs/_docker/info.py
+#COPY ../../../xdg/openshot-arrow.png docs/_docker/openshot-arrow.png
+#COPY ../../../xdg/openshot-qt.ico docs/_docker/openshot-qt.ico
+#COPY ../../../xdg/icon/512/openshot-qt.png docs/_docker/openshot-qt.png
```
ferdnyc: (The whole reason these are commented out is, they can't work. In fact I should just remove them.)
```suggestion
```

ferdnyc:
Thoughts on this, anyone? @JacksonRG ? @jonoomph ?

I'm reluctant to just merge it unobserved, not so much for the documentation changes but for fear of breaking the official doc builds. Due to the addition of MarkDown into the mix, it will likely require at least a one-time update of the dependencies on the GitLab Linux builder — most likely, just grabbing the `requirements.txt` contents from this branch:
```python3
Sphinx ~= 4.3.2
Pillow
sphinx_rtd_theme
sphinx_copybutton
sphinxcontrib-svg2pdfconverter[Cairosvg]
cairosvg
myst_parser
```
And running a `pip install -r requirements.txt`. (I'm going to assume that `libcairo.so` is already available on the machine, which is the only new external dependency. Because it's almost impossible to _not_ have libcairo available, on Linux.)

Once all of the Python packages are installed, doc builds should be the same as before, though now we also have the option of autogenerating PDF docs on the project builders as well.

As mentioned, you can check out the generated _output_ files from the new process by downloading a zip Artifact containing either the HTML source or the PDF from any of the CI runs, say for example these:
* [OpenShot 2.6.1-dev User Guide (HTML)](https://github.com/OpenShot/openshot-qt/suites/4870251921/artifacts/138924431)
* [OpenShot 2.6.1-dev User Guide (PDF)](https://github.com/OpenShot/openshot-qt/suites/4870251921/artifacts/138924432)

...That is, until they expire, which I think happens after 90 days (so, April 8).

ferdnyc:
Artifacts from the most recent build [can be found here](https://github.com/OpenShot/openshot-qt/actions/runs/1981772023)

PR #4639, opened 2022-01-07
Don't start the region selector at frame 1 if a clip is sliced

### LINKED ISSUE #4635: Tracker Effect does not recognize a video clipp is sliced(keep right side only) and starts from Frame 1 when performing Region selection.
**Describe the bug:**
The Tracker Effect does not properly recognize that a video clip has been sliced.  Meaning, whey you apply the Tracker Effect to a video clip and then click on "Click to Select" to select the Region, the entire clip is shown instead starting from frame 1.

1. Start a new project
2. Import a video clip.  In my case .mp4 file.
3. Slice the video clip in half and keep the right side only.
4. Drag the Tracker effect onto this clip.
5. From the Tracker:Initialize Effect windows, Click on "Click to Select".
6. You will now notice that the "Select Region" window is showing you the left half of the clip which is the half you didn't keep during the Slice process.
7.  If you go ahead and select a region with the mouse and then click on "Select Region".
8. Then click on "Process Effect".  It completes the process but never achieve what you actually want to track.

This is the main issue.

**Expected behavior:**
The "Select Region" window should be displaying the right half of the sliced video clip to allow me to select the region I want to track and then process it successfully to achieve the desired tracking.

**System Details:**
- OpenShot Version 2.6.1-dev, libopenshot: 0.2.7, Openshot-v2.6.1-dev-daily-8946
- Operating System / Distro: Windows 10 with the latest updates.
Windows 10 with 1TB NVMe, Nvidia GeForce 2060 w/6GB RAM, and 32GB RAM, AMD Ryzen 7 3700x 8-Core processor

**Log Files:**
If you are experiencing a crash, please collect and attach logs of the problem.  Log files can be found in the `.openshot_qt` folder in your user home directory. Log files over 2 MB in size will need to be compressed, please attach a ZIP file instead of the raw logs.

**Exception / Stacktrace:**
If you are experiencing a crash or strange output on the command line, please attach a small snippet of the exception here, but please keep it as small as possible (for readability).

**Screenshots:** (Optional)
If applicable, add screenshots to help explain your problem. To include screenshots, copy/paste them here, drag the image(s) into this form, or click the link below this edit field to load images saved to your computer. All images are public, so please don't post screenshots containing personal information.

JacksonRG:
Fixes #4635

Our tracker effect was always starting at frame 1. This change uses the clip's `start` property if it exists.

Still to test: make sure region selector doesn't show removed frames at the end either.

### DIFF
--- a/src/windows/region.py
+++ b/src/windows/region.py
@@ -172,8 +172,9 @@ class SelectRegion(QDialog):
 
         # Determine if a start or end attribute is in this file
         start_frame = 1
-        # if 'start' in self.file.data.keys():
-        #     start_frame = (float(self.file.data['start']) * self.fps) + 1
+        # For sliced clips, use start*fps as the first frame
+        if clip and clip.Json():
+            start_frame = (json.loads(clip.Json()).get("start",0) * self.fps) + 1
 
         # Display start frame (and then the previous frame)
         QTimer.singleShot(500, functools.partial(self.sliderVideo.setValue, start_frame + 1))
--- a/src/windows/region.py
+++ b/src/windows/region.py
@@ -96,7 +96,6 @@ class SelectRegion(QDialog):
         # Get profile settings
         c_info = clip.Reader().info
         self.fps = c_info.fps.ToInt() #float(self.fps_num) / float(self.fps_den)
-        self.clipStartFrame = round(json.loads(clip.Json()).get('start',0) * self.fps) + 1
         self.fps_num = self.fps #int(file.data['fps']['num'])
         self.fps_den = 1 #int(file.data['fps']['den'])
         self.width = c_info.width #int(file.data['width'])
--- a/src/windows/region.py
+++ b/src/windows/region.py
@@ -82,15 +82,12 @@ class SelectRegion(QDialog):
 
         # Create region clip with Reader
         self.clip = openshot.Clip(clip.Reader())
-
         self.clip.Open()
 
         # Set region clip start and end
         self.clip.Start(clip.Start())
         self.clip.End(clip.End())
         self.clip.Id( get_app().project.generate_id() )
-
-        print("IDS {} {}".format(clip.Id(), self.clip.Id()))
 
         # Keep track of file object
         self.file = file
@@ -135,9 +132,6 @@ class SelectRegion(QDialog):
         self.r.SetMaxSize(self.viewport_rect.width(), self.viewport_rect.height())
 
         try:
-            # Add clip for current preview file
-            self.clip = openshot.Clip(self.file_path)
-
             # Show waveform for audio files
             if not self.clip.Reader().info.has_video and self.clip.Reader().info.has_audio:
                 self.clip.Waveform(True)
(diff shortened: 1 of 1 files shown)


src/windows/region.py
```diff
@@ -172,8 +172,9 @@ def __init__(self, file=None, clip=None):
...
         # Determine if a start or end attribute is in this file
         start_frame = 1
-        # if 'start' in self.file.data.keys():
-        #     start_frame = (float(self.file.data['start']) * self.fps) + 1
+        # For sliced clips, use start*fps as the first frame
+        if clip and clip.Json():
+            start_frame = (json.loads(clip.Json()).get("start",0) * self.fps) + 1
 
         # Display start frame (and then the previous frame)
         QTimer.singleShot(500, functools.partial(self.sliderVideo.setValue, start_frame + 1))
```
JacksonRG: Just curious why we display the second then the first frame if anyone can tell me.

jonoomph: Looks like a bug, since we already increment start_frame + 1 earlier

src/windows/region.py
```diff
@@ -133,9 +130,6 @@ def __init__(self, file=None, clip=None):
         self.r.SetMaxSize(self.viewport_rect.width(), self.viewport_rect.height())
 
         try:
-            # Add clip for current preview file
-            self.clip = openshot.Clip(self.file_path)
```
jonoomph, line 137: This was recreating the attached Clip object for no reason (at least no reason I could tell). By removing this line, we can still access the `self.clip` object, which already has the Start() and End() properties set. However, we then add this clip to a new Timeline at `position: 0.0`. So our slider always starts at frame 1 and ends at the `clip.Duration() * clip.fps()`.

src/windows/region.py
```diff
@@ -167,17 +161,11 @@ def __init__(self, file=None, clip=None):
         self.sliderVideo.setMinimum(1)
         self.sliderVideo.setMaximum(self.video_length)
         self.sliderVideo.setSingleStep(1)
-        self.sliderVideo.setSingleStep(1)
```
jonoomph, line 170: Duplicate line

src/windows/region.py
```diff
@@ -167,17 +161,11 @@ def __init__(self, file=None, clip=None):
...
 
-        # Determine if a start or end attribute is in this file
-        start_frame = 1
-        # if 'start' in self.file.data.keys():
-        #     start_frame = (float(self.file.data['start']) * self.fps) + 1
-
         # Display start frame (and then the previous frame)
-        QTimer.singleShot(500, functools.partial(self.sliderVideo.setValue, start_frame + 1))
-        QTimer.singleShot(600, functools.partial(self.sliderVideo.setValue, start_frame))
+        QTimer.singleShot(500, functools.partial(self.sliderVideo.setValue, 2))
```
jonoomph, line 167: Not sure why we start at frame 2 and then jump to 1... I think this is a hack to correctly trigger the `valueChanged()` signal. But I see no reason to change this at the moment, until we have time to investigate it further.

PR #4640, opened 2022-01-08
Add required libqt5svg5 library into developer documentation

bieli:
Running `sudo apt install libqt5svg5-dev` before compiling `libopenshot` is required for success in building all components for developement.

### DIFF
--- a/doc/developers.rst
+++ b/doc/developers.rst
@@ -98,6 +98,7 @@ unofficial Ubuntu repository, which has our software packages available to downl
                         pkg-config \
                         python3-dev \
                         qtbase5-dev \
+                        libqt5svg5-dev \
                         qtmultimedia5-dev \
                         swig
 
--- a/doc/developers.rst
+++ b/doc/developers.rst
@@ -99,6 +99,7 @@ unofficial Ubuntu repository, which has our software packages available to downl
                         python3-dev \
                         qtbase5-dev \
                         libqt5svg5-dev \
+                        libxcb-xfixes0-dev \
                         qtmultimedia5-dev \
                         swig
 


doc/developers.rst
```diff
@@ -99,6 +99,7 @@ unofficial Ubuntu repository, which has our software packages available to downl
                         python3-dev \
                         qtbase5-dev \
                         libqt5svg5-dev \
+                        libxcb-xfixes0-dev \
```
bieli, line 102: Without this package we can't finish `make install` for qt5 library with success

ferdnyc:
@bieli True, if not using Resvg, then QtSVG is required. (With Resvg it's unnecessary.) Good call, makes sense to add that in.

doc/developers.rst
ferdnyc: I've never even heard of `libxcb-xfixes0-dev`, and it's certainly not anything we explicitly install in either our CI builds or the official PPA builds.

Hmm. In [our recent PPA builds](https://launchpadlibrarian.net/583137755/buildlog.txt.gz) I can see where `libxcb-xfixes0` itself is pulled in, along with the rest of the xcb libs, but the `-dev` packages are never pulled in and the build runs fine.

What build failures were you getting without `libxcb-xfixes0-dev`? Can you post the output from `cmake`/`g++`/`clang++`?

ferdnyc: Oh, sorry, you said `make install`... So the build is successful, but the _install_ fails without `libxcb-xfixes0-dev`? I'm even more confused, if that's the case.

bieli: @ferdnyc Yes. My patch is useful for developers who would like to replicate `make install` case from the documentation :-) Is it something, that we can merge?

jonoomph:
Thanks! LGTM

PR #4644, opened 2022-01-12
Prevent removing clips, effects or transitions if they're on a locked track

JacksonRG:
# The Issue
Currently locking a track doesn't prevent the clips on it from being deleted (or edited for that matter) it only locks them to a position on the timeline.

To me it makes sense that if you care enough to want a clip to stay where it is, you likely don't want to accidentally delete it.
# The fix
Before deleting any of those items, filter out any that are on locked tracks

# Question for reviewers:
How much locking do we want the lock feature to do?

By this I mean that if you lock a track how you like it, you can still edit properties of the clips (like alpha value and rotation). Unless we want to completely freeze all editing on locked clips, I thought it made sense still allow adding and removing of effects.

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -1747,10 +1747,15 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
     def actionRemoveClip_trigger(self):
         log.debug('actionRemoveClip_trigger')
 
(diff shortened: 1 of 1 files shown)


jonoomph:
@JacksonRG LGTM!

PR #4653, opened 2022-01-23
STY: Using bool on comparisons is not necessary

MartinThoma:


ferdnyc:
@MartinThoma Fair 'nuff. I sometimes like the readability of being a little more explicit... but some quick tests with `timeit.repeat` show that this:

```python3
>>> repeat(stmt="n += 1; trythis(n)", setup="""n = 0
def trythis(x):
    return bool(x % 14 == 0)""")
```
Takes roughly 10-20% longer to execute (somewhat surprisingly, at least to me), compared to this:
```python3
>>> repeat(stmt="n += 1; trythis(n)", setup="""n = 0
def trythis(x):
    return x % 14 == 0""")
```
So I certainly can't make the argument that it's an inconsequential syntactic flourish.

MartinThoma:
Thanks for the benchmark! That is also surprising to me, but I guess the numbers are so low that it doesn't really matter. Readability counts (although we seem to disagree what the more readable version is in this case 😅 )

PR #4656, opened 2022-01-24
Freeze and unfreeze toolbar along with other windows.

### LINKED ISSUE #4255: Freeze View does not lock toolbar
OpenShot 2.5.1-dev3
Windows 7

Seems like the toolbar and the views are not considered the same so when i use Freeze Views the toolbar can still be moved around.

Why not freeze/lock everything including the toolbar when freeze is used?
That would be more consistent.
Then perhaps move Freeze View into the Views menu and call it Lock instead of Freeze.

JacksonRG:
Closes #4255

**Issue:**
When freezing the view, toolbar could still be dragged around.

**Fix:**
Toggle `toolBar.movable` along with the other dock widgets.

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -2063,6 +2063,7 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
     def freezeDocks(self):
         """ Freeze all dockable widgets on the main screen
             (prevent them being closed, floated, or moved) """
+        self.toolBar.setMovable(False)
         for dock in self.getDocks():
             if self.dockWidgetArea(dock) != Qt.NoDockWidgetArea:
                 dock.setFeatures(QDockWidget.NoDockWidgetFeatures)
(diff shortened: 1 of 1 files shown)


ferdnyc:
@JacksonRG Makes sense to me! The only concern I have is that freezing the view while the toolbar is _floating_ makes it non-movable. (Which isn't the case with a floated dock, because it's still a window that can be dragged by its titlebar.)

So I wonder if maybe there should be a check of `toolBar.floating`, first, before it's potentially frozen? Just so we don't stick people with any non-draggable screen elements.

And then we'd want to link the movability to the floating status, so that it freezes once it gets docked (if the interface is frozen)... actually, I came up with some things, I'll open a PR against this branch for you to look at.

JacksonRG:
@ferdnyc
Good thinking! I have window dragging bound to a key, so I hadn't noticed. That's a reminder for me to get out of my window manager occasionally.

ferdnyc:
@JacksonRG Heh. Yeah, GNOME Shell _can_ drag any window (or other onscreen element) using <kbd>Super</kbd>+LMB, but that's cheating!

ferdnyc (approved):
Once the CI build finishes, I'll go ahead and merge this.

JacksonRG:
> Once the CI build finishes, I'll go ahead and merge this.
Ty, I was about to merge on Friday, but had merge conflicts for some reason.

ferdnyc:
@JacksonRG Yeah, I had to do a little history rewriting to get the post-my-PR-merge branch properly rebased on top of current `develop` and mergeable, but I eventually sorted it out. (Hopefully, we'll see what `develop`'s history looks like now that it's merged.)

ferdnyc:
Huh. Looks fine, actually.

image

Well, cool.

PR #4660, opened 2022-01-26
Fix font size issues in title editor

hexcone:
If you create a 'Bar 2' title (in the title editor), and change the font... the following things fail:
We are looking at the wrong SVG attribute (font_size instead of font-size)
We don't deal with titles like 'bar 2', which contain multiple font sizes

Solution:
We need to determine the % change in font-size... before and after the font dialog is used. If the user has increased/decreased the font size by X%, we need to apply that change for all font-sizes mentioned in the SVG file.

### DIFF
--- a/src/windows/title_editor.py
+++ b/src/windows/title_editor.py
@@ -436,10 +436,8 @@ class TitleEditor(QDialog):
             self.font_family = fontinfo.family()
             self.font_style = fontinfo.styleName()
             self.font_weight = fontinfo.weight()
-            if (oldfontinfo.pixelSize() != 0):
+            if (oldfontinfo.pixelSize() > 0):
                 self.font_size_ratio = fontinfo.pixelSize() / oldfontinfo.pixelSize()
-            else:
-                self.font_size_ratio = 0
             self.set_font_style()
             self.save_and_reload()
 
--- a/src/windows/title_editor.py
+++ b/src/windows/title_editor.py
@@ -436,7 +436,10 @@ class TitleEditor(QDialog):
             self.font_family = fontinfo.family()
             self.font_style = fontinfo.styleName()
             self.font_weight = fontinfo.weight()
-            self.font_size_ratio = fontinfo.pixelSize() / oldfontinfo.pixelSize() if oldfontinfo.pixelSize() else 0
+            if (oldfontinfo.pixelSize() != 0):
+                self.font_size_ratio = fontinfo.pixelSize() / oldfontinfo.pixelSize()
+            else:
+                self.font_size_ratio = 0
             self.set_font_style()
             self.save_and_reload()
 
--- a/src/windows/title_editor.py
+++ b/src/windows/title_editor.py
@@ -99,7 +99,7 @@ class TitleEditor(QDialog):
 
         self.font_weight = 'normal'
         self.font_style = 'normal'
-        self.font_size_pixel = 20
+        self.font_size_ratio = 1
 
         self.new_title_text = ""
         self.sub_title_text = ""
@@ -288,7 +288,7 @@ class TitleEditor(QDialog):
             # Set font size (for possible font dialog)
             s = node.attributes["style"].value
             ard = style_to_dict(s)
-            fs = ard.get("font_size")
+            fs = ard.get("font-size")
             if fs and fs.endswith("px"):
                 self.qfont.setPixelSize(float(fs[:-2]))
             elif fs and fs.endswith("pt"):
@@ -432,10 +432,11 @@ class TitleEditor(QDialog):
         if ok and font is not oldfont:
             self.qfont = font
             fontinfo = QtGui.QFontInfo(font)
+            oldfontinfo = QtGui.QFontInfo(oldfont)
             self.font_family = fontinfo.family()
             self.font_style = fontinfo.styleName()
             self.font_weight = fontinfo.weight()
-            self.font_size_pixel = fontinfo.pixelSize()
+            self.font_size_ratio = fontinfo.pixelSize() / oldfontinfo.pixelSize() if oldfontinfo.pixelSize() else 0
             self.set_font_style()
             self.save_and_reload()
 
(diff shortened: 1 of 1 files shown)


src/windows/title_editor.py
```diff
@@ -432,10 +432,11 @@ def btnFont_clicked(self):
         if ok and font is not oldfont:
             self.qfont = font
             fontinfo = QtGui.QFontInfo(font)
+            oldfontinfo = QtGui.QFontInfo(oldfont)
             self.font_family = fontinfo.family()
             self.font_style = fontinfo.styleName()
             self.font_weight = fontinfo.weight()
-            self.font_size_pixel = fontinfo.pixelSize()
+            self.font_size_ratio = fontinfo.pixelSize() / oldfontinfo.pixelSize() if oldfontinfo.pixelSize() else 0
```
jonoomph: Will this cause a potential divide by zero error? I would suggest splitting this logic onto multiple lines for clarity. :+1:

src/windows/title_editor.py
```diff
@@ -288,7 +288,7 @@ def load_svg_template(self, filename_field=None):
             # Set font size (for possible font dialog)
             s = node.attributes["style"].value
             ard = style_to_dict(s)
-            fs = ard.get("font_size")
+            fs = ard.get("font-size")
```
jonoomph, line 291: Were you able to confirm this attribute is simply misspelled? Did you find any mention of it in our SVG files? Or in the SVG file format spec?

jonoomph:
Looking good. I love the idea of adjusting the font-size based on a % changed. I'm still not sure how this will work correctly with different font-sizes in a single SVG. For example, let's say an SVG has 2 font sizes:   30pt and 52pt. The user opens the font dialog and chooses size 40pt.  Which of the "old" font sizes are we basing the % change from?

I would assume we need to be consistent, and always choose the smallest or largest (which ever makes the most sense). The multiple font-sizes in an SVG make this one a harder task for sure, lol.

src/windows/title_editor.py
hexcone: Referring to the latest SVG specifications (https://www.w3.org/TR/SVG2/) > G.2 .Presentation attributes (https://www.w3.org/TR/SVG2/attindex.html#PresentationAttributes), we can see that the correct spelling of the attribute is indeed "font-size". Running `grep -rni "font_size" --include="*.svg"` will return no results, whereas `grep -rni "font-size" --include="*.svg"` will generate results. All the .svg files in the project are using the correct spelling of "font-size" :)

src/windows/title_editor.py
hexcone: This if-else one liner is indeed catering for the case whereby `fontinfo.pixelSize()` is 0, we will only perform the division if `fontinfo.pixelSize()` is a non-zero value. If `fontinfo.pixelSize()` is 0, we set `font_size_ratio` to 0, any new font size multiplied by this ratio will also be 0. Currently `self.font_size_ratio` is only used for multiplication with the new font size. Are you refering to the chance whereby `self.font_size_ratio` is used as the denominator for a division operation in future?

And sure! We can definitely split up the one-liner to make it more readable :)

hexcone:
> Looking good. I love the idea of adjusting the font-size based on a % changed. I'm still not sure how this will work correctly with different font-sizes in a single SVG. For example, let's say an SVG has 2 font sizes: 30pt and 52pt. The user opens the font dialog and chooses size 40pt. Which of the "old" font sizes are we basing the % change from?
> 
> I would assume we need to be consistent, and always choose the smallest or largest (which ever makes the most sense). The multiple font-sizes in an SVG make this one a harder task for sure, lol.

The current situation is that the font size displayed on the font dialog is simply the last textual node from the svg file. In my opinion, to choose the smallest/largest font size might be a little confusing to the end user, it might be hard for them to correlate which is the font size they are changing, to the text on the screen.

A simple solution might be to change the font dialog from 30pt, 52pt, etc to a concept of percentage. eg 0%, 100%, 200% perhaps. Which is what our backend logic is currently doing :)

Another solution is to have individual "Change font" option for each of the text lines. This gives the user more flexiblility to customize each of the text individually. However, the downside is that the interface might look more cluttered.

Let me know what you think!

src/windows/title_editor.py
```diff
@@ -432,10 +432,14 @@ def btnFont_clicked(self):
...
             fontinfo = QtGui.QFontInfo(font)
+            oldfontinfo = QtGui.QFontInfo(oldfont)
             self.font_family = fontinfo.family()
             self.font_style = fontinfo.styleName()
             self.font_weight = fontinfo.weight()
-            self.font_size_pixel = fontinfo.pixelSize()
+            if (oldfontinfo.pixelSize() != 0):
+                self.font_size_ratio = fontinfo.pixelSize() / oldfontinfo.pixelSize()
+            else:
+                self.font_size_ratio = 0
```
jonoomph: If we set `font_size_ratio` = 0, won't this eventually lead to a 0 sized `new_font_size_pixel`? I'm unclear why the oldfontinfo.pixelSize would be zero to begin with. If we don't detect a valid oldfontsize, should the ratio default to 1, and just stay the same size? What do you think?

jonoomph:
> The current situation is that the font size displayed on the font dialog is simply the last textual node from the svg file.

This is fine for now. :+1:

src/windows/title_editor.py
jonoomph: I'm really only afraid of a user launching the editor and accidentally making the font-size 0 (and disappearing), lol

hexcone: My initial thinking was that svg files can be modified by users, or user supplied, and might contain invalid values like 0, which will result in an error when we use it as a denominator. But it might not be an issue since we are handling pre-defined templates.

Sure, let's default the font size to 1. Since it is already initialized to 1, we will perform the division when `oldfontinfo.pixelSize() > 0`

JacksonRG:
> A simple solution might be to change the font dialog from 30pt, 52pt, etc to a concept of percentage. eg 0%, 100%, 200% perhaps. Which is what our backend logic is currently doing :)

I like this... Might need some changing to the wording in the title editor. Like "Scale font size by:"

jonoomph:
LGTM, Merging

PR #4671, opened 2022-02-03
Link toolbar movability to floating status

ferdnyc:
This PR, targeting @JacksonRG 's PR #4656, will link freezing of the main toolbar to its floating status, so that we don't create an immovable floating toolbar when we freeze the window.

The toolbar's `topLevelChanged` signal (which indicates floating status) is connected to a slot that applies current frozen status if it's docked, and ensures it's always movable if it's floating. The same slot ensures that if a floating toolbar gets docked while the window is frozen, it will freeze in place and no longer be movable until the freeze is deactivated.

I also slightly re-worked how initial freeze state is applied when loading settings, as well as how the docks get frozen/unfrozen, to reduce code duplication.

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -733,7 +733,7 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
             _("Would you like to import %s as an image sequence?") % filename,
             QMessageBox.No | QMessageBox.Yes
         )
-        return bool(ret == QMessageBox.Yes)
+        return ret == QMessageBox.Yes
 
     def actionAdd_to_Timeline_trigger(self, checked=False):
         # Loop through selected files
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -2063,6 +2063,7 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
     def freezeDocks(self):
         """ Freeze all dockable widgets on the main screen
             (prevent them being closed, floated, or moved) """
+        self.toolBar.setMovable(False)
         for dock in self.getDocks():
             if self.dockWidgetArea(dock) != Qt.NoDockWidgetArea:
                 dock.setFeatures(QDockWidget.NoDockWidgetFeatures)
@@ -2070,6 +2071,7 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
     def unFreezeDocks(self):
         """ Un-freeze all dockable widgets on the main screen
             (allow them to be closed, floated, or moved, as appropriate) """
+        self.toolBar.setMovable(True)
         for dock in self.getDocks():
             if self.dockWidgetArea(dock) != Qt.NoDockWidgetArea:
                 if dock is self.dockTimeline:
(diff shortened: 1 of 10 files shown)


src/windows/main_window.py
```diff
@@ -2060,39 +2061,41 @@ def showDocks(self, docks):
...
+            # Don't freeze undockable widgets
+            return
+        if frozen:
+            dock.setFeatures(QDockWidget.NoDockWidgetFeatures)
+        else:
+            features = (
+                QDockWidget.DockWidgetFloatable
+                | QDockWidget.DockWidgetMovable)
+            if dock is not self.dockTimeline:
+                features |= QDockWidget.DockWidgetClosable
```
JacksonRG, line 2076: I had questions whether a `|=` was really called for. But with how Qt stores features, it looks like best approach.

src/windows/main_window.py
```diff
@@ -3061,6 +3065,10 @@ def __init__(self, *args):
         self.SelectionAdded.connect(self.addSelection)
         self.SelectionRemoved.connect(self.removeSelection)
 
+        # Connect toolbar float changes to movability/frozenness
```
JacksonRG: Imagining reading this without reading `freezeMainToolBar`, the function name implies that we freeze the tool bar whenever it gets docked/undocked. Including "Freeze toolbar IF the view is frozen" in the comment might clarify.

JacksonRG (approved):
Just a comment suggestion. The code looks awesome!

src/windows/main_window.py
ferdnyc: Fair enough. I sort of take it as an _invitation_ to read `freezeMainToolBar` to learn the details (I mean, there's a lot that's implicit here, including how the `topLevelChanged` signal works), but point taken. How about...
```suggestion
        # Ensure toolbar is movable when floated (even with docks frozen)
```
That's not the _entirety_ of what `freezeMainToolBar()`  does (since it also freezes it when _re_-docked), but it's the main reason it exists.

src/windows/main_window.py
ferdnyc: Yeah, all the alternatives are uglier / more redundant.Either, more like the old code,
```python3
if dock is self.dockTimeline:
    features = (
        QDockWidget.DockWidgetFloatable
        | QDockWidget.DockWidgetMovable)
else:
    features = (
        QDockWidget.DockWidgetFloatable
        | QDockWidget.DockWidgetMovable
        | QDockWidget.DockWidgetClosable)
dock.setFeatures(features)
```
(or just two separate calls to `dock.setFeatures()` directly, without a `features` variable...)

Or else, something weird like:
```python3
features = (
    QDockWidget.DockWidgetFloatable
    | QDockWidget.DockWidgetMovable)
if dock is self.dockTimeline:
    dock.setFeatures(features)
else:
    dock.setFeatures(features | QDockWidget.DockWidgetClosable)
```

ferdnyc:
Huh. Looks fine, actually.

image

Well, cool.

PR #4672, opened 2022-02-10
Add check such that clicking "Play" only advance playhead up to end of last clip

hexcone:
Problem:
Launching OpenShot, with no project open, and clicking Play over and over again, causes the playhead to move 1 frame at a time. This is invalid behavior, and playhead should not move past the end of a project, or at all in an empty project.

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -904,8 +904,6 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
 
     def should_play(self, play_forward=True):
         max_frame = get_app().window.timeline_sync.timeline.GetMaxFrame()
-        log.debug("juliana: max_frame = %d", max_frame)
-        log.debug("juliana: self.preview_thread.current_frame = %d", self.preview_thread.current_frame)
         if play_forward and self.preview_thread.current_frame < max_frame:
             return True
         if not play_forward and self.preview_thread.current_frame <= max_frame:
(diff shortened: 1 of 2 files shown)


jonoomph:
@hexcone This is looking close! However, I think we are handling the symptom here (instead of the cause). For example, `onPositionChanged` is emitted after the actual position has changed and you are simply hiding the change from the UI.

This same basic logic could be moved to `actionPlay_trigger`, where we actually calculate the length of the clips and emit the following: `self.PlaySignal.emit(max_frame)`. I think the idea is to never emit the Play signal if you are >= calculated max_frame. Then the position will not change, and the onPositionChanged handler will never fire.

Hopefully this makes sense! :wink:

jonoomph:
In fact, the fast forward / rewind buttons might need some similar logic. Perhaps you can refactor our the `max_frame` calculation into it's own method, and protect all these buttons when we are out of range for valid Play/Rewind/Fast actions.

hexcone:
@jonoomph Ahhh, I see what you mean! I have moved the logic into main_window.py `actionPlay_trigger` `actionFastForward_trigger` and `actionRewind_trigger`.

One thing to note is, by default, the play function will play to the end of clip and stop after the last frame `max_frame+1`. As this is beyond the end of the last clip, the logic defines that the rewind button will not play anything. However, when we are within the clip, up to `max_frame`, the rewind button will play the clip backwards as expected.

jonoomph:
@hexcone While testing this PR I have actually run into a bunch of other issues, related to Play/Pause toggles, updating icon, speed, and mode. I'm going to merge a much larger refactor of the Play/Pause button handling soon... which unfortunately means this PR will not be merged. I do appreciate the help on it, and I'll mention this PR in the commit.

hexcone:
@jonoomph No worries! Awesome!

PR #4678, opened 2022-02-13
Caching & Playback Timing Improvements (UI)

jonoomph:
Related to https://github.com/OpenShot/libopenshot/pull/796/. (will break without 796 being merged first)

**Fixes in this PR:**
 - Fixing some **audio playback issues** where opening a project / creating a project would not use the correct sample rate
 - Fixing **cache rendering issue**, which would leave blanks unrendered (i.e. blue bar)
 - Increase **default cache** size to **1024 MB**
 - Change of **default sample rate** to **48000** (which is the default on Windows 10+ now)
 - Adding **Playback FPS** to the UI (for troubleshooting)
 - Logging **default system sample rate**, for troubleshooting (especially on Windows if audio plays back at the wrong rate, too slow or too fast). Sample rate must match OpenShot's preferences if you experience an issue with audio pitch/speed.
 - Always **reset playhead to 0** when opening / creating a project
 - **Refactor to Play/Pause buttons**, and cleaned up a bunch of sloppy icon, toggle playback code.
 - Fixed a bug which would **increment +1 timeline position** by clicking Play on an empty timeline

### DIFF
--- a/freeze.py
+++ b/freeze.py
@@ -202,7 +202,7 @@ if sys.platform == "win32":
     extra_exe = {"base": None, "name": exe_name + "-cli.exe"}
 
     # Standard graphical Win32 launcher
-    base = "Win32GUI"
+    #base = "Win32GUI"
     build_exe_options["include_msvcr"] = True
     exe_name += ".exe"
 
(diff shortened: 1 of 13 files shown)


jonoomph:
Merging this now. We have had some users helping us test this branch, and we are finally getting good feedback from this PR and https://github.com/OpenShot/libopenshot/pull/796. Need to get this into the daily build and expand testing.

PR #4683, opened 2022-02-16
Update to the YouTube export presets

### LINKED ISSUE #4675: Update "YouTube" and "YouTube-HD" target options in the "Simple" export video tab
**Describe the new feature:**

I think it would be a nice idea to update and expand the `YouTube` and `YouTube-HD` target profiles in the `Simple` export video tab to follow the official [recommended upload encoding settings](https://support.google.com/youtube/answer/1722171?hl=en) guide since for example at the moment 2K and 4K video export options are not present in the video profile selection in the `Simple` tab.

- Move the 720p export options from the `YouTube-HD` to the `YouTube` target menu since YouTube has removed the HD tag from 720p videos in 2020
- Update the already present export options if necessary
- Add 2K and 4K export options
- Add 48, 50 and 60 FPS export options for all video profiles under the `YouTube` and `YouTube-HD` targets

In alternative I think users could continue to manually set the export options (for example for 4K videos) using the `Advanced` tab following the beforementioned guide for recommended upload encoding, but since the paramenters are well explained I think it would be nice to add them all in the `Simple` tab

Trainax:
This PR closes #4675

- Moved the 720p export options from the `YouTube-HD` to the `YouTube` target menu since YouTube has removed the HD tag from 720p videos in 2020
- Updated the `YouTube` target to use the MP4 format and H.264 video codec as recommended in the guide
- Added 2K and 4K export options
- Added High Frame Rate (50, 59.94 and 60 FPS) export presets. To do this:
    - For 720p options I simply added them in the `YouTube` preset since the video bitrates for the preset were high enough (the "high" option is 8 Mb/s and the recommended bitrate for 720p 60 FPS is 7.5 Mb/s)
    - For HD, 2K and 4K options a separate file with higher bitrates for each resolution had to be created since there's currently no way to make export bitrate dependent on frame rate (see https://github.com/OpenShot/openshot-qt/commit/440b13e738caa470c745426c5864bcc4cd78a88b)

I made all these changes looking at the [recommended upload encoding settings](https://support.google.com/youtube/answer/1722171?hl=en#zippy=%2Ccontainer-mp%2Caudio-codec-aac-lc%2Cvideo-codec-h%2Cbitrate) guide. In particular for bitrates:
- when a single bitrate value is recommended in the guide, the "high" option is set to the recommended value
- when a range of bitrates is recommended (see 4K recommended export bitrates in the guide) the "low" bitrate option is the lowest value of the range and the "high" bitrate option is the highest value of the range. The "med" bitrate option is in between

Edit to add: I couldn't create a single `YouTube-HFR` profile because the bitrate between 1080p 50/60 FPS and 4K 50/60 FPS is very different and putting everything into a single profile would mean being able to choose only 3 bitrate options (low, med, high) to fit values which go from 12 Mb/s for 1080p 50/60 FPS up to 68 Mb/s for 4K 50/60 FPS and for that reason I don't think it's possibile to cover everything with a single profile and 3 bitrate options


jonoomph:
Thanks for this PR! A few **comments**:

- Update the names to more "translatable" friendly names: For example, instead of "YouTube-2k-HFR" - "YouTube HD" or "YouTube Ultra HD" or "YouTube HD (2K)"
- What is the HFR refer to? High frame rate? i.e. more than 24 FPS?  
 
I would **suggest** we don't divide things by HFR, since most users would be confused. I vote for a simple 3 divisions:  "YouTube Standard", "YouTube HD (2K)", and "YouTube HD (4K)". How does that sound?

Trainax:
> - Update the names to more "translatable" friendly names: For example, instead of "YouTube-2k-HFR" - "YouTube HD" or "YouTube Ultra HD" or "YouTube HD (2K)"

OK, so something like:

- "YouTube" => "YouTube Standard" 
- "YouTube-HD" => "YouTube HD" 
- "YouTube-2K" => "YouTube (W)QHD" 
- "YouTube-4K" => "YouTube Ultra HD" 

Is it OK in your opinion? 

> - What is the HFR refer to? High frame rate? i.e. more than 24 FPS?  

Exactly, HFR stands for "High Frame Rate", which is more that the Standard Frame Rate (SFR) of up to 30 FPS. In this case HFR is 50, 59.94 and 60 FPS. 

> I would **suggest** we don't divide things by HFR, since most users would be confused. I vote for a simple 3 divisions:  "YouTube Standard", "YouTube HD (2K)", and "YouTube HD (4K)". How does that sound?

It sounds good to me, however the *bitrate option problem* arises. What bitrate options do I insert? Let me make some examples:

- For "standard" quality (which is up to 720p) everything is OK as the "high" bitrate option of 8  Mb/s is higher than the recommended value for 720p 60 FPS video which is 7.5 Mb/s
- For example for "HD" quality (1080p) the recommended bitrate value for SFR videos is 8 Mb/s and for HFR videos is 12 Mb/s. Which bitrate value do I have to put in each bitrate option? Do I put 8 Mb/s in "low", 12 Mb/s in "high" and I put some value in between those two for the "med" option? This would mean that the "high" option would be excessive for SFR videos and the "low" option would be insufficient for HFR videos. Do we have a way to easily explain to the users what option to use and when?
- Same applies for (W)QHD and Ultra HD profiles (please see the [guide](https://support.google.com/youtube/answer/1722171?hl=en#zippy=%2Cbitrate))

In any case I think we should keep at least 4 different export presets:

- "YouTube Standard" for profiles up to 720p
- "YouTube HD" for 1080p profiles
- "YouTube (W)QHD" for 1440p (2K) profiles
- "YouTube Ultra HD" for 2160p (4K) profiles

Condensing the first 2 elements of this list into a single preset would mean increasing even more the *bitrate option problem* in my opinion as we would have to consider even more different values for all the different video export resolutions that preset would have inside (please see the guide to see all the different recommended bitrate options). 

Let me know what you think!

jonoomph:
@Trainax Thanks for the explanation and detailed response! Let me ponder this a bit. The more accurate we make this (according to the YouTube guide), the more complex it will appear to our users. So, this might be a balancing act between bitrate accuracy and usability.

jonoomph:
Okay, how about this (trying to find the balance, and maybe lean towards not using too much industry "lingo"):
 
- YouTube Standard
- YouTube HD
- YouTube HD (2K)
- YouTube HD (4K)

Let's go with your plan on bitrates, using the HRF bitrate on the "high" quality value, and the non-HRF bitrate on the "low" quality value, with a middle value for "med" quality value. This might result in some edge cases that are invalid bitrates, and file sizes that might be larger than needed, but I think it's a fair balance. For example, if I choose "med" or "low" quality on a HRF video, should I be surprised to see compression, etc..? Probably not. I think in most cases the user would choose "high" quality if they were planning on uploading the video to YouTube (just a theory).

How does that sound?

Trainax:
> How does that sound?

Sounds very good!

I will merge SFR and HFR presets into a single preset for each resolution making 4 total presets and I will update the preset names to be more easily translatable.

I will take another look at the guide to find the most appropriate value for each bitrate option of the 4 presets as suggested.

I will update you when I've done it 👍🏻

Trainax:
@jonoomph Done!

I have modified the preset names to be more easily translatable. The preset names now are:

- YouTube Standard
- YouTube HD
- YouTube HD (2K)
- YouTube HD (4K)

I have merged the HFR and the SFR presets into a single preset for each resolution. The video bitrate options now are:

| Preset name | Low (Mb/s) | Med (Mb/s) | High (Mb/s) |
| :---: | :---: | :---: | :---: |
| YouTube Standard | 2 | 5 | 8 |
| YouTube HD | 8 | 10 | 12 |
| YouTube HD (2K) | 16 | 20 | 24 |
| YouTube HD (4K) | 45 | 56 | 68 |

jonoomph:
LGTM! Thanks for the contribution, really appreciate it!

PR #4691, opened 2022-02-22
WIP: Make caption timestamps relative to the start of the clip

JacksonRG:
# Issue:
Timestamps took absolute times for when to start and stop displaying.

This meant that they had to be adjusted any time a clip was moved.

# Solution:
Subtract the `position` (the time at the left edge of a clip) before creating the timestamp.

Note: `libopenshot`'s caption code already accounted for the beginning clips, so no change to libopenshot is needed.


JacksonRG:
Closing. See below PR
https://github.com/OpenShot/openshot-qt/pull/4694

PR #4694, opened 2022-02-23
Caption relative timestamps rebase

JacksonRG:
# Issue:
Timestamps took absolute times for when to start and stop displaying.

This meant that they had to be adjusted any time a clip was moved.

# Solution:
Subtract the `position` (the time at the left edge of a clip) before creating the timestamp.

Note: `libopenshot`'s caption code already accounted for the beginning clips, so no change to libopenshot is needed.

Other Improvements:
- Prevent timestamps outside the time of the clip
- Prevent `add timestamp` button from adding after a line has an `ending` timestamp

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -2232,6 +2232,30 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
         app = get_app()
         _ = app._tr
 
+        if len(self.selected_effects) == 0:
+            log.error("No caption effect selected")
+            return
+        effect_data = Effect.filter(id=self.selected_effects[0])[0].data
+        effect_id = effect_data.get("id")
+        if effect_data.get("type") != "Caption":
+            log.error("Captioning an effect that is not a Caption")
+            return
+
+        # Get the Clip that owns this caption effect
+        clip_data = None
+        for clip in Clip.filter():
+            for effect in clip.data.get('effects'):
+                if effect.get("id") == effect_id:
+                    clip_data = clip.data
+                    break
+            if clip_data != None:
+                break
+
+        if clip_data == None:
+            log.error("No clip owns this caption effect")
+            return
+
+
         if self.captionTextEdit.isReadOnly():
             return
 
@@ -2239,6 +2263,7 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
         fps = get_app().project.get("fps")
         fps_float = float(fps["num"]) / float(fps["den"])
         current_position = (self.preview_thread.current_frame - 1) / fps_float
+        relative_position = current_position - clip_data.get("position")
 
         # Get cursor / current line of text (where cursor is located)
         cursor = self.captionTextEdit.textCursor()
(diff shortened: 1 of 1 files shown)


src/windows/main_window.py
```diff
@@ -2232,25 +2232,65 @@ def actionInsertTimestamp_trigger(self, event):
         app = get_app()
         _ = app._tr
 
+        if len(self.selected_effects) == 0:
+            log.error("No caption effect selected")
```
jonoomph: I think this might send data to Sentry.io, probably 'error' is to strong

src/windows/main_window.py
```diff
@@ -2232,25 +2232,65 @@ def actionInsertTimestamp_trigger(self, event):
...
+            return
+        effect_data = Effect.filter(id=self.selected_effects[0])[0].data
+        effect_id = effect_data.get("id")
+        if effect_data.get("type") != "Caption":
+            log.error("Captioning an effect that is not a Caption")
```
jonoomph: ditto

src/windows/main_window.py
```diff
@@ -2232,25 +2232,65 @@ def actionInsertTimestamp_trigger(self, event):
...
+        for clip in Clip.filter():
+            for effect in clip.data.get('effects'):
+                if effect.get("id") == effect_id:
+                    clip_data = clip.data
+                    break
+            if clip_data != None:
+                break
+
+        if clip_data == None:
+            log.error("No clip owns this caption effect")
```
jonoomph: Ditto

src/windows/main_window.py
```diff
@@ -2232,25 +2232,65 @@ def actionInsertTimestamp_trigger(self, event):
         app = get_app()
         _ = app._tr
 
+        if len(self.selected_effects) == 0:
```
jonoomph: `if self.selected_effects:` does the same thing

src/windows/main_window.py
```diff
@@ -2232,25 +2232,65 @@ def actionInsertTimestamp_trigger(self, event):
...
+            return
+        effect_data = Effect.filter(id=self.selected_effects[0])[0].data
+        effect_id = effect_data.get("id")
+        if effect_data.get("type") != "Caption":
+            log.info("Captioning an effect that is not a Caption")
+            return
+
+        # Get the Clip that owns this caption effect
+        clip_data = None
+        for clip in Clip.filter():
```
jonoomph, line 2246: Clip.filter() will return all clips in the project. It would probably be better if you could get the Clip ID of the parent clip, and pass it to the filter() function. Not sure if we have the Clip ID in the effect data though.

src/windows/main_window.py
```diff
@@ -2232,25 +2232,65 @@ def actionInsertTimestamp_trigger(self, event):
...
+                if effect.get("id") == effect_id:
+                    clip_data = clip.data
+                    break
+            if clip_data != None:
+                break
+
+        if clip_data == None:
+            log.info("No clip owns this caption effect")
+            return
+
```
jonoomph, line 2257: Extra blank line

src/windows/main_window.py
```diff
@@ -2232,25 +2232,65 @@ def actionInsertTimestamp_trigger(self, event):
...
         # Get cursor / current line of text (where cursor is located)
         cursor = self.captionTextEdit.textCursor()
         self.captionTextEdit.moveCursor(QTextCursor.StartOfLine)
         line_text = cursor.block().text()
         self.captionTextEdit.moveCursor(QTextCursor.EndOfLine)
 
+        # If position hasn't changed, and creating the ending timestamp, add 5 seconds
+        if "-->"  in line_text and line_text.count(':') == 3 and self.last_caption_timestamp_position == current_position:
+            # prevent caption with 0 duration
+            relative_position += 5
```
jonoomph: Should this be `+= 5.0`?

jonoomph:
I feel like it should be easier to get the Clip ID from the selected effect. But off the top of my head, I don't recall. Otherwise looks good.

src/windows/main_window.py
JacksonRG: I just checked `clip_data.get("id") in str(effect_data)` -> `false`. Not seeing a way to get clip id out of the effect

jonoomph:
LGTM

PR #4695, opened 2022-02-23
Revert "detect duration too long"

jonoomph:
Reverts OpenShot/openshot-qt#4539

The duration detection causes an infinite loop. Not sure how we missed this in testing. Reverting until we can rework this fix.

### DIFF
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -742,12 +742,9 @@ App.controller("TimelineCtrl", function ($scope) {
     return Math.min(32767, desired_width);
   };
 
-// Find the furthest right edge on the timeline (and resize it if too small or too large)
+// Find the furthest right edge on the timeline (and resize it if too small)
   $scope.resizeTimeline = function () {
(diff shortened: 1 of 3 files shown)


src/timeline/js/controllers.js
```diff
@@ -757,11 +754,10 @@ App.controller("TimelineCtrl", function ($scope) {
       }
     }
     // Resize timeline
-    if (furthest_right_edge > $scope.project.duration - min_timeline_padding || $scope.project.duration - furthest_right_edge > max_timeline_padding) {
+    if (furthest_right_edge > $scope.project.duration) {
       if ($scope.Qt) {
-        let new_timeline_length = Math.max(min_timeline_length, furthest_right_edge + min_timeline_padding);
-        timeline.resizeTimeline(new_timeline_length);
```
jonoomph, line 763: This line causes an update to the project's "duration" attribute, which triggers/propagates the change back `controllers::applyJsonDiff`, which then calls this function again. Infinity!

PR #4696, opened 2022-02-23
Fixing trimming support (which got broken due to a regression)

jonoomph:
Fixing trimming support (which got broken due to a regression). Trimming fully supports snapping and clip boundaries (0 to duration).

**Trimming features:**
 - Don't allow left edge before zero
 - Don't allow right edge after duration
 - Support snapping hints
 - Preview the frame as the user moves the cursor (while dragging the edge of the clip)

### DIFF
--- a/src/timeline/js/directives/clip.js
+++ b/src/timeline/js/directives/clip.js
@@ -180,8 +180,7 @@ App.directive("tlClip", function ($timeout) {
             cursor_position, cursor_position, "trimming");
 
           // Calculate delta from current mouse position
-          let new_position = cursor_position;
-          new_position = results.position.left;
+          let new_position = results.position.left;
           let delta_x = 0;
           if (dragLoc === "left") {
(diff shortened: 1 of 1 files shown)


src/timeline/js/directives/clip.js
```diff
@@ -214,21 +214,18 @@ App.directive("tlClip", function ($timeout) {
...
-              ui.element.width((new_right - new_left));
             }
+            ui.element.width(new_right - new_left);
           }
 
           // Preview frame during resize
           if (dragLoc === "left") {
             // Preview the left side of the clip
-            scope.previewClipFrame(scope.clip.id, new_left);
+            scope.previewClipFrame(scope.clip.id, new_left / scope.pixelsPerSecond);
```
jonoomph, line 224: Convert pixels to time (in seconds)

JacksonRG (approved):
Tested. LGTM

PR #4698, opened 2022-02-24
Fix Title Editor Regression (infinite timer loop) - causing slowdowns

jonoomph:
Fix an infinite timer loop on the Title Editor. This makes editing titles less CPU intensive, and delays the preview until typing has stopped for a moment. We were starting a timer, but never stopping it... so it just continued to generate temp SVG files and preview them forever.

### DIFF
--- a/src/windows/title_editor.py
+++ b/src/windows/title_editor.py
@@ -67,6 +67,7 @@ class TitleEditor(QDialog):
 
         # A timer to pause until user input stops before updating the svg
         self.update_timer = QTimer()
+        self.update_timer.setInterval(300)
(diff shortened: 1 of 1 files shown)


src/windows/title_editor.py
```diff
@@ -368,6 +369,7 @@ def writeToFile(self, xmldoc):
 
     def save_and_reload(self):
         """Something changed, so update temp SVG and redisplay"""
+        self.update_timer.stop()
```
JacksonRG, line 372: Oh! Now I see the difference between timers and `singleShot` timers.

JacksonRG (approved):
Looks Great!

PR #4702, opened 2022-02-28
Don't prompt to import as image sequence for titles

JacksonRG:
# Bug:
In the title editor, certain formats of file names would trigger the "Import as an Image Sequence?" dialog, but titles can only be individual files.

# Solution:
Added an optional argument to prevent the image sequence prompt


jonoomph:
LGTM

Colorjet3:
I tested this in 9208 and can confirm that it is fixed and working properly.  Additionally, I tested importing a sequence of images (outside of the titles), just to make sure that process still works and there were no issue with that either.  Nicely done!

PR #4703, opened 2022-03-02
Detect shorten project duration rebase

JacksonRG:
# Bug
When a very long clip is added to a project, the timeline expands to fit. If that clip is removed or shortened, the timeline was not shortening back down.
# Fix
If the timeline is beyond maximum extra length, the timeline will be shortened.
# Testing
1) Set a clip's position higher than the ending of the timeline
2) Move that clip to the far left end of the timeline
3) Observe that the timeline has shortened.

Example:

https://user-images.githubusercontent.com/42394129/156685929-882504e6-ad99-4d5e-980e-53fd63e9fc59.mp4

### DIFF
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -760,6 +760,11 @@ App.controller("TimelineCtrl", function ($scope) {
         timeline.resizeTimeline(furthest_right_edge + 10);
         $scope.project.duration = furthest_right_edge + 10;
       }
+    } else if ( $scope.project.duration - furthest_right_edge > 20) {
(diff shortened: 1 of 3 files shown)


jonoomph:
@JacksonRG LGTM. Given this created issues last time, you might want to do a quick video capture, showing OpenShot, the stdout/log output, and the behavior (both adding clips - extending timeline duration, and removing clips - shortening timeline duration). Also, if there are any edge cases, such as a minimum duration, explain how that works as well. I would just attach that video to this PR. Hope that makes sense!

PR #4706, opened 2022-03-02
Scale time keyframes to match changing fps

JacksonRG:
# Bug
When fps changes, the frame stored in a time keyframe doesn't change.
### Example:
1) Set profile to 30 fps
2) Add a 10 second clip
3) Reverse the clip (clip now starts at frame 300 ends at frame 1)
4) Change the profile to 60 fps
### Problem: 
The clip still at frame 300 and ends at frame 1, but the 10 second clip is now 600 frames, so the clip now starts in the middle.
# Test
Repeat the example above, and see the video starts and ends in the same place before and after changing FPS.

Also to confirm this code is running:
open a terminal and run `watch 'grep "time keyframes" $HOME/.openshot_qt/openshot-qt.log'`
If you run openshot in `--debug` mode, you should see openshot report scaling time keyframes

### DIFF
--- a/src/classes/keyframe_scaler.py
+++ b/src/classes/keyframe_scaler.py
@@ -25,6 +25,7 @@
  You should have received a copy of the GNU General Public License
  along with OpenShot Library.  If not, see <http://www.gnu.org/licenses/>.
  """
+from classes.logger import log
 
 
 class KeyframeScaler:
@@ -34,14 +35,14 @@ class KeyframeScaler:
     multiplied by the scaling factor, except X=1 (because the first
     frame never changes)"""
 
-    def _scale_x_value(self, value: float) -> int:
+    def _scale_value(self, value: float) -> int:
         """Scale value by some factor, except for 1 (leave that alone)"""
         if value == 1.0:
             return value
         # Round to nearest INT
         return round(value * self._scale_factor)
 
-    def _update_prop(self, prop: dict):
+    def _update_prop(self, prop: dict, is_time = False):
         """Find the keyframe points in a property and scale"""
         # Create a list of lists of keyframe points for this prop
         if "red" in prop:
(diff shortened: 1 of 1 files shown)


src/classes/keyframe_scaler.py
```diff
@@ -34,14 +35,14 @@ class KeyframeScaler:
...
-    def _scale_x_value(self, value: float) -> int:
+    def _scale_value(self, value: float) -> int:
         """Scale value by some factor, except for 1 (leave that alone)"""
         if value == 1.0:
             return value
         # Round to nearest INT
         return round(value * self._scale_factor)
 
-    def _update_prop(self, prop: dict):
+    def _update_prop(self, prop: dict, is_time = False):
```
jonoomph: Let's mention the args in the doc string, and explain them a bit

src/classes/keyframe_scaler.py
```diff
@@ -52,19 +53,23 @@ def _update_prop(self, prop: dict):
             keyframes = [prop.get("Points", [])]
         for k in keyframes:
             # Scale the X coordinate (frame #) by the stored factor
```
jonoomph: Invalid comment. Let's double check all the comments in the function, and update any that are misleading or wrong

src/classes/keyframe_scaler.py
```diff
@@ -52,19 +53,23 @@ def _update_prop(self, prop: dict):
...
-            [point["co"].update({
-                "X": self._scale_x_value(point["co"].get("X", 0.0))
-                })
-                for point in k if "co" in point]
+            if (is_time):
+                log.debug("Updating both values of time keyframes")
```
jonoomph: Maybe mention the "X" and "Y" coordinates  vs  "both values", just to be sure the log output is crystal clear

jonoomph:
@JacksonRG LG otherwise -  Oh, and we mention to "check the log entry" when a time keyframe is scaled (to verify it in the testing instructions)

PR #4707, opened 2022-03-03
Fixing Razor Cursor (which was broken on Windows & webkit backends)

jonoomph:
Fixing razor cursor, which was broken on Windows (i.e. webkit backend) when we moved from PNG to SVG icon file. Also, adding a new way to target CSS for webkit VS webengine.

We will eventually move away from a JS/CSS approach to the timeline, so this is really a bandage for now, to allow some slight differences on how we CSS style the timeline (depending on the webkit / webengine backend used).

**Testing Instructions:**
- Launch this build on Windows (or run with `--web-backend webkit` backend)
- Drop an emoji or video on the timeline
- Click the 'Razor' toolbar icon
- Move your mouse over the clip, and you should see the razor cursor


src/timeline/js/mixin_webengine.js
```diff
@@ -39,6 +39,7 @@ function init_mixin() {
     angular.element(document).ready(function () {
       timeline.qt_log("INFO", "Angular Ready");
       $("body").scope().enableQt();
+      $("body").addClass("webengine");
```
jonoomph, line 42: Adding a custom class to the `<body class="webengine">`, so we can use it to target the difference backends using CSS.

src/timeline/media/css/main.css
```diff
@@ -303,7 +303,11 @@ img {
   margin-right: 1px;
 }
 
-.razor_cursor {
+.webkit .razor_cursor {
+  cursor: url(../images/razor_line_with_razor.png), default !important;
```
jonoomph, line 307: Apparently webkit does not support SVG files for CSS cursors. I tried a bunch of different test files, different sizes, and never got one to appear. So using a PNG instead (but only for webkit).

JacksonRG: Bizarre! But good to know!

JacksonRG (approved):
Merge away!

PR #4709, opened 2022-03-04
Formatting change to Caption Effect

JacksonRG:
# Issue:
In the case that a user adds captions in an unusual order, the caption formatting interferes.
1) Add a starting caption timestamp, and the caption string below it.
2) Play the video, until the next audio to caption
3) Add a caption for that new audio
4) Go back to the first caption to mark it's end.
Demonstrated below:

https://user-images.githubusercontent.com/42394129/156821659-c0acc38b-e2e9-4472-8ee2-e3d4da2484cf.mp4

The logic to prevent a caption of 0 duration checks the current playhead position to where it was on the last timestamp click. Since it hasn't moved between `step 3` and `step 4`, 5 seconds is wrongly added.

(Smaller issue: adding the second timestamp adds `Enter caption text...` where the user has already typed.)

# Solution:
1) Check that the actual timecode isn't in the cursor's line.
2) Add placeholder text: (`Enter caption text...`) with the _first_ timecode, then move the cursor back up a line for the user to specify the end time (or click again for the 5 second default)

# Example with fix:

https://user-images.githubusercontent.com/42394129/156822430-810d9eec-1776-425d-9c38-3e3f70a8f681.mp4

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -2300,7 +2300,7 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
                 self.captionTextEdit.moveCursor(QTextCursor.End)
                 self.captionTextEdit.insertPlainText("\n\n")
             # Add timestamp, and placeholder caption
-            self.captionTextEdit.insertPlainText("%s --> \n %s" % (current_timestamp, _("Enter caption text...")))
+            self.captionTextEdit.insertPlainText("%s --> \n%s" % (current_timestamp, _("Enter caption text...")))
             # Return to timestamp line, to await ending timestamp
             self.captionTextEdit.moveCursor(QTextCursor.Up)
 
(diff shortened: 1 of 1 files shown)


src/windows/main_window.py
```diff
@@ -2276,21 +2277,32 @@ def actionInsertTimestamp_trigger(self, event):
...
-            # If the last line isn't blank, add two blank lines
+            # Current line isn't a starting timestamp, so add a starting timestamp
+
+            # If the current line isn't blank, go to end and add two blank lines
             if (self.captionTextEdit.textCursor().block().text().strip() != ""):
                 self.captionTextEdit.moveCursor(QTextCursor.End)
                 self.captionTextEdit.insertPlainText("\n\n")
-            self.captionTextEdit.insertPlainText("%s --> " % (current_timestamp))
+            # Add timestamp, and placeholder caption
+            self.captionTextEdit.insertPlainText("%s --> \n %s" % (current_timestamp, _("Enter caption text...")))
```
jonoomph: You are inserting an extra space after the `\n`

jonoomph:
1 Comment above (about an extra space). Otherwise, LGTM

JacksonRG:
Took care of that empty space. Merging

PR #4719, opened 2022-03-11
Trigger a redraw of the zoom slider when the timeline auto-shortens

JacksonRG:
# issue
My update to (#4703) lengthens or shortens the timeline to fit the clips in the project. But it wasn't updating the zoom slider with that new duration.

# Solution
I added a signal (`resizeTimeline`) and have the zoom slider update itself in response to that signal.

# Possible issue:
The zoom slider keeps its handles at the same positions they were at.
This means if your zoom slider was covering the first half of your project, and your project doubled in length,
your zoom slider will still be covering half of your project. (Now twice as much time)

https://user-images.githubusercontent.com/42394129/157952043-e7ef532a-da73-4f1c-bfd5-4e18077a87c2.mp4

# Testing:
1) Run in debug mode
2) Start a new project.
3) Add a clip, and set it's `position` higher than 300.
4) Watch the debug console.
    - Should see `DEBUG webview: Updating duration`
    - Then the debug window shouldn't keep moving until you interact with OpenShot again.

### DIFF
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -760,7 +760,7 @@ App.controller("TimelineCtrl", function ($scope) {
     if (furthest_right_edge > $scope.project.duration - min_timeline_padding || furthest_right_edge < $scope.project.duration - max_timeline_padding) {
       if ($scope.Qt) {
         let new_timeline_length = Math.max(min_timeline_length, furthest_right_edge + min_timeline_padding);
-        $scope.project.duration = new_timeline_length;
+        timeline.resizeTimeline(new_timeline_length);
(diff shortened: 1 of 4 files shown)


Colorjet3:
The test looks good and matches the video recording you provided.  However, I don't know what to do about this:  

Watch the debug console.

    Should see DEBUG webview: Updating duration
    Then the debug window shouldn't keep moving until you interact with OpenShot again.

...since I am not sure how to enable the DEBUG webview to watch the debug console.

JacksonRG:
@Colorjet3 No problem! Thanks for the help

PR #4725, opened 2022-03-14
Integrate Sentry.io `sample_rate` and `traces_sample_rate` correctly

jonoomph:
Integrate Sentry.io `sample_rate` and `traces_sample_rate` correctly from openshot.org. This should dramatically reduce the # of errors sent to Sentry.io, and setup the potential to use transactions to monitor performance.

We were previously only setting the transaction reporting %, and thus, our error sample_rate was defaulting to 100% of all errors... oops. This was blowing past our Sentry.io limits, and ignoring our web service (which was designed to dynamically set these values so we can adjust them without doing a release)

### DIFF
--- a/src/classes/info.py
+++ b/src/classes/info.py
@@ -109,15 +109,19 @@ LOG_LEVEL_CONSOLE = 'INFO'
 # Web backend selection, overridable at launch
 WEB_BACKEND = 'auto'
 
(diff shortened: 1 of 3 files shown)


JacksonRG (approved):
LGTM! This might make the repeated errors less of an issue too! Though I'm still looking into that

PR #4732, opened 2022-03-17
Make sure keyframes appear at the right place on the timeline

JacksonRG:
# Issue:
When cutting the beginning off of a clip, clicking a keyframe would send the playhead to a location that wasn't the frame the keyframe affected.

https://user-images.githubusercontent.com/42394129/158725801-2ae34f46-c0e4-4ade-8252-97d488d0eaf0.mp4

# Fix:
Subtracting the start value from the position of each keyframe

https://user-images.githubusercontent.com/42394129/158725840-66b93004-f1d0-4f7f-8228-ae9a38efc961.mp4

### DIFF
--- a/src/timeline/js/controllers.js
+++ b/src/timeline/js/controllers.js
@@ -154,7 +154,7 @@ App.controller("TimelineCtrl", function ($scope) {
   $scope.selectPoint = function(object, point) {
     var frames_per_second = $scope.project.fps.num / $scope.project.fps.den;
     var clip_position_frames = object.position * frames_per_second;
-    var absolute_seek_frames = clip_position_frames + parseInt(point);
+    var absolute_seek_frames = clip_position_frames + parseInt(point) - (object.start * frames_per_second);
 
     if ($scope.Qt) {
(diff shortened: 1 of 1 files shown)


jonoomph:
@JacksonRG I don't think this is correct. `object.position` represents an arbitrary position (i.e. a list of keyframe positions on the clip, for example). I think the real problem might be where we generate those list of points used for jumping around (`$scope.getKeyframes`).

jonoomph:
I just want to make sure we aren't applying invalid logic to all keyframe points... but if my assumption is invalid, please let me know.

JacksonRG:
> I think the real problem might be where we generate those list of points used for 
jumping around (`$scope.getKeyframes`).

@jonoomph

That's actually where I started with this! But I had to subtract `clip.start` in `getKeyframes` and also take that subtraction out of the math in `index.html`

I changed to this solution because it's just changing in once place. And it allows the `point` attribute to still represent the same `x` value as it does on the python side.

jonoomph:
LGTM

PR #4733, opened 2022-03-17
Client side error filtering

JacksonRG:
Prevent errors from being reported too rapidly, or if it's a repeated error.

# Example
I added errors (not in included in the PR) to three buttons to show how errors in under a second, or errors of the same name will be prevented. 

https://user-images.githubusercontent.com/42394129/158903687-591bd94e-2ac2-493a-ab94-ea5eeaabdfe2.mp4

### DIFF
--- a/src/classes/sentry.py
+++ b/src/classes/sentry.py
@@ -44,7 +44,9 @@ except ModuleNotFoundError:
     sdk = None
 
 
-min_error_freq = 10**6 # 1 second (in micro seconds.)
+min_error_freq = 1 # seconds required between errors
+last_send_time = None
+last_event_message = None
 def init_tracing():
     """Init all Sentry tracing"""
     if not sdk:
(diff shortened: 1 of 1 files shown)


jonoomph:
@JacksonRG A few thoughts here... in general, this looks a bit complex. Here are a few suggestions:

1) Don't use globals(): 
Can we simply use module scoped variables (at the top of the `sentry.py` file)? Then you don't need local variables either (i.e. `last_send_time = ...`, because you can just access the module variable.

2) Why micro seconds? How about just using # of seconds between errors.
```
>>> a = datetime.datetime.now()
>>> b = datetime.datetime.now()
>>> (b - a).total_seconds()
5.593407
>>> 
```

In general, I think the logic flow is fine, but seems like this could be simplified quite a bit.

JacksonRG:
2. Good point! I added the microseconds for precision but we don't need it if we're using whole seconds.

1. The globals came from when I couldn't figure out where the callbacks were being run. (Turns out they weren't) but the fact that I can access `min_error_frequency` means I don't need globals.

I'll get that changed.

src/classes/sentry.py
```diff
@@ -42,6 +44,9 @@
     sdk = None
 
 
+min_error_freq = 1 # seconds required between errors
+last_send_time = None
+last_event_message = None
 def init_tracing():
```
jonoomph, line 51: Nitpick: extra line before function

src/classes/sentry.py
```diff
@@ -42,6 +44,9 @@
     sdk = None
 
 
+min_error_freq = 1 # seconds required between errors
```
jonoomph: Nitpick: Move comment above variable on it's own line

src/classes/sentry.py
```diff
@@ -65,14 +70,43 @@ def init_tracing():
                                                                                        sample_rate,
                                                                                        traces_sample_rate))
 
+    def before_send(event,hint):
```
jonoomph, line 74: Add docstring to this function
```
"""
Function to filter out repetitive Sentry.io errors before sending them
"""
```

jonoomph:
LGTM

JacksonRG:
Cool. Just noticed I was retrieving the event's message twice. Once to compare, and once to store it in in the `last_event_message`. Just changed that. (and tested) Merging

PR #4738, opened 2022-03-22
Changing floats to integers to prevent crash with Python 3.10

rjnewsham:
The colour selector in the Chroma Key effect causes a crash due to the
change in Python 3.10 for no longer automatically converting floats to integers.
Changing the inputs to currentColor to intgegers to correct this.


ferdnyc:
@rjnewsham I knew there'd be at least one I missed! Thanks.

PR #4746, opened 2022-03-26
WIP: Preserve user choice of waveform display on project reload.

JacksonRG:
### Issue
1) Right click on a clip with audio
2) Click `Display` > `Show Waveform`
3) Save
3) Click `File` > `Recent` and re-load the current project.
4) See that the waveform no longer shows

### Solution
Store the user setting on each clip, and generate waveforms when loading projects.

### Test
Repeat the steps in _Issue_, and confirm that the waveform does display on reload.

### DIFF
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -41,6 +41,7 @@ from classes.json_data import JsonDataStore
 from classes.logger import log
 from classes.updates import UpdateInterface
 from classes.assets import get_assets_path
+from .waveform import get_audio_data
 from windows.views.find_file import find_missing_file
 
 from .keyframe_scaler import KeyframeScaler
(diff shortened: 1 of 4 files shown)


jonoomph:
My primary concern with this approach is the cost of generating all waveforms for an entire project "on project load". For a very similar solution (I think), we could actually store the waveform array on the clip JSON, and then it would "just work" when the project was re-opened. The cost of re-opening an entire project would only be the cost of "drawing" the waveform, not decoding the all project assets again. Hopefully that makes sense.

JacksonRG:
@jonoomph Right, the trouble I was having is incorporating channel filters. My first attempt was creating a big dictionary of
```json
{
"sound_data": [
    {"file_id":
        {1: "<channel 1's data>"},
        {2: "<channel 2's data>"},
        //...
        {-1: "<all channel's data averaged>"}
    },
    {"file2_id": "..."}
]
}
```
Then when constructing the waveform for a given clip, only the relevant channel would be grabbed, or `-1` if there's no filter. Then, before getting the audio from libopenshot, I'd first check if it was in that object.

Does that make sense, or am I way over complicating the process.

jonoomph:
@JacksonRG There are some suspicious deletions in this PR. Please carefully review the changes here, and double check any deletions are on purpose, and perhaps add some comments to the PR explaining why we are deleting the code. Looks similar to a merge regression to me :grimacing:

JacksonRG:
This PR had some weird merge issues. see #4769

PR #4753, opened 2022-03-30
Fix rewind at project start and end.

JacksonRG:
### Issue
If the playhead is at the end of the project, and you press rewind, the play button wouldn't switch to a pause button. And sometimes, the rewind wouldn't start at all.

### Cause
We checking whether to toggle the play button without taking into account that the playhead is going to move backwards.

I made a change so that the auto-pause jumps the playhead back to the first or last frame of the project. I also made it take the direction into account when toggling the pause button.

### Test 1
Use `ctrl+left_arrow` to move to the end of your project and rewind.
Your player should rewind, but the play button won't toggle.

### Test 2
Set the playhead back from the end and let it play until it stops automatically. (Best way to make sure this happens is with fast forward) This should run past the last frame before Openshot catches it and pauses. Now if you press rewind nothing should happen. (Because even going back a frame is still outside the project)

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -916,7 +916,7 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
         current_frame = self.preview_thread.current_frame
         if current_frame is not None:
             next_frame = current_frame + requested_speed
-            return next_frame < max_frame
+            return next_frame <= max_frame and next_frame > 0
         return False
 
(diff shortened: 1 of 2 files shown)


jonoomph (approved):
LGTM. I did notice that it glitches a bit (i.e. clicking play on an empty project - jumping to frame 2 and then back to frame 1). But overall, performs better than our current code. :+1:

PR #4755, opened 2022-03-30
Display playback speed

JacksonRG:
# Enhancement
Show some visual feedback that fast forward or rewind has been applied, even if the frames take a moment to load.

# Example
https://user-images.githubusercontent.com/42394129/160917948-29a08ce5-87f3-4d7f-8e90-1f8d9714e8af.mp4

> Note: Unlike this example, the time now only shows if its not 1x, 0x, or -1x

# Testing
This one's pretty easy. Just fast forward, rewind, pause, and play several times and see that the displayed playback speed changes accordingly.

### DIFF
--- a/src/windows/video_widget.py
+++ b/src/windows/video_widget.py
@@ -275,6 +275,19 @@ class VideoWidget(QWidget, updates.UpdateInterface):
         rect = self.centeredViewport(self.width(), self.height())
         scale = self.devicePixelRatioF()
 
+        # Display the playback speed in widget title
+        speed_str = ""
+        try:
+            win = get_app().window
+            mode = win.preview_thread.player.Mode()
+            if mode == openshot.PLAYBACK_PAUSED:
+                speed = 0.0
+            else:
+                speed = win.preview_thread.player.Speed()
+            speed_str = f" {speed}x "
+        except NameError:
+            log.error("Couldn't get player speed for window title")
+
         # Find parent dockWidget (if any)
         dock = None
         if self.parent() and self.parent().parent():
@@ -286,12 +299,12 @@ class VideoWidget(QWidget, updates.UpdateInterface):
 
         if self.settings.get("preview-fps"):
             # Update window title with FPS output
-            dock.setWindowTitle(_("Video Preview") + " " + _("(Paint: %d FPS, Render: %d FPS, %dx%d)")
+            dock.setWindowTitle(_("Video Preview") + speed_str + _("(Paint: %d FPS, Render: %d FPS, %dx%d)")
                                                       % (self.paint_fps, self.present_fps,
                                                          rect.width() * scale, rect.height() * scale))
         else:
             # Restore window title
-            dock.setWindowTitle(_("Video Preview"))
+            dock.setWindowTitle(_("Video Preview") + speed_str)
 
     def paintEvent(self, event, *args):
         """ Custom paint event """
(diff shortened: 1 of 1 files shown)


src/windows/video_widget.py
```diff
@@ -286,12 +298,12 @@ def update_title(self):
 
         if self.settings.get("preview-fps"):
             # Update window title with FPS output
-            dock.setWindowTitle(_("Video Preview") + " " + _("(Paint: %d FPS, Render: %d FPS, %dx%d)")
+            dock.setWindowTitle(_("Video Preview") + speed_str + _("(Paint: %d FPS, Render: %d FPS, %dx%d)")
```
jonoomph: Let's move this into the parenthesis, for consistency (when the debug FPS is displayed)

src/windows/video_widget.py
```diff
@@ -286,12 +298,12 @@ def update_title(self):
...
                                                       % (self.paint_fps, self.present_fps,
                                                          rect.width() * scale, rect.height() * scale))
         else:
             # Restore window title
-            dock.setWindowTitle(_("Video Preview"))
+            dock.setWindowTitle(_("Video Preview") + speed_str)
```
jonoomph: I'm thinking we keep this in parenthesis, similar to the FPS above: `Video Preview (2x)`. Also, let's omit `0x` and `+-1x` speeds (since they are just noise). I mostly want to display the **speed** when it's not in a normal speed.

For example:
 - SPEED 0:  Video Preview
 - SPEED 1:  Video Preview
 - SPEED 2:  Video Preview (2x)
 - SPEED 4:  Video Preview (4x)
 - SPEED -1:  Video Preview
 - SPEED -2:  Video Preview (-2x)
 - SPEED -4:  Video Preview (-4x)

src/windows/video_widget.py
```diff
@@ -1465,6 +1474,13 @@ def __init__(self, watch_project=True, *args):
         # Get a reference to the window object
         self.win = get_app().window
 
+        # Update title whenever playback speed changes.
+        self.win.PlaySignal.connect(self.update_title, Qt.QueuedConnection)
+        self.win.PlaySignal.connect(self.update_title, Qt.QueuedConnection)
+        self.win.PauseSignal.connect(self.update_title, Qt.QueuedConnection)
+        self.win.SpeedSignal.connect(self.update_title, Qt.QueuedConnection)
+        self.win.StopSignal.connect(self.update_title, Qt.QueuedConnection)
+
```
ferdnyc, line 1483: I'd really like to avoid adding all these signal connections to the preview widget _just_ to add the speed to its title, since they're otherwise unnecessary.

As an alternative, my #4697 (which I've just updated to clean up some of the tangential stuff) moves the code that sets the preview title into the `MainWindow` class, which would make these connections unnecessary. (I would argue that code belongs there anyway, as the preview widget doesn't know anything about the dock it lives in. That's part of the child hierarchy for MainWindow, its owner.) 

Adding in speed display when the code lives in MainWindow is a snap, since there's already code there to keep track of the playback speed. (I don't care about the formatting of the stats display, I can change it to match the format here if that's preferred.)

jonoomph: I have no issues with these signal connections inside the widget. However, I'm also not opposed to #4697, which looks very nice as well. Sorry, a little late with my feedback on this one.

PR #4758, opened 2022-04-02
webview: Check id before showing track menu

ferdnyc:
Just hit a traceback with this; we already have these same protections on the `Clip` and `Effect` references, but it seems invalid (just-deleted) `Track`s can also be the target of new menu requests.


ferdnyc:
Merging as painfully uncontroversial.

PR #4757, opened 2022-03-31
WIP: File Location Settings

JacksonRG:
Closes #4590

This PR adds an option in the settings to set a default location to import files, export videos, and save projects. This could support users creating many projects, and reusing source files.

### Feature
In settings, there is a new "Locations" tab.

Locations Window

Here you can select the starting locations for saving/opening projects, importing resource files, and exporting videos.

Locations Window 2

### Testing
1) Create a new project, and save it to a folder.
    - Folder picker should still start in home directory
2) Import a file
    - File picker should start in home directory, or wherever your last file was imported from
3) Export the the project
    - Should default to the project folder.
4) Open preferences, and set a folder in you `Locations` tab.
    - For instance, I used `$HOME/Videos/Resources`, `$HOME/Videos/Projects`, and `$HOME/Videos/Exports`
5) Repeat steps 1-3. The file dialog should start on the folders you specified.

PR #4766, opened 2022-04-08
Folder locations mk2

### LINKED ISSUE #4590: Option to set Folder paths
**Describe the new feature:**
The ability to set default paths for the following:  New Project/Open Project, Save Project, Import Files, Export Folder Path.  I had submitted this request via email long time ago and see no reference to it at this time.

1.  In Edit | Preference there would be a tab named "Folder Paths".

2.   In this tab there would be the following options to override:
a.  New Project/Open Project path:  (default to c:\users\username)
b.  Save Project:                               (default to c:\users\username)
c.  Import Files:                                (default to c:\users\username)
- Music:                                           (default to c:\users\username)
- Sound:                                          (default to c:\users\username)
- Images:                                          (default to c:\users\username)
- Videos:                                          (default to c:\users\username)
d.  Import Project:                           (default to c:\users\username)
e.  Export Project:
- Video:                       (default to c:\users\username)
- EDL:                          (default to c:\users\username)
- XM (Final Cut Pro):    (default to c:\users\username)
g.  Export Folder path:  (defaults to c:\users\username)

The default paths obviously would be different for the different operating systems.

OpenShot should continue to maintain certain folders paths settings as it does currently.

Implement it in phases if needed but I don't see it couldn't be implemented all at once.

**Screenshots:** (Optional)
If applicable, add screenshots to help explain your request.

JacksonRG:
closes #4590
# Enhancement
Currently openshot's memory of folder locations is tied to projects. Some users may keep lots (or all) of their resource files in the same folder. If they do, this change will start them off in the folder they last imported from. If not, this can be switched off in `Settings` > `Locations`

# Example
1) Open a project
2) Start to import a file. (Ctrl+f)
3) Go into a sub folder and choose an image to import.
3) Create a new project
4) start to Import another file (Ctrl+f)

Notice that you are not in the same subfolder.

# Test
Repeat the steps from example, and on the last step, You should be in the subfolder navigated to in step 3. (Unless you've changed `Settings` > `Locations`)

### DIFF
--- a/src/classes/settings.py
+++ b/src/classes/settings.py
@@ -43,9 +43,8 @@ class SettingStore(JsonDataStore):
 
     @unique
     class pathType(Enum):
-        # Using literals to match the setting values
-        RECENT=0 # Return the last used path for the same action
-        PROJECT=1 # Return the current project path
+        RECENT=auto() # Return the last used path for the same action
+        PROJECT=auto() # Return the current project path
 
     @unique
     class actionType(Enum):
--- a/src/classes/settings.py
+++ b/src/classes/settings.py
@@ -184,10 +184,8 @@ class SettingStore(JsonDataStore):
                 log.error("getting setting %s" % setting["path"])
 
         if os.path.isfile(default_path):
-            # Default set to a file, not a dir. Fixing it.
-            log.info("Default path is a file. Correcting.")
+            # Make sure default_path is a directory
             default_path = os.path.dirname(default_path)
-            self.setDefaultPath(action,default_path)
 
         if (not default_path) or (not os.path.exists(default_path)):
             log.debug("Default path invalid. Falling back to home directory")
@@ -195,7 +193,7 @@ class SettingStore(JsonDataStore):
 
         if action == self.actions.SAVE:
             log.info("Adding file name for project path")
-            # Only get project name if one is currently active.
+            # Default name unless a project is currently active.
             project_name = os.path.basename(self.app.project.current_filepath or "") or "%s.osp" % _("Untitled Project")
             default_path = os.path.join(default_path, project_name)
         return default_path
--- a/src/classes/settings.py
+++ b/src/classes/settings.py
@@ -29,8 +29,10 @@
 
 # SettingStore - class which allows getting/storing of settings, loading and saving to json
 import os
+from enum import Enum, unique, auto
 
 from classes import info
+from classes.app import get_app
 from classes.logger import log
 from classes.json_data import JsonDataStore
 
@@ -38,6 +40,18 @@ from classes.json_data import JsonDataStore
 class SettingStore(JsonDataStore):
     """ This class only allows setting pre-existing keys taken from default settings file, and merges user settings
     on load, assumes default OS dir."""
+
+    @unique
+    class pathType(Enum):
+        RECENT=0 # Return the last used path for the same action
+        PROJECT=1 # Return the current project path
+
+    @unique
+    class actions(Enum):
+        IMPORT=auto()
+        EXPORT=auto()
+        LOAD=auto()
+        SAVE=auto()
 
     def __init__(self, parent=None):
         super().__init__()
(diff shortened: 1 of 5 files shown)


JacksonRG:
[Import/Export Folders in Settings](https://trello.com/c/ABZbyHRd/48-import-export-folders-in-settings)

Colorjet3:
What was the reason PR #4757 closed?  I thought having at minimum those three file paths in the Locations tab is great.  Here is my thoughts on this:
1.  PR #4757 should be implemented.
2.  The user has the option to populate any of the fields.
3.  If "File Imports:" = /home/jackson/Videos/Resources, then regardless of where the last import was made, OpenShot always defaults to this path for imports.  However, if this field is left empty, then OpenShot resorts to the above logic and defaults to the last location that an import was made.
4. This logic should apply to "Video Exports" and "Save Projects" (as I suggested this should be "Open/Save Projects" for clarity).

This gives the user the option/freedom to influence the behavior of "File Imports", "Video Exports", and "Open/Save Projects".

JacksonRG:
@Colorjet3 In talking it over with @jonoomph we realized that if the user does have a specific folders they always import from and export to, this logic will work for them without them having to find and change the setting. The only difference would be navigating back if they import from a different folder.

I wouldn't mind adding the folder settings back in. But for now, changing the default behavior benefits all users, not just the ones who find out about setting the folders.

src/classes/settings.py
```diff
@@ -39,6 +41,18 @@ class SettingStore(JsonDataStore):
     """ This class only allows setting pre-existing keys taken from default settings file, and merges user settings
     on load, assumes default OS dir."""
 
+    @unique
+    class pathType(Enum):
+        RECENT=0 # Return the last used path for the same action
```
jonoomph: Is there a reason we are starting with 1 here, but using auto() in other places?

src/classes/settings.py
```diff
@@ -39,6 +41,18 @@ class SettingStore(JsonDataStore):
...
+        PROJECT=1 # Return the current project path
+
+    @unique
+    class actions(Enum):
```
jonoomph: action => actionType: what do you think about renaming this to clearly be a 'type'?

src/classes/settings.py
```diff
@@ -111,3 +125,75 @@ def save(self):
...
+        if action in [self.actions.SAVE, self.actions.LOAD]:
+            return {'type': "locationProjectType", 'path': "locationProjectPath"}
+        log.info("Action not recognized")
+        log.error("Given action is not valid %s" % action)
+        return None
+
+    def setDefaultPath(self, action: pathType, recent_path: str):
+        """ Change the path setting corresponding to the given action """
+        if not os.path.exists(recent_path):
+            return
```
jonoomph, line 146: Should we **log** a warning here if the path is invalid?

src/classes/settings.py
```diff
@@ -111,3 +125,75 @@ def save(self):
...
+        if os.path.isfile(recent_path):
+            recent_path=os.path.dirname(recent_path)
+        try:
+            setting = self.pathSettings(action)
+            log.debug(f"Setting %s to %s" % (action.value, recent_path))
+            self.set(setting["path"], recent_path)
+        except:
+            log.error(f"Error setting the path of %s" % setting["path"])
```
jonoomph: Will `setting` even be defined at this point in the `except` clause?  Perhaps the log.error could be a safer message, that will definitely work?

src/classes/settings.py
```diff
@@ -111,3 +125,75 @@ def save(self):
...
+                log.debug("Using Recent Path")
+                default_path = self.get(setting["path"])
+            except:
+                log.error("getting setting %s" % setting["path"])
+
+        if os.path.isfile(default_path):
+            # Make sure default_path is a directory
+            default_path = os.path.dirname(default_path)
+
+        if (not default_path) or (not os.path.exists(default_path)):
```
jonoomph: I don't think you need the () and ()'s

src/classes/settings.py
```diff
@@ -111,3 +125,75 @@ def save(self):
...
+            log.debug("Default path invalid. Falling back to home directory")
+            return os.path.expanduser("~")
+
+        if action == self.actions.SAVE:
+            log.info("Adding file name for project path")
+            # Default name unless a project is currently active.
+            project_name = os.path.basename(self.app.project.current_filepath or "") or "%s.osp" % _("Untitled Project")
```
jonoomph: This looks sus:
`project_name = os.path.basename(self.app.project.current_filepath or "") or "%s.osp" % _("Untitled Project")`

What does `basename("")` return? Are you trying to prevent `basename(None)`?

jonoomph: Also, I'm not sure I understand adding the `project_name` to this function. It seems like the caller would add on the project_name or default name to the path returned.

src/windows/main_window.py
```diff
@@ -565,22 +564,26 @@ def actionOpen_trigger(self):
...
     def actionSave_trigger(self):
         app = get_app()
+        s = app.get_settings()
         _ = app._tr
 
         # Get current filepath if any, otherwise ask user
         file_path = app.project.current_filepath
         if not file_path:
-            recommended_path = os.path.join(info.HOME_PATH, "%s.osp" % _("Untitled Project"))
+            recommended_path = s.getDefaultPath(settings.actions.SAVE)
```
jonoomph: As mentioned above, I would add on the project_name or "Untitled" name here, instead of inside the function.

src/windows/main_window.py
```diff
@@ -565,22 +564,26 @@ def actionOpen_trigger(self):
...
             file_path = QFileDialog.getSaveFileName(
                 self,
                 _("Save Project..."),
                 recommended_path,
                 _("OpenShot Project (*.osp)"))[0]
+            s.setDefaultPath(s.actions.SAVE, file_path)
```
jonoomph, line 586: Same: add project name here

jonoomph:
This mostly looks good. Just a little bit of clean-up, and some nitpicks. The biggest change is to never return the project name in the `getDefaultPath()` function... let the caller do that. That function is all about the `directory`, and nothing else.

src/classes/settings.py
JacksonRG: My reason was just to ensure that the value of RECENT and PROJECT were the same as the value of their respective settings.

src/classes/settings.py
JacksonRG: Makes sense. I'll make sure the error msg is safe

src/classes/settings.py
JacksonRG: @jonoomph what if I simplified it like this: `if not (default_path and os.path.exists(default_path))`  ?

I still think it's good to be clear what the`not` is affecting

src/classes/settings.py
jonoomph: But we could change the settings to use 1, and 2 as values, right?

src/settings/_default.settings
```diff
@@ -967,5 +967,83 @@
...
+  },
+  {
+    "category": "Location",
+    "title": "File Import",
+    "setting": "locationImportType",
+    "value": 0,
+    "type": "dropdown",
+    "values": [
+      {
+        "value": 0,
```
jonoomph, line 979: For example, this could be 1 instead of 0 (so we can use `auto()` in the enum)

src/windows/main_window.py
```diff
@@ -649,18 +658,21 @@ def auto_save_project(self):
...
     def actionSaveAs_trigger(self):
         app = get_app()
+        s = app.get_settings()
         _ = app._tr
 
-        recommended_path = app.project.current_filepath
-        if not recommended_path:
-            recommended_path = os.path.join(
-                info.HOME_PATH, "%s.osp" % _("Untitled Project"))
+        recommended_folder = s.getDefaultPath(s.actionType.SAVE)
```
jonoomph, line 669: This behavior is different now, right? We were using the current_filepath previously, and only using "Untitled Project" when there is no current_filepath. I'm not sure about this though... just curious.

src/windows/main_window.py
```diff
@@ -691,6 +702,8 @@ def actionImportFiles_trigger(self):
             start_location,
             )[0]
 
+        if len(qurl_list):
```
jonoomph, line 710: what is this code for? `qurl_list`? Is there a reason we aren't just using our recommended_path from above?

src/classes/settings.py
JacksonRG: Yeah, No reason the setting values can't be 1 and 2

src/windows/main_window.py
JacksonRG: Fixed this. Now it'll check for a project name

src/windows/main_window.py
JacksonRG: I figured the most logical `new default` path is the folder of the *last* file selected.

pseudocode:
```
if any files were selected
    setDefault <- <basename of LAST file selected>
```
I'll add a comment to explain that.

jonoomph:
LGTM!

PR #4769, opened 2022-04-11
Waveform Improvements

JacksonRG:
# Issues
1) Clips that had `show-waveform` enabled didn't keep that setting when a project was closed and opened again.
2) If volume was changed after the waveform was generated, the change wouldn't display in the waveform.
3) (Performance) Even for clips from same file, OpenShot would get retrieve the file's audio data again for each waveform.

# Solution
1) Set `show_waveform` on clips, and check that setting when loading a project
2) Call the waveform function when changes to volume are detected
3) Store audio data in project, and check for stored data before retrieving it.

# Test
1) Open a project, and add clips with sound.
2) Enable waveform
3) Add some volume keyframes. (For instance, volume: 1.0 at the beginning and 0.0 at the end.)
    - Observe that the shape changes.
4) Save and close the project.
5) Reopen and see that the waveforms reappear automatically.

### DIFF
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -387,6 +387,8 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
         from classes.app import get_app
         get_app().updates.load(self._data)
 
+        self.show_waveforms()
+
     def rescale_keyframes(self, scale_factor):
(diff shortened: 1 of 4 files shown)


JacksonRG:

PR #4784, opened 2022-04-29
WIP: Show waveform project load mk3

JacksonRG:


src/classes/query.py
```diff
@@ -170,6 +172,49 @@ def title(self):
         path = self.data.get("reader", {}).get("path")
         return os.path.basename(path)
 
+    def showAudioData(self):
+        
+        if "ui" not in self.data:
+            self.data["ui"] = {}
+            self.save()
```
jonoomph, line 179: Let's make sure we don't have 2 undos for a single "show waveform" action

src/classes/query.py
```diff
@@ -170,6 +172,49 @@ def title(self):
...
+        # Get File's audio data
+        file_audio_data = file.data.get("ui",{}).get("audio_data", False)
+        if not file_audio_data:
+            log.info("clip.showAudioData was called, but file has no audio data")
+            return
+
+        sample_count = len(file_audio_data)
+        file_duration = file.data.get("duration")
+        time_per_sample = file_duration / sample_count
+        def sample_from_time(time):
```
jonoomph, line 193: Needs a doc string - explaining it's purpose

src/classes/query.py
jonoomph: Maybe get rid of this `.save()`

src/classes/query.py
```diff
@@ -257,6 +302,48 @@ def relative_path(self):
...
+            self.data["ui"] = {}
+            self.save()
+
+        audio_data = self.data["ui"].get("audio_data", False)
+        if audio_data and len(audio_data) > 1:
+            log.info("Audio Data already retrieved.")
+            return
+        if not audio_data:
+            # Placeholder value. Communicates that data is being retrieved
+            self.data["ui"]["audio_data"] = [-999]
```
JacksonRG, line 317: Save here, not after creating `ui`

src/classes/query.py
```diff
@@ -257,6 +302,48 @@ def relative_path(self):
         # Convert path to relative (based on current working directory of Python)
         return os.path.relpath(file_path, info.CWD)
 
+    def getAudioData(self):
+        # Ensure that UI attribute exists
+        if "ui" not in self.data:
+            self.data["ui"] = {}
+            self.save()
```
jonoomph, line 309: Maybe delete this `.save()`?

src/classes/query.py
jonoomph, line 317: Let's handle the [-999] only value - and protect against from accidentally generating the audio_data again

jonoomph: Possibly return `True` and `False` based on if things are success, or need to be tried again

src/classes/query.py
```diff
@@ -257,6 +302,48 @@ def relative_path(self):
...
+
+        audio_data = self.data["ui"].get("audio_data", False)
```
jonoomph, line 311: `audio_data = self.data["ui"].get("audio_data", [-999])` ??? to go ahead and init?

src/classes/query.py
jonoomph:
```
        if "ui" not in self.data:
            self.data["ui"] = { "audio_data": [] }
```

src/classes/query.py
```diff
@@ -257,6 +302,48 @@ def relative_path(self):
...
+        # Do the audio data stuff.
+        temp_clip = openshot.Clip(self.data["path"])
+        temp_clip.Open()
+        temp_clip.Reader().info.has_video = False
+
+        sample_rate = temp_clip.Reader().info.sample_rate
+        samples_per_second = 20
+        sample_divisor = round(sample_rate / samples_per_second)
+
+        audio_data = []
```
jonoomph, line 328: Why reinitialize `audio_data` variable?

src/classes/query.py
```diff
@@ -257,6 +302,48 @@ def relative_path(self):
...
+        for frame_num in range(1, temp_clip.Reader().info.video_length):
+            print(f"FileShow: frame {frame_num}")
```
jonoomph, line 330: Remove print - or make it a `log.debug`

src/classes/waveform.py
```diff
@@ -25,73 +25,63 @@
...
+    """
+    For the given file, update audio data.
+
+    arg1: file id to get the audio data of.
+    arg2: list of clips to update when the audio data is ready.
+    """
+
+    file = File.get(id=file_id)
+
+    file_audio_data = file.data.get("ui", {}).get("audio_data", False)
```
jonoomph, line 67: `file_audio_data = file.data.get("ui", {}).get("audio_data", [])`

src/classes/waveform.py
```diff
@@ -25,73 +25,63 @@
...
-
-    # Close reader
-    clip.Close()
-
-    # Emit signal when done
-    log.info("get_waveform_thread completed")
-    get_app().window.WaveformReady.emit(clip_id, audio_data)
+            # Loop until audio data is ready.
+            sleep(1)
+            if file.data.get("ui").get("audio_data", False) != [-999]:
```
jonoomph, line 75: `if file.data.get("ui").get("audio_data", []) != [-999]`

src/classes/waveform.py
```diff
@@ -25,73 +25,63 @@
...
+                break
+            count += 1
+            if count >= 15:
```
jonoomph, line 78: Let's use 300 seconds instead... to give large files time to generate waveforms

src/classes/waveform.py
```diff
@@ -25,73 +25,63 @@
...
+                # After 15 seconds. Abandon
```
jonoomph, line 79: Don't use the literal in the comment

src/windows/views/webview.py
```diff
@@ -1004,7 +995,7 @@ def Waveform_Ready(self, clip_id, audio_data):
         self.waveform_cache[clip_id] = serialized_audio_data
 
         # Pass to javascript timeline (and render)
-        self.run_js(JS_SCOPE_SELECTOR + ".setAudioData('" + clip_id + "', " + serialized_audio_data + ");")
+        self.run_js("drawAudio('" + clip_id + "');")
```
jonoomph, line 998: Removing direct calls to `drawAudio`, since the Angular scope/watch will handle when to draw audio

src/timeline/js/functions.js
```diff
@@ -58,7 +58,7 @@ function drawAudio(scope, clip_id) {
   //get the clip in the scope
   var clip = findElement(scope.project.clips, "id", clip_id);
 
-  if (clip.show_audio) {
+  if (clip.ui && clip.ui.audio_data) {
```
jonoomph, line 61: I think this function `drawAudio` should be converted to a Angular `watch`, in the `clip` directive. 

Watch:
- scale changes
- clip UI : audio_data changes

src/classes/query.py
JacksonRG: If I init `audio_data` to [-999], I'll have to empty the list again to store data. I think [-999] should just be saved to the file.

src/classes/query.py
JacksonRG: Removed. I don't think logging every frame is useful information.

ferdnyc:
One thing I've been thinking about for a _while_ now — if we converted the audio peak values into an SVG plot, instead of manually painting them onto a canvas, the result would be resolution-independent and could rescale automatically. The WebView would just handle the resizing and redrawing, without spending a _lot_ of time in `drawAudio()` every time the scale changes.

JacksonRG:
@ferdnyc 

I do like the idea of that... Also, I've been having a ton of trouble getting the webview to notice when audio values has been added to it.

PR #4795, opened 2022-05-10
WIP: Patch for folder feature

JacksonRG:
1) Update all path settings to use correct Enums.
- Cause paths to update in accordance with the settings.
2) Don't reject paths if the file doesn't exist
- Paths of new projects weren't being saved, because the full path (filename included) didn't exist yet.
3) Don't change save/load path when opening a project.
- Updating "Recent Path" for saving and loading projects on project load makes this option work the same as the "Project Folder" option. Only update on save.

### DIFF
--- a/src/classes/settings.py
+++ b/src/classes/settings.py
@@ -141,11 +141,11 @@ class SettingStore(JsonDataStore):
         """
         Change the path setting corresponding to the given action
         """
+        if os.path.isfile(recent_path):
+            recent_path=os.path.dirname(recent_path)
         if not os.path.exists(recent_path):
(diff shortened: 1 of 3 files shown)


jonoomph:
LGTM! Does this also include ignoring canceled File Import dialogs?  For example, click "Import File", go back a few directories, and hit ESC. Then open up "Import File" again and see that it changed.

JacksonRG:
I'm checking out the previous commit to recreate that. I did test that pattern with these changes and didn't see that issue, but I didn't see anything that should have caused that. So I'll recreate to make sure.

PR #4797, opened 2022-05-11
Fix blender integration on AppImage (for newer distros)

jonoomph:
On Ubuntu 22.04 (and probably other newish distros), the system version of `libtiff.so.5` contains new symbols... compared to the version we include in our AppImage. Thus, when launching Blender from our AppImage, we get a missing symbols error: `blender: symbol lookup error: /lib/libgdal.so.30: undefined symbol: TIFFGetStrileByteCount, version LIBTIFF_4.0`

This can be fixed by simply including the `/lib/x86_64-linux-gnu/` path when launching Blender, to force it to use the system libraries instead of our AppImage libraries. This seems to fix the issue nicely, both on older systems and newer systems.

Also, I have simplified the **regex** to support text before and after the Blender version, which I have found during testing.

### DIFF
--- a/src/windows/views/blender_listview.py
+++ b/src/windows/views/blender_listview.py
@@ -735,6 +735,13 @@ class Worker(QObject):
         self.version = None
         self.process = None
         self.canceled = False
+        self.env = dict(os.environ)
+        if sys.platform == "linux" and os.path.exists('/lib/x86_64-linux-gnu/'):
+            # If on Linux, verify we have the following LD library path defined.
+            # This is needed for Blender to use the system libraries instead
+            # of our AppImage libraries (i.e. our libtiff.so.5 is missing symbols,
+            # compared to libtiff.so.5 on newer distros for some reason).
+            self.env['LD_LIBRARY_PATH'] = '/lib/x86_64-linux-gnu/'
 
         self.startupinfo = None
         if sys.platform == 'win32':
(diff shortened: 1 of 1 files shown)


jonoomph:
[Blender fails to launch on Ubuntu 22.04 using AppImage (daily build)](https://trello.com/c/F1fk2vLc/104-blender-fails-to-launch-on-ubuntu-2204-using-appimage-daily-build)

src/windows/views/blender_listview.py
```diff
@@ -735,6 +735,13 @@ def __init__(self, blend_file_path, target_script, preview_frame=0):
         self.version = None
         self.process = None
         self.canceled = False
+        self.env = dict(os.environ)
+        if sys.platform == "linux" and os.path.exists('/lib/x86_64-linux-gnu/'):
+            # If on Linux, verify we have the following LD library path defined.
+            # This is needed for Blender to use the system libraries instead
+            # of our AppImage libraries (i.e. our libtiff.so.5 is missing symbols,
+            # compared to libtiff.so.5 on newer distros for some reason).
+            self.env['LD_LIBRARY_PATH'] = '/lib/x86_64-linux-gnu/'
```
JacksonRG, line 744: Is there any significance to this `x86_64` part? I'm wondering if that's correct on AMD systems.

JacksonRG (approved):
LGTM. Fix makes sense.

My only question is if this problem is present outside of x86 systems. If it is nothing changes for those users, so merge away.

src/windows/views/blender_listview.py
jonoomph: Great question. We might need to handle things differently for different CPUs, just depending on how distros handle this.

PR #4798, opened 2022-05-12
Set starting directory for QT file dialog

JacksonRG:
Default folder for file imports wasn't being updated. Always starting in the same folder.

Fixed by using the `setDirectory` function rather than the `getOpenFileUrls`'s `directory` argument which some file dialogs interpret as a search term rather than Folder Path.


JacksonRG:
[Import/Export Folders in Settings](https://trello.com/c/ABZbyHRd/48-import-export-folders-in-settings)

JacksonRG:
@jonoomph Figured out why the setting wasn't changing on *initial save* of a new project.

I cutting off the filename only IF `os.path.isfile()` Which returns true for *existing* file names. Anyway, now I remove the file name if `not os.path.isdir()`

src/classes/project_data.py
```diff
@@ -947,7 +942,8 @@ def check_if_paths_are_valid(self):
                 if path and is_modified and not is_skipped:
                     # Found file, update path
                     file["path"] = path
-                    get_app().updates.update_untracked(["import_path"], os.path.dirname(path))
+                    settings
```
jonoomph: What is this?

src/classes/project_data.py
```diff
@@ -947,7 +942,8 @@ def check_if_paths_are_valid(self):
...
+                    settings.setDefaultPath(settings.actionType.LOAD, os.path.dirname(path))
```
jonoomph: Should this use `IMPORT` instead of `LOAD` - just based on the previous behavior?

src/classes/settings.py
```diff
@@ -141,10 +141,12 @@ def setDefaultPath(self, action: pathType, recent_path: str):
         """
         Change the path setting corresponding to the given action
         """
-        if os.path.isfile(recent_path):
-            recent_path=os.path.dirname(recent_path)
+        if not os.path.isdir(recent_path):
+            # log.info("File path. reducing to directory")
```
jonoomph: Fix commented out debug lines - seems useful to keep this logged (file path => folder path)

src/windows/export.py
```diff
@@ -790,7 +790,8 @@ def titlestring(sec, fps, mess):
                 video_settings["vcodec"] = image_ext
 
         # Store updated export folder path in project file
-        get_app().updates.update_untracked(["export_path"], os.path.dirname(export_file_path))
+        settings = get_app().get_settings()
+        settings.setDefaultPath(settings.actionType.EXPORT, os.path.dirname(export_file_path))
```
jonoomph: Doesn't the setDefaultPath method already handle file_paths?  i.e. by reducing them to the folder path?

src/windows/main_window.py
```diff
@@ -429,6 +429,9 @@ def open_project(self, file_path, clear_thumbnails=True):
         """ Open a project from a file path, and refresh the screen """
 
         app = get_app()
+        settings = app.get_settings()
+        # Set default load-project path to this path.
+        settings.setDefaultPath(settings.actionType.LOAD, os.path.dirname(file_path))
```
jonoomph: Ditto on file_path being reduced to folder name here - already handled in setDefaultPath()

src/windows/models/files_model.py
```diff
@@ -350,10 +351,10 @@ def add_files(self, files, image_seq_details=None, quiet=False, prevent_image_se
 
                 # Let the event loop run to update the status bar
                 get_app().processEvents()
+                prev_path = settings.getDefaultPath(settings.actionType.IMPORT)
 
-                prev_path = app.project.get("import_path")
                 if dir_path != prev_path:
```
jonoomph: I'm not sure we need this logic anymore: `if dir_path != prev_path`... since we are only updating a setting... and no longer updating the project.

jonoomph:
Some comments and a few suggestions. Otherwise looks good.

src/classes/project_data.py
JacksonRG: :grimacing: an unfinished change. It was going to be `settings.setDefaultPath...` fixing

src/windows/export.py
JacksonRG: It does. I can take out the `dirname` for simplicity

jonoomph:
LGTM

PR #4819, opened 2022-06-02
Improved Waveforms (highly optimized)

jonoomph:
Building off some experimental work started by @JacksonRG (https://github.com/OpenShot/openshot-qt/pull/4769). Essentially, we are optimizing the waveform generation in OpenShot, and storing the summarized **audio_data** used to draw the waveforms in the *.osp project files.

**Use Cases Improved:**
- Selecting more than 1 clip from the same file, and choosing "Display->Waveform" - instead of duplicating a ton of work, we only generate each file's waveform once, and then share the data to each child clip.
- Re-opening a project with waveforms will now maintain the waveform (since we are storing the audio data)
- Once a file has been processed (i.e. audio_data stored), it is super fast to display waveforms in any clip from that file.

**My Changes:**
Fixed many issues related to waveform generation and reuse. Added an Angular directive to watch for **audio_data** changes. Simplified the waveform process also. When no waveforms are present, *.osp file size does not change. When huge waveforms are displayed on many clips, the size of *.osp project will grow quite a bit. In fact, we might want to consider moving audio_data into some separate cache files, vs baking the audio_data into our *.osp file. Or at a minimum, allow the user to clear all audio data from an *.osp project (to reduce the size).

### DIFF
--- a/src/classes/waveform.py
+++ b/src/classes/waveform.py
@@ -37,8 +37,7 @@ s = get_app().get_settings()
 
 
 def get_audio_data(files: dict):
-    """Get a Clip object form libopenshot, and grab audio data"""
-    """
+    """Get a Clip object form libopenshot, and grab audio data
         For for the given files and clips, start threads to gather audio data.
 
         arg1: a dict of clip_ids grouped by their file_id
--- a/src/classes/waveform.py
+++ b/src/classes/waveform.py
@@ -25,73 +25,63 @@
  along with OpenShot Library.  If not, see <http://www.gnu.org/licenses/>.
  """
 
-import platform
 import threading
-from copy import deepcopy
-from classes import info
 from classes.app import get_app
 from classes.logger import log
-import openshot
+from classes.query import File, Clip
+from time import sleep
 
 
 # Get settings
 s = get_app().get_settings()
 
 
-def get_audio_data(clip_id, file_path, channel_filter, volume_keyframe):
+def get_audio_data(files: dict):
     """Get a Clip object form libopenshot, and grab audio data"""
-    clip = openshot.Clip(file_path)
-    clip.Open()
+    """
+        For for the given files and clips, start threads to gather audio data.
 
-    # Disable video stream (for speed improvement)
-    clip.Reader().info.has_video = False
+        arg1: a dict of clip_ids grouped by their file_id
+    """
 
-    log.info("Clip loaded, start thread")
-    t = threading.Thread(target=get_waveform_thread, args=[clip, clip_id, file_path, channel_filter, volume_keyframe])
-    t.daemon = True
-    t.start()
+    for file_id in files:
+        clip_list = files[file_id]
 
-def get_waveform_thread(clip, clip_id, file_path, channel_filter=-1, volume_keyframe=None):
-    """Get the audio data from a clip in a separate thread"""
-    audio_data = []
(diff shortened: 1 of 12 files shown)


jonoomph:

src/classes/waveform.py
```diff
@@ -25,73 +25,164 @@
...
-    # Emit signal when done
-    log.info("get_waveform_thread completed")
-    get_app().window.WaveformReady.emit(clip_id, audio_data)
+            # Loop until audio data is ready.
+            sleep(1)
+            if file.data.get("ui", {}).get("audio_data", []) != [-999]:
+                break
+            count += 1
+            if count >= 15:
+                # After 15 seconds. Abandon
```
JacksonRG: I know you had a nitpick (which I addressed) about avoiding literal numbers in comments. But that must've gotten undone in some version changing.

src/classes/waveform.py
```diff
@@ -25,73 +25,164 @@
...
+    file = File.get(id=file_id)
+
+    # Loop through each selected clip (which uses this file)
+    for clip_id in clip_list:
+        clip = Clip.get(id=clip_id)
+        clip_reader = clip.data.get("reader")
+
+        # If another thread is already getting audio data for this clip, bail out.
+        clip_audio_data = clip.data.get("ui", {}).get("audio_data", [])
+        if clip_audio_data and clip_audio_data == [-999]:
```
JacksonRG: I know I came up with this, but I think it has to go.

If we ctrl-z after a `show waveform` I think this will set us back to `-999`. But this time there's no thread working on it. So we can't re-show the waveform.

Would love to say I'm just *that* good at reading code. But I found this with a quick test.

For now I think ctrl-z is a more likely use case than pressing `show waveform` twice in under a second. Also without this we have the chance of repeating work. But with it we have the chance of a clip which *can't* show waveform.

JacksonRG (changes requested):
A couple changes. But they're changes you've asked me to make on earlier PRs.

So here's a patch file with the fixes I'm asking for. (Plus it updates the waveform if you edit volume in the properties window)

[waveforms.zip](https://github.com/OpenShot/openshot-qt/files/8830809/waveforms.zip)

jonoomph:
@JacksonRG LGTM - But before we merge these changes, I still have a few ideas / thoughts.  Based on the primary concern of *.osp files become very large, due primarily to waveform data. 

**Solution 1:** We add a menu option: "Edit -> Clear Waveforms", which clears all file **audio_data** and clip **audio_data** in the project JSON. Since the *.osp file size is only affected when generating large amounts of waveform data, giving users an easy way to remove this data seems important.

**Solution 2:** We store the **audio_data** outside of the *.osp (for example, in a separate `audio_data.json` file). Although this sounds great on the surface, it means that the data would not be easy to "bind" using Angular though... and would change this PR a bunch.

JacksonRG:
@jonoomph At this point, this PR's goals have been accomplished. There's still a little weird behavior.

1. Dragging sliders in the properties window when `Show Waveform` is activated is a little weird. The values change, but the sliders don't show it.
2. `Show Waveform` changes get added to Action History (though I have some stashed changes that wrap every change to `audio_data` in an `ignore_history`.

I understand if you want to get these changes merged, get feedback, and address those issues in their own tickets.

JacksonRG (approved):

PR #4821, opened 2022-06-06
End blender render process when cancel button is pressed.

JacksonRG:
# Bug:
When rendering an animated title, clicking cancel doesn't end the render process.

# Test:
1) If you have blender running, close it.
1) Open your task manager, and search blender.
    - You should see only one task
1) Open the Animated Titles Dialog
1) Click render on a title. (I recommend `Fly Towards` title, as it renders pretty quickly)
1) In your task manager, see a process appear.
1) Click cancel on the Animated Title Dialog.
1) The blender process should close soon (a few seconds) after you click cancel.

### DIFF
--- a/src/windows/views/blender_listview.py
+++ b/src/windows/views/blender_listview.py
@@ -585,7 +585,7 @@ Blender Path: {}
     def Cancel(self):
         """Cancel the current render, if any"""
         #QMetaObject.invokeMethod(self.worker, 'Cancel', Qt.DirectConnection)
-        self.cancel_render.emit()
+        self.worker.Cancel()
 
     def Render(self, frame=None):
         """ Render an images sequence of the current template using Blender 2.62+ and the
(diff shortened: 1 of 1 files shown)


JacksonRG:
[Blender can not be canceled (continues running in background)](https://trello.com/c/8Tnr7DMk/119-blender-can-not-be-canceled-continues-running-in-background)

jonoomph:
LGTM, but I'll be honest: I don't really follow the logic (based on just this PR). Let's just make sure we've really tested this fully:
1. That rendering a title still works (without any canceling)
2. Canceling a title kills the blender process (and exits the dialog), etc...

JacksonRG:
Still testing (testing slowed down because the code that was working on day started crashing the next day.)

And also I noticed the thread doesn't exactly get *killed* (It just hangs around using 0 resources.) even though I think I'm using the same signals as when the render finishes.

I'll update you when I can test more, and get the process to really end.

jonoomph:
LGTM

PR #4787, opened 2022-05-02
Fixing QtWebEngine Black/not-rendered Timeline

jonoomph:
Disable sandbox support for QtWebEngine (required on some Linux distros for the QtWebEngineWidgets to be rendered, otherwise no timeline is visible). This was initially reported on OpenSUSE and the latest Ubuntu 22.04.
https://doc.qt.io/qt-5/qtwebengine-platform-notes.html#sandboxing-support

Closes [#4676](https://github.com/OpenShot/openshot-qt/issues/4676)

PR #4827, opened 2022-06-20
Add dissable sandbox to the environment when openshot launches.

JacksonRG:
PR #4787 fixed this issue in our AppImages, but certain other setups were still encountering the issue of the timeline never showing up.

This PR moves this definition to `app.py`, so that it can apply to all ways of running openshot.

Tested the appimage, and it still shows timeline.

### DIFF
--- a/installer/launch-linux.sh
+++ b/installer/launch-linux.sh
@@ -32,11 +32,6 @@ export LD_LIBRARY_PATH="${HERE}"
 # Set some environment variables
 export QT_PLUGIN_PATH="${HERE}"
 
-# Disable sandbox support for QtWebEngine (required on some Linux distros
-# for the QtWebEngineWidgets to be rendered, otherwise no timeline is visible).
-# https://doc.qt.io/qt-5/qtwebengine-platform-notes.html#sandboxing-support
-export QTWEBENGINE_DISABLE_SANDBOX="1"
-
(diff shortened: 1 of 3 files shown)


JacksonRG:
[Dissable Sandbox when building from source](https://trello.com/c/co7s5EcL/148-dissable-sandbox-when-building-from-source)

jonoomph:
A few comments/questions:
* Does this need to be set before the `imports` of QtWebEngine*?
* Can we add some additional documentation to this (and some spacing), to give future devs an understanding of why this line of code exists? (example comment below)
* Should we remove our fix from https://github.com/OpenShot/openshot-qt/pull/4787/files, since including the ENV VAR in our `launch.py` script should work for all our builds - of course, this needs to be tested and verified though. But seems like 1 place in the code would be better than 2 places.

```
# Disable sandbox support for QtWebEngine (required on some Linux distros
# for the QtWebEngineWidgets to be rendered, otherwise no timeline is visible).
# https://doc.qt.io/qt-5/qtwebengine-platform-notes.html#sandboxing-support
```

jonoomph:
I do think `launch.py` is an appropriate place for this code. But also, we might try `app.py`, unless that is too-late in the lifecycle for this to be useful. `app.py` might be even better than `launch.py`.

jonoomph:
LGTM

PR #4833, opened 2022-06-22
(Py)Qt6 readiness: Use typed enums

ferdnyc:
@jonoomph @JacksonRG et al..

I've been plugging away, the past few weeks, at trying to come up with a PyQt6-compatible OpenShot. It's **very** slow going. (Libopenshot was _easy_, literally **one line** in the GUI code for the `openshot-player` tool had to be changed. That's it! Other than that, it's already fully Qt6-compatible. OpenShot? Not so much.)

The changes necessary to make it work are _huge_, and because Riverbank chose to make a clean break and drop **everything** that's deprecated from PyQt5 (even the stuff that Qt, itself, has kept in the Qt5 Compatibility Layer module), it may not be (realistically) possible to simultaneously support PyQt5 and PyQt6 in the same codebase. Branching and maintaining parallel versions, at least for a period, may be the only realistic option, unfortunately.

Because let's face it, _eventually_ OpenShot will need to make the move to PyQt6. So, there are plenty of things that can be done **now** to get it at least "more ready", since a lot of mandatory-in-PyQt6 features are already available today in PyQt5, and have been for some time. We just haven't adopted them yet.

I'm going to make an effort to push as many of those adaptations as possible back into the PyQt5 code, so that the PyQ6-compatibility transition is at least **less** painful and world-breaking. This PR is the first of those changes.

I'm also going to make a plea for a bit of fast-tracking on this one, because while it's **very** focused, it's also _very_ sweeping, and edit conflicts will crop up quickly if it's left open and unmerged while other changes get committed ahead of it. Obviously some conflicts are likely to come up, and I'll fix those as they come, but there is a certain degree of time-sensitivity to this _PR_, if no these changes. (Which are, in one sense, admittedly "totally unnecessary", but I hope I've adequately sold the reasons why they're still a good idea.)

### Adoption of typed enums.

In PyQt6, in order to provide better type safety, enum values are no longer all lumped together under their owner classes, or  thrown in a big pile inside `PyQt6.QtCore.Qt`.

Now, every enum has its own, distinct type, which means that enum
values can be type checked. It's no longer possible to do something
crazy like call `QImage.scale(newSize, Qt.SolidPattern)` and have
that work, because `Qt.SolidPattern` has the value `1` which is
interpreted by `QImage::scale()` as the mode `Qt.KeepAspectRatio`.
Mypy won't catch it either, because in PyQt5
`Qt.SolidPattern == Qt.KeepAspectRatio` returns `True`.
After all, they both have the integer value `1`.

### Safety through precise typing

In PyQt6 the enums have all been separated from each other.
They're partitioned into type silos, each type holding only
a set of members that represent all the possible values for
that specific enum.So even though `KeepAspectRatio` has the value `1` in the `Qt.AspectRatioMode` enum, and `SolidPattern` is `1` in the `Qt.BrushStyle` enum, in PyQt6 they **won't** be equal or interchangeable with each other, because they'll have their own, separate type classes.

* Aspect ratio modes have the type `Qt.AspectRatioMode`.

* Brush styles have the distinct type `Qt.BrushStyle`.
 
...Which means they have to be _accessed_ that way. **Exclusively** that way, in PyQt6.

So, come PyQt6:
- `Qt.KeepAspectRatio` becomes `Qt.AspectRatioMode.KeepAspectRatio`
- `Qt.SolidPattern` becomes `Qt.BrushStyle.SolidPattern`
- `QImage.Format_ARGB32` becomes `QImage.ImageFormat.Format_ARGB32`

And so on, for Every. Single. Enum!

- Like `Qt.green`? Instead cozy up to `Qt.GlobalColor.green`.

- Friends with `Qt.Key_Escape`? Embrace `Qt.Key.Key_Escape`.

- Want `Qt.AA_ShareOpenGLContexts`? Hope you're ready for
  `Qt.ApplicationAttribute.AA_ShareOpenGLContexts`.

It does make for much stronger typing, which when combined with
mypy and code that's type hinted means stronger code, period.

This comparison:
`Qt.AspectRatioMode.KeepAspectRatio == Qt.BrushStyle.SolidPattern`
will now properly return `False`, yay!

The per-enum classes are already present in PyQt5, they're just
not type-siloed into distinct, non-equivalent classes. (That
previous comparison still returns `True`, in PyQt5.) But they're
there, so we can choose to start using them now, since we'll
HAVE to eventually.

But it's definitely a big adjustment. (At least by making the
switch before it's forced, we have a safety net in the event
that we've missed any old-style enum accesses. Instead of a
traceback, the code will still function as expected. Only in
PyQt6 do the old-style "short" enums get taken away.)

### Secondary changes

The PR also carries a couple of minor additional changes that got swept up in the enum typing. (Meaning, they were changes that also appeared on lines where enums changed, so they got committed together.)

1. It replaces all of our remaining `QRegExp` usage with `QRegularExpression`. QRegExp has been deprecated long time now, and it's removed from PyQt6. QRegularExpression has been the official replacement since Qt 5.12.

1. It replaces _some_ uses of outdated, aliased, trailing-underscore methods with their unaliased counterparts. (Mostly `QDialog`/`QMessageBox` `.exec_()` calls, which haven't required the trailing underscore for, again, a long time. Since Python 3.0, in fact. Python 2 was picky about builtin words as class methods, but Python 3 isn't, so it's fine to just call `QDialog.exec()` and those aliases have  been deprecated for a while.  They're also removed in PyQt6.)

    I'll submit a separate PR at some point that fixes the rest of them, but a few got swept up in this change.

    Note, BTW, that _**certain**_ methods are still aliased, if they're Python KEYWORDS. Like, `QWidget::raise()` is still `PyQt{5,6}.QtWidgets.QWidget.raise_()`, because calling a method `raise` is still a terrible idea in any version of Python.

### DIFF
--- a/src/windows/region.py
+++ b/src/windows/region.py
@@ -113,7 +113,8 @@ class SelectRegion(QDialog):
 
         # Add Video Widget
         self.videoPreview = VideoWidget()
-        self.videoPreview.setSizePolicy(QSizePolicy.Preferred, QSizePolicy.Expanding)
+        self.videoPreview.setSizePolicy(
+            QSizePolicy.Policy.Preferred, QSizePolicy.Policy.Expanding)
         self.verticalLayout.insertWidget(0, self.videoPreview)
 
         # Set aspect ratio to match source content
@@ -122,7 +123,8 @@ class SelectRegion(QDialog):
         self.videoPreview.aspect_ratio = aspect_ratio
 
         # Set max size of video preview (for speed)
-        self.viewport_rect = self.videoPreview.centeredViewport(self.width, self.height)
+        self.viewport_rect = self.videoPreview.centeredViewport(
+            self.width, self.height)
 
         # Create an instance of a libopenshot Timeline object
         self.r = openshot.Timeline(self.viewport_rect.width(), self.viewport_rect.height(), openshot.Fraction(self.fps_num, self.fps_den), self.sample_rate, self.channels, self.channel_layout)
@@ -168,10 +170,10 @@ class SelectRegion(QDialog):
         QTimer.singleShot(600, functools.partial(self.sliderVideo.setValue, 1))
 
         # Add buttons
-        self.cancel_button = QPushButton(_('Cancel'))
-        self.process_button = QPushButton(_('Select Region'))
-        self.buttonBox.addButton(self.process_button, QDialogButtonBox.AcceptRole)
-        self.buttonBox.addButton(self.cancel_button, QDialogButtonBox.RejectRole)
+        self.cancel_button = self.buttonBox.addButton(
+            QDialogButtonBox.StandardButtons.Cancel)
+        self.process_button = self.buttonBox.addButton(
+            _('Select Region'), QDialogButtonBox.ButtonRole.AcceptRole)
 
         # Connect signals
         self.actionPlay.triggered.connect(self.actionPlay_Triggered)
--- a/src/windows/views/add_to_timeline_treeview.py
+++ b/src/windows/views/add_to_timeline_treeview.py
@@ -77,8 +77,8 @@ class TimelineTreeView(QTreeView):
         self.setModel(self.timeline_model.model)
         self.setIconSize(info.TREE_ICON_SIZE)
         self.setIndentation(0)
-        self.setSelectionBehavior(QTreeView.SelectRows)
-        self.setSelectionBehavior(QAbstractItemView.SelectRows)
+        self.setSelectionBehavior(QAbstractItemView.SelectionBehavior.SelectRows)
+        self.setSelectionBehavior(QAbstractItemView.SelectionBehavior.SelectRows)
         self.setWordWrap(True)
         self.setStyleSheet('QTreeView::item { padding-top: 2px; }')
 
--- a/src/windows/views/blender_listview.py
+++ b/src/windows/views/blender_listview.py
@@ -270,7 +270,7 @@ class BlenderListView(QListView):
 
         # Show 'Wait' cursor
         if cursor:
-            QApplication.setOverrideCursor(Qt.WaitCursor)
+            QApplication.setOverrideCursor(Qt.CursorShape.WaitCursor)
 
     @pyqtSlot()
     def end_processing(self):
@@ -588,7 +588,7 @@ Blender Path: {}
 
     def Cancel(self):
         """Cancel the current render, if any"""
-        #QMetaObject.invokeMethod(self.worker, 'Cancel', Qt.DirectConnection)
+        #QMetaObject.invokeMethod(self.worker, 'Cancel', Qt.ConnectionType.DirectConnection)
         self.cancel_render.emit()
 
     def Render(self, frame=None):
@@ -637,7 +637,7 @@ Blender Path: {}
 
         # Cleanup signals all 'round
         self.worker.finished.connect(self.worker.deleteLater)
-        self.worker.finished.connect(self.background.quit, Qt.DirectConnection)
+        self.worker.finished.connect(self.background.quit, Qt.ConnectionType.DirectConnection)
         self.background.finished.connect(self.background.deleteLater)
         self.background.finished.connect(self.worker.deleteLater)
 
@@ -685,11 +685,11 @@ Blender Path: {}
         self.setModel(self.blender_model.model)
         self.setIconSize(info.LIST_ICON_SIZE)
         self.setGridSize(info.LIST_GRID_SIZE)
-        self.setViewMode(QListView.IconMode)
-        self.setResizeMode(QListView.Adjust)
+        self.setViewMode(QListView.ViewMode.IconMode)
+        self.setResizeMode(QListView.ResizeMode.Adjust)
         self.setUniformItemSizes(True)
         self.setWordWrap(True)
-        self.setTextElideMode(Qt.ElideRight)
+        self.setTextElideMode(Qt.TextElideMode.ElideRight)
 
         # Hook up controls
         self.win.btnRefresh.clicked.connect(self.preview_timer.start)
--- a/src/windows/views/changelog_treeview.py
+++ b/src/windows/views/changelog_treeview.py
@@ -50,8 +50,8 @@ class ChangelogTreeView(QTreeView):
         self.changelog_model.update_model(filter=filter)
 
         # Format columns
-        self.header().setSectionResizeMode(0, QHeaderView.Fixed)
-        self.header().setSectionResizeMode(1, QHeaderView.Fixed)
+        self.header().setSectionResizeMode(0, QHeaderView.ResizeMode.Fixed)
+        self.header().setSectionResizeMode(1, QHeaderView.ResizeMode.Fixed)
         self.setColumnWidth(0, 70)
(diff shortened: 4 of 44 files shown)


jonoomph:
[PR: Support for PyQt6 (initial steps)](https://trello.com/c/qhKzSLUJ/149-pr-support-for-pyqt6-initial-steps)

src/windows/region.py
```diff
@@ -168,10 +170,10 @@ def __init__(self, file=None, clip=None):
         QTimer.singleShot(600, functools.partial(self.sliderVideo.setValue, 1))
 
         # Add buttons
-        self.cancel_button = QPushButton(_('Cancel'))
-        self.process_button = QPushButton(_('Select Region'))
-        self.buttonBox.addButton(self.process_button, QDialogButtonBox.AcceptRole)
-        self.buttonBox.addButton(self.cancel_button, QDialogButtonBox.RejectRole)
+        self.cancel_button = self.buttonBox.addButton(
+            QDialogButtonBox.StandardButtons.Cancel)
```
jonoomph, line 174: Will `StanardButtons.Cancel` be translated? Our previous code had a manually translated _('Cancel')

src/windows/views/add_to_timeline_treeview.py
```diff
@@ -77,8 +77,8 @@ def __init__(self, *args):
         self.setModel(self.timeline_model.model)
         self.setIconSize(info.TREE_ICON_SIZE)
         self.setIndentation(0)
-        self.setSelectionBehavior(QTreeView.SelectRows)
-        self.setSelectionBehavior(QAbstractItemView.SelectRows)
+        self.setSelectionBehavior(QAbstractItemView.SelectionBehavior.SelectRows)
```
jonoomph, line 80: This line looks like a duplicate

src/windows/views/blender_listview.py
```diff
@@ -588,7 +588,7 @@ def update_image(self, image_path):
 
     def Cancel(self):
         """Cancel the current render, if any"""
-        #QMetaObject.invokeMethod(self.worker, 'Cancel', Qt.DirectConnection)
+        #QMetaObject.invokeMethod(self.worker, 'Cancel', Qt.ConnectionType.DirectConnection)
```
jonoomph, line 591: Commented line can be removed

src/windows/views/changelog_treeview.py
```diff
@@ -101,10 +101,10 @@ def __init__(self, commits, commit_url, *args):
         # Setup header columns
         self.setModel(self.changelog_model.model)
         self.setIndentation(0)
-        self.setSelectionBehavior(QTreeView.SelectRows)
-        self.setSelectionBehavior(QAbstractItemView.SelectRows)
-        self.setSelectionMode(QAbstractItemView.ExtendedSelection)
-        self.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Expanding)
+        self.setSelectionBehavior(QAbstractItemView.SelectionBehavior.SelectRows)
```
jonoomph, line 104: Duplicate line of code

src/windows/views/credits_treeview.py
```diff
@@ -110,10 +110,10 @@ def __init__(self, credits, columns, *args):
         # Setup header columns
         self.setModel(self.credits_model.model)
         self.setIndentation(0)
-        self.setSelectionBehavior(QTreeView.SelectRows)
-        self.setSelectionBehavior(QAbstractItemView.SelectRows)
-        self.setSelectionMode(QAbstractItemView.ExtendedSelection)
-        self.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Expanding)
+        self.setSelectionBehavior(QAbstractItemView.SelectionBehavior.SelectRows)
```
jonoomph, line 113: Duplicate line of code

src/windows/views/tutorial.py
```diff
@@ -430,18 +430,21 @@ def __init__(self, win):
...
         self.dock.setFloating(True)
 
         # Connect to interface dock widgets
-        self.win.dockFiles.visibilityChanged.connect(functools.partial(self.process, "dockFiles"))
-        self.win.dockTransitions.visibilityChanged.connect(functools.partial(self.process, "dockTransitions"))
-        self.win.dockEffects.visibilityChanged.connect(functools.partial(self.process, "dockEffects"))
-        self.win.dockProperties.visibilityChanged.connect(functools.partial(self.process, "dockProperties"))
-        self.win.dockVideo.visibilityChanged.connect(functools.partial(self.process, "dockVideo"))
-        self.win.dockEmojis.visibilityChanged.connect(functools.partial(self.process, "dockEmojis"))
+        for d in [
```
jonoomph, line 439: This is actually more lines of code than he original, with the humble brag of not repeating any code, lol

jonoomph: I ask you... which version is easier to glance at and understand?

jonoomph:
A few **Codacy issues**:

- `src/windows/process_effect.py`: `Undefined variable 'Qt'`
- `src/windows/process_effect.py`: `Undefined variable 'QSizePolicy'`
- `src/windows/process_effect.py`: `Undefined variable 'QDialogButtonBox'`
- `src/windows/process_effect.py`: `Undefined variable 'QDialogButtonBox'`
- `src/windows/region.py`: `Undefined variable 'QSizePolicy'`
- `src/windows/region.py`: `Undefined variable 'QDialogButtonBox'`
- `src/windows/title_editor.py`: `Undefined variable 'QFontInfo'`

jonoomph:
@ferdnyc Thanks for taking a shot at this one! It's a big one, which seems fairly high-risk (as with any PR that changes 40+ files I suppose, lol). I'll give this a try locally (and hopefully @JacksonRG can do the same local testing). Let us run through a full set of regression tests, to verify we don't see a bunch of stack-traces. :+1: We'll keep you posted!

jonoomph:
@ferdnyc Moving this conversation to https://github.com/OpenShot/openshot-qt/pull/4836 (closing this one)

src/windows/views/add_to_timeline_treeview.py
ferdnyc: Good call, my regular-expression-based updates even turned them into identical duplicates, one can clearly go.

src/windows/views/tutorial.py
ferdnyc: Fair question, but I'll throw one back at you: Which version will crash if `self.win` doesn't have a `dockProperties` attribute, or has one that's not a `QDockWidget` / lacks a `visibilityChanged` signal, and which version won't? The intent isn't merely to avoid cut-and-paste, it's to avoid assumptions about the structure of external classes.

jonoomph:
Re-opening this PR, and waiting for changes from branch: https://github.com/OpenShot/openshot-qt/tree/qt6-prep-enums to be applied to this one.

pva:
@ferdnyc so what's the status of this pull request? I found only two MR both closed and it looks like job was stalled. Still, some distributions (e.g. Gentoo) already try to avoid qt5 (https://bugs.gentoo.org/925718) for a good reason, so this job is appreciated!

ferdnyc:
@pva There is no status. I am no longer involved in the project.

PR #4834, opened 2022-06-23
Add button to remove audio data from a project.

JacksonRG:
Example:

https://user-images.githubusercontent.com/42394129/176049611-8e54a41b-4174-48ee-a6cb-937253caf08b.mp4

Clear audio data from all files and clips in the project

Our redesign of how waveforms work in #4819 can make project files pretty big. Here's a button to remove all data for displaying waveforms from a project.

### DIFF
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -322,6 +322,7 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
     def load(self, file_path, clear_thumbnails=True):
         """ Load project from file """
 
+        from classes.app import get_app
         self.new()
 
         if file_path:
@@ -337,6 +338,13 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
                 # Fix history (if broken)
                 if not project_data.get("history"):
                     project_data["history"] = {"undo": [], "redo": []}
+
+                # If project has waveforms, enable removing waveforms
+                get_app().window.actionClearWaveformData.setEnabled(False)
+                for file in project_data["files"]:
+                    if file.get("ui",{}).get("audio_data", []):
+                        get_app().window.actionClearWaveformData.setEnabled(True)
+                        break
 
             except Exception:
                 try:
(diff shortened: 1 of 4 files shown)


Colorjet3:
I somewhat understand what this is but I do have some questions for further clarification:
1.  When you say "Clear audio data from all **files** and **clips** in the project" how does OpenShot differentiate a "file" from a "clip".  I've seen these two terms used interchangeably by many users.
2. Is the files size related to the size of the .osp file only or does this also impact the rendered/exported created file?  I ask, because the size of the exported file often comes up as an issue in Reddit(too large)?
3.  In relation to PR #4819, what happens if I clear the audio data in a project and then save the project?  Does this mean I will no longer hear audio while previewing in OpenShot for a specific project?
4.   If I Clear the data and then Export the project, will the Exported file still play the audio?
5.   Can I "Undo" this while I have not yet saved the project(in case I made a mistake)?
I will probably have some follow up questions based on the answers to these questions and more when I see the step-by-step test plan.

JacksonRG:
@Colorjet3 Good questions! The "audio data" I'm talking about is only the data openshot gathers to display waveforms on the timeline. No impact on preview or export.

### Files/Clips:
Let's say I open a project and import 1 video file. Then I drag it to the timeline in 3 different places.
I now have 3 clips, but they all come from the same file. I can set the first to half volume and not affect the others.

So if select `Show Waveform` on two of those clips, openshot will read the audio samples from the file once. Then the two clips will both get a copy of the audio data, but specially calculated with their own volume values.

jonoomph:
To prevent confusion... I suggestion the following wording:

`Clear->History (undo / redo data)` (only enabled if we have history in the project)
`Clear->Waveform (cached audio data)` (only enabled if we have cached audio_data in the project)

Alternatively, we can remove the parentheses:

`Clear->History`
`Clear->Waveform`

jonoomph:
But the key change is... let's not repeat the word "Clear" multiple times... and only allow a user to clear each option, if there is data of that type in the project. So for a new project, both would be disabled.

jonoomph:
LGTM

PR #4835, opened 2022-06-23
Animated titles (Blender) should use current project's FPS

jonoomph:
When using Animated Titles in OpenShot, we need to make some adjustments for the current project's FPS, and be sure to generate the correct # of frames, for silky smooth animations that match our current FPS. For example, even though all our animated title templates are created at an "assumed" 25 FPS (for length calculations), if we are editing a 50 FPS project, and generate an Animated Title in Blender, we need 2X the frames.

So if our 'Fly Toward Camera' title uses 80 FPS (by default, at the assumed 25 FPS), when we click on this title using a 50 FPS project, it should default to 160 FPS (for the same "length", but 2X the frames).

### DIFF
--- a/src/windows/views/blender_listview.py
+++ b/src/windows/views/blender_listview.py
@@ -219,7 +219,7 @@ class BlenderListView(QListView):
         if param["name"] == "length_multiplier":
             # Convert value to float (and multiply with project FPS diff)
             # This converts all length_multipliers to the correct project FPS reference.
-            # For example, a 1X multiplier would be 1.25X for a 30 FPS project
+            # For example, a 1X multiplier would be 1.2X for a 30 FPS project
             # using a 25 FPS animated title - to scale up to the correct # of frames.
             self.params[param["name"]] = float(value) * self.project_fps_diff
             self.init_slider_values()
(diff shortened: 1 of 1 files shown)


jonoomph:
[Blender Animated Titles do not use project's FPS (they are static 25 fps)](https://trello.com/c/Q9j3VcZt/150-blender-animated-titles-do-not-use-projects-fps-they-are-static-25-fps)

src/windows/views/blender_listview.py
```diff
@@ -217,6 +217,11 @@ def dropdown_index_changed(self, widget, param, index):
         self.params[param["name"]] = value
         log.info('Animation param %s set to %s' % (param["name"], value))
         if param["name"] == "length_multiplier":
+            # Convert value to float (and multiply with project FPS diff)
+            # This converts all length_multipliers to the correct project FPS reference.
+            # For example, a 1X multiplier would be 1.25X for a 30 FPS project
+            # using a 25 FPS animated title - to scale up to the correct # of frames.
```
JacksonRG, line 223: Nitpick: In this example of `25` -> `30` fps the multiplier would be `1.2`. It _would_ be 1.25 if its `24` -> `30` fps

src/windows/views/blender_listview.py
```diff
@@ -672,6 +677,13 @@ def __init__(self, parent, *args):
         self.preview_timer.setSingleShot(True)
         self.preview_timer.timeout.connect(self.preview_timer_onTimeout)
 
+        # Calculate diff between project FPS and title FPS
+        # All animated titles are created at an assumed default 25.0 FPS
+        self.project_fps_diff = 1.0
```
JacksonRG: Unless we put a conditional (`project_fps == 25.0`) in here, this variable will always be overwritten.

JacksonRG (approved):
Logic all looks good. Just noticed a math error in a comment.

Approving anyway, because it would still work as is.

PR #4848, opened 2022-07-07
When all clips are exported set progress to 100%

JacksonRG:
# Issue
In some cases when exporting clips, the progress bar would stay at 99%. Even if the export was complete, and the "Done" button had appeared.

## Steps to reproduce
1) import a file
1) Right click > Split Clip.
1) Create 3 clips
1) Highlight all three > right click > export clips
1) Hit Export

# Solution
When done looping through each clip, force the progress bar to 100%.

### DIFF
--- a/src/windows/export_clips.py
+++ b/src/windows/export_clips.py
@@ -263,9 +263,10 @@ class clipExportWindow(QDialog):
             # If within 2 frames of complete, show 100 percent.
             self.progressExportVideo.setValue(100)
             return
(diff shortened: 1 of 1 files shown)


JacksonRG:
[Export Clips: 3 Clips stops at 99% (instead of 100%)](https://trello.com/c/tu4nUSVl/114-export-clips-3-clips-stops-at-99-instead-of-100)

jonoomph:
LGTM!

PR #4857, opened 2022-07-13
Use Signal to save audio data to clips

JacksonRG:
Update to the previous `improved-waveforms` PR.

Now using a signal so that the main thread edits the clip data. This prevents a crash on windows


JacksonRG:

PR #4898, opened 2022-08-28
Trimming + Snapping Improvements for Clips and Transitions

jonoomph:
There are many snapping issues addressed in this PR. 

- Trimming + snapping (Clips and Transitions) has been fixed and now has improved precision (exactly matching the edges suggested with snapping)
- Next Marker + Previous Marker logic has been improved to correctly detect the right edge of clips/transitions. It was previously adding a +1 frame which is a bug
- Auto set the direction of transitions, based on which side of a clip it intersects. On the left side we now fade-in, and on the right side, we fade out (by default) - Custom mask/transitions brightness and contrast are preserved (at least the basic range of values is preserved)
- Fixed a bounding box computation rounding error, caused by JQuery. We now compute width based on `time` data from our scope.

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -1360,10 +1360,12 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
 
             fps = get_app().project.get("fps")
             fps_float = float(fps["num"]) / float(fps["den"])
+            frame_duration = float(fps["den"]) / float(fps["num"])
 
             clip_start_time = obj.data["position"]
             clip_orig_time = clip_start_time - obj.data["start"]
-            clip_stop_time = clip_orig_time + obj.data["end"]
+            # Last frame on clip is -1 frame's duration
+            clip_stop_time = clip_orig_time + obj.data["end"] - frame_duration
 
             # add clip boundaries
             positions.append(clip_start_time)
@@ -1413,11 +1415,14 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
 
         # We can always jump to the beginning of the timeline
         all_marker_positions = [0]
+        fps = get_app().project.get("fps")
+        frame_duration = float(fps["den"]) / float(fps["num"])
 
         # If nothing is selected, also add the end of the last clip
         if not self.selected_clips + self.selected_transitions:
             all_marker_positions.append(
-                get_app().window.timeline_sync.timeline.GetMaxTime())
+                # last frame is -1 frame's duration
+                get_app().window.timeline_sync.timeline.GetMaxTime() - frame_duration)
 
         # Get list of marker and important positions (like selected clip bounds)
         for marker in Marker.filter():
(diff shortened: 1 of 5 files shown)


jonoomph:
[Clip/Transition Snapping is not precise enough (causing misalignments)](https://trello.com/c/heJ2rDOf/160-clip-transition-snapping-is-not-precise-enough-causing-misalignments)

src/windows/main_window.py
```diff
@@ -1360,10 +1360,12 @@ def getTimelineObjectPositions(obj):
 
             fps = get_app().project.get("fps")
             fps_float = float(fps["num"]) / float(fps["den"])
+            frame_duration = float(fps["den"]) / float(fps["num"])
 
             clip_start_time = obj.data["position"]
             clip_orig_time = clip_start_time - obj.data["start"]
-            clip_stop_time = clip_orig_time + obj.data["end"]
+            # Last frame on clip is -1 frame's duration
+            clip_stop_time = clip_orig_time + obj.data["end"] - frame_duration
```
JacksonRG, line 1368: Looks good! Walked through the logic with backwards and slowed down clips, and it makes sense.

src/windows/views/webview.py
```diff
@@ -329,40 +329,65 @@ def update_transition_data(self, transition_json, only_basic_props=True, ignore_
...
+            c.AddPoint(round(duration * fps_float), sorted(contrast_range)[0], openshot.BEZIER)
         else:
-            # Create new brightness and contrast Keyframes
-            b = openshot.Keyframe()
-            b.AddPoint(1, 1.0, openshot.BEZIER)
-            b.AddPoint(round(duration * fps_float) + 1, -1.0, openshot.BEZIER)
-            brightness = json.loads(b.Json())
+            c.AddPoint(1, sorted(contrast_range)[0], openshot.BEZIER)
+            c.AddPoint(round(duration * fps_float), sorted(contrast_range)[-1], openshot.BEZIER)
+        contrast = json.loads(c.Json())
```
JacksonRG, line 390: So rather than overwriting contrast and brightness keyframes, we follow along, and then continue the fade from the first keyframe to the first frame, or from the last keyframe to the last frame?

Makes sense as a way to do it, just making sure that's the intent here.

JacksonRG (approved):
Looks good, Just clarifying that I'm reading correctly

src/windows/views/webview.py
jonoomph: `contrast_range`/`brightness_range` contains the actual values used on the Y axis for brightness - for example, imagine a user set the following values [3.0, -3.0].  Those would be sorted first, and our new brightness keyframe would either be [3.0, -3.0] if it's on the left side of a clip, or [-3.0, 3.0] if it's on the right of a clip. The idea though, is to preserve the user's actual values if the keyframe already exists, and recreate a new keyframe (going in the correct wipe direction).

jonoomph: Most average users do not customize the `brightness` and `contrast` keyframes used by transitions. So, this is kind of a quick and dirty way to preserve the customized values (if any) when resizing transitions / moving transitions. This isn't perfect, because we are only preserving the min/max range of values used in the keyframe... not actually all the points in between (if there are more than 2).

JacksonRG: Ok.... So if I'm reading this right, then if there's only 1 brightness keyframe, then there will be no change in brightness.

Because `brightness_range[0]` and `brightness_range[-1]` will be the same value.

jonoomph: Exactly... it works with a range of 1 to n

PR #4903, opened 2022-09-02
Check if folder exists before removing whitespace from folder name

JacksonRG:
# Issue
Openshot currently removes whitespace form the end of all export folder names. But a user recently pointed out that it's possible (on some operating systems) to have a folder name ending in a space.

# Solution
Since this doesn't mean that all spaces at the end of folder names are correct, we should check if the folder name exists *as is* then try again with the space removed.

### DIFF
--- a/src/windows/export.py
+++ b/src/windows/export.py
@@ -722,7 +722,15 @@ class Export(QDialog):
             if not file_name_with_ext.endswith(file_ext):
                 file_name_with_ext = '{}.{}'.format(file_name_with_ext, file_ext)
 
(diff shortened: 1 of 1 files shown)


JacksonRG:
[Export fails if folder ends with whitespace](https://trello.com/c/K4r0cBXT/161-export-fails-if-folder-ends-with-whitespace)

jonoomph:
LGTM

jonoomph:
[Cache bar fragmentation & optimization](https://trello.com/c/BpoBxnSQ/84-cache-bar-fragmentation-optimization)

PR #4909, opened 2022-09-07
Don't leave a one frame gap between clips.

JacksonRG:
To avoid showing a frame twice, or overlapping clips when slicing, we were adding a frame to the position of the right clip when slicing and keeping both sides.

The solution was simplifying, and using the play-head position to set the clip positions wherever possible.


JacksonRG:
[Slicing clip (Keep Both Sides) results in black frame between clips](https://trello.com/c/iPPE0lyW/164-slicing-clip-keep-both-sides-results-in-black-frame-between-clips)

jonoomph:
LGTM!

PR #4919, opened 2022-09-16
Improved Caching Support (openshot-qt)

jonoomph:
This PR contains the UI components and improvements for better caching support in OpenShot, and is related to libopenshot PR: https://github.com/OpenShot/libopenshot/pull/847.

**TLDR:** 
Make playback smoother by improving caching during pause (and disabling caching during certain UI operations)

**UI Changes supported by this PR:**
 - Scrub timeline
    - disable caching thread until mouse-up
    - clear cache if seeking outside the cached range
 - Transform / Property Editor
    - disable caching thread until mouse-up
 - Fast Forward
    - cache every frame in the forward direction
 - Rewind
    - cache every frame in the reverse direction
    - KNOWN ISSUE: frames cached in reverse direction will contain "temporary" audio pops, use new context menu: **Playhead->Cache->Clear All** to fix
 - Playhead
    - Cache context menu -> Clear All
    - For use when cache contains audio pops or glitches - and the user wants to clear and rebuild cached frames
 - Load Project
    - pre-roll and cache frames when paused, right after opening a project
 - Pause
    - cache more aggressively - more frames cached on timeline (50% in front, 50% behind current playhead position)
    - NOTE: the cached frames to the left of the playhead position are cached during playback, and are not immediately available
 - Playback
    - cache less frames ahead - to conserve CPU for displaying frames rapidly

**Bug Fixes:**
 - Jump to Start / Jump to End toolbar buttons would not correctly update the frame # on the Property Editor - causing edits to affect the wrong frame #
 - Javascript Error when dragging the Playhead (or scrubbing the ruler) - regression from a previous accuracy improvement to `functions.js`
 - Fix a build-server error, caused by a soon-to-expire code signing cert
 - Fixed a bug causing empty cache objects to crash (or not render the cache progress on the timeline)

### DIFF
--- a/installer/build_server.py
+++ b/installer/build_server.py
@@ -576,7 +576,7 @@ def main():
             # Sign MSI
             for line in run_command(key_sign_command):
                 output(line)
-                if line:
+                if line and "will expire" not in line:
                     key_sign_success = False
                     key_sign_output = line
 
--- a/installer/build_server.py
+++ b/installer/build_server.py
@@ -576,7 +576,7 @@ def main():
             # Sign MSI
             for line in run_command(key_sign_command):
                 output(line)
-                if line and "will expire" not in line:
+                if line and "will expire" not in line.decode('UTF-8'):
                     key_sign_success = False
                     key_sign_output = line
 
--- a/src/timeline/js/functions.js
+++ b/src/timeline/js/functions.js
@@ -247,7 +247,12 @@ function setBoundingBox(scope, item, item_type="clip") {
 
     // Compute width from `time` duration (for more accuracy). Getting the width from
     // JQuery is not accurate, and is occasionally rounded.
-    let item_width = (item_object.end - item_object.start) * scope.pixelsPerSecond;
+    let item_width = 1;
+    if (item_object) {
+        item_width = (item_object.end - item_object.start) * scope.pixelsPerSecond;
+    } else {
+        item_width = item.width();
+    }
 
     item_bottom = item.position().top + item.height() + vert_scroll_offset;
     item_top = item.position().top + vert_scroll_offset;
--- a/src/windows/models/properties_model.py
+++ b/src/windows/models/properties_model.py
@@ -148,23 +148,26 @@ class PropertiesModel(updates.UpdateInterface):
 
             # Determine the frame needed for this clip (based on the position on the timeline)
             time_diff = (requested_time - clip.Position()) + clip.Start()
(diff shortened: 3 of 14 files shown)


jonoomph:
[Cache bar fragmentation & optimization](https://trello.com/c/BpoBxnSQ/84-cache-bar-fragmentation-optimization)

installer/build_server.py
```diff
@@ -576,7 +576,7 @@ def main():
             # Sign MSI
             for line in run_command(key_sign_command):
                 output(line)
-                if line:
+                if line and "will expire" not in line.decode('UTF-8'):
```
jonoomph, line 579: Temporarily fix build server error on Windows

src/timeline/js/functions.js
```diff
@@ -247,7 +247,12 @@ function setBoundingBox(scope, item, item_type="clip") {
 
     // Compute width from `time` duration (for more accuracy). Getting the width from
     // JQuery is not accurate, and is occasionally rounded.
```
jonoomph, line 249: item_object can also be NULL, and we need to fallback to the JQuery element width (pixels)

src/timeline/js/functions.js
```diff
@@ -247,7 +247,12 @@ function setBoundingBox(scope, item, item_type="clip") {
...
-    let item_width = (item_object.end - item_object.start) * scope.pixelsPerSecond;
+    let item_width = 1;
```
jonoomph, line 250: item_object can also be NULL, and we need to fallback to the JQuery element width (pixels)

src/windows/models/properties_model.py
```diff
@@ -568,6 +571,7 @@ def value_updated(self, item, interpolation=-1, value=None, interpolation_detail
             # Save changes
             if clip_updated:
                 # Save
+                c.data = clip_data
```
jonoomph, line 574: `clip_data` is updated heavily before this moment - but never added back to the Clip object before saving

src/windows/views/webview.py
```diff
@@ -3156,7 +3176,7 @@ def render_cache_json(self):
         try:
             if self.window.timeline_sync and self.window.timeline_sync.timeline:
                 cache_object = self.window.timeline_sync.timeline.GetCache()
-                if not cache_object or cache_object.Count() <= 0:
```
jonoomph, line 3159: Now that the crash caused by empty caches is fixed in libopenshot, we actually want to render empty caches (otherwise we can't show the user when the cache is emptied)

PR #4925, opened 2022-09-22
Refresh final frame after Scrubbing

jonoomph:
For some reason, after scrubbing a bunch of random frames on the Timeline, perhaps due to a race condition, the final frame in our video preview is not the final frame requested. This PR adds a "refresh frame" signal after scrubbing has completed, to ensure the final frame is correctly displayed. Seems to work fine.


jonoomph:
[`Frame Number` clip property displays the wrong frame after Slice](https://trello.com/c/4FKzfnZk/168-frame-number-clip-property-displays-the-wrong-frame-after-slice)

JacksonRG (approved):
LGTM

PR #4930, opened 2022-09-24
Fixing the detection of max timeline length (during playback)

jonoomph:
Fixing the detection of max timeline length. Any change to the project data (i.e. a clip moving for example) will re-calculate the max frame/max length of the timeline, so our playback stops at the correct moment. Also, improved the clean-up code on our video preview widget and preview thread, so we don't accumulate unneeded project data watchers.

### DIFF
--- a/src/classes/updates.py
+++ b/src/classes/updates.py
@@ -202,6 +202,12 @@ class UpdateManager:
         # Notify watchers of new history state
         self.update_watchers()
 
+    def disconnect_listener(self, listener):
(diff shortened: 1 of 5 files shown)


jonoomph:
["Timeline Length" not calculated correctly during playback + moving clip](https://trello.com/c/Nm2r3Ypi/154-timeline-length-not-calculated-correctly-during-playback-moving-clip)

JacksonRG (approved):
LGTM!

PR #4932, opened 2022-09-28
Updating the regex used for Mac notarizations

jonoomph:
Updating the regex used for Mac notarizations, to get the correct status, and prevent pre-mature notarizations. It appears to be capturing with the dotall setting, causing the expression to match the next line as well... 

The response from Apple is now:
```
b'       Date: 2022-09-28 01:33:05 +0000\n'
b'RequestUUID: ee50bebd-bb83-4fa9-8d9d-c0f70832b16e\n'
b'     Status: in progress\n'
b'Status Code: 0\n'
```

### DIFF
--- a/installer/build-mac-dmg.sh
+++ b/installer/build-mac-dmg.sh
@@ -88,12 +88,13 @@ while [ "$(( $(date +%s) - 3600 ))" -lt "$START" ]; do
     notarize_info=$(xcrun altool --notarization-info "$REQUEST_UUID" -u "jonathan@openshot.org" -p "@keychain:NOTARIZE_AUTH")
     echo "$notarize_info"
 
-    pat='Status: (.*)'
+    pat='Status: (.*)\n'
     [[ "$notarize_info" =~ $pat ]]
     notarize_status="${BASH_REMATCH[1]}"
+    echo "Notarization Status Found: $notarize_status"
 
     if [ "$notarize_status" != "in progress" ] && [ "$notarize_status" != "" ]; then
-      echo "Notarization Status Found: $notarize_status. Wait for notarization to appear in --notarization-history/"
+      echo "Wait for notarization to appear in --notarization-history/"
       verify_output=$(xcrun altool --notarization-history 0 -u "jonathan@openshot.org" -p "@keychain:NOTARIZE_AUTH" | grep "$REQUEST_UUID")
       if [ "$verify_output" != "" ]; then
         echo "Notarization record found, and ready for stapling!"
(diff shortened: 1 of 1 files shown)


jonoomph:
[Randomly failing Mac builds on GitLab](https://trello.com/c/vWFEUmBT/176-randomly-failing-mac-builds-on-gitlab)

jonoomph:
Merging this before review - needed to fix our Mac builds

installer/build-mac-dmg.sh
```diff
@@ -88,23 +88,28 @@ while [ "$(( $(date +%s) - 3600 ))" -lt "$START" ]; do
...
     fi
 
-    # Wait a few seconds (otherwise the stapler can sometimes fail to find the ticket)
-    sleep 180
+    # Wait a few seconds (so we don't spam the API)
+    sleep 60
 done
 
+# Wait a few more seconds (otherwise the stapler can sometimes fail to find the ticket)
+sleep 180
```
jonoomph, line 111: This is probably not needed now that the regex bug is fixed, but keeping for now...just in case

installer/build-mac-dmg.sh
```diff
@@ -88,23 +88,28 @@ while [ "$(( $(date +%s) - 3600 ))" -lt "$START" ]; do
     notarize_info=$(xcrun altool --notarization-info "$REQUEST_UUID" -u "jonathan@openshot.org" -p "@keychain:NOTARIZE_AUTH")
     echo "$notarize_info"
 
-    pat='Status: (.*)'
```
jonoomph, line 91: `.` matches all new lines after `Status: ` as well... and due to Apple changing the response syntax... this started breaking. For example, $notarize_status would equal:

```
in progress\n'
b'Status Code: 0\n
```

installer/build-mac-dmg.sh
```diff
@@ -88,23 +88,28 @@ while [ "$(( $(date +%s) - 3600 ))" -lt "$START" ]; do
...
+    # Match status (stop at newline)
+    pat='Status: ([^'$'\n'']*)'
     [[ "$notarize_info" =~ $pat ]]
     notarize_status="${BASH_REMATCH[1]}"
+    echo "Notarization Status Found: $notarize_status"
 
     if [ "$notarize_status" != "in progress" ] && [ "$notarize_status" != "" ]; then
```
JacksonRG, line 97: so `$notarize_status` was never `== "in progress"` because the string was "in progress `\n` and another line" Right?

JacksonRG (commented):
Makes sense (If I understand it correctly). Looks good.

PR #4934, opened 2022-09-29
Prevent invalid paths in the File Properties Dialog

JacksonRG:
Users could enter invalid paths when editing file properties. Also a blank path would be entered if a user canceled the file dialog. In addition, this PR includes fixes for replacing file paths in the File Dialog (for example, say you have an image file used in many places on the Timeline - and you want to replace it with a video file) - this requires updating the File JSON for potentially lots of clips, update thumbnails, and requires fully supporting image sequence detection. Basically, an identical flow to adding new files to OpenShot, with the added task of updating existing related items and thumbnails.

### DIFF
--- a/src/windows/file_properties.py
+++ b/src/windows/file_properties.py
@@ -66,6 +66,9 @@ class FileProperties(QDialog):
         # Get settings
         self.s = app.get_settings()
 
+        # Keep track of whether this file is an image sequence
+        self.is_sequence = False
+
         # Track metrics
         track_metric_screen("file-properties-screen")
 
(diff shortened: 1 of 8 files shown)


JacksonRG:
[File Properties allows the user to change the file path, but canceling breaks the file (empty path)](https://trello.com/c/5Uh7hOEE/126-file-properties-allows-the-user-to-change-the-file-path-but-canceling-breaks-the-file-empty-path)

src/windows/file_properties.py
```diff
@@ -171,15 +174,58 @@ def browsePath(self):
...
+
+
+        # don't update if dialog was canceled
+        if newFilePath:
+            # If this path could be an image sequence, get that info and prompt user.
+            seq_info = get_app().window.files_model.get_image_sequence_details(newFilePath)
+            get_app().window.files_model.ignore_image_sequence_paths = []
+
+            # create the proper path for an image sequence
+            if seq_info:
```
JacksonRG: So only allowing paths that pass `os.path.isfile()` would exclude image sequences.

So I check if the new path *could* be an image sequence (the code for that automatically prompts the user) and if they accept, the below code will format the path correctly

jonoomph: Correct, isfile() is too restrictive

jonoomph:
@JacksonRG Big refactor of `add_file()` function, to standardize some functionality around image sequence detection.

**File properties dialog:**
- Update `duration` and `video_length` when changing file
- Update `thumbnail` and file model after changing file
- Ignore invalid file paths (logged)
- Update `start frame` and `end frame` textboxes after changing files
- Refactored to support image seq detection on `browse` button AND on `update` button
- Fixed bug which would always add `start` and `end` properties to files after clicking `Update`

**File Model changes:**
- Refactored some image seq detection code
- Added logging for failed img seq parsing

jonoomph:
After testing this a bit more, found more issues with replacing the file path in the File Dialog. Essentially, this allows a user to replace any file in OpenShot, and thus requires us to completely replace the `file.data`, and then **propagate** that data to all **related clips**, and then update all **thumbnails** (among other cached data).

PR #4944, opened 2022-10-13
Memory Clean-up, Safer Shutdown, and Cache Improvements

jonoomph:
- Fixed caching direction on "Jump to Start" during a rewind operation
- Updating default cache settings (24 min pre-roll, 48 max pre-roll, 300 max frames, 512 MB memory cache)
- Protect against accessing deleted clips (property window, preview widget, new/open dialogs, etc...)
- Disabling video cache during shutdown

### DIFF
--- a/src/classes/info.py
+++ b/src/classes/info.py
@@ -29,7 +29,7 @@ import os
 from time import strftime
 
 VERSION = "2.6.1-dev"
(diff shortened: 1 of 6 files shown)


jonoomph:
[OpenShot Stuttering During Multi-Track Preview](https://trello.com/c/y1pcdWLF/182-openshot-stuttering-during-multi-track-preview)

PR #4945, opened 2022-10-13
Fixing the resolution of China Video Disc (CVD)

jonoomph:
Fixing the resolution of China Video Disc (CVD), which is described as a **horizontal** format - but most websites list the resolution backwards for some reason.

NTSC: `480x352`
PAL: `576x352`


jonoomph:
[CVD (China Video Disk) format has the wrong resolution](https://trello.com/c/GQKYiKnq/181-cvd-china-video-disk-format-has-the-wrong-resolution)

PR #4950, opened 2022-10-19
Freeze BABL extensions for all 3 OSes

jonoomph:
All 3 OSes (in official installers) are missing the BABL extensions, used by our ChromaKey effect to use different color algorithms - for color distance. We need to both include these missing libraries (DLL, SO, and DYLIB which are loaded at runtime), and set the `BABL_PATH` environment variable when launching OpenShot on Windows - so we load the correct folder.


jonoomph:
[Chromakey effect crashes Windows (when not using Basic keying)](https://trello.com/c/tvdkqDHy/122-chromakey-effect-crashes-windows-when-not-using-basic-keying)

uglycoder99:
Looks okay too me.

PR #4952, opened 2022-10-20
Fix `Remove from Project` Crash when a related clip is selected

jonoomph:
Remove selected clip from selections when removing a file from the project - preventing a crash. OpenShot actually deletes the clip and removes it from memory now, so any old references to the clip (for example, in our selections array) will cause a crash eventually.


jonoomph:
[OpenShot Crashes when "Remove from Project" deletes a "selected" Clip](https://trello.com/c/nwFJJfcu/189-openshot-crashes-when-remove-from-project-deletes-a-selected-clip)

PR #4957, opened 2022-10-23
Reduce verbosity for Sentry (on a few methods)

jonoomph:
Related to https://github.com/OpenShot/libopenshot/pull/861

This is a minor change, but reducing the verbosity of a few logging calls, and removing a copy.deepcopy() call around all query object data, which is unneeded and slows down openshot-qt.


jonoomph:
[Crash inside Timeline::ClearAllCache() method](https://trello.com/c/24HO8Uwd/191-crash-inside-timelineclearallcache-method)

PR #4960, opened 2022-10-23
Faster Serialization (Replace deepcopy with json)

jonoomph:
After profiling OpenShot with CProfile, the deepcopy calls to our internal project data structures is painfully inefficient. Replacing with `json.loads(json.dumps())`, which appears to be much lighter and requires far less CPU. Also, moving a slow call to `Timeline::ApplyJsonDiff` to it's own thread, to not block our main thread.

### DIFF
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -229,7 +229,7 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
 
 
         # After processing each key, we've found object and parent, return former value/s on update
-        ret = copy.deepcopy(obj)
+        ret = json.loads(json.dumps(obj))
 
         # Apply the correct action to the found item
         if remove:
@@ -400,7 +400,7 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
         # Create a scaler instance
         scaler = KeyframeScaler(factor=scale_factor)
         # Create copy of active project data and scale
-        scaled = scaler(copy.deepcopy(self._data))
+        scaled = scaler(json.loads(json.dumps(self._data)))
         return scaled
 
     def read_legacy_project_file(self, file_path):
@@ -482,7 +482,7 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
                             failed_files.append(item.name)
 
                 # Delete all tracks
-                track_list = copy.deepcopy(Track.filter())
+                track_list = json.loads(json.dumps(Track.filter()))
                 for track in track_list:
                     track.delete()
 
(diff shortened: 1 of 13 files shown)


jonoomph:
[Improve `DeepCopy` Performance](https://trello.com/c/u6x8DzZD/192-improve-deepcopy-performance)

src/classes/project_data.py
```diff
@@ -482,7 +482,7 @@ def read_legacy_project_file(self, file_path):
                             failed_files.append(item.name)
 
                 # Delete all tracks
-                track_list = copy.deepcopy(Track.filter())
+                track_list = json.loads(json.dumps(Track.filter()))
```
uglycoder99: Could filter()  produce a result that cannot be dumped.

src/classes/timeline.py
```diff
@@ -73,35 +74,41 @@ def __init__(self, window):
...
-                self.window.refreshFrameSignal.emit()
+            except Exception as e:
+                log.info("Error applying JSON to timeline object in libopenshot: %s. %s" %
+                         (e, update_action.json(is_array=True)))
 
-            else:
-                # This JSON DIFF is passed to libopenshot to update the timeline
-                self.timeline.ApplyJsonDiff(action.json(is_array=True))
+        # Ignore changes that don't affect libopenshot
+        if len(action.key) >= 1 and action.key[0].lower() in ["files", "history", "markers", "layers", "scale", "profile"]:
```
uglycoder99: Potential issue if key[0] is not a string or does not exist

src/classes/updates.py
```diff
@@ -70,13 +70,13 @@ def json(self, is_array=False, only_value=False):
 
         # Build the dictionary to be serialized
         if only_value:
-            data_dict = copy.deepcopy(self.values)
+            data_dict = json.loads(json.dumps(self.values))
```
uglycoder99, line 73: Possibly make a method that combines the dumps and loads method so you could more easily fix weird copy issues if they come up.

uglycoder99 (approved):
Looks good.  I would move the dumps loads json statements to a function so you could hanlde off cases.

src/classes/project_data.py
jonoomph: Yup, this line does not need the json.loads/json.dumps - because the below calls to `.delete()` don't actually modify anything.

src/classes/updates.py
jonoomph: Yeah, I considered that - however this is a common pattern, and I kind of want to make it obvious what is happening. But it's a good suggestion.

Issue #4404, opened 2021-09-07
Undo function doesnt undo previous item and undoes item many steps prior (add project file)


SoftwareTesterScott:
**Describe the bug**
Undo functionality doenst undo previous actions. A file edit where approximately 15 splits and deletes was undone three or four times and this resulted in the only project file present being removed. 

Note: bug may be a result of a cascading issue as this bug was seen immediately after another bug (raised seperately) which deleted two segments after the program was commanded to remove one.  See video for a recording of bug below

**Steps to reproduce the behavior:**
1. Load and import a file, add to a track
2. Split trim and condense file approx 15 times. 
3. Trigger undo function

**Expected behavior**
Undo feature would undo split and trims before removing the file from the Project Files. 

**System Details**
- OpenShot Version: 2.6.0-dev
- libopenshot Version:  0.2.6-dev
- Platform: Linux-5.11.0-27-generic-x86_64-with-glibc2.29
- Processor: x86_64
- Machine: x86_64
- Python version: 3.8.10
- Qt5 version: 5.12.8
- PyQt5 version: 5.14.1
- Qt Detected Languages: ['en-AU', 'en']
- LANG Environment Variable: en_AU.UTF-8
- LOCALE Environment Variable: 
- Daily Build: Not able to verify issue still exists in daily build

**Log Files**
- [openshot-qt.log](https://s3.amazonaws.com/logs.openshot.org/ig6wi9u40l57ad4soy6uthg00b3uhfe4/openshot-qt.zip) (304 KB)
- libopenshot.log: File too small or empty (not uploaded)

**Exception / Stacktrace**
```No stacktrace found in log files```

**Screenshots** (Optional)
A screen recording was done showing the issue, see https://youtu.be/swaMuWVQtkw?t=1369  The video contains a complete recording from opening to this bug and shows it behaving normally,, me doing a split and trim, then doing an undo after the program removes two clips instead of one (two clip removal already raised as seperate bug), Undo is triggered. 

I should be able to spend some time investigating this further and focusing on getting steps to quickly reproduce the Undo bug if requested. 

This recording doesnt contain any personal information and several other anomalies were identified in the recording but video hasnt been trimmed in case it can be used to help determine complete steps to reproduce.

JacksonRG:
Not sure if you've opened that second bug in another issue, but it looks like you didn't deselect other files before right clicking to delete a file.

[...]

PR #4965, opened 2022-10-26
Fix regression which broke Undo & Redo

jonoomph:
Fix regression which broke undo/redo - this adds a helper copy() function for UpdateAction objects, to easily and safely copy them. Also, remove selections for clips/effect which get deleted in the undo()/redo() functions.

### DIFF
--- a/src/classes/metrics.py
+++ b/src/classes/metrics.py
@@ -41,7 +41,8 @@ from classes.logger import log
 
 import openshot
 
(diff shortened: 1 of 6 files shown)


jonoomph:
[Improve `DeepCopy` Performance](https://trello.com/c/u6x8DzZD/192-improve-deepcopy-performance)

drking-c:
Does this fix #4404 ?

PR #4969, opened 2022-10-31
Fixing regression to Animated Titles

jonoomph:
Fixing regression to animated titles caused by recent image sequence changes. We are missing a "path" key, which is causing our image sequence to fail detection after it is created by Blender.


jonoomph:
[Animated Titles are broken on Daily Build](https://trello.com/c/OQPV5rDf/195-animated-titles-are-broken-on-daily-build)

PR #4971, opened 2022-11-01
Integrate New AudioWaveformer Class (for generating fast, graph-friendly audio datasets)

jonoomph:
Use Root Mean Square to calculate both the "average" and "max" samples per period. Integrate new class from libopenshot. This PR is related to https://github.com/OpenShot/libopenshot/pull/868. Instead of iterating via Python, over all our Frames and building Python lists of audio samples... we call a C++ function and retrieve the data directly from libopenshot. This also fixes some bugs in our applying of Clip Keyframes (time and volume) to our file audio data set.

### DIFF
--- a/src/classes/waveform.py
+++ b/src/classes/waveform.py
@@ -29,7 +29,8 @@ import threading
 from classes.app import get_app
 from classes.logger import log
 from classes.query import File, Clip
-from time import sleep
+from PyQt5.QtGui import QCursor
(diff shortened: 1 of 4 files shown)


jonoomph:
[Waveform + Volume keyframe issues](https://trello.com/c/P88Tv4kW/194-waveform-volume-keyframe-issues)

PR #4974, opened 2022-11-03
Remove $scope.$apply from right click directive

jonoomph:
Remove `$scope.$apply` from right click directive, to prevent double `$apply()` invokations when calling `ApplyJsonDiff()` from context menu.

Related to https://github.com/OpenShot/openshot-qt/pull/4971

### DIFF
--- a/src/timeline/js/directives/misc.js
+++ b/src/timeline/js/directives/misc.js
@@ -33,10 +33,8 @@ App.directive("ngRightClick", function ($parse) {
   return function (scope, element, attrs) {
     var fn = $parse(attrs.ngRightClick);
     element.bind("contextmenu", function (event) {
-      scope.$apply(function () {
-        event.preventDefault();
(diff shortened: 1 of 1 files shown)


jonoomph:
[JS Error on Context Menu (Show Thumbnail)](https://trello.com/c/K9LG4j2v/197-js-error-on-context-menu-show-thumbnail)

src/timeline/js/directives/misc.js
```diff
@@ -33,10 +33,8 @@ App.directive("ngRightClick", function ($parse) {
   return function (scope, element, attrs) {
     var fn = $parse(attrs.ngRightClick);
     element.bind("contextmenu", function (event) {
-      scope.$apply(function () {
```
jonoomph, line 36: No need to apply the entire scope (for Angular) after calling a context menu. This was conflicting with another call to `$scope.$apply()`

PR #4979, opened 2022-11-05
Export Dialog - Re-select Current Profile (if possible)

jonoomph:
Fix for export dialog to always show "All Formats" first in list, and refactor how we maintain state of the current selected profile (i.e. default to project profile, always re-select it if available in list, if not available switch context to first available profile)


jonoomph:
[Export Dialog Issues w/ Non-English](https://trello.com/c/PpmODfo5/199-export-dialog-issues-w-non-english)

PR #4981, opened 2022-11-07
Fixes for Common Translation UI Issues (sorting, missing strings, etc...)

jonoomph:
Enable **sorting** on UI views at runtime, with translated text:

- Transitions
- Effects
- Titles
- Animated Titles
- Project Files
- Emojis

Fix **missing** translatable strings:

- Default Locations (Preferences)

Fix invalid thumbnail **column size** (which would reveal some text next to the icon)

- Transitions
- Effects
- Project Files


jonoomph:
[Translation Issues - Sorting, Thumbnail Size, etc...](https://trello.com/c/hgND30iV/200-translation-issues-sorting-thumbnail-size-etc)

PR #4982, opened 2022-11-07
Removing obsolete translations (including #-#-# characters)

jonoomph:
Removing older, obsolete translations, and fixing some invalid `#-#-#` characters found in many translations due to current translations conflicting with older obsolete ones.


jonoomph:
[Translations contain #-#-# characters](https://trello.com/c/G8myNP7q/201-translations-contain-characters)

PR #4983, opened 2022-11-08
Improve Tutorial Visibility Logic

jonoomph:
Improving tutorial processing logic, so it's processed less often. Restoring views (Simple/Advanced) causes many widgets to change visibility, triggering a huge amount of tutorial processing events. This has been moved to a timer, so it's only processed once for all these changes.

Also, fixing an issue when a widget becomes invisible after the tutorial is visible - so the tutorial hides correctly, until the widget becomes visible again. Before this fix, a hidden widget would cause the tutorial dialog to jump to coordinate 0,0 (top left of screen).

### DIFF
--- a/src/windows/views/tutorial.py
+++ b/src/windows/views/tutorial.py
@@ -27,7 +27,7 @@
 
 import functools
 
-from PyQt5.QtCore import Qt, QPoint, QRectF
+from PyQt5.QtCore import Qt, QPoint, QRectF, QTimer, QObject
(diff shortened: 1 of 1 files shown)


jonoomph:
[Tutorial Jumps to coordinate 0,0 (top left) when widget becomes invisible](https://trello.com/c/xcIpuUF1/202-tutorial-jumps-to-coordinate-00-top-left-when-widget-becomes-invisible)

PR #4984, opened 2022-11-09
Translate Preferences Dropdowns

jonoomph:
Fix a mistake which causes certain dropdowns in the OpenShot preferences to be left untranslated. Also, fix a small bug on launch of OpenShot, which occasionally causes a "has_changes" * save asterisk, due to a race condition of the initial screen resizing, and the loading of the default blank project.

### DIFF
--- a/src/settings/_default.project
+++ b/src/settings/_default.project
@@ -22,7 +22,7 @@
   "effects": [],
   "files": [],
   "duration": 300,
-  "scale": 15,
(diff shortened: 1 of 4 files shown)


jonoomph:
["Location" Preferences are NOT translated correctly](https://trello.com/c/jWKl66lO/204-location-preferences-are-not-translated-correctly)

PR #4985, opened 2022-11-09
Fix Blender Animated Titles for <= 24 FPS Projects

jonoomph:
Rounding length multiplier for all animated title scripts - to allow for <= 24 FPS, without truncating to 0. Currently, if you select a 24 FPS profile and try and generate a Blender Animated Title, it will only export a single frame.


jonoomph:
[Animated titles do not work in <= 24 FPS](https://trello.com/c/Wv6puJER/205-animated-titles-do-not-work-in-24-fps)

PR #4987, opened 2022-11-10
"Recent Folder" Import File logic should Ignore Titles/Animated Titles

jonoomph:
Improving the "Recent Folder" Import File logic to ignore "Title Editor" and "Animated Title Editor" files, and only use recent paths of assets imported into OpenShot directly. It's confusing when a user creates a title, to all of a sudden have that `.openshot_qt/title` folder set as their default import folder.

### DIFF
--- a/src/windows/main_window.py
+++ b/src/windows/main_window.py
@@ -750,10 +750,6 @@ class MainWindow(updates.UpdateWatcher, QMainWindow):
             _("Import Files...")
         )[0]
 
-        if len(qurl_list):
-            # If any files were imported,
-            # Use the folder of the LAST one as the new default path.
-            s.setDefaultPath(s.actionType.IMPORT, qurl_list[-1].toLocalFile())
         # Set cursor to waiting
         app.setOverrideCursor(QCursor(Qt.WaitCursor))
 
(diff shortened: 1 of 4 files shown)


jonoomph:
[`Location->File Import->Recent Folder` should ignore `.openshot_qt` folder](https://trello.com/c/lB1onP69/208-location-file-import-recent-folder-should-ignore-openshotqt-folder)

src/windows/main_window.py
```diff
@@ -750,10 +750,6 @@ def actionImportFiles_trigger(self):
             _("Import Files...")
         )[0]
 
-        if len(qurl_list):
```
jonoomph, line 753: Not needed, since all file importing eventually calls `add_files()`

src/windows/models/files_model.py
```diff
@@ -343,7 +344,8 @@ def add_files(self, files, image_seq_details=None, quiet=False, prevent_image_se
                 # Let the event loop run to update the status bar
                 get_app().processEvents()
                 # Update the recent import path
-                settings.setDefaultPath(settings.actionType.IMPORT, dir_path)
+                if not prevent_recent_folder:
```
uglycoder99, line 347: Kinda reminds me of taking the opposite of an opposite.  Like a double negative,  Nothing wrong with it just seems like it would be cleaner to say something like "use_recent_folder". Struggling to find the exact right words here.

uglycoder99 (commented):
Everything else looks okay.

PR #4988, opened 2022-11-10
Cache Fragmentation Fixes (Preferences & Profile Dialogs)

jonoomph:
Cache can become fragmented when the Preferences or Profile dialogs are opened during the video cache building, since both of those dialogs can adjust the cache.

- Disable video caching on Preferences and Profile windows.
- Re-enable video caching when exiting Preferences and Profile windows
- Exiting Profile dialog will refresh current frame (even if changing FPS)
- Changing Caching settings in Preferences will never fragment cache


jonoomph:
[Cache Fragmentation Fixes (Preferences & Profile Dialog)](https://trello.com/c/Eae4I8z4/209-cache-fragmentation-fixes-preferences-profile-dialog)

PR #5009, opened 2022-11-29
Support for New Linux Build Server + Update AppImage Support

jonoomph:
Related to: https://github.com/OpenShot/libopenshot-audio/pull/146 and https://github.com/OpenShot/libopenshot/pull/878

Adding support for a new Linux build server (Ubuntu 20.04), which has updated dependencies, but also a newer glibc version. This only affects our **AppImage** builds.

**Pros:** 
- Newer FFmpeg 
- Newer Qt
- Newer SVG / ReSVG
- Newer AppImage Format
- New AppStream XML Verification / AppImageLauncher support

**Cons:**
- Older distros (older than April 2020) will likely be incompatible due to glibc. However, we did add some glibc wrapping for OpenShot related executables - although I think more work is needed in this area (i.e. cx_Freeze, Qt, FFmpeg, and other dependencies which are not compiled, but still packaged). However, older Debian-based distros can still install our PPA and get daily updates: https://www.openshot.org/ppa/)

### DIFF
--- a/installer/build_server.py
+++ b/installer/build_server.py
@@ -365,7 +365,8 @@ def main():
             # Copy modified .desktop file to usr/share/applciations
             dest = os.path.join(app_dir_path, "usr", "share", "applications")
             os.makedirs(dest, exist_ok=True)
-            shutil.copyfile(desk_out, os.path.join(app_dir_path, "usr", "share", "applications"))
+            shutil.copyfile(os.path.join(app_dir_path, "org.openshot.OpenShot.desktop"),
+                            os.path.join(app_dir_path, "usr", "share", "applications", "org.openshot.OpenShot.desktop"))
 
             # Rename executable launcher script
             launcher_path = os.path.join(app_dir_path, "usr", "bin", "openshot-qt-launch")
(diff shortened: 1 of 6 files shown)


jonoomph:
[New AppImage + New Linux Builder](https://trello.com/c/u8jCeev8/240-new-appimage-new-linux-builder)

installer/launch-linux.sh
```diff
@@ -1,30 +1,5 @@
 #!/bin/bash
 
-# Query the current display DPI
```
jonoomph, line 3: High DPI is now supported by default in the AppImage format

installer/build_server.py
```diff
@@ -340,54 +340,61 @@ def main():
...
             st = os.stat(app_run_path)
             os.chmod(app_run_path, st.st_mode | stat.S_IEXEC)
-            os.chmod(desktop_wrapper, st.st_mode | stat.S_IEXEC)
             os.chmod(launcher_path, st.st_mode | stat.S_IEXEC)
 
             # Create AppImage (OpenShot-%s-x86_64.AppImage)
             app_image_success = False
             for line in run_command(" ".join([
-                '/home/ubuntu/apps/AppImageKit/AppImageAssistant',
+                '/home/ubuntu/apps/AppImageKit/appimagetool-x86_64.AppImage',
```
jonoomph, line 387: Updating from very old AppImage creation process to latest / best practice

PR #5010, opened 2022-11-30
Improve user-guide embedded javascript (resize iFrame, external link targets)

jonoomph:
Improve user guide embedded javascript, to communicate back to openshot.org/user-guide correctly. Since we are hosting these docs on a cdn, this is the only secure way to resize the contents of these docs inside an iframe. Also, making external links use target _blank

### DIFF
--- a/doc/_templates/layout.html
+++ b/doc/_templates/layout.html
@@ -1,2 +1,22 @@
 {% extends "!layout.html" %}
 {% set css_files = css_files + ["_static/tablefix.css"] %}
+{%- block footer %}
+  <script>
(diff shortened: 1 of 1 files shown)


jonoomph:
[Improve User-Guide Webpage (iFrame)](https://trello.com/c/ucHUzdVn/243-improve-user-guide-webpage-iframe)

jonoomph:
[Improve User-Guide Webpage (iFrame)](https://trello.com/c/ucHUzdVn/243-improve-user-guide-webpage-iframe)

PR #5011, opened 2022-11-30
Fix changelog parsing (simplify regex, code cleanup)

jonoomph:
Simplify the changelog parsing to only look in the `/settings` folder, and use a simpler regex to match the various parts. Also, removing an older format we were supporting.


jonoomph:
[Changelog data is offset - View on GitHub  link broken](https://trello.com/c/sxG0Fl62/242-changelog-data-is-offset-view-on-github-link-broken)

PR #5014, opened 2022-12-01
More attempts to communicate between cdn hosted documentation embedde…

jonoomph:
More attempts to communicate between CDN hosted documentation embedded on openshot.org iframe


jonoomph:
[Improve User-Guide Webpage (iFrame)](https://trello.com/c/ucHUzdVn/243-improve-user-guide-webpage-iframe)

PR #5018, opened 2022-12-02
Fixing Inkscape/Blender launching in AppImage

jonoomph:
Fixing inkscape launching in AppImage on newer / different distros, use **system libraries** for missing libraries.

### DIFF
--- a/src/windows/title_editor.py
+++ b/src/windows/title_editor.py
@@ -29,6 +29,7 @@
 import os
 import re
 import shutil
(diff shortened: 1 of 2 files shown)


jonoomph:
[Blender support from AppImage does not need to include `LD_LIBRARY_PATH`](https://trello.com/c/DzZQLIXg/136-blender-support-from-appimage-does-not-need-to-include-ldlibrarypath)

PR #5051, opened 2022-12-18
Fixing many bugs reported via Sentry.io, and refactoring Audio Device detection

jonoomph:
Related to https://github.com/OpenShot/libopenshot/pull/883

- Refactor of local HTTP thumbnail server initialization
  - Retry logic (in case we picked an unavailable port)
  - Loud error message (prompt user if impossible to start HTTP localhost server)
  - If failure, exit OpenShot after error message is displayed
- Store both the requested audio device `name` and `type` as a double pipe delimited value: `"PulseAudio Sound Server||ALSA")`. And during launch of OpenShot, set both values in libopenshot settings.
- Refactor of audio device callback (after launch), to update OpenShot preferences to match the actual device opened, and the actual sample rate opened. This also adds some debug logs to notate when things happen. But hopefully, this will prevent OpenShot from running in a different sample rate than the OS, and prevent audio drift.
- Misc Sentry fixes:
  - Improved shutdown of OpenShot (and better protected some clean-up code in the closeEvent method)
  - Prevent error when adding new files to the timeline (clip's that do not contain an 'id' property yet)

### DIFF
--- a/src/windows/preview_thread.py
+++ b/src/windows/preview_thread.py
@@ -87,31 +87,11 @@ class PreviewParent(QObject, UpdateInterface):
 
     # Signal when the playback encounters an error
     def onError(self, error):
-        log.warning('Player error: %s', error)
-
         # Get translation object
         _ = get_app()._tr
 
         # Only JUCE audio errors bubble up here now
         QMessageBox.warning(self.parent, _("Audio Error"), _("Please fix the following error and restart OpenShot\n%s") % error)
-
-    def onDefaultSampleRate(self, detected_rate):
-        """The default audio sample rate (as detected on the default audio device)"""
-        log.info('Detected default audio device sample rate: %d' % detected_rate)
-
-        s = get_app().get_settings()
-        rate = int(s.get("default-samplerate") or 48000)
-        if detected_rate != rate:
-            log.warning("Your sample rate (%d) does not match OpenShot (%d)."
-                        "Adjusting your 'Preferences->Preview->Default Sample Rate to match your "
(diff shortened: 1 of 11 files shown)


src/windows/preview_thread.py
```diff
@@ -214,8 +227,7 @@ def Start(self):
         # Check for any Player initialization errors (only JUCE errors bubble up here now)
         # But slightly delay, to allow for correct audio thread initialization with the
         # correct number of channels and sample rate
-        QTimer.singleShot(100, self.CheckAudioError)
-        QTimer.singleShot(100, self.CheckDefaultSampleRate)
+        QTimer.singleShot(1000, self.CheckAudioDevice)
```
jonoomph, line 230: Slow this down, to ensure audio device manager has already been started before we call it's singleton

src/windows/views/webview.py
```diff
@@ -247,11 +247,13 @@ def update_clip_data(self, clip_json, only_basic_props=True, ignore_reader=False
         except Exception:
             # Failed to parse json, do nothing
             log.warning('Failed to parse clip JSON data', exc_info=1)
+            return
 
         # Search for matching clip in project data (if any)
-        existing_clip = Clip.get(id=clip_data["id"])
+        existing_clip = Clip.get(id=clip_data.get("id"))
```
jonoomph, line 253: "id" does not always exist, for example when dropping a new file into the timeline

jonoomph:
[Fix top sentry bugs / exceptions](https://trello.com/c/V0yZMg1g/16-fix-top-sentry-bugs-exceptions)

PR #5055, opened 2022-12-20
Delete clips / transitions with invalid or negative duration

jonoomph:
Delete clips / transitions with invalid or negative duration. For example, trim the left edge past the right edge, or the right edge past the left edge, and the item will now be deleted.

Previously, clips and transitions with a negative duration would just flicker and re-appear on the timeline, but are completely broken (no longer render on the timeline)


jonoomph:
[Dragging the end of a clip before the start breaks the clip](https://trello.com/c/TReo5Wzu/11-dragging-the-end-of-a-clip-before-the-start-breaks-the-clip)

PR #5057, opened 2022-12-20
Add tolerance to JavaScript track position calculation

jonoomph:
Add tolerance to JavaScript track position calculation - to better determine which track is under the cursor (i.e. used when dropping effects on a clip). This makes it the accuracy of dropping effects on a clip feel much better, and users should have less frustration with this.


jonoomph:
[Effect Drop on Clip is Inaccurate](https://trello.com/c/qzLqEMcH/19-effect-drop-on-clip-is-inaccurate)

PR #5058, opened 2022-12-20
Calculate z-index for clips and transitions (selected items use larger z-index)

jonoomph:
Calculate z-index for clips and transitions on the timeline, and ensure selected z-indexes are higher than unselected ones, to allow a selected clip to be easily resized/trimmed. This makes editing clips/transitions that are snapped together much easier to grab the edge of the selected item.

### DIFF
--- a/src/timeline/index.html
+++ b/src/timeline/index.html
@@ -83,7 +83,7 @@
 				</div>
 				
 				<!-- CLIPS -->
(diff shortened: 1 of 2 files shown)


jonoomph:
[Selected Clips/Transitions should be easy to Trim/Resize](https://trello.com/c/n8WoqnnJ/20-selected-clips-transitions-should-be-easy-to-trim-resize)

PR #5092, opened 2023-01-13
Bug Fix: Multiple Clips/Transitions Dropped on New Layer

jonoomph:
Also keeping selected items (clips / transitions) after drop... instead of clearing the selection. This is especially useful when a bunch of specific items are selected, and the user might require many attempts to position them correctly.


jonoomph:
[Moving multiple clips from one track to another track does not work.](https://trello.com/c/xWtnv1fJ/29-moving-multiple-clips-from-one-track-to-another-track-does-not-work)

Issue #5007, opened 2022-11-28
ImportError: libkeyutils missing, python version "mismatch"


petaflot:
**Describe the bug:**
ImportError when starting the app, app crashes on startup, gui does not even show for a fraction of a second

**Steps to reproduce the behavior:**
Download appimage, chmod+x, execute.

**Expected behavior:**
obviously the app should start.

neither sys-apps/keyutils nor dev-python/python-keyutils are installed on the system ; however, backtrace suggests python 3.6 which has been removed from the system around a year ago or so (system now has python 3.10.8 and 3.11.0 installed)

does it mean I need to reinstall the entire python 3.6 environment just for openshot? I thought the Appimage had "everything included" to make it work

**System Details:**
- OpenShot Version : /opt/OpenShot-v2.6.1-x86_64.AppImage
- Operating System / Distro: Gentoo Base System Release 2.9

**Log Files:**
see backtrace below.

**Exception / Stacktrace:**

```
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-meta'
INFO app: ------------------------------------------------
INFO app:             Mon Nov 28 12:51:25 2022            
INFO app:               Starting new session              
ERROR app: OpenShotApp::Import Error
Traceback (most recent call last):
  File "/home/gitlab-runner/builds/d0022447/0/OpenShot/openshot-qt/openshot.py", line 14, in swig_import_helper
  File "/usr/lib/python3.6/importlib/__init__.py", line 126, in import_module
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 658, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 571, in module_from_spec
  File "<frozen importlib._bootstrap_external>", line 922, in create_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
ImportError: libkeyutils.so.1: cannot open shared object file: No such file or directory

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/tmp/.mount_esaaGf/usr/bin/classes/app.py", line 99, in __init__
    from classes import settings, project_data, updates
  File "/tmp/.mount_esaaGf/usr/bin/classes/project_data.py", line 48, in <module>
    import openshot
[...]

PR #5093, opened 2023-01-13
Experimental freeze changes to include libthai.so (Slackware 15)

### LINKED ISSUE #5060: AppImage of v3.0 fails if `libthai.so` is missing
I am running the AppImage of v3.0 on a Gentoo system without `libthai` installed. Openshot fails to open in this case, saying it cannot find `libthai.so`. After installing `libthai`, Openshot will launch, but it seems like this should be an optional dependency, or bundled with the AppImage.

jonoomph:
Experimental freeze changes, to include libthai, required for some font rasterizing - needed by Slackware distro in AppImage.

Closes #5060, #5007

### DIFF
--- a/freeze.py
+++ b/freeze.py
@@ -391,7 +391,7 @@ elif sys.platform == "linux":
                     "libfreetype.so.6",
                     "libfontconfig.so.1",
                     "libharfbuzz.so.0",
-                    "libthai.so.0",
+                    #"libthai.so.0",
                     ]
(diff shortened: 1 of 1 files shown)


jonoomph:
[AppImage: Slackware 15 Requires libthai.so](https://trello.com/c/hBF6IL7z/30-appimage-slackware-15-requires-libthaiso)

PR #5096, opened 2023-01-16
Fix transition direction detection bug, when clips are overlapping

jonoomph:
fix-transition-direction

In this example (seen in attached image), the direction of the transition would be set incorrectly - because we were not filtering out clips on the same layer.


jonoomph:

PR #5099, opened 2023-01-16
Fix transition direction (overlapping clips)

jonoomph:
When clips are overlapping (i.e. auto created transitions), replacing or moving those did not work correctly, and often, the direction would be set incorrectly... causing the 2nd clips to fade out, and then snap back to visible.


jonoomph:

PR #5102, opened 2023-01-17
Transition Direction: Ignore certain clips which don't affect transitions

jonoomph:
When determining transition direction, ignore clips that end too close to the beginning of the transition. This prevents an edge case where the End of Clip A and the Start of Clip B are both 0.0 seconds from the start of the transition.


jonoomph:

PR #5127, opened 2023-02-03
Default Clip Keyframes (when removing all keyframes)

jonoomph:
Add **sane** default clip keyframe values, for all common properties (i.e alpha, scale, location, rotation, etc...). For example, when removing the default `alpha` keyframe, the clip should not suddenly become invisible... but rather default back to the default value... which is fully visible). Same for all other default clip properties which are Keyframe objects.

remove-all-keyframes


jonoomph:
[Clip turns black in "Video Preview" window after your remove the Alpha keyframe.](https://trello.com/c/UnyLEol5/35-clip-turns-black-in-video-preview-window-after-your-remove-the-alpha-keyframe)

PR #5146, opened 2023-02-14
Group all 'Separate Audio' context menu updates in a single undo/redo transaction

jonoomph:


jonoomph:
[Undo of Group "Separate Audio" does not work properly.  You have to Undo multiple times to undo.](https://trello.com/c/Y75JWj0v/39-undo-of-group-separate-audio-does-not-work-properly-you-have-to-undo-multiple-times-to-undo)

PR #5147, opened 2023-02-14
Expand Default Emojis (adding in Arrows, Shapes, and Brands)

jonoomph:
Also, updating the OpenMoji SVG icons (where changed), and including a few new emojis which are useful for overlays. 

- Updating SVG icon cache
- Updating POT translations
- Updating SVG emojis (adding Symbol emojis)
- Fixing bug with Emoji Group Filter (now saves a non-translated ID to the settings)


jonoomph:
[Additional emojis](https://trello.com/c/0xk6faZb/38-additional-emojis)

PR #5169, opened 2023-03-11
Time Remapping - Invalid `End` frame #

jonoomph:
When using the "Time" context menu to reverse or change the speed of a clip, we were calculating the "End" frame # incorrectly, resulting in occasionally duplicating the last or first frame #.


jonoomph:
[Time Remapping Crashes (due to Audio Resampling)](https://trello.com/c/myJ5A44q/34-time-remapping-crashes-due-to-audio-resampling)

PR #5183, opened 2023-03-27
Fix Lag in Title Editor on UI Thread

jonoomph:
On slower computers, the main UI thread bogs down when rendering previews of the SVG title, in the Title Editor. This work needs to be moved to it's own thread, in a non blocking way.


jonoomph:
[Title Editor lag on certain SVG templates](https://trello.com/c/PO5ISq3a/51-title-editor-lag-on-certain-svg-templates)

PR #5200, opened 2023-04-12
Duplicate Profile Names can not be selected

jonoomph:
Fix duplicate profile names, including a unique name for those with duplicates. For example, their are many similar profiles that go by a common name, for example `PAL SD Anamorphic 576i 25 fps` has both a `4:3` and a `16:9` ratio version. These profiles must have a unique name for OpenShot to function correctly. 

Also, adding a "Search Profile" button for our Preferences (when updating the default profile):
Default-Profile-OpenShot

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,8 +23,5 @@ jobs:
         pip3 install setuptools wheel sentry-sdk
         pip3 install cx_Freeze==6.1 distro defusedxml requests certifi chardet urllib3
 
-    - name: Build Python package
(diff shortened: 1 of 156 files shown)


jonoomph:
[Chosen Profile does not stay selected.](https://trello.com/c/vdRqG8DO/23-chosen-profile-does-not-stay-selected)

jonoomph:
[Edit Title does not correct updated thumbnails on timeline](https://trello.com/c/VBDNBL2K/35-edit-title-does-not-correct-updated-thumbnails-on-timeline)

PR #5201, opened 2023-04-13
`Edit Title` does not update thumbnails on timeline

jonoomph:
Fix for editing a title and trying to update a thumbnail on the timeline - was missing a required arg. This error was caught in Sentry.


jonoomph:
[Edit Title does not correct updated thumbnails on timeline](https://trello.com/c/VBDNBL2K/35-edit-title-does-not-correct-updated-thumbnails-on-timeline)

PR #5202, opened 2023-04-13
Protecting timeline args from non-Integer values

jonoomph:
Protecting timeline args from non-Integer values. Detected on Sentry: OPENSHOT-245G. Occasionally Sentry reports a float sample rate, which breaks our Timeline() constructor


jonoomph:
[Non-Integer sample rate breaks openshot::Timeline() constructor](https://trello.com/c/nSEZSZVP/36-non-integer-sample-rate-breaks-openshottimeline-constructor)

PR #5205, opened 2023-04-13
Prevent Crash in Object Detector and Tracker effects (due to child clip id)

jonoomph:
There have been many bad recent regressions which have broken our Object Detector and Tracker effects - mostly related to optimizations before we save project JSON - accidentally deleting some bits that are important, such as our "tracked" objects, among other Property window errors (missing or invalid child clip ids or parent ids), etc...

This PR fixes all of this, and protects the code much better against missing or invalid IDs.

### DIFF
--- a/src/classes/timeline.py
+++ b/src/classes/timeline.py
@@ -79,6 +79,7 @@ class TimelineSync(UpdateInterface):
             return
 
         # Disable video caching temporarily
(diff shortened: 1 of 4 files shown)


jonoomph:
[`Child_Clip_ID` in the Tracker effect causes OpenShot to crash](https://trello.com/c/vFGMVC9t/20-childclipid-in-the-tracker-effect-causes-openshot-to-crash)

PR #5210, opened 2023-04-16
Show friendly error to user when Export dialog fails to parse project JSON

jonoomph:
Show friendly error to user when Export dialog fails to parse project JSON data: Reported in Sentry: OPENSHOT-6X. I do not know the root cause of this, but hopefully this error will help users resolve or identify the issue in more detail.


jonoomph:
[Show friendly error to user when Export dialog fails to parse project JSON](https://trello.com/c/2ioIFlk6/40-show-friendly-error-to-user-when-export-dialog-fails-to-parse-project-json)

PR #5211, opened 2023-04-16
Protect from generating a waveform for a NULL clip

jonoomph:
Protect from generating a waveform for a NULL clip (we should skip clips that we can't find or query). Reported in Sentry: OPENSHOT-1WJ5


jonoomph:
[`Show Waveform` throws exception to log if Clip is not found (or removed)](https://trello.com/c/2G4rGJkH/41-show-waveform-throws-exception-to-log-if-clip-is-not-found-or-removed)

PR #5260, opened 2023-06-10
Effect sequencing - update documentation

jonoomph:
Update our documentation with new Effect properties, plus a new section on effect sequencing (allowing the user to control when an effect is applied - before or after clip keyframes are processed)


jonoomph:
[Improve Crop Effect to allow "Resize Image"](https://trello.com/c/XAQpA4Ok/51-improve-crop-effect-to-allow-resize-image)

Issue #5264, opened 2023-06-14
Title text lines are translated


MateuszChm:
Hello,
First of all - thank you for creating such a great tool. I've been using the Openshot for a few years, and it's incredible how the application has improved. Unfortunately, I believe I found a minor bug, but the good news is that it should be easy to fix.

**Describe the bug:**
A title text lines, which user can set, are translated, but they should not.

**Steps to reproduce the behavior:**
1. Change the language to any other than English.
2. Create a title and type a text to any of the text lines, which is also defined as a translation key (e.g. "Center").
3. Save the title.
4. Open the title modification modal - the text will change.

**Expected behavior:**
Title text lines should not be translated.

**System Details:**
- OpenShot Version: 3.1.1
- Operating System / Distro: Ubuntu 20.04

**Log Files:**
N/A

**Exception / Stacktrace:**
N/A

**Screenshots:** (Optional)
https://github.com/OpenShot/openshot-qt/assets/18557918/d14ece80-2e73-4309-b9e7-4ae05d515554

ferdnyc:
@MateuszChm 

Not sure why your PR is still unmerged, your fix is the only possible solution to the situation as it currently stands.

This is partly an issue of translation context — specifically [the lack of any](https://github.com/OpenShot/openshot-qt/issues/4326#issuecomment-904622180) in the OpenShot i18n. If translations used contexts, then a translation for title template strings could be distinct from a translation for dialog buttons, and "Close" wouldn't have to translate the same everywhere.

But even _with_ context, the code would need to make a distinction between **new** title data being created (when it would be reasonable to translate the initial text contents of the templates, so they're useful in all locales) and _existing_ user data.

Existing user data should **never** be (re-)translated, under any circumstances. As soon as the user hits "Save", it becomes _their content_. The assumption then is (or should be) that whatever text the user committed into the project, that's the exact text they expect to see when they return. Any additional translation passes on user-modifiable data risk corrupting it, as you've seen.

PR #5265, opened 2023-06-14
#5264 Unnecessary title text lines translation removed

MateuszChm:
This simple change should fix the problem reported in #5264 issue


jonoomph:
Thanks for the patch, and sorry this took so long to get to.

PR #5336, opened 2023-10-08
Migrate from altool to notarytool (for Mac Notarization)

jonoomph:
Since Apple has now blocked the old method of notarizing, our Mac builds have been broken for a little bit. This replaces the old notarize command with the new `notarytool` command, which is much easier to use in all honesty (i.e. --wait keyword).


jonoomph:
[Mac Notarization Failing, needs migration to new notarytool](https://trello.com/c/I9HcUzP0/59-mac-notarization-failing-needs-migration-to-new-notarytool)

PR #5424, opened 2024-02-15
Copy Effects Between Clips (bug fix)

jonoomph:
Fixing bug when copying effects from 1 clip to another clip. We were copying the "id" from the first clip, instead of generating a new one. The created all sorts of chaos, and linked the 2 effects together incorrectly.


jonoomph:
[Effects are linked when copied from one clip to another](https://trello.com/c/IsqSHXkJ/69-effects-are-linked-when-copied-from-one-clip-to-another)

PR #5425, opened 2024-02-15
Fix Copy/Paste Shortcuts (when Timeline has focus)

jonoomph:
This is a temporary work-around, until our new Timeline widget is ready. But this fixes certain webview widgets from eating our Ctrl+C / Ctrl+V keypresses when the timeline is in focus.


jonoomph:
[CTRL+C does not work.](https://trello.com/c/JamFnb78/65-ctrlc-does-not-work)

PR #5428, opened 2024-02-19
Add Transaction around Copy/Paste on Timeline

jonoomph:
This allows all items pasted on the timeline to be removed with a single **undo**, or added back with a single **redo**.


jonoomph:
[Undo requires multiple action to undo a Copy and Paste of multiple clips.](https://trello.com/c/Ike4W9Wn/64-undo-requires-multiple-action-to-undo-a-copy-and-paste-of-multiple-clips)

PR #5430, opened 2024-02-19
Improved Tracker & Object Detector Effects (Refactor, Bug-Fixes, Enhancements)

jonoomph:
Related PR: https://github.com/OpenShot/libopenshot/pull/950

**Improvements:**
- Fixed timeline tooltips on clip titles, effects and parent badges (bootstrap js update)
- Disable "Process Effect" button after clicked
- Fixed right click menu on "Selected Object" property editor (it was broken). Added "class_name: object id" as format.
- Added ability to clear class_filter (on Object Detector effect)
- Removing child_clip_id property, replacing with parent property for the actual child clips (I.e. parent any clip on a tracked object I'd)
- Fix sorting of tracked object properties in the Property Editor
- Added lots of new documentation for Effects, including Tracker & Object Detector effects. New docs on Parenting also include.
- Large refactor of context menus in OpenShot (when they contain more than 20 items of sub-items). Nested menus now contain a range of items (I.e. 1-20, 21-40, etc...). Applied this format to all Property Editor context menus used in OpenShot.
- Added 'class_filter' context menus, to choose between any visible tracked object classes (I.e. car, truck, etc...)
- Fixed undo/redo issues related to effect property's being updated. For example, some effect updates (or tracked object property updates) would not undo correctly.
- No longer show context menus for read-only properties
- Migrate old *.osp projects that include Tracker or Object Detector effects, to fix alpha values, move child_clip_id to the parent property, etc...
- Bump version of OpenShot to 3.1.1-dev

**Prompting** user if Tracker effect region is invalid (i.e. not at the beginning of the clip), and then jumping the preview to the first frame. Otherwise, if a user chooses a region further into a clip, the Tracker can often fail to ever correctly track this region, since it always begins tracking on the 1st frame of the clip.

OpenShot-Choose-Valid-Region

**Improved Selected Object menu** (Object Detection Effect)
OpenShot-Object-Selector-Menu

**New Parent Badge** (for any clip which has a parent set):
OpenShot-Parent-Badge

---
**Refactor Object Detector** with QPainter (support for corner radius, faster drawing performance)
OpenShot-Object-Detector-Radius

**Refactor Tracker** with QPainter (support for corner radius, faster drawing performance)
OpenShot-Tracker-Radius


jonoomph:
[Tracker Effect has serious regressions (different `types` crash, attributes don't work, poor performance)](https://trello.com/c/zxNARDjF/27-tracker-effect-has-serious-regressions-different-types-crash-attributes-dont-work-poor-performance)

PR #5484, opened 2024-03-24
fix issue with parsing relative filepath

d-vasilev:
Conversion of asset path from relative to absolute does not work, as the used regexp is selecting the entire json data. Proposed change is to select the minimum string which fits the pattern.


jonoomph:
@d-vasilev Thanks for this patch, it looks good to me.

PR #5485, opened 2024-03-26
Verify ALL selections on Undo, and clear property window (if selection is invalid)

jonoomph:
Verify ALL selections on Undo, and clear property window (if selection is invalid). For example, if you drop an effect on a clip, select the effect (which makes the Properties dock visible), and then UNDO until the effect is no longer valid, the property window was remaining on the screen, even though the effect was deleted.
OpenShot-Empty-Properties


jonoomph:
[Effects Undo function not working correctly.](https://trello.com/c/Uqu6KKH1/13-effects-undo-function-not-working-correctly)

PR #5487, opened 2024-03-27
Theme Refactoring (with new Theme: Cosmic Dusk)

jonoomph:
In an effort to update the user-interface for OpenShot, first our theme engine needs some major refactoring. 

### Theme Engine
The current theme engine for OpenShot is really just a few lines of hard-coded stylesheet that is very difficult to work with. We need to refactor this into Classes, which can inherit some base theme properties, etc... We want new themes to be very easy to switch between, and we want real-time theme switching.

### Timeline Refactor
We currently support 2 different webview-based timeline widgets (QWebKit and QWebEngine), and we will eventually replace both of these with a QWidget-based timeline widget. In the meantime, we need to refactor the timeline code to support a 3rd widget (experimental placeholder currently), and clean-up the timeline-related code as much as possible. This will be a gradual change-over, since much work remains on the new QWidget timeline widget. This will eventually be a huge simplification on how our timeline works, improved performance when editing projects, huge reduction in our download size, and a more feature-full timeline widget, which of course is also theme-able (due to this new theme engine).

### New Theme
Part of this PR includes some initial work on a new UI theme for OpenShot. This is very early and should not be used by normal users yet. We have very detailed mock-ups and work-flows that we are moving towards, but it will be an iterative process before we switch the default theme.


jonoomph:
[New UI Theme (refactor, theme manager, new timeline)](https://trello.com/c/5Z8M5NCa/72-new-ui-theme-refactor-theme-manager-new-timeline)

PR #5523, opened 2024-06-01
Fix shortcuts for Copy, Paste, and Slice

jonoomph:
Fix shortcuts for Copy, Paste, and Slice, which broke due to Timeline back-end refactor.


jonoomph:
[CTRL+C shortcut key is not working and causing a crash.](https://trello.com/c/Fu4dcHOB/110-ctrlc-shortcut-key-is-not-working-and-causing-a-crash)

PR #5548, opened 2024-06-25
Move to typed enums from QtCore.Qt

Vekhir:
The codebase currently accesses the fields of `QtCore.Qt` directly, e.g. `Qt.white`. `Qt.white` is part of the `Qt.GlobalColor` enum, so can also be accessed via `Qt.GlobalColor.white`. The latter access method is mandatory with PyQt6 to achieve better type safety.

The full list of enums is available at https://www.riverbankcomputing.com/static/Docs/PyQt5/api/qtcore/qt.html. I've checked for each and every literal name whether it is used in OpenShot, and then committed the changes by enum type. The commit header and message are essentially the same each time, mostly noting the enum type contained within it and a note that it's required for PyQt6 compatibility. The split into 33 commits is supposed to make the code review easier; if you'd rather like a single commit, I can squash them.

The idea is more thoroughly explained in #4833. This PR is a subset of the changes there - this is by no means complete. I've tried to keep the changes as small and simple as possible, being basically find and replace.

Feedback and comments are appreciated.

-- Vekhir


Vekhir:
@jonoomph Is there something that prevents this PR from being merged? It's a small step toward PyQt6 support and referencing the scope of the enums is recommended by PyQt5 aswell.

PR #5552, opened 2024-06-27
Protect windows.statusBar on launch of OpenShot

jonoomph:


jonoomph:
[Fixing Sentry bug on app launch (statusBar.hide())](https://trello.com/c/incwuxrk/133-fixing-sentry-bug-on-app-launch-statusbarhide)

PR #5556, opened 2024-06-28
Fixing many misc Sentry reports on OpenShot 3.2

jonoomph:


jonoomph:
[Fixing Sentry bug on app launch (statusBar.hide())](https://trello.com/c/incwuxrk/133-fixing-sentry-bug-on-app-launch-statusbarhide)

PR #5573, opened 2024-07-07
Theme Manager refactor (prevent freezing on Windows)

### LINKED ISSUE #5549: App won't load
**Describe the bug:**
I just downloaded the new update (version 3.2). The app won't load. When I try to open OpenShot, it opens, but then it freezes and windows says that the app is not responding. I have tried several times and the same thing happens each time. I have even uninstalled the app and reinstalled it and the problem persists. (I had no problem running the previous version of OpenShot.)

Happens whenever I try to start the app.

**Expected behavior:**
I expect the app to open properly so that I can use it.

**System Details:**
- OpenShot Version: 3.2
- Operating System: Windows 10, 64 bit

**Log Files:**
[libopenshot.log](https://github.com/user-attachments/files/15972847/libopenshot.log)
[openshot-qt.log](https://github.com/user-attachments/files/15972848/openshot-qt.log)

jonoomph:
This is a refactor of the ThemeManager, making it more thread-safe and less likely to freeze certain versions of Windows. This limits the # of times ThemeManager is imported, and maintains a single reference to the manager instance. It also removes ThemeManager access from the video widget paint method, which was directly responsible for the crash / freeze in certain versions of Windows.

The freeze would manifest itself as an issue with `oleaut32.dll` in a debugger, which was the last line logged when the freeze would occur. While I never fully found the root cause of the freeze, removing ThemeManager from the video preview widget paint method seems to resolved the issue for testers.

Closes #5549

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -153,7 +153,7 @@ class OpenShotApp(QApplication):
 
         # Instantiate Theme Manager (Singleton)
         from themes.manager import ThemeManager
-        ThemeManager(self)
+        self.theme_manager = ThemeManager(self)
 
     def show_environment(self, info, openshot):
         log = self.log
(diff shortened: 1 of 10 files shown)


jonoomph:
[Add Logging to Debug Windows 10 Freeze](https://trello.com/c/v2oBaXqh/134-add-logging-to-debug-windows-10-freeze)

PR #5574, opened 2024-07-08
Delay Showing Main Windows (until after Theme is applied)

jonoomph:
Delay the showing of the main window until all themes are applied fully. This prevents the "popping" / rapid changing of the theme engine, which first clears any theme, and then applies the selected theme. In other words, this makes the initial launch of OpenShot smoother.

### DIFF
--- a/src/classes/app.py
+++ b/src/classes/app.py
@@ -273,9 +273,6 @@ class OpenShotApp(QApplication):
         # Connect our exit signals
         self.aboutToQuit.connect(self.cleanup)
 
-        # Process any queued events
-        QApplication.processEvents()
-
         # Show main window
         self.window.show()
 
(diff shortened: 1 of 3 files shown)


jonoomph:
[Delay showing main window until all themes are loaded](https://trello.com/c/S1P9lCrN/137-delay-showing-main-window-until-all-themes-are-loaded)

src/classes/app.py
```diff
@@ -273,6 +273,12 @@ def gui(self):
         # Connect our exit signals
         self.aboutToQuit.connect(self.cleanup)
 
+        # Process any queued events
+        QApplication.processEvents()
```
jonoomph: The idea here is to fully process the singleShot timers (for applying geometry and position state) before showing the window, to avoid the window "jumping" right after we show it.

PR #5576, opened 2024-07-08
Restore geometry and window state (on Windows High DPI monitors)

jonoomph:
Restore geometry and window state twice (once on creation of main windows, and once after window is shown). This fixes an issue where on Windows, Qt does not yet know about the scaling pixel ratio until the window is shown, causing the geometry to get HUGE, due to the window size increasing on each launch of OpenShot. But on other OSes (Linux, Mac), it seems to work on window creation just fine.


jonoomph:
[Window geometry is incorrect on High DPI Windows screens (huge screen)](https://trello.com/c/CB43ljtC/138-window-geometry-is-incorrect-on-high-dpi-windows-screens-huge-screen)

PR #5577, opened 2024-07-09
Fixing more Sentry Errors

jonoomph:
- Empty context menus
- Index out of range (Add to Timeline dialog, up/down buttons)
- Index out of range (populating Properties model)
- Update listeners failing when action.key is empty list (i.e. "load" action type)


jonoomph:
[Fixing more Sentry errors](https://trello.com/c/JWNZkaTq/139-fixing-more-sentry-errors)

PR #5584, opened 2024-07-31
Detect font (and style: italics and bold) in Title Editor

jonoomph:
Correctly detect font family, italics, bold, and font-size when editing previous titles, so the font selector defaults correctly, and the font's don't revert back to the default font.

### DIFF
--- a/src/classes/info.py
+++ b/src/classes/info.py
@@ -28,7 +28,7 @@
 import os
 from time import strftime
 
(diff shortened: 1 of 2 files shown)


jonoomph:
[Static Title Fonts do not stick when duplicating or editing.](https://trello.com/c/zWMayO4W/143-static-title-fonts-do-not-stick-when-duplicating-or-editing)

PR #5592, opened 2024-08-09
Fix "SyntaxWarning: invalid escape sequence"

nteodosio:
Reported downstream at https://launchpad.net/bugs/2066375.


joo4mart:
The simple solution: in the file _language/generate_translations.py_ you must only change '\\' to '\\\\' in two lines and the warning goes away.

a-detiste:
patched in Debian

PR #5610, opened 2024-09-07
Prevent Gaps when Editing / Exporting / Changing Profiles

jonoomph:
This is a pretty major change to OpenShot, and is attempting to solve a really big "precision" issue when editing clips and transitions in a specific profile FPS (framerate). We want to ensure that **ALL** clip **position**, **start**, and **end** values are snapped to the current profile's FPS grid / precision. Another way of saying this, we want all of the key position values to land exactly on a real frame, and NEVER between frames. So if we are editing a 30 FPS video project, all our our position, start, and end values much be in increments of 1/30 of a second.

When changing profiles and/or exporting to a different FPS profile, we need to ensure that ALL clips and transitions are converted to this new FPS precision also. The challenge of this, some FPS changes cause unavoidable gaps between clips (+ or - 1 to 3 frames) depending on the change. We want to avoid "random black" gaps on our timeline, so when we change FPS in OpenShot, we are now removing TINY gaps that might have been caused by the FPS change (1 to 3 frames).

Since there are 3 ways to remove a gap (position, start trim, or end trim), we have decided to only adjust the "end" trim, since this method maintains and protects the original timing and start positions of all clips. This might even extend the end trim past the duration of a video clip, which will simply repeat the last frame or two. While some edge cases caused by this are not ideal, it's the most reasonable way we can think of to adapt an entire timeline of edits in a different FPS precision, without loosing the original timing or the original start trim of clips.

**Snapping Changes:**
- Moving clips / transitions on the timeline now snaps to accurate frame positions
- Resizing clips / transitions on the timeline now snaps to accurate frame positions
- Manual Property Editing: position, start, end now snaps to accurate frame positions
- Changing Profiles: updates ALL position, start, and end values to be accurate frame positions - and then removes any 1 to 3 frame gaps detected
- Exporting to a different profile FPS: updates ALL position, start, and end values to be accurate frame positions - and then removes any 1 to 3 frame gaps detected

### DIFF
--- a/doc/profiles.rst
+++ b/doc/profiles.rst
@@ -61,6 +61,19 @@ Choose Profile Dialog
 4   Accept Profile      Click the :guilabel:`OK` button to switch to the selected profile.
 ==  ==================  ============
 
+Converting Profiles
+-------------------
+When switching profiles (or exporting to a different profile), OpenShot will do it's best to convert all clip,
+transition, and keyframe data to the new framerate (FPS). Certain properties, such as `position`, `start`, `end`, and `keyframes`
+will be updated to match the new framerate precision. For example, if moving from 30 FPS to 25 FPS, these properties will be changed
+from increments of 1/30 seconds to increments of 1/25 seconds. In order to preserve the overall timing accuracy of the
+timeline, OpenShot will match the `position` and `start` trim as closely as possible, and any tiny gaps (1-3 frames big)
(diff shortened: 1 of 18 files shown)


jonoomph:
[Prevent GAPS from appearing when snapping, changing Profile, and Exporting](https://trello.com/c/UfsTO5vH/4-prevent-gaps-from-appearing-when-snapping-changing-profile-and-exporting)

PR #5611, opened 2024-09-08
Selection + Copy / Paste Improvements

jonoomph:
- Adding new clipboard manager (for copy/paste support on timeline).
- Improving selections when right-clicking on clips/transitions/effects. Supports CTRL modifier.
- Added new 'refreshTimeline' controller method to timeline to force update of timeline selections
- Added new Copy context menu for individual Effects
- Added JSON syntax for all clipboard items (clips, transitions, effects, keyframes) to make it easy to copy/paste this data into an IDE
- Added support for pasting a clip or transition over other clips and transitions (replacing all attributes except position, layer, start, and end)
- Improved Paste_Triggered to always get the cursor and track position of the mouse (no longer uses playhead position)
- Improved Keybinds for Copy/Paste - so they work correctly with newly refactored methods


jonoomph:
[Paste should replace duplicate effects, add add new ones (not replace the original effects)](https://trello.com/c/Ur1v2GrW/6-paste-should-replace-duplicate-effects-add-add-new-ones-not-replace-the-original-effects)

Issue #5616, opened 2024-09-13
Colorpicker "Pick Screen Color" Does Not Work


kaihilbourne:
**UPDATE: Looks like this is an issue with the imported QColorDialog widget:**
Here is the documentation for QColorDialog: [https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QColorDialog.html#PySide2.QtWidgets.PySide2.QtWidgets.QColorDialog](url)
**Describe the bug:**
When trying to select a color using the "Pick Screen Color", the color picked does not match anything, and the cursor is not tracked outside of the colorpicker window.
![image](https://github.com/user-attachments/assets/164d2c0d-d1d3-4f5a-bdc2-63f32fb66168)

**Steps to reproduce the behavior:**
1. Apply the ChromaKey video effect to one of your clips. Alternatively, pick another effect that uses a color as a property. The color picker seems to be broken, not Chromakey.
2. Right click on the effect icon added to the clip and open "Properties"
3. Open the colorpicker menu by right-clicking the "Key Color" property.
4. Choose "Pick Screen Color" and move your mouse around. You should see that the number tracking the position of the cursor stops updating when you leave the pop-up window. Also, the color selected once you click to select does not seem to correspond to an actual color you have clicked.

**Expected behavior:**
Once "Pick Screen Color" has been selected, the large color selection box (bottom middle) should show the exact color that the cursor is hovering over. 
Additionally, the cursor should be tracked outside of the pop-up window so that you can select colors from your preview window, which I expect is the intended purpose of this tool. The color should be tracked everywhere, not just inside of OpenShot.
Once a left-click happens, the color should be locked in to the color value from underneath the cursor.

**System Details:**
- OpenShot Version 3.2.1-dev
- Operating System / Distro: Ubuntu 24.04.1

kaihilbourne:
I am attempting to work on this, and will post progress updates here as well as questions.

Colorjet3:
Hello @kaihilbourne I have assigned this to @jonoomph and also submitted this in the queue as a bug to be looked at.

kaihilbourne:
Looks like someone else has encountered this bug with QColorDialog:
[https://github.com/pyqtgraph/pyqtgraph/issues/2161](url)

kaihilbourne:
[...]

PR #5619, opened 2024-09-15
Fixing #5616 - NEEDS LOCAL TESTING - (Read description) - ColorPicker Select Screen Color 

kaihilbourne:
The fix is simple, and removes the `DontUseNativeDialog` option from the color_picker.py file. This results in the program using the native pop-up color picker. I am running Ubuntu 24.04.1 and the colorpicker looks like this:
image
Once you press the `+` button, the pop-up changes to one where you can use the eyedropper tool.
image

I have a feeling that using the native pop-up color picker might not be the most desirable solution, but it does work better than the current option.

### DIFF
--- a/src/windows/color_picker.py
+++ b/src/windows/color_picker.py
@@ -41,6 +41,7 @@ class ColorPicker(QWidget):
         super().__init__(parent=parent, *args, **kwargs)
         self.setObjectName("ColorPicker")
         # Merge any additional user-supplied options with our own
+        options = QColorDialog.ColorDialogOptions()
         if extra_options > 0:
             options = options | extra_options
         # Set up non-modal color dialog (to avoid blocking the eyedropper)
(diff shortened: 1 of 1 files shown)


src/windows/color_picker.py
```diff
@@ -41,7 +41,6 @@ def __init__(self, initial_color: QColor, callback, extra_options=0,
         super().__init__(parent=parent, *args, **kwargs)
         self.setObjectName("ColorPicker")
         # Merge any additional user-supplied options with our own
-        options = QColorDialog.DontUseNativeDialog
```
Vekhir, line 44: You can't just remove this variable, as it is used for custom options too (see lines below). When `extra_options` gets used, the program will crash.

It would probably be better to set `options` to `QColorDialog.ColorDialogOptions()`.

src/windows/color_picker.py
```diff
@@ -53,7 +52,6 @@ def __init__(self, initial_color: QColor, callback, extra_options=0,
         if title:
             self.dialog.setWindowTitle(title)
         self.dialog.setWindowFlags(Qt.Tool)
-        self.dialog.setOptions(options)
```
Vekhir: Similar to above, this cannot be removed as it's needed for custom options.

Vekhir (changes requested):
Without prejudice on its usefulness, the current suggestion would introduce a bug that can crash the program.

Vekhir (commented):
Comments are resolved.

PR #5621, opened 2024-09-17
Allow clip drag past edge of Timeline (without resetting position)

jonoomph:
Remove Droppable Track from Angular timeline code. Replace with static function called on Draggable End (to update UI data). Allows for clips to be dragged to edge of timeline without resetting back to original positions. Also fixed a race condition (on Web Engine) that caused a ng-click to happen after a drag (randomly) clearing selections.


jonoomph:
[Dragging playhead, clips, or transitions past the start of the timeline, should stick to 0.0 (not jump back)](https://trello.com/c/uphcXd8s/14-dragging-playhead-clips-or-transitions-past-the-start-of-the-timeline-should-stick-to-00-not-jump-back)

PR #5634, opened 2024-10-08
Adding new "Choose Profile" context menu to Project Files

jonoomph:
Adding new "Choose Profile" context menu to Project Files, to make it easy to edit using your source file width+height+FPS profile. Large refactor of how profile switching happens (moving to UpdateManger - so it will support undo/redo system). Add new profile() method to File Query class, to make it easy to generate/find a profile object for any File object.

Plus... avoid error message prompts when importing multiple files. 

image

image


jonoomph:
[Need simple way to set Project Profile from Project File](https://trello.com/c/XuPltHAa/17-need-simple-way-to-set-project-profile-from-project-file)

PR #5720, opened 2025-03-10
Adding alpha input back to color picker

jonoomph:
Adding alpha input back to color picker, so transparency can be set in the title editor and property editors. This also fixes an issue which preventing the previous color from initializing correctly in the color picker.

image
Screenshot from 2025-03-11 11-24-20


jonoomph:
[Color Picker screen is missing the "Alpha Channel:" field.](https://trello.com/c/WjymjHRs/39-color-picker-screen-is-missing-the-alpha-channel-field)

PR #5722, opened 2025-03-11
Fix upgrade logic for legacy profile (prevent accidental matches)

jonoomph:
Improve upgrade logic for legacy profiles, to prevent the accidentally matching of a new profile description with a legacy one (such as HD Vertical 1080p 50 fps)

Screenshot from 2025-03-11 12-49-30


jonoomph:
[Unable to set the default profile 1080 x 1350.](https://trello.com/c/Xp8yEmjj/26-unable-to-set-the-default-profile-1080-x-1350)

PR #5724, opened 2025-03-11
Making actionSave_trigger use a synchronous Save function

jonoomph:
Making actionSave_trigger use a synchronous save method, instead of a Thread, which prevents many race conditions related to "Recent Projects/Open Project" while saving unsaved changes at the same time.

image


jonoomph:
[Opening a Recent file and answering Yes to the Save prompt empties out the project and must be recovered.](https://trello.com/c/vxUl0oIS/24-opening-a-recent-file-and-answering-yes-to-the-save-prompt-empties-out-the-project-and-must-be-recovered)

PR #5725, opened 2025-03-12
Correctly access "rotation" metadata from vertical videos

jonoomph:
Fixing the access to "rotate" metadata (which was crashing) and adding better support for transform handles of rotated clips (the math got a bit broken when a clip was rotated). Related to https://github.com/OpenShot/libopenshot/pull/995


jonoomph:
[Vertical video is being cropped when imported.  Additionally, the Transform tool (blue handles) do not work when trying to adjust Scale X & Y](https://trello.com/c/0gwIonbl/6-vertical-video-is-being-cropped-when-imported-additionally-the-transform-tool-blue-handles-do-not-work-when-trying-to-adjust-sca)

PR #5754, opened 2025-05-06
Fixing Color bug when choosing effect colors (i.e. ChromaKey effect)

jonoomph:
Fixing color keyframes which were broken in the previous merge (i.e. chromakey and similar handling)


jonoomph:
[Chroma Key attributes are missing.](https://trello.com/c/2OqBLRfw/50-chroma-key-attributes-are-missing)

jonoomph:
[Chroma Key attributes are missing.](https://trello.com/c/2OqBLRfw/50-chroma-key-attributes-are-missing)

PR #5757, opened 2025-05-10
Fix misaligned tracks/track labels, and incorrect playhead height

jonoomph:
Refactor how scrolling track divs are resized to fit the webview widget, and how the playhead-line & snapping-line height is kept in-sync with the track heights. Also, this fixes a bug where clips don't instantly render when a track is removed, and also fixes a bug where track Y coordinates were out of sync with the UI (when adding/removing tracks), causing clips to jump to nearby tracks incorrectly.

Screenshot from 2025-05-09 16-19-21


jonoomph:
[Track misalignment.](https://trello.com/c/A9wnfxEx/41-track-misalignment)

PR #5758, opened 2025-05-10
Adding "None" option for Mask/Transition Source Property

jonoomph:
Adding "None" option for Mask/Transition source attribute, and really any "Reader" based property of any object. This clears the current reader.


jonoomph:
[Clear Source for Alpha Mask Effect & Transition](https://trello.com/c/4hHFj3Yv/36-clear-source-for-alpha-mask-effect-transition)

PR #5759, opened 2025-05-10
Snap Markers to FPS Precision (when changing profiles)

jonoomph:
Fixing marker data to snap to FPS precision when changing profiles (similar to clips, transitions, effects, and keyframes).

image


jonoomph:
[Previous/Next Marker icons stops working after changing profile.](https://trello.com/c/wAiZAKNw/43-previous-next-marker-icons-stops-working-after-changing-profile)

PR #5760, opened 2025-05-11
Use 3 digit milliseconds on Caption Editor

jonoomph:
Force the insert timestamp button in the Caption editor to use 3 digit milliseconds (which is the standard).

Screenshot from 2025-05-10 18-12-19


jonoomph:
[Captions Effect Timestamp (3 digits vs 2 digits millisecond)](https://trello.com/c/mPUSCFdJ/33-captions-effect-timestamp-3-digits-vs-2-digits-millisecond)

PR #5761, opened 2025-05-14
Fixing Regression when deleting an item (items were not really being deleted)

jonoomph:
Fixed a small regression which caused deleted clips/transitions/effects to be left in the project, even though the UI showed them gone. Also fixed a small issue outputting errors in Javascript when moving items (on occasion)


jonoomph:
[OpenShot does not play the correct video file when dragged onto a clip that was occupied by another clip.](https://trello.com/c/kCmLCL6M/92-openshot-does-not-play-the-correct-video-file-when-dragged-onto-a-clip-that-was-occupied-by-another-clip)

PR #5763, opened 2025-05-15
Protect Undo and Redo from Timers (prevent crashes)

jonoomph:
Protect undo and redo actions from selection and property timers, and prevent crashes in some situations (i.e. undoing 2 new selected clips). This was a tricky one, caused by a race condition between timers for our selection and property models and our libopenshot C++ updates. So when a Undo caused a selected clip to be deleted, the race was on, often causing a crash in Python.

### DIFF
--- a/src/classes/timeline.py
+++ b/src/classes/timeline.py
@@ -104,11 +104,6 @@ class TimelineSync(UpdateInterface):
                 self.window.refreshFrameSignal.emit()
 
             else:
-                if action.type == "delete":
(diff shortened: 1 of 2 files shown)


jonoomph:
[CTRL+Z does not remove/undo fade in and out with one press of CTRL+Z.  It requires 2 undos.](https://trello.com/c/5YWUCQdW/3-ctrlz-does-not-remove-undo-fade-in-and-out-with-one-press-of-ctrlz-it-requires-2-undos)

PR #5780, opened 2025-06-01
Fix for Color Picker on Mac (crash when choosing colors)

jonoomph:
Fixing a color picker bug which causes a crash on Mac due to 2 QPainters colliding. Also fixing a parent issue on the color picker used on the property dialog, to correctly size the dialog.

image


jonoomph:
[Color Picker crashing on Mac](https://trello.com/c/sWVK57O0/113-color-picker-crashing-on-mac)

PR #5783, opened 2025-06-08
Multi-Selection Editing (large refactor of selections)

jonoomph:
Full support for multiple-selection editing in the property dock and context menus. Added properties on the main window to emulate the original separate lists, to minimize code changes.


jonoomph:
[Investigate Multiple Selections in Properties dock (Multiple drags, context menus, multiple Effects, Clips, and Transitions)](https://trello.com/c/3oUGymKf/2-investigate-multiple-selections-in-properties-dock-multiple-drags-context-menus-multiple-effects-clips-and-transitions)

PR #5885, opened 2025-12-18
PyQt5, PyQt6, PySide6 Shim (Switch between different Python bindings for Qt versions)

jonoomph:
Related to https://github.com/OpenShot/libopenshot/pull/1027

  ### openshot-qt runtime

  - openshot-qt uses the env var `OPENSHOT_QT_API`.
  - Values:
      - `auto`: try bindings in order and use the first one that imports
      - `pyqt6`: *requires libopenshot to use Qt6*
      - `pyside6`:  *requires libopenshot to use Qt6*
      - `pyqt5`
  - In auto mode, it tries: `pyqt6`, then `pyside6`, then `pyqt5`.

  ### libopenshot build-time

  - libopenshot uses the CMake cache variable `-DUSE_QT6`.
  - Values:
      - `AUTO`: prefer Qt6 if found, otherwise fall back to Qt5
      - `ON`: force Qt6
      - `OFF`: force Qt5
  - After that, CMake sets QT_VERSION_MAJOR to 6 or 5 and links Qt${QT_VERSION_MAJOR}.


jonoomph:
[Support PyQt5, PyQt6, PySide6 Shim (Switch between different Python bindings for Qt versions)](https://trello.com/c/XEGaTv2E/204-support-pyside-for-qt6-shim-to-easily-toggle-between-pyqt6-and-pyside-without-breaking-pyqt5)

bgermann:
You could spare yourself the qt_api.py file by using [QtPy](https://github.com/spyder-ide/qtpy).

jonoomph:
> You could spare yourself the qt_api.py file by using [QtPy](https://github.com/spyder-ide/qtpy).

I did not know this exists, lol.  Thanks, I'll take a look.

philocalyst:
Does this also migrate to QtWebEngine? @jonoomph 

From my understanding QtWebKit is largely unmaintained.

bgermann:
There is already a QtWebEngine codepath additionally to QtWebKit. This PR does not change that.

philocalyst:
Is there a need for the dual code paths? Are there users relying on a particular renderer?

▰▰▰▰▰
Miles Wirth 🙃

philocalyst:
Or the addition of a compile-time flag? Just makes it impossible to package for Nix as the old engine is marked as vulnerable for 24 CVEs :(

jonoomph:
We currently support 3 timeline backends: QtWebkit, QtWebEngine, and a custom qwidget built one.  The webview backends will soon be removed - as they are only here for legacy reasons now. I wanted to keep them in the 3.5.x releases - but I plan on removing them from develop branch very soon.

PR #5913, opened 2026-02-03
Reduce size of Emoji Grid (horizontal margins)

jonoomph:
... also reduce some log noise around theme loading


jonoomph:
[Reduce Emoji Grid size (horizontal margins should be tighter)'](https://trello.com/c/I5Bp0tJs/239-reduce-emoji-grid-size-horizontal-margins-should-be-tighter)

PR #5915, opened 2026-02-04
Prevent crash on exit (when using --web-backend qwidget)

jonoomph:
Prevent a crash on shutdown when using the qwidget timeline backend (by ensuring the thumbnail thread is properly shut down first)


jonoomph:
[Crash on shutdown when using qwidget timeline backend](https://trello.com/c/2uh1s0yf/244-crash-on-shutdown-when-using-qwidget-timeline-backend)

PR #5916, opened 2026-02-04
Adding CTRL and SHIFT modifiers to Rotation

jonoomph:
Adding CTRL and SHIFT modifiers to snap rotation in the video preview transform handles. We snap by 15 degree increments only when pressing the modifier key.


jonoomph:
[Rotation Animation - Add CTRL+mouse control when rotating.](https://trello.com/c/zARFLhys/242-rotation-animation-add-ctrlmouse-control-when-rotating)

JMii63:
so you use the control and shift keys to rotate the video or image? Gonna have to try that out.

Colorjet3:
> so you use the control and shift keys to rotate the video or image? Gonna have to try that out.

Not "And".  It is CTRL + using the mouse OR SHIFT + using the mouse

PR #5917, opened 2026-02-05
Fix Undo for Adding Effect to Clip (regression)

jonoomph:
Fixing crash on deleting clips with effects, and fixing regression that causes effects to break the UNDO/REDO system - not allowing the user to undo an effect added to a clip.


jonoomph:
[Effects on clip cannot be always undone (UNDO/REDO) - this is a regression](https://trello.com/c/aks6m1Nd/249-effects-on-clip-cannot-be-always-undone-undo-redo-this-is-a-regression)

PR #5922, opened 2026-02-10
Enable "Use Blender GPU rendering for Animated Titles" as Default setting

jonoomph:
Enable "Use Blender GPU rendering for Animated Titles" setting by default (seems very stable now)


jonoomph:
[For Blender Performance, remove the "Experimental" and Enable it (checkbox) by default.](https://trello.com/c/eJWykXEW/246-for-blender-performance-remove-the-experimental-and-enable-it-checkbox-by-default)

PR #5925, opened 2026-02-13
Experimental Timeline Improvements (styles, keyframes, transitions, themes, misc bug fixes)

jonoomph:
This ended up being a really large PR, affecting many different things - oops. 

- bug with opening the Preview dialog or Split File dialog - so we need to fix that as well.
- I included a tiny improvement to show the emoji name in Project Files, instead of the ugly SVG file name.
- Removing a few icons from the Experimental timeline (seem a bit redundant and add complexity and annoying accidental clicks):
  - Add Track Above
  - Add Track Below
  - Delete Track
- Added Lock functionality into Experimental Timeline (undraggable, undeletable) - and new lock style (dimmer version of normal theme track and borders and clip styles). Matching JS styles to this also.
- Moving thumbnails downward on Humanity Dark / Retro (Experimental timeline), and improving all the themes in the Experimental Timeline (to more closely match the JS themes)
- Keyframe Panel Improvements / fixes
  - Snapping issues between clip keyframe icon and keyframe panel icons
  - Support for multiple selected clips
  - Auto select panel keyframes when selecting or dragging top clip keyframe icon
  - DELETE key support when keyframe panel icons are selected, same if a clip’s keyframe is selected (DELETE keyframe before the clip). Sequence:  DELETE keyframe panel selections, DELETE all keyframes at a specific frame, DELETE clip
  - Dragging a clip to tracks hides/shows the keyframe panel as you would expect
  - Inserting keyframes works with multiple selected clips, where the playhead overlaps it
  - Use ↔︎ cursor icons when hovering over points in the keyframe panel
  - Improve selection box on keyframe panel, so the tiny gaps between the property rows do not unselect the clip.
- Transition Auto-Direction fixes (auto reverses when moved to each side of a clip), or when initially dropped
- Transition now has transparency on the qwidget backend
- Transition now moves keyframe icons when resizing it (matches JS behavior)


jonoomph:
[Experimental Timeline Improvements (styles, keyframes, transitions, themes)](https://trello.com/c/DWnp0PNW/259-experimental-timeline-improvements-styles-keyframes-transitions-themes)

PR #5926, opened 2026-02-13
Fix Play/Pause button state when playback reaches end of timeline

jonoomph:
Fix playback state checks and refresh behavior (related to https://github.com/OpenShot/libopenshot/pull/1036)

  - Use robust play-mode checks for play/ff/rewind (PLAYBACK_PLAY + non-zero speed).
  - Skip refreshFrame() seeks during active playback to avoid UI/selection refresh perturbing frame progression.


jonoomph:
[OpenShot does not recognize that it has reached at the end of the timeline and stays in "Play"/"Preview" mode.](https://trello.com/c/RIeQxaoM/260-openshot-does-not-recognize-that-it-has-reached-at-the-end-of-the-timeline-and-stays-in-play-preview-mode)

PR #5939, opened 2026-02-24
Fixing Duplicate Imports

jonoomph:
This PR fixes some issues with duplicate file paths, which import would not catch due to URL absolute vs relative logic. Also, fixed an issue with timeline focus (Experimental timeline backend) after dropping a new file on the timeline, so DELETE only removes the new clip.  Lastly, batched Project Files model updates during "Open Project", to improve the speed of opening a large project with lots of files.

### DIFF
--- a/src/classes/path_utils.py
+++ b/src/classes/path_utils.py
@@ -137,3 +137,18 @@ def normalize_path(path_value):
     if not path_value:
         return ""
     return path_value.replace("\\", "/")
+
+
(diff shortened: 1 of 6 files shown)


jonoomph:
[v3.4.0 Production Release allows importing of duplicate files.  You can import the same files many times over.](https://trello.com/c/oLfY788i/271-v340-production-release-allows-importing-of-duplicate-files-you-can-import-the-same-files-many-times-over)

PR #5952, opened 2026-03-13
Add fade_audio_hint to transitions created automatically

jonoomph:
Add fade_audio_hint to transitions created automatically, which requests the Timeline to fade audio as well as the image. If the transition is on the left side of a clip, it will fade in audio. If the transition is on the right side of a clip, it will fade out audio. And if the transition is overlapping 2 clips, it will fade out one, and fade in the other, using an equal-power curve.


jonoomph:
[Audio Transitions (Equal-Power Cross-Fade)](https://trello.com/c/V6hLgfvv/293-audio-transitions-equal-power-cross-fade)

PR #5953, opened 2026-03-13
Use Constant Interpolation when inserting keyframes from Dropdowns

jonoomph:
When dropdowns are used to keyframe a value, use constant interpolation, since these values always represent binary choices (like types, enums, modes), and never ranges (linear, bezier). This was done in response to a bug with "Draw Box" on the Tracker effect, which was using linear interpolation between 2 boolean (0 to 1) values.


jonoomph:
[Tracker Effect - Draw Box not working properly.](https://trello.com/c/zw97Iazf/189-tracker-effect-draw-box-not-working-properly)

PR #5954, opened 2026-03-13
Improving dragging multiple selections on qwidget timeline backend

jonoomph:
Improving dragging multiple selections on qwidget timeline backend, to ensure they are moved as an atomic unit, with no drift during fast dragging and dropping.


jonoomph:
[Fix dragging multiple selections in qwidget timeline (they should not be shifting during drop)](https://trello.com/c/yTAX4eJW/294-fix-dragging-multiple-selections-in-qwidget-timeline-they-should-not-be-shifting-during-drop)

PR #5955, opened 2026-03-13
Stop playback when trimming starts, and reset play button correctly

jonoomph:


jonoomph:
[Adjusting the right edge of a video file during preview causes audio issues.](https://trello.com/c/g6BikqRP/85-adjusting-the-right-edge-of-a-video-file-during-preview-causes-audio-issues)

PR #5956, opened 2026-03-14
Vertical Video support for Tracker Effect (rotated Coordinates)

jonoomph:
Adding vertical video coordinate mapping to region dialog, when choosing a region for the Tracker effect (so it unrotates coordinates correctly)


jonoomph:
[Vertical videos + tracker and object detection](https://trello.com/c/gYorQLTl/181-vertical-videos-tracker-and-object-detection)

PR #5958, opened 2026-03-14
Fix Missing File Dialog (only choose the directory)

jonoomph:
Fix regression which caused missing file dialog to choose a specific missing file, instead of just the directory.


jonoomph:
["Missing File" process is now forcing to select the file instead of just selecting the folder.](https://trello.com/c/pv2mlaRK/295-missing-file-process-is-now-forcing-to-select-the-file-instead-of-just-selecting-the-folder)

PR #5968, opened 2026-03-21
Fix user-guide/docs for ComfyUI

jonoomph:
We had some incorrect instructions in our ComfyUI instructions, and also needed retranslation.


jonoomph:
[ComfyUI Docs has a few typos (mentions Experimetnal tab in preferences)](https://trello.com/c/lmk43ZZr/11-comfyui-docs-has-a-few-typos-mentions-experimetnal-tab-in-preferences)

PR #5969, opened 2026-03-21
Fixing unit tests discovery issues (failing on launchpad builders)

jonoomph:
- Added a shared test helper at src/qt_test_app.py so tests can retrofit an existing Qt app instance with the OpenShot-like methods/state they need.
- Changed src/language/test_translations.py to create its QCoreApplication lazily instead of at import time, which removes the discovery-order poison pill.

### DIFF
--- a/src/language/test_translations.py
+++ b/src/language/test_translations.py
@@ -68,8 +68,12 @@ class Color:
         return cls._green + str(*args) + cls._reset
 
 
-# Get app instance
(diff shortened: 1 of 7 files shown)


jonoomph:
[Python unit tests are failing on Gitlab builders (for PPA builds)](https://trello.com/c/RpSgqlTi/15-python-unit-tests-are-failing-on-gitlab-builders-for-ppa-builds)

PR #5975, opened 2026-03-22
Fixing Framerate in File Properties, Profile Editing, Export Dialog (for extreme large FPS values)

jonoomph:
- Also limiting framerate editing to image sequences only (as that was the initial intent, and editing fps fractions for videos breaks libopenshot pretty bad (when searching for PTS)


jonoomph:
[File Profile (FPS) does not match source file once imported.  Also causing bad exported file.](https://trello.com/c/SdAfWQ4H/3-file-profile-fps-does-not-match-source-file-once-imported-also-causing-bad-exported-file)

PR #5978, opened 2026-03-22
Fixing keyframe detection logic to better detect keyframe heuristics

jonoomph:
Fixing keyframe detection logic to better detect keyframes (even empty ones with no points), for example when you delete all keyframes for a clip using the keyframe panel. Now, all keyframes available still appear when right clicking on the keyframe panel. Also, some log adjustments have been made around deleting keyframes.


jonoomph:
[Deleting all keyframes makes clip disappear (instead of restoring defaults)](https://trello.com/c/1CHFCkiO/16-deleting-all-keyframes-makes-clip-disappear-instead-of-restoring-defaults)

PR #5980, opened 2026-03-23
Attempt to fix spinning cursor on Windows (mp3 files)

jonoomph:
Attempt to fix spinning cursor on Windows - that appears to spin forever after dropping an mp3 file onto the timeline.


jonoomph:
[In Windows, import a audio(.mp3) and drag it onto track 5 position 0:00... the mouse is a twirling circle forever](https://trello.com/c/omMeW4Z7/18-in-windows-import-a-audiomp3-and-drag-it-onto-track-5-position-000-the-mouse-is-a-twirling-circle-forever)

jonoomph:
No good - didn't fix the issue

PR #5981, opened 2026-03-24
Don't mark project as dirty on launch (due to scale equivalence checks)

jonoomph:


jonoomph:
[Why Prompted to save a project when nothing has happened?  (default sample rate changing on launch)](https://trello.com/c/kpt4QETx/17-why-prompted-to-save-a-project-when-nothing-has-happened-default-sample-rate-changing-on-launch)

PR #5982, opened 2026-03-24
Fixing multi-selection drags of transitions to not apply any auto-direction

jonoomph:
Fixing multi-selection drags of transitions to not apply any auto-direction updates (just move them). Also, apply clip moves first, and then apply transition moves, to keep things in batches and process them in a better order. This fixes a bug where by moving a bunch of clips + transitions, it could reverse the transitions due to invalid ordering.

### DIFF
--- a/src/tests/test_timeline_helpers.py
+++ b/src/tests/test_timeline_helpers.py
@@ -208,6 +208,67 @@ class TimelineHelperTests(unittest.TestCase):
 
         return Helper()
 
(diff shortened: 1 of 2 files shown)


jonoomph:
[Transitions lose the "Reverse Transition" function when they are shifted off the clip and then back onto the clip.](https://trello.com/c/qBcilnRm/8-transitions-lose-the-reverse-transition-function-when-they-are-shifted-off-the-clip-and-then-back-onto-the-clip)

PR #5997, opened 2026-04-09
Fixing file importing to fallback with inspect_reader=True (if fails)

jonoomph:
Fixing file importing to fallback with inspect_reader=True if fail happens. This forces the CreateReader() to attempt to open each reader and calls a more gracefull/complete fallback sequence. So a failed flac import in QtImageReader will successfully be tried by FFmpegReader next.


jonoomph:
[Fix Flac Extension Support](https://trello.com/c/1Y2SEXxu/221-fix-flac-importing)

PR #6004, opened 2026-04-16
Make macOS DMG filename arch-aware (arm64 / x86_64)

geekrebel:
The `Darwin` branch of `installer/build_server.py` currently hard-codes
`-x86_64.dmg` as the DMG suffix. That name is fine on Intel Macs but
mis-labels any Apple Silicon build, and (more importantly) would prevent an
arm64 and an Intel DMG from coexisting under `releases.openshot.org/mac/`.

This PR detects the arch via `platform.machine()`:

* `arm64` produces `OpenShot-<version>-arm64.dmg`.
* Anything else (including Rosetta-translated Python on Apple Silicon)
  produces `OpenShot-<version>-x86_64.dmg`, which is byte-identical to
  today's name.

So there is no observable change for existing Intel Mac build pipelines.
Once an Apple Silicon runner is available, it will produce a clearly
labelled `-arm64.dmg` artifact that can sit alongside the Intel one.

The Linux AppImage and Windows `.exe` suffixes are left untouched; they
are still `x86_64`-only in practice, and changing them is out of scope
for this PR.

This is the second in a small series of Apple Silicon PRs (following
#6003, the "Apple VideoToolbox" label rename). The next planned change
is replacing the hard-coded `/usr/local/Cellar/python@3.7/3.7.9_2/...`
PATH in `installer/build-mac-dmg.sh`, which is required before any arm64
Homebrew (`/opt/homebrew`) machine can run the mac packaging script
end-to-end.

### DIFF
--- a/installer/build_server.py
+++ b/installer/build_server.py
@@ -385,7 +385,10 @@ def main():
             app_name += "-x86_64.AppImage"
             app_upload_bucket = "releases.openshot.org/linux"
         elif platform.system() == "Darwin":
-            app_name += "-x86_64.dmg"
+            # Apple Silicon (arm64) and Intel (x86_64) Macs produce arch-suffixed
+            # DMGs so both can be published side-by-side to the mac release bucket.
+            mac_arch = "arm64" if platform.machine() == "arm64" else "x86_64"
+            app_name += "-%s.dmg" % mac_arch
             app_upload_bucket = "releases.openshot.org/mac"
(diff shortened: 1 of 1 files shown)


jonoomph:
[Make macOS DMG filename arch-aware (arm64 / x86_64)- #6004](https://trello.com/c/zuHCYie9/239-make-macos-dmg-filename-arch-aware-arm64-x8664-6004)

jonoomph:
LGTM! Thanks!

PR #6007, opened 2026-04-17
Displacement Effect (Icons and Docs)

jonoomph:
Related to https://github.com/OpenShot/libopenshot/pull/1059


jonoomph:
[New Effect: Displacement Map](https://trello.com/c/4mPb4fQh/234-new-effect-displacement-map)

PR #6009, opened 2026-04-17
Remove Webview Timeline (QtWebKit & QtWebEngine)

jonoomph:
Including all packaging, freeze, preferences, documentation assets/mentions, and our `--web-backend` arg. The webview timelines are now replaced with a custom qwidget timeline which is much more featureful, less laggy, and better in every way.


jonoomph:
[Removing webview timelines (code, packaging, and docs)](https://trello.com/c/jQrwn8hT/235-removing-webview-timelines-code-packaging-and-docs)

PR #6010, opened 2026-04-17
Drop Shadow and Glow Icons & Documentation

jonoomph:
Related to https://github.com/OpenShot/libopenshot/pull/1060


jonoomph:
[New Drop Shadow Effect](https://trello.com/c/kgpvGiYa/237-new-drop-shadow-effect)

PR #6013, opened 2026-04-20
Adding fallback sip import for our PyQt5 build servers

jonoomph:
Adding fallback sip import for our PyQt5 build servers (Linux AppImage was failing to import sip)


jonoomph:
[Audio Background Denoiser (audio effect)](https://trello.com/c/VkWTj13n/220-audio-background-denoiser-audio-effect)

PR #6016, opened 2026-04-25
Color Grade Effect and Workflow (Scopes, Correction, Grading, Curve Editing, Wheels)

jonoomph:
color-view-grading-skin-tone


jonoomph:
[First Class Color Grading and Correction!](https://trello.com/c/gQ2l7Vqe/240-first-class-color-grading-and-correction)

PR #6029, opened 2026-05-06
Improved screen dock/widget restoring (at > 100% screen scale)

jonoomph:
Improved screen dock restoring at > 100% screen scales, regardless of OS scaling or OpenShot's internal scaling. The goal is to restore the screen as accurately as possible on launch, and without this change, the video widget kept getting smaller and smaller with each launch.

### DIFF
--- a/src/classes/settings.py
+++ b/src/classes/settings.py
@@ -114,11 +114,11 @@ class SettingStore(JsonDataStore):
         if key in user_values:
             user_values[key].update({"value": value})
         else:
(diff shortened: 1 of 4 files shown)


jonoomph:
[Video Widget gets smaller with each launch (when UI Scale > 100%)](https://trello.com/c/KMWSKrQD/262-video-widget-gets-smaller-with-each-launch-when-ui-scale-100)

PR #6031, opened 2026-05-07
Denoise Effect

jonoomph:
denoiseimage@2x


jonoomph:
[Denoise Image effect (low light, sensor noise, noisy renders)](https://trello.com/c/NdeNGaBy/261-denoise-image-effect-low-light-sensor-noise-noisy-renders)

PR #6034, opened 2026-05-08
Misc Updates to Docs

jonoomph:
- Updated clip docs: sorted Clip Properties and added Waveform Mode.
- Updated effects docs: sorted the Effects list to match OpenShot and added common mask properties.
- Updated transition docs: added mask properties and descriptions.
- Replaced Preferences Timeline and Performance screenshots with native JPEG versions from the new captures.
- Verified docs with make html; build succeeds.


jonoomph:
[Feedback on Documentation](https://trello.com/c/p3z7MNo7/257-feedback-on-documentation)

PR #6037, opened 2026-05-09
Use reader-applied orientation in preview, region, thumbnails (i.e. let libopenshot rotate videos/photos)

jonoomph:
Use reader-applied orientation in preview, region, thumbnails, and proxy paths (source media is now rotated prior to Clip)

- Pass the internal reader_orientation_mode flag when creating temporary preview and region clips, remove duplicate thumbnail/proxy rotation handling, and update preview dialog sizing from oriented reader dimensions.

- Also make preview shutdown wait for the worker/player stop path to avoid crashes when closing the dialog during playback.

Related to https://github.com/OpenShot/libopenshot/pull/1073

### DIFF
--- a/src/classes/proxy_service.py
+++ b/src/classes/proxy_service.py
@@ -732,7 +732,6 @@ class ProxyService(QObject):
                      max_frame = int(source_reader.get("video_length", 0) or 0)
                      if max_frame <= 0:
                          raise RuntimeError("invalid source frame count")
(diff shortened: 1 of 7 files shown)


jonoomph:
[Vertical videos + tracker and object detection](https://trello.com/c/gYorQLTl/78-vertical-videos-tracker-and-object-detection)

PR #6042, opened 2026-05-09
Fix timeline marker and cache bar vertical scrolling

jonoomph:
- Keep timeline marker icons and hit targets fixed in the ruler when vertically scrolling tracks, while preserving horizontal timeline scrolling. Also repaint the playback cache lane background under the ruler so the cache bar no longer appears to float over clips when track backgrounds scroll away. 
- Add regression coverage for both the marker viewport behavior and the fixed cache lane background.


jonoomph:
[Markers do not stay fixed below on the timline ruler when you scroll the tracks.](https://trello.com/c/L2zDJ3sp/268-markers-do-not-stay-fixed-below-on-the-timline-ruler-when-you-scroll-the-tracks)

PR #6047, opened 2026-05-11
Object detection migration to Yolo v5 (ONNX) and new Download UI

jonoomph:
Plus lots of UI improvments, global object style editing, and lots more.


jonoomph:
[Object Detector effect attributes not manageable once the effect is applied.](https://trello.com/c/jFdpu2LX/37-object-detector-effect-attributes-not-manageable-once-the-effect-is-applied)

PR #6051, opened 2026-05-24
Preserve import selection order

jonoomph:
- Keep top-level file URL import order intact instead of globally sorting all paths.
- Sort recursively imported directory names and filenames during os.walk() for deterministic directory imports.
- Add regression coverage for mixed file/directory imports to ensure directory contents expand in place.
- Guard thumbnail view refresh sorting when the files proxy is intentionally unsorted.


jonoomph:
[Imported Source file are now automatically sorted instead of being saved in the order they are imported.](https://trello.com/c/eTDowor4/249-imported-source-file-are-now-automatically-sorted-instead-of-being-saved-in-the-order-they-are-imported)

PR #6052, opened 2026-05-24
Allow mouse wheel tilt (left/right) to scroll timeline

jonoomph:
- Restored horizontal timeline scrolling via mouse-wheel tilt by handling angleDelta().x() / pixelDelta().x() in wheelEvent() and routing that input through the existing deferred horizontal scroll accumulator/timer path. This allows left/right wheel tilt to scroll without requiring Shift.
- Kept existing Shift + wheel behavior intact (still mapped to horizontal scrolling using Y delta).
- Extended timeline helper wheel-event stubs to support X-axis deltas, enabling regression tests for tilt input in unit tests.
- Added a regression test that verifies horizontal scroll occurs from tilt-wheel input with no modifier keys.

### DIFF
--- a/src/tests/test_timeline_helpers.py
+++ b/src/tests/test_timeline_helpers.py
@@ -546,9 +546,13 @@ class TimelineHelperTests(unittest.TestCase):
                 return self.active
 
         class DeltaStub:
-            def __init__(self, y=0.0, is_null=False):
+            def __init__(self, x=0.0, y=0.0, is_null=False):
+                self._x = float(x)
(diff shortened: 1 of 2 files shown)


jonoomph:
[Unable to scroll the timeline forward or backward using the left/right tilt on the mouse wheel.](https://trello.com/c/N8nAE2Ou/270-unable-to-scroll-the-timeline-forward-or-backward-using-the-left-right-tilt-on-the-mouse-wheel)

PR #6109, opened 2026-09-04
Preserve clip positions and transform alignment across the 4.0.0 upgrade

jonoomph:
Related to https://github.com/OpenShot/libopenshot/pull/1093

This complements the libopenshot location fix by keeping OpenShot’s transform handles, rendered output, and saved projects consistent.

  ### Changes

  - Use geometry-aware positioning only for Crop clips.
  - Restore canvas-relative transform positioning for Fit, Stretch, and None.
  - Preserve the existing migration for legacy Crop projects.
  - Convert affected non-Crop projects saved by the official 4.0.0 release once, preserving their 4.0.0 visual
    positions.

  - Avoid repeated conversion when those projects are reopened in 4.0.1 or later.

  ### Compatibility

  - Projects last saved in 3.5.1 and earlier retain their original positions.
  - Projects adjusted and saved in 4.0.0 do not jump again after upgrading.
  - Transform handles now match libopenshot’s rendered clip positions.

  Tests cover legacy projects, exact 4.0.0 migration, 4.0.1 reopening, all relevant scale modes, keyframed values, gravity, margins, and reported customer geometry.

### DIFF
--- a/src/classes/project_data.py
+++ b/src/classes/project_data.py
@@ -939,18 +939,30 @@ class ProjectDataStore(JsonDataStore, UpdateInterface):
         ):
             self._migrate_legacy_crop_locations()
 
+        # OpenShot 4.0.0 saved non-Crop locations using libopenshot 1.0's
+        # temporary geometry-relative behavior. Convert those values back to
+        # canvas-relative units so the positions remain unchanged with the
+        # corrected libopenshot behavior.
+        if (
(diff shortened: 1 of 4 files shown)


jonoomph:
[Project created in v3.5.1-Release-15816 then opened in v4.0.0-RC is not consistent.](https://trello.com/c/VEbIfDsu/128-project-created-in-v351-release-15816-then-opened-in-v400-rc-is-not-consistent)

PR #6126, opened 2026-09-13
Trigger feedback banner after editing experience and release usage

jonoomph:
Require three lifetime actions across two categories and 20 minutes of active editing per release. Show the banner during a quiet moment and acknowledge clicks or dismissal for that release.

Keep editing hooks small, preserve Qt signal behavior, and remove the input filter on shutdown. Cover eligibility, editing commands, 154 live menu actions, undo/redo, and shutdown with regression tests.

### DIFF
--- a/src/classes/feedback.py
+++ b/src/classes/feedback.py
@@ -1,12 +1,18 @@
-"""One-time survey policy and URL metadata (no automatic network requests)."""
+"""Feedback eligibility and survey URL metadata (no automatic network requests)."""
 
 import base64
 import json
 import math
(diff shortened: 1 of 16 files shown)


jonoomph:
[Feedback Banner Trigger improvements (after 3 editing actions, and 20 minutes of editing use) - never appear after dismissal.](https://trello.com/c/dYK7UrWG/240-feedback-banner-trigger-improvements-after-3-editing-actions-and-20-minutes-of-editing-use-never-appear-after-dismissal)

PR #6127, opened 2026-09-13
Fix feedback snapshot syntax for Python 3.7 macOS builds

jonoomph:
Replace the Python 3.8 assignment expression with a plain loop, preserving snapshot behavior.


jonoomph:
[Feedback Banner Trigger improvements (after 3 editing actions, and 20 minutes of editing use) - never appear after dismissal.](https://trello.com/c/dYK7UrWG/240-feedback-banner-trigger-improvements-after-3-editing-actions-and-20-minutes-of-editing-use-never-appear-after-dismissal)

PR #6130, opened 2026-09-13
Improve keyframe targeting and preserve playhead position after dragging

jonoomph:
Makes keyframe icons easier to click and drag, especially near clip edges, with slightly larger hit areas.

Fixes the playhead jumping back after dragging a keyframe. It now stays on the moved keyframe, including in the keyframe panel for trimmed or repositioned clips.


jonoomph:
[Improve keyframe targeting and keep the playhead in place after dragging](https://trello.com/c/DAsaOVX4/243-improve-keyframe-targeting-and-keep-the-playhead-in-place-after-dragging)

PR #6134, opened 2026-09-14
Restore native AppImage file dialogs with correct starting folders

jonoomph:
Restore native file dialogs when the desktop portal supports the requested starting folder. Older or unavailable portals fall back to Qt’s picker, preserving folder preferences.

Also prevent menus from briefly graying out before the Qt fallback opens.


jonoomph:
[Restore native AppImage file dialogs with correct starting folders- #6134 #6134](https://trello.com/c/qehij1Jt/246-restore-native-appimage-file-dialogs-with-correct-starting-folders-6134-6134)

PR #6141, opened 2026-09-19
Fix feedback crashes, banner icons, and startup with older libopenshot

jonoomph:
Guard feedback mouse-event access, reuse tinted menu icons in banners without double high-DPI scaling, and allow startup with older libopenshot logging bindings.

Cover event-wrapper mismatches, logging lifecycle compatibility, and banner artwork at 100%, 150%, and 200% display scales.


jonoomph:
[Fix feedback crashes, banner icons, and startup with older libopenshot](https://trello.com/c/lsaTQp1T/250-fix-feedback-crashes-banner-icons-and-startup-with-older-libopenshot)

PR #6142, opened 2026-09-19
Use native camera discovery for Linux webcam recording

jonoomph:
Related to https://github.com/OpenShot/libopenshot/pull/1100


jonoomph:
[Linux camera discovery built-in to libopenshot (no external app needed)](https://trello.com/c/2imAfUqI/251-linux-camera-discovery-built-in-to-libopenshot-no-external-app-needed)

